PR #540, opened 2014-08-04
Resteasy-1025 async http client engine

MarkusKull:
Adds a new interface "AsyncHttpClientEngine" deriving from ClientHttpEngine to support truly async communication model. ClientInvocation uses this optional interface if available. 
There is also a behaviour change regarding InvocationCallbacks: the callback is now called in parallel/after the future, similar to jersey. Thus the callback and the future will now see always the same result (and the future will see no exception thrown by callback).

Also included is a implementation for apache httpcomponents httpasyncclient. Care has been taken to not change the existing maven dependencies except of a new optional dependency.


eleduardo:
This is a very old but very desirable capability on RestEasy specially when consumers need to use their own engine. Netty based clients could really leverage this new interface. Is there any change that this PR or at least this feature would be added to RestEasy?

avram:
As @MarkusKull previously noted, bringing this PR up to date and helping it get a complete review would be a huge step towards that. I know that my teams would be interested in vetting a full-async Netty implementation of this interface and I'm sure others would as well.

MarkusKull:
I will try to bring this PR up-to-date in the next weeks.

MarkusKull:
Updated the PR. Just some minor conflicts.

ronsigal:
Thank you, Marcus. You have not been forgotten.

stevenschlansker:
We're looking into RESTEasy's asynchronous support -- is this PR going to move anytime soon?  Sounds like currently the client does not actually support async calls.

liweinan:
I'm working on migrating this PR: https://github.com/liweinan/Resteasy/tree/resteasy-async-client

liweinan:
https://github.com/resteasy/Resteasy/pull/1228

asoldano:
@MarkusKull , we're eventually moving forward with https://github.com/resteasy/Resteasy/pull/1228 , which is an updated/rebased version of this PR. Unless you already did that in the past, would you mind explicitly writing here that you agree your contributions being made under the ASL 2.0 License http://www.apache.org/licenses/LICENSE-2.0 ? Thanks!

MarkusKull:
@asoldano Good Work! Yes, i agree for my contributions being made under Apache License 2.0 https://www.apache.org/licenses/LICENSE-2.0 .

asoldano:
@MarkusKull , thanks, I've merged the other PR, closing this now.

PR #584, opened 2014-09-22
WFLY-2859 Don't turn CDI beans into application and request scoped beans

stuartwdouglas:
Instead a listener is used to manually control the lifecycle


stuartwdouglas:
Sorry, I forgot about this. I have updated the PR, although I can't compile upstream at the moment to test.

patriot1burke:
Had to revert this as it was breaking a cdi test

PR #615, opened 2014-12-01
SpringBeanProcessor didn't find methods in super-class any more 

mihxil:
It still did in 3.0.9. This makes the application broken, e.g. if you also use org.springframework.data.neo4j.aspects.config.Neo4jAspectConfiguration, which is a bean extending from another one.


VanRoy:
Hi @mihxil, @wilkinsona 

This problem seems always be present in 3.0.11.Final.
This problem break the "Spring Boot" intégration too.

Do you have any news about this PR or a fix of this issue ?

mihxil:
@VanRoy I haven't. I noticed the problem, and made a fix and a pull request, but afaik it wasn't pulled or fixed otherwise, and we're still using 3.0.9 because of that.

mihxil:
See https://issues.jboss.org/browse/RESTEASY-1212

mihxil:
New pull request in 
https://github.com/resteasy/Resteasy/pull/659

PR #620, opened 2015-01-15
fixes RESTEASY-1149

krisskross:
Tests fails on my environment without this fix so its hard to judge if the fix is correct. Please review.


patriot1burke:
This is incorrect fix.  You are wiping out any existing context data.

PR #633, opened 2015-03-20
RESTEASY-1163: use RestEasy's own UriBuilder and clone on resolveTemplate()

eeichinger:
ensure to use RestEasy's own UriBuilder and proper cloning uriBuilder instances on resolveTemplateXXXX()

see https://issues.jboss.org/browse/RESTEASY-1163

### DIFF
--- a/jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientWebTarget.java
+++ b/jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientWebTarget.java
@@ -35,19 +35,19 @@ public class ClientWebTarget implements ResteasyWebTarget
    public ClientWebTarget(ResteasyClient client, String uri, ClientConfiguration configuration) throws IllegalArgumentException, NullPointerException
    {
       this(client, configuration);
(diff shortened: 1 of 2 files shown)


ronsigal:
1. Dependency issues should be resolved outside of Resteasy.
2. The call to UriBuilder.clone() seems to be unnecessary. ResteasyUriBuilder.resolveTemplate() returns a new UriBuilder.

eeichinger:
@ronsigal  you missed the point: it's not about the dependency issue, but in short about the different behaviour of resolveTemplate() in RestEasy and Jersey. Jersey's resolveTemplate() does *not* return a new UriBuilder. If you use RestEasy with Jersey's UriBuilder you end up with very hard to find & debug url problems.

ronsigal:
I'm still missing the point. Why would you use Resteasy with Jersey's UriBuilder?

asoldano:
I'm closing this PR as I basically agree with Ron. The RESTEasy client should not end up being used Jersey stuff, many other weird problem could possibly happen.

eeichinger:
then please at least give out a warning or an error - it happened to us accidentally with maven to have both libs on the classpath, unfortunately jersey's uribuilder got pulled in first. It took as a day to figure out what was wrong.

PR #636, opened 2015-04-01
Truncate seconds before comparing HTTP-date based preconditions.

JessThrysoee:
The HTTP-date (RFC7231 Section 7.1.1.1) format used for headers,

Last-Modified
If-Modified-Since
If-Unmodified-Since

can not express millliseconds. So to avoid unexpected mismatches of
java.util.Date objects given as argument to evaluatePreconditions
it should be rounded down to the nearest second.


ronsigal:
These changes don't seem right. DateUtil.parseDate("Mon, 1 Jan 2007 00:00:00 GMT").getTime() is
1167609600000 and that plus 117 milliseconds is 1167609600117. If a file was modified at 1167609600117, then "if modified since 1167609600000" should be true and "if unmodified since 1167609600000" should be false. But if you truncate, you get false and true, respectively.

asoldano:
Having heard nothing from @JessThrysoee , after @ronsigal comment, I'm closing this. Feel free to reopen if there's more to discuss.

PR #639, opened 2015-04-07
[Netty4] Correct handling for absolute URIs in HTTP request lines

avram:
Correctly handle requests with an absolute URI in the request line, like:

```
GET http://www.example.com/testContent HTTP/1.1
Host: www.example.com
```

This is required per the spec: http://tools.ietf.org/html/rfc2616#section-5, specifically: 

> To allow for transition to absoluteURIs in all requests in future versions of HTTP, all HTTP/1.1 servers MUST accept the absoluteURI form in requests, even though HTTP/1.1 clients will only generate them in requests to proxies.

And:

> If Request-URI is an absoluteURI, the host is part of the Request-URI. Any Host header field value in the request MUST be ignored.

Issue: https://issues.jboss.org/browse/RESTEASY-1193

### DIFF
--- a/jaxrs/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/NettyUtil.java
+++ b/jaxrs/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/NettyUtil.java
@@ -8,12 +8,10 @@ import org.jboss.resteasy.spi.ResteasyUriInfo;
 import org.jboss.resteasy.util.CookieParser;
 import org.jboss.resteasy.util.HttpHeaderNames;
 import org.jboss.resteasy.util.MediaTypeHelper;
-import org.jboss.resteasy.util.PathHelper;
 
 import javax.ws.rs.core.Cookie;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
-import javax.ws.rs.core.UriBuilder;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -31,9 +29,16 @@ public class NettyUtil
       String host = HttpHeaders.getHost(request, "unknown");
       String uri = request.getUri();
 
-      String uriString = protocol + "://" + host + uri;
+      String uriString;
+
+      // If we appear to have an absolute URL, don't try to recreate it from the host and request line.
+      if (uri.startsWith(protocol + "://")) {
+         uriString = uri;
+      } else {
+         uriString = protocol + "://" + host + uri;
+      }
+
       URI absoluteURI = URI.create(uriString);
-      URI noQuery = UriBuilder.fromUri(uriString).replaceQuery(null).build();
       return new ResteasyUriInfo(uriString, absoluteURI.getRawQuery(), contextPath);
    }
 
--- a/jaxrs/server-adapters/resteasy-netty4/src/test/java/org/jboss/resteasy/test/NettyTest.java
+++ b/jaxrs/server-adapters/resteasy-netty4/src/test/java/org/jboss/resteasy/test/NettyTest.java
@@ -1,7 +1,10 @@
 package org.jboss.resteasy.test;
 
-import static org.jboss.resteasy.test.TestPortProvider.generateURL;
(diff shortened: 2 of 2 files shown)


jaxrs/server-adapters/resteasy-netty4/src/test/java/org/jboss/resteasy/test/NettyTest.java
```diff
@@ -190,4 +197,36 @@ public void testPost() {
...
+    form in requests, even though HTTP/1.1 clients will only generate
+    them in requests to proxies.
+    </blockquote>
+    * @throws Exception
+    */
+   @Test
+   public void testAbsoluteURI() throws Exception {
+      String uri = generateURL("/test");
+
+      Socket client = new Socket(getHost(), getPort());
```
avram: This tests uses sockets directly since the Resteasy client builders don't support making requests with absolute URIs in the request line.

jaxrs/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/NettyUtil.java
```diff
@@ -31,9 +29,16 @@ public static ResteasyUriInfo extractUriInfo(HttpRequest request, String context
...
+
+      // If we appear to have an absolute URL, don't try to recreate it from the host and request line.
+      if (uri.startsWith(protocol + "://")) {
+         uriString = uri;
+      } else {
+         uriString = protocol + "://" + host + uri;
+      }
+
       URI absoluteURI = URI.create(uriString);
-      URI noQuery = UriBuilder.fromUri(uriString).replaceQuery(null).build();
```
avram: Since this wasn't being used, I went ahead and deleted this line. Let me know if you would prefer that this unrelated change be in a separate PR.

avram:
@patriot1burke Can you take a look at this change? It looked to me like a divergence from the HTTP spec, and reproducible easily with RxNetty.

jaxrs/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/NettyUtil.java
ronsigal: Anything that simplifies reading the changes is appreciated. Even a note. Thanks, Avram.

avram: Glad to help. And very happy to see you working through the PR backlog on this project!

ronsigal:
Hi Avram,

I have a suggestion for a test:

      @GET
      @Path("/test/absolute")
      @Produces("text/plain")
      public String absolute(@Context UriInfo info)
      {
          return "uri: " + info.getRequestUri().toString();
      }

and

    @Test
    public void testAbsoluteURI() throws Exception {
	   String uri = generateURL("/test/absolute");

	   Socket client = new Socket(getHost(), getPort());
	   PrintWriter out = new PrintWriter(client.getOutputStream(), true);
	   BufferedReader in = new BufferedReader(new InputStreamReader(client.getInputStream()));
	   out.printf(Locale.US, "GET %s HTTP/1.1\nHost: %s:%d\n\n", uri, "127.0.0.1", getPort());
	   String statusLine = in.readLine();
	   String response = in.readLine();
	   while (!response.startsWith("uri")) 
	   {
		   response = in.readLine();
	   }
	   client.close();
	   Assert.assertEquals("HTTP/1.1 200 OK", statusLine);
	   Assert.assertEquals(uri, response.subSequence(5, response.length()));
   }

This test proves that NettyUtil is using the given absolute URI in the GET line rather than constructing a URI.

ronsigal:
Other than that, the patch looks good. Avram, could you create a JIRA issue here: https://jira.jboss.org/browse/RESTEASY ?

avram:
Fixed and rebased vs master. Issue incoming.

avram:
Issue: https://issues.jboss.org/browse/RESTEASY-1193

ronsigal:
Thanks, Avram. Looks good to merge.

PR #641, opened 2015-04-11
resteasy-html: Add Viewable

JessThrysoee:
As opposed to a View, a Viewable makes it possible to prepare a Response
with any header or status code, while importing only the body content from
e.g. a JSP template.

This is very useful for resources that produce application/xhtml+xml.

This commit also replaces the test-resteasy-html module with a suite of
maven-failsafe integration tests.


ronsigal:
Hi Jess,

Why not have Viewable in addition to, rather than instead of, View? If you want to pursue this, can you create a JIRA issue at https://github.com/resteasy/Resteasy?

Thanks,
Ron

JessThrysoee:
@ronsigal The patch does indeed have both View and Viewable. Viewable inherits from View

ronsigal:
re: "Viewable inherits from View"

Ah, missed that.

Jess, could you do a couple of things for me?

1) Create a JIRA issue.
2) Add some documentation to the Resteasy user's guide.

Thanks,
Ron

asoldano:
I know this is quite an old PR, but I'd rather give it a chance before closing as out-of-date. So, besides for the JIRA and doc thing, would it be possible to have this PR rebased on top of current master? The testsuite structure changed quite a bit ...
Thanks!

asoldano:
As we got now answer, I'm closing this PR. Feel free to re-open (a new one) with the required info at any time.

PR #642, opened 2015-04-17
Resteasy 1171 add hooks for http handlers

Iker-Jimenez:
I added a hook to insert http related handlers between the Netty and the RestEasy ones.
I also refactored the start method to remove code duplication and made it easier to read by extracting some code into separate methods.

Please note that setting the channelOptions and the childChannelOptions was previously only being executed for Https requests. I believe this is a bug and I have moved it so it executes for Http and Https requests alike.


hypnopadge:
@patriot1burke would it be possible to get this merged if you think it is appropriate to do so?

PR #645, opened 2015-05-07
Update DocumentProvider

rambertmaurice:
Logger.getLogger(DocumentProvider.class) should be Logger.getLogger(DocumentProvider.class.getName()) else JBoss give : 
ERROR [stderr] (ServerService Thread Pool -- 63) log4j:WARN No appenders could be found for logger (org.jboss.resteasy.plugins.providers.DocumentProvider).


ronsigal:
Hi Maurice,

Adding a log4j.properties or log4j.xml file should eliminate that message. See, for example, http://stackoverflow.com/questions/7685510/log4j-warning-while-initializing.

Anyway, I should internationalize all the Resteasy logging, so that line will go away.

-Ron

liweinan:
not a bug.

PR #646, opened 2015-05-09
Fix scaning rest path annotated classes-exclude interfaces

pgozdzikowski:
My problem was that when I wanted to use in same application rest easy client proxy and rest controllers then I got exception that there is absent no-args public constructor. This fix solved my issue.


ronsigal:
Hi pgozdzikowski,

Can you create a JIRA issue at https://jira.jboss.org/browse/RESTEASY so we can discuss further.?

What exactly did the exception look like? I'd like to see where it's coming from.

Thanks,
Ron

pgozdzikowski:
https://issues.jboss.org/browse/RESTEASY-1189
Done.

ronsigal:
Thanks. I'll continue the discussion at the JIRA issue.

asoldano:
Closing as per comment on JIRA.

PR #648, opened 2015-05-13
resteasy-jaxrs: attempt to use ParamConverter on null input Strings

scottbessler:
The jaxrs spec is unclear exactly what to do in regards to ParamExtractor
when the incoming value to fromString is null. They say it should return
an IllegalArgumentException. Resteasy appears to have taken that to mean
that they should not even call the converter with the null input.

This ends up being a limiting factor when you want to, for example, take
null in and convert that parameter to Optional.empty().

I came across this while trying to write something akin to
dropwizard-java8 for gwizard, which uses Resteasy.

Further, I am swallowing the NPE that this change possibly introduces because previous Resteasy users may depend on fromString not being called with null inputs. However, its still possible that a ParamConverter.fromString that returns an IllegalArgumentException will begin erroring after this push, so an argument could be made that that should be caught as well, or perhaps all

### DIFF
--- a/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
+++ b/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
@@ -278,16 +278,33 @@ public class StringParameterInjector
    {
       if (strVal == null)
       {
-         if (defaultValue == null)
+         if (defaultValue != null)
          {
-            //System.out.println("NO DEFAULT VALUE");
-            if (!baseType.isPrimitive()) return null;
+            strVal = defaultValue;
+         }
+      }
+      if (paramConverter != null)
+      {
+         if (strVal == null)
+         {
+            try
+            {
+               return paramConverter.fromString(null);
+            }
+            catch (Exception e)
+            {
+               // fall through in this scenario to maintain old behavior that was not sending null
+               // strings to custom paramconverters
+            }
          }
          else
          {
-            strVal = defaultValue;
-            //System.out.println("DEFAULT VAULUE: " + strVal);
+            return paramConverter.fromString(strVal);
          }
+      }
+      if (strVal == null && !baseType.isPrimitive())
+      {
+         return null;
       }
       try
       {
@@ -296,10 +313,6 @@ public class StringParameterInjector
       catch (Exception e)
       {
          throwProcessingException("Unable to extract parameter from http request for " + getParamSignature() + " value is '" + strVal + "'" + " for " + target, e);
-      }
-      if (paramConverter != null)
-      {
-         return paramConverter.fromString(strVal);
       }
       if (converter != null)
       {
--- a/jaxrs/resteasy-jaxrs/src/test/java/org/jboss/resteasy/test/finegrain/StringParameterInjectorTest.java
+++ b/jaxrs/resteasy-jaxrs/src/test/java/org/jboss/resteasy/test/finegrain/StringParameterInjectorTest.java
@@ -7,6 +7,8 @@ import org.jboss.resteasy.spi.StringParameterUnmarshaller;
 import org.junit.Test;
 
 import javax.ws.rs.core.Context;
+import javax.ws.rs.ext.ParamConverter;
+import javax.ws.rs.ext.ParamConverterProvider;
 import java.lang.annotation.Annotation;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
@@ -14,6 +16,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Type;
 import java.util.List;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertSame;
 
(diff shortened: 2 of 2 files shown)


scottbessler:
any feedback?

ronsigal:
Hi Scott,

I see your point, up to a point.

1. According to the javadoc for ParamConverter, Resteasy should call ParamConverter.fromString() on any default value, which implies that if a default value is declared as null, Resteasy should throw an IllegalArgumentException. It's not doing that. [I'm not convinced that it should, though. See 4 below.]

2. If no default value is declared and the input parameter is null, then, as you say, the javadoc doesn't specify whether ParamConverter.fromString() should be called. It follows that Resteasy isn't breaking the spec by not calling it.

3. If no default value is declared, and the input parameter is null, and ParamConverter.fromString() IS called, then, according to the javadoc, an IllegalArgumentException should be thrown. So your code isn't correct in that case. [Again, see 4, though.]

4. I don't really understand why the javadoc for ParamConverter makes null values illegal. Maybe it would be useful to allow them.

To summarize,

1. I think that StringParameterInjector is incorrect in its treatment of default values. You could create a RESTEASY JIRA for that, if you like. 

2. I don't think that StringParameterInjector.extractValue() is incorrect in not calling ParamConverter.fromString() on a null parameter.

3. I don't think your version of StringParameterInjector.extract() is incorrect in calling ParamConverter.fromString(), but, having made the call, I think your code is incorrect in swallowing the exception.

4. I think it wouldn't be unreasonable to start a discussion about the outlawing of null in ParamConverter inputs at the JSR-339 Users Mailing List: https://java.net/projects/jax-rs-spec/lists/users/archive. Maybe it would be better to redefine today's incorrect code as correct than to change it to conform to questionable specs.

5. I don't see an argument for adopting your Pull Request.

-Ron

scottbessler:
> If no default value is declared and the input parameter is null, then, as you say, the javadoc doesn't specify whether ParamConverter.fromString() should be called. It follows that Resteasy isn't breaking the spec by not calling it.

It may not be breaking the spec, but it wouldn't be breaking the spec to call it either. And it currently does make it impossible to convert null values, e.g. to Optional.empty.

> If no default value is declared, and the input parameter is null, and ParamConverter.fromString() IS called, then, according to the javadoc, an IllegalArgumentException should be thrown. So your code isn't correct in that case. [Again, see 4, though.]

However, it is the ParamConverter's job to throw that exception.

> I don't really understand why the javadoc for ParamConverter makes null values illegal. Maybe it would be useful to allow them.

It would be. And Jersey, the reference implementation, does what I've set up here, it leaves it up to the ParamConverter to either throw IllegalArgumentException or handle the null itself.

> I think that StringParameterInjector is incorrect in its treatment of default values. You could create a RESTEASY JIRA for that, if you like.


> I don't think that StringParameterInjector.extractValue() is incorrect in not calling ParamConverter.fromString() on a null parameter.

I think it absolutely is incorrect in that its job is not to handle anything in the presence of a ParamConverter, again, how Jersey does it.

> I don't think your version of StringParameterInjector.extract() is incorrect in calling ParamConverter.fromString(), but, having made the call, I think your code is incorrect in swallowing the exception.

I swallowed that exception so that if someone had created a paramconverter that isn't null-safe only for RESTEasy (and not for jersey, since it would NPE there) would not be backwards incompatible. I would be more than happy to remove this, but felt it was more appropriate to not surprise anyone.

> I think it wouldn't be unreasonable to start a discussion about the outlawing of null in ParamConverter inputs at the JSR-339 Users Mailing List: https://java.net/projects/jax-rs-spec/lists/users/archive. Maybe it would be better to redefine today's incorrect code as correct than to change it to conform to questionable specs.

Agreed, but the reference implementation still passes nulls even tho the spec says the ParamConverter should throw. Further, a very popular framework (dropwizard), built on top of Jersey, uses this to be able to turn Null into Optional.Empty for Java 8.

> I don't see an argument for adopting your Pull Request.

My argument is to remove the hidden surprise that null values are never passed to ParamConverters and to bring RESTEasy in line with the jax-rs reference implementation. Further, this would allow an idiomatic way to allow a Java8 Optional paramconverter, setting RESTEasy up for the future with no downside.

ronsigal:
Looking at the spec and the javadoc again, I have a somewhat different take than before. The spec defines a parameter type that may be legally annotated with @PathParam, etc. to be a type which has a constructor that takes a String, a type for which a ParamConverter is available, etc. It never specifies which of these should be used, and it never specifies that a ParamConverter MUST be used. So I disagree with your assertion that "[StringParameterInjector.extractValue()] absolutely is incorrect in that its job is not to handle anything in the presence of a ParamConverter". I can't fnd any support for that assertion. However, I do think it might be better if the spec declared the order in which conversions are selected.

re: "It may not be breaking the spec, but it wouldn't be breaking the spec to call it either."

I agree, but, as a general philosophy, I wouldn't change default behavior unless it is incorrect. I.e., I would want there to be a configurable parameter that could entail new behavior.

re: "I don't think your version of StringParameterInjector.extract() is incorrect in calling ParamConverter.fromString(), but, having made the call, I think your code is incorrect in swallowing the exception."

I take that back. The spec says ParamConverter.fromString() must throw an exception, but it is silent on how the exception is handled.

So, this time around, my conclusion is that 1) StringParameterInjector is correct as is (except for the treatment of default values) , and 2) your version is also correct. But your fundamental goal of allowing ParamConverter.fromString() to do something other than throw an exception depends on illegal behavior in order to be useful. That doesn't really make sense to me, regardless of what Jersey does. I think it makes more sense to lobby for making that illegal behavior legal.

-Ron

scottbessler:
If the opinion is that the reference implementation is not reference-able and the only way to fix Resteasy is to lobby a java standards body then we might as well close this.

PR #649, opened 2015-05-20
RESTEASY-1151 Upgraded Jackson to version 2.5.3 (from 2.4.1).

awhitford:


ronsigal:
The Resteasy testsuite passes with this upgrade.

ronsigal:
Wildfly 9.0.0.CR3-SNAPSHOT uses Jackson 2.5.1, so we'll probably use that.

liweinan:
Thanks Ron :-) Please let me know if you want to merge it.

ronsigal:
I was probably unclear. Shouldn't the dependency be on 2.5.1?

liweinan:
It has been upgraded to 2.6.3

PR #650, opened 2015-06-12
JBEAP-246 Resteasy 3 cannot be build with JDK 1.8

liweinan:


ronsigal:
Weinan, I'll take your word for it. ;-)

PR #651, opened 2015-06-17
RESTEASY-1184: Cannot remove property from client configuration

kanovotn:
client.property(some_property, null) should remove property from client configuration.


ronsigal:
Resteasy testsuite passes with this change.

PR #652, opened 2015-06-17
RESTEASY-1185 - Missing warning 2nd registration of the *class* provider

kanovotn:
The user warning when the 2nd registration of the *class* provider is attempted.
This is documented as user should get warning in the javax.ws.rs.core.Configurable interface.
The 2nd registration of the *instance* provider gives the warning.


ronsigal:
This change is a one liner. Ready to merge.

PR #658, opened 2015-08-04
RESTEASY-1207

liweinan:
@ronsigal Could you please help to review whether this fix is reasonable? Thanks!


ronsigal:
I've added a note to RESTEASY-1207.

PR #667, opened 2015-08-07
bump version

liweinan:
@ronsigal We have tagged 3.0.12.Final: https://github.com/resteasy/Resteasy/releases/tag/3.0.12.Final

So let's move upstream master to 3.0.13.Final-SNAPSHOT, do you agree?


ronsigal:
Hey Wei,

I think either 3.0.13.Final or 3.0.13.Final-SNAPSHOT would be fine.

By the way, that travis failure is interesting. I often have the same problem, where building jboss-module fails.

-Ron

liweinan:
Thank you Ron!

For jboss-modules building failures, Chao has addressed in https://github.com/resteasy/Resteasy/pull/668

PR #668, opened 2015-08-07
RESTEASY-1213, fix build problem with JDK 8.

soul2zimate:
https://issues.jboss.org/browse/RESTEASY-1213
fix Resteasy master build problem with JDK 8. After this change, both JDK7 and JDK8 can BUILD SUCCESS with "mvn clean install"

Also set a bigger connection timeout to avoid ApacheHttpClient4Test intermittent failure as reported in one of travis build https://s3.amazonaws.com/archive.travis-ci.org/jobs/74869505/log.txt 
Test failed with "ConnectionPoolTimeoutException: Timeout waiting for connection from pool", I think the connection timeout it too short to let garbage collector to finish itself. (I never see this test failure in my local test run)

I disabled two tests to avoid intermittent failures on travis-ci build as well, now that latest build output is "All checks have passed". I propose to use https://issues.jboss.org/browse/RESTEASY-1215 in order to track these random failures.

### DIFF
--- a/jaxrs/jboss-modules/lib.xml
+++ b/jaxrs/jboss-modules/lib.xml
@@ -99,7 +99,7 @@
         </sequential>
     </macrodef>
 
-    <scriptdef name="define-module-dir" language="javascript">
+    <scriptdef name="define-module-dir" language="javascript" manager="bsf">
         <attribute name="name"/>
         <attribute name="slot"/>
         <![CDATA[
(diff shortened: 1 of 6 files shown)


liweinan:
@soul2zimate Thanks! your patch worked and the travis build goes further. Let's work on the left errors for next step.

liweinan:
@ronsigal I propose to merge this request. Do you agree?

soul2zimate:
I still see different random failures in the CI build, which can not be reproduced in local machine. Is that doable to disable these intermittent ones?

ronsigal:
@liweinan, this PR looks good to merge.

liweinan:
Thanks Ron!

PR #671, opened 2015-08-14
resteasy 1216 part2

liweinan:


liweinan:
@ronsigal Chao and I found additional randomly failed tests, so here is the PR. Could you please help to check it? Thanks!

PR #672, opened 2015-08-20
Remove jaxrs-api module and use jboss-jaxrs-api_2.0_spec instead

ctomc:
Use official jboss jaxrs-api artifact instead of one provided by this project itself.
WildFly already uses jaxrs-api.


liweinan:
I don't think I can decide for this change by myself. @ronsigal  Could you please help to check this PR?

ronsigal:
The travis test failures are

   Caused by: java.lang.NoClassDefFoundError: javax/annotation/Priority

javax.annotation.Priority exists in Resteasy's jaxrs-api module. Tomaz, could there be a versioning problem? Where is the source code for jboss-jaxrs-api_2.0_spec?

-Ron

ctomc:
Source for jboss-jaxrs-api can be found at https://github.com/jboss/jboss-jaxrs-api_spec
the javax.annotation.Priority was wrongly in jaxrs-api module, it is actually part of common annotations for java ee and as such part of JSR 250, https://www.jcp.org/en/jsr/detail?id=250
and we mentain that in repo: https://github.com/jboss/jboss-annotations-api_spec

ctomc:
Updated PR so that tests pass.

ronsigal:
Hmm, @Priority didn't seem to be in jsr250-api-2.1.1.jar.

Anyway, this is great. Thanks, Tomaz.

ronsigal:
I think we have a problem. I've been running some i18n tests, and I've realized that tne other thing this change needs is updating jboss-modules with dependencies on javax.annotation.api. I'm still testing, but

 1. I think the wildfly zip file will be ok

 2. AS7 (I'm looking at jboss-as-7.1.1.Final in particular) includes jboss-annotations-api_1.1_spec-1.0.0.Final.jar, which doesn't have the Priority class.

I don't know what to do about 2. Tomaz, any thoughts?

ronsigal:
Maybe we should include jboss-annotations-api_1.2_spec:1.0.0.Final in the AS7 zip file.

ctomc:
Yes. That would be the best solution.

ronsigal:
Ok, thanks, Tomaz. I'll take care of it.

PR #673, opened 2015-08-21
RESTEASY-1223 Added better POJO handling and upgraded library.

johnament:


ronsigal:
Given the comment "After digging around a bit, this is a classloader issue mostly. It passes fine in local unit tests, but when you have a modular classloader (like in the app server), this fails." on https://issues.jboss.org/browse/RESTEASY-1223, we really should have an Arquillian test for this one.

liweinan:
Thanks John and Ron! I have created and merged https://github.com/resteasy/Resteasy/pull/718  which contains the fix + arqullian tests and solved the conflicts.

PR #676, opened 2015-08-28
Multiple improvements of client

hmorch:
Hi RESTeasy Team,

I have some small improvements of the client:

3281089: Support body in case of server side exception (http status >= 400).

e7b0959: client sends only the first value of @Produces in accept header. The client is more flexible if it sends all values in accept header.

935fa81: MatrixParams didn't work if sub resources are used. Now the client supports this and so a parent resource can be reused with different sub resources without setting the MatrixParams all the time in the last resource.

Kind regards,
Holger


liweinan:
Thanks @hmorch, Ron and I will check this.

ronsigal:
Hi Holger,

You've done some interesting things, and I've made some comments.

Also, we need

1. a JIRA issue
2. unit tests for the changes

Thanks for your work,
Ron

asoldano:
I've merged #930 which covers second and third topics here. On the first topic, Ron said he does not see why the changes are needed. Considering this is quite an old PR, with no reply on last question from Ron back in Feb, and almost duplicated by another PR, I'm closing it.

hmorch:
Hi all,

there where no responses for quite a while an so I forgot about this and didn't recognize the comments in February. As you already saw we updated the changes for the current version and my colleague MarvGilb submitted the new PR. Thanks for integration that so fast this time.

Kind regards,
Holger

asoldano:
:-)

PR #683, opened 2015-09-21
ProxyBuilder/ProxyClient: do not proxy default methods

jwgmeligmeyling:
I have been playing around with this for a while and wanted to see if other people might be interested in this and want to see this merged. This allows to use default methods in shared resource interfaces that can be used for a [HttpClient proxy](https://docs.jboss.org/resteasy/docs/3.0-beta-3/userguide/html/RESTEasy_Client_Framework.html#d4e2049). This allows you to create method aliases with default values set. Approach based on https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/

Example usage:

``` java

@Consumes(MediaType.WILDCARD)
interface ExampleResource {

    default Response test() {
        return test("5");
    }

    @GET
    Response test(@QueryParam("lupa") @DefaultValue("5") String value);
}

```

And:

``` java
            Client client = ClientFactory.newClient();
            WebTarget target = client.target("http://example.com/base/uri");
            ResteasyWebTarget rtarget = (ResteasyWebTarget)target;

            ExampleResource simple = rtarget.proxy(ExampleResource.class);
            simple.test(); // instead of simple.test(5).
```

_Edit_ I just see that it does not completely match the code style. If there is interest in merging this I will correct the errors.

### DIFF
--- a/jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
+++ b/jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
@@ -1,8 +1,12 @@
 package org.jboss.resteasy.client.jaxrs.internal.proxy;
 
 
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodHandles.Lookup;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.Map;
 
 import org.jboss.resteasy.client.jaxrs.i18n.Messages;
@@ -32,6 +36,11 @@ public class ClientProxy implements InvocationHandler
 		this.clazz = clazz;
 	}
 
+	private static boolean isDefault(Method method) {
+		return ((method.getModifiers() & (Modifier.ABSTRACT | Modifier.PUBLIC | Modifier.STATIC)) ==
+			Modifier.PUBLIC) && method.getDeclaringClass().isInterface();
+	}
+
 	public Object invoke(Object o, Method method, Object[] args)
            throws Throwable
    {
@@ -43,7 +52,21 @@ public class ClientProxy implements InvocationHandler
       MethodInvoker clientInvoker = methodMap.get(method);
       if (clientInvoker == null)
       {
-         if (method.getName().equals("equals"))
+	      if (isDefault(method)) {
+		      // Call default methods as-is.
+		      // Approach from https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/
+		      final Constructor<Lookup> constructor = MethodHandles.Lookup.class.getDeclaredConstructor(Class.class, int.class);
+		      if (!constructor.isAccessible()) {
+			      constructor.setAccessible(true);
+		      }
+
+		      final Class<?> declaringClass = method.getDeclaringClass();
+		      return constructor.newInstance(declaringClass, MethodHandles.Lookup.PRIVATE)
+			      .unreflectSpecial(method, declaringClass)
+			      .bindTo(o)
+			      .invokeWithArguments(args);
+	      }
+         else if (method.getName().equals("equals"))
          {
             return this.equals(o);
          }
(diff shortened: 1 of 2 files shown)


tpgmeligmeyling:
:+1:

ronsigal:
Hi Jan-Willem,

I see what you're doing, but I don't see the point. The call to simple.test is a stub test, so to speak, that doesn't reach the implementation of ExampleResource. As a general philosophy, isn't it better to build tests that approximate, as much as possible, real situations? Am I missing something? Could you make a case for including this facility?

Thanks,
Ron

JWGmeligMeyling:
What I'm trying to do here is not specifically for test. Say you have a resource method that consumes a lot of optional `Matrix` parameters:

```java
public interface CarService {

	@GET
	@Path("cars")
	public List<Car> getCars(
			@MatrixParam("color") String color,
			@MatrixParam("brand") String brand,
                        @MatrixParam("keywords") String... keywords);

}
```

Say these parameters are all optional, and are only provided by the API consumer when needed. Then for general cases (using a Resteasy client proxy) you find yourself writing:

```java
CarService carservice = ...
carservice.getCars(null, "Ford", "Focus");
carservice.getCars("Blue", null);
```

Because the Matrix parameters are optional, it feels a bit cluttered to have to specify all of them with null values. Even worse, would you add an additional `Matrix` parameter for other filter properties, then you would have to rewrite all your client usage.

What I'm trying to achieve is to create `default` aliases for my methods, that translate it to the right actual HTTP call. For the above API this would look like the following:

```java
public interface CarService {

	public default List<Car> getCars(@MatrixParam("color") String color,
                       @MatrixParam("keywords") String... keywords) {
                 return getCars(color, null, keywords);
       }

```

Which would allow you to call `carservice.getCars("Blue")` from client code.

The problem is that this currently does not work, as the proxy is trying to transform the method into a request, and misses the JAX-RS annotations. Therefore, in order to make above code work, one has to write:

```java
public interface CarService {

        @GET
        @Path("default-alias/cars")
	public default List<Car> getCars(@MatrixParam("color") String color,
                       @MatrixParam("keywords") String... keywords) {
                 return getCars(color, null, keywords);
       }

```

So that the message ends up at a different Resource Method in the server, which then performs the call to `getCars(string, string, string... )` internally. This has two downsides: (1) we have to create a strange alias for the resource, (2) this alias is responding to requests server side, allowing API users to use unwanted URLs.

Would the proxy itself call the real interface method, and then call the proxy method on the `getCars(string, string, string... )` method itself, then we don't have these problems.

I would by the way be willing to create some test code / example, but is there a test package that runs under J1.8?

ronsigal:
Hi Jan-Willem,

1. I think I remember the JAX-RS expert group talking about a default resource method (https://java.net/projects/jax-rs-spec/lists). Can't find the discussion, though. 

2. Maybe you could achieve your goal using an extra method

        @GET
        @Path("cars")
        List<Car> getCarsByColor(@MatrixParam("color") String color);

on the client side interface and @DefaultValue annotations on the server side. Haven't tried it, though.

-Ron

jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
```diff
@@ -43,7 +52,21 @@ public Object invoke(Object o, Method method, Object[] args)
       MethodInvoker clientInvoker = methodMap.get(method);
       if (clientInvoker == null)
       {
-         if (method.getName().equals("equals"))
+	      if (isDefault(method)) {
```
stevenschlansker: Tabbing here isn't consistent with the existing file.

jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
```diff
@@ -43,7 +52,21 @@ public Object invoke(Object o, Method method, Object[] args)
...
+		      // Call default methods as-is.
+		      // Approach from https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/
```
stevenschlansker: Note that this is likely to break in Java 9 (or so we've been told)

stevenschlansker: The new `privateLookupIn` method fixes this, although it is not available with 8 so you have to do it reflectively: https://github.com/jdbi/jdbi/pull/1066

JWGmeligMeyling:
@asoldano @stevenschlansker This PR went unnoticed for me, because after the original feedback I abandoned the idea and worked around the issue. If there is interest in merging it, I am willing to rebase it 😄

stevenschlansker:
Similarly, we work around the lack of this feature by wrapping our proxy interfaces in more feature-rich facades.

jamezp:

PR #684, opened 2015-09-25
Support for not defined Matrix Parameter

MarvGilb:
When sending matrixParams that are not defined (like ;a;b=1;c=2, where a is not defined) a 404 Not found will be returned. Instead it should just ignore the value and fullfill the reuqest, therefor a check for an empty String is necessary.


JWGmeligMeyling:
I think this is not the right place to fix this. And it breaks several tests.

MarvGilb:
You were right. I moved the change to the point where the values get initialized. 
Now it only sets the value, when something is passed, otherwise null will be stored.

JWGmeligMeyling:
This looks good! :+1: Now its up to the resteasy devs :smile:

PR #685, opened 2015-09-28
Add META-INF service for RestEASY ClientBuilder

ahaeber:
Needed to make service lookup for RestEASY work correctly. Otherwise it
will fallback to org.glassfish.jersey.client.JerseyClientBuilder.

Signed-off-by: Andreas Häber <andreas.haber@intele.com>


ahaeber:
Sorry, missed the resteasy-client dependency.

PR #689, opened 2015-10-12
Migrating RESTEASY-610 patch to new ProxyBuilder, ignore Java8 default methods

shanedav:
A patch was submitted to ProxyBuilder for RestEasy 2.3.4.  The original ProxyBuilder has now been deprecated and the new one did not include similar patch code.  That was copied that over.

https://issues.jboss.org/browse/RESTEASY-610

Also, we found that Java 8 generates default methods which don't include the original annotations causing RestEasy to fail.  This patch ignores those.

http://stackoverflow.com/questions/31962433/classgetdeclaredmethods-returns-inherited-methods
http://stackoverflow.com/questions/27013542/getdeclaredmethods-behaving-differently-in-java-7-vs-java-8
http://stackoverflow.com/questions/1961350/problem-in-the-getdeclaredmethods-java
http://stackoverflow.com/questions/6204339/java-class-getmethods-behavior-on-overridden-methods

### DIFF
--- a/jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ProxyBuilder.java
+++ b/jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ProxyBuilder.java
@@ -18,20 +18,20 @@ import java.util.Set;
 
 public class ProxyBuilder<T>
 {
-   private final Class<T> iface;
-   private final ResteasyWebTarget webTarget;
-   private ClassLoader loader = Thread.currentThread().getContextClassLoader();
-   private MediaType serverConsumes;
-   private MediaType serverProduces;
+	private final Class<T> iface;
+	private final ResteasyWebTarget webTarget;
+	private ClassLoader loader = Thread.currentThread().getContextClassLoader();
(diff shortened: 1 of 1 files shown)


JWGmeligMeyling:
Method::isDefault is not available on Java 7, therefore I came up with https://github.com/resteasy/Resteasy/pull/683

shanedav:
Sorry about that. This is my first time doing a PR (and haven't done much with git or GitHub yet). Do I need to create a new PR for the first part or does the patch get modified (by you/me) and continue? Thanks!

JWGmeligMeyling:
Not sure, I am not one of the collaborators of this project and have not had any response on my pull request either.

But anyways your changes need to be compatible with the Java versions that Resteasy currently supports, so you might want to copy these changes for isDefault back in. It's a common practise to then squash the two commits of your pull requests together into a single commit.

shanedav:
I've modified the code for this PR to include the first part (migrate RESTEASY-610 patch) but removed the Java8 part for now.

shanedav:
More investigation needs to be done here to get everything that was working before to the point of all working again.  So closing this for now.

PR #693, opened 2015-11-03
Guice fix: Register generic @Providers

fzakaria:
For guice, if you want to bind() a generic type you have to use the
TypeLiteral class. The type however is not of Class and the registration of class's annotated with
@Provider were being ommited.

see: https://github.com/google/guice/wiki/FrequentlyAskedQuestions#how-to-inject-class-with-generic-type


fzakaria:
JIRA created: https://issues.jboss.org/browse/RESTEASY-1257

fzakaria:
awesome!

fzakaria:
@liweinan should I close the JIRA myself?

PR #696, opened 2015-11-11
RESTEASY-1261 Use queryStringEncoding for the fragment

Edubits:
Use queryStringEncoding instead of queryNameValueEncoding for encoding
the fragment part of a URI


liweinan:
@Edubits The fix looks great! :-)  Could you please also add a test case for this?

Edubits:
@liweinan Sure, I've just added a test case!

liweinan:
@kanovotn I decide to merge this because it conforms with RFC standard. Could you please help to double check whether this conforms to TCK tests also?

Edubits:
Thanks @liweinan!

PR #702, opened 2015-11-18
[RESTEASY-1191] : Backport JBEAP-246 to 2_3

panossot:
[RESTEASY-1191] : https://issues.jboss.org/browse/RESTEASY-1191 (adding the PR for 2_3)
[Upstream] : https://github.com/resteasy/Resteasy/pull/650

### DIFF
--- a/jboss-modules/lib.xml
+++ b/jboss-modules/lib.xml
@@ -70,7 +70,7 @@
         <attribute name="slot"/>
         <![CDATA[
             name = attributes.get("name");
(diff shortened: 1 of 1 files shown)


liweinan:
@panossot Thanks for the addition!

However Branch_2_3 is totally dead, stopped, frozen and not supported. So in the future please don't bother to commit anything to Branch_2_3 :-)

We only use our internal maintenance repo for 2.3.x.

PR #704, opened 2015-11-18
RESTEASY-1202 Bump Jackson to 2.6.3

choweiyuan:


liweinan:
This is not enough because we should also fix the deprecated APIs. But let me merge it firstly.

PR #710, opened 2015-12-08
Test case and fix for RESTEASY-222

mihxil:
It seems that RESTEASY-222 should be reopened again, because when I use 'lastModified' the actual Last-Modified header is not correctly formatted. This is a fix, plus a test-case.

### DIFF
--- a/jaxrs/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/client/ResponseTest.java
+++ b/jaxrs/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/client/ResponseTest.java
@@ -22,7 +22,7 @@ public class ResponseTest
    @Test
    public void getLastModifiedTest()
    {
-      Date date = Calendar.getInstance().getTime();
+      Date date = new Date(1455097347000L);
       Response response = Response.ok().lastModified(date).build();
       Date responseDate = response.getLastModified();
       System.out.println(date);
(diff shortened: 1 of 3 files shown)


ronsigal:
Hi Michiel,

Have you looked at the failing test (https://s3.amazonaws.com/archive.travis-ci.org/jobs/95601252/log.txt)? org.jboss.resteasy.test.client.ResponseTest.getLastModifiedTest()

    @Test
    public void getLastModifiedTest()
    {
       Date date = Calendar.getInstance().getTime();
       Response response = Response.ok().lastModified(date).build();
       Date responseDate = response.getLastModified();
       System.out.println(date);
       System.out.println(responseDate);
       Assert.assertTrue(date.equals(responseDate));
    }

fails because

    metadata.putSingle(HttpHeaderNames.LAST_MODIFIED, getDateFormatRFC822().format(lastModified));

truncates milliseconds and

    metadata.putSingle(HttpHeaderNames.LAST_MODIFIED, lastModified);

doesn't.

-Ron

mihxil:
I'll look into it.

mihxil:
Well, I think since dates are always truncated in http headers, the test case is actually wrong.  Also I doubt if whether it is a good idea to depend the test on the current time. It won't test the same every time like that.  So I propose to simply use the same, rounded, date every time in this test.

ronsigal:
I've created RESTEASY-1299, a clone of RESTEASY-222.

This PR is ready to merge. Thanks, Michiel.

liweinan:
Confirmed with Ron this change failed TCK. I'll revert it.

PR #712, opened 2015-12-09
[JBEAP-2138] The stream never gets reset after Jaxb exception occurs …

psakar:
…when parsing server response


liweinan:
Thanks @psakar  :-) @ronsigal @kanovotn  I think this fix is correct. Could you please help to review it?

liweinan:
@psakar btw, could you please help to add an test case in 'resteasy-jaxrs-testsuite'?

PR #713, opened 2015-12-10
Fixes for resteasy-links #833, #1271

locosmac:
#833
- In RESTUtils: Load existing value of injection field from entity to
  avoid overwriting manually specified links. 
- In RESTServiceDiscovery: Prevent insertion of duplicate links into
  injection field by performing duplicate check in addLinks. 
- In Messages: Add additional error message when injection field cannot
  be read which will result in overwritten manually specified links.
#1271
- In RESTUtils: Replace calls to Method.findAnnotation and
  Method.isAnnotationPresent with implementations that also look at
  inherited annotations.


ronsigal:
See comment on RESTEASY-1271.

PR #715, opened 2015-12-13
Fix RESTEASY-1201

ian4hu:
resteasy-multipart-provider/AbstractMultipartWriter: Wrap entityStream into a no closable stream when writeParts()

The cause of the issue https://issues.jboss.org/browse/RESTEASY-1201 is like that, when write the content of parts, the JSON parts will use the org.jboss.resteasy.plugins.providers.jackson.ResteasyJackson2Provider to write, and in https://github.com/resteasy/Resteasy/blob/master/jaxrs/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java#L209, the jq.close() closed the stream, that cause the entityStream closed unexpectly.

I'm not sure that jq.close() is really need to close the entityStream, but I am really sure that the implements of MessageBodyWriter shoud never close the entityStream as it mentioned in http://docs.jboss.org/resteasy/docs/3.0-rc-1/javadocs/javax/ws/rs/ext/MessageBodyWriter.html#writeTo%28T,%20java.lang.Class,%20java.lang.reflect.Type,%20java.lang.annotation.Annotation[],%20javax.ws.rs.core.MediaType,%20javax.ws.rs.core.MultivaluedMap,%20java.io.OutputStream%29 about the entityStream param

### DIFF
--- a/jaxrs/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
+++ b/jaxrs/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
@@ -8,6 +8,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.MessageBodyWriter;
 import javax.ws.rs.ext.Providers;
+import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.lang.reflect.Type;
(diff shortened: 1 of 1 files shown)


ronsigal:
Hi Ian,

1. You are quite right that ResteasyJackson2Provider is incorrectly closing the OutputStream. I think that a more direct solution, then, is just to remove that call to jg.close().

2. Note that 

    public void write(byte[] b, int off, int len) throws IOException

of FilterOutputStream writes each byte individually. The javadoc says, "Subclasses of FilterOutputStream should provide a more efficient implementation of this method."

3. Could you commit a unit test? It doesn't have to be an arquillian test.

Thank you,
Ron

ian4hu:
sorry for the latecy of feedback.

I think remove the jq.close() is not a good idea. if you use the ResteasyJackson2Provider as the only provider, the jq should close. if you use ResteasyJackson2Provider  as a sub-field provider then the jq should not close, because the whole entity may not end of parse or output

ronsigal:
Hey Ian,

re: "I think remove the jq.close() is not a good idea. "

Yes, I don't remember what I was thinking, but now I think you're right.

ronsigal:
Simple unit test?

ian4hu:
@ronsigal my unit test need the jackson2provider could i add the dependency in test scope to the pom?

ronsigal:
Hi Ian,

You're putting the test in the multipart module, right? Sure, adding jackson2provider in test scope sounds fine. Thank you.

-Ron

ian4hu:
Hi ronsigal,

I have written the unit tests code and tried to re-produce this issue but failed. It seems that there are some difference between running jetty in unit tests and running jetty in standalone, but I can't figure out where is the difference. the result is that I can not make this issue in my unit tests code. I need some help to work it out.

-Ian

ian4hu:
there are some conflicts, temporally close this, I'll create a new one.

ian4hu:
new PR is at #905

PR #716, opened 2015-12-13
Make multipart/form-data more compatible

ian4hu:
resteasy-multipart-provider/AbstractMultipartWriter: Make multipart/form-data more compatible

As some web server not compatibe with this writer but working well with curl, I diffed the package that write by AbstractMultipartWriter and curl, I found AbstractMultipartWriter did not write a trialling CRLF but the curl did. and when there is no trailling CRLF some server may miss the whole body.


ronsigal:
Hi Ian,

Chapter 4 "HTTP Message" of RFC2616 (https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4) says:

To restate what is explicitly forbidden by the BNF, an HTTP/1.1 client MUST NOT preface or follow a request with an extra CRLF. 

Your change seems to violate that requirement.

-Ron

PR #724, opened 2016-01-08
Fix Httpclient-Connectionleak for InputStream or Reader and Content-Length 0

MarkusKull:
Fixing a HttpClient-ConnectionLeak we encountered when downloading 0Byte-Files with the Restclient.

It seems this bug was originally introduced in a large commit f846687fced7 "tck fixes and oauth2 fixes", probably when changing similar non-streaming MessageBodyReaderWriter but without regard to the special streaming MessageBodyReaderWriter for InputStream or Readers.


liweinan:
Great! Thanks!

PR #733, opened 2016-02-05
RESTEASY-1294: Fixed comparison bug in JaxrsInterceptorRegistry.

ronsigal:


jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/interception/JaxrsInterceptorRegistry.java
```diff
@@ -360,7 +360,15 @@ public JaxrsInterceptorRegistry(ResteasyProviderFactory providerFactory, Class<T
    {
       public int compare(Match match, Match match2)
       {
-         return match.order - match2.order;
```
JWGmeligMeyling: This implementation is equal to `Integer.compareTo(match.order, match2.order);`

liweinan:
@ronsigal @JWGmeligMeyling Can I merge this one?

JWGmeligMeyling:
The new implementation is defenitely allright, but is already floating somewhere in the JRE. But changing that would be quite some hassle for a few lines less, so feel free to merge!

ronsigal:
re: "Can I merge this one?"

Yeah. Not as pretty as Integer.compare(), but it's fine.

PR #736, opened 2016-02-10
RESTEASY-1212

mihxil:


ronsigal:
Hi Michiel,

Looking at the history of SpringBeanProcessor, I found commit https://github.com/resteasy/Resteasy/commit/2fb71377db365c2db6b5da2f32141dc34633236f in Resteasy 3.0.14.Final. Does that fix the problem?

Thanks,
Ron

mihxil:
It seems that it does.  So, I rolled back my change in the SpringBeanProcessor. I still propose to merge the pull request then, because it includes a test case, to ensure that it won't happen again.

Also the associated JIRA issue can be closed for 3.0.14.

PR #738, opened 2016-02-12
RESTEASY-1296

ronsigal:
Changed treatment of Weld proxies in GeneralValidatorImpl in resteasy-validator-provider-11.


ronsigal:
Some kind of intermittent failure in RESTEASY-923-WF8.

PR #745, opened 2016-02-26
RESTEASY-1300

ronsigal:
Added warning message when multiple resource methods match request.


ronsigal:
For one of the builds, maven says it passed, and then

>The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising >the same exception over and over).

>The build has been terminated.

I'll do another push and try again ...

PR #747, opened 2016-02-28
for Unauthorized send 401 and for wrong authentication send 403

jkandasa:
Fix for TJWS embedded server.

- Currently for wrong authentication  returns 401 and for no `Basic` auth details returns 405. Which is not correct.

- This fix will return 403 for wrong authentication and 401 for no `Basic` auth details on request.

### DIFF
--- a/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/tjws/TJWSRequestPreProcessor.java
+++ b/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/tjws/TJWSRequestPreProcessor.java
@@ -60,11 +60,16 @@ public class TJWSRequestPreProcessor
                }
                catch (SecurityException e)
                {
-                  response.sendError(HttpResponseCodes.SC_UNAUTHORIZED);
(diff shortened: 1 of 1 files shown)


ronsigal:
Hi Jeeva,

- I don't see why it would be incorrect to send 401. RFC2616 (https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) says:

>If the request already included Authorization credentials, then the 401 response indicates that authorization has been refused for those credentials. 

whereas for 403 it says

>Authorization will not help and the request SHOULD NOT be repeated.

So I think 401 makes sense here.

- I think you're right that, in the absence of an Authorization header, or in the presence of an Authorization header that is not a request for BASIC authorization, 401 should be returned.

- Have you looked at the test failures in the resteasy-jaxrs module:

Failed tests: 

  BasicAuthTest.testProxyFailure:200 expected:<401> but was:<403>
  BasicAuthTest.testSecurity:246 expected:<403> but was:<401>
  BasicAuthTest.testSecurityFailure:286 expected:<403> but was:<401>

These need to be dealt with.

Thanks,
Ron

asoldano:
Given Ron's comment above (and no response to him) and considering we're working on actually removing the TJWS embedded server from RESTEasy, I'm closing this PR:

PR #752, opened 2016-03-12
RESTEASY-1321: Support non-path-parameters in LinkResource

alexrashed:
Implementation of feature request [RESTEASY-1321](https://issues.jboss.org/browse/RESTEASY-1321).

Currently the LinkResource annotation only supports path parameters for the atom link definition. This is enough when only adding default atom links (self, list, CRUD). But in order to add more advanced customized links (f.e. for paging) it is necessary to support the other types of parameters (query and matrix).

Therefore I added a new annotation called ParamBinding. Typically I would add the type as an enum in the ParamBinding, but just added two more members to the annotation (queryParameters, matrixParameters) in order not to break the API (this would remove / rename LinkResource#pathParameters).

### DIFF
--- a/jaxrs/resteasy-links/src/main/java/org/jboss/resteasy/links/LinkResource.java
+++ b/jaxrs/resteasy-links/src/main/java/org/jboss/resteasy/links/LinkResource.java
@@ -69,6 +69,20 @@ public @interface LinkResource {
 	
 	/**
 	 * <p>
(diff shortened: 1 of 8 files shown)


ronsigal:
Added comment to https://jira.jboss.org/browse/RESTEASY-1321

ronsigal:
This PR is ready to merge. Thanks, Alexander.

PR #757, opened 2016-03-17
Resteasy 1325

hrqiang:
RETEASY-1325: Netty server should close idle connections.
Even NettyJaxrsServer has SO_KEEPALIVE socket option, I would like to add idle timeout option to cleanup idle TCP clients.
By defaul, this feature doesn't turn on, so it's identical to previous versions.
If call to setIdleTimeout with number of seconds, Http client connection will be closed after specific idle time.

Description of this JIRA from https://issues.jboss.org/browse/RESTEASY-1325

"
Netty server adapter should handle client idle when there is no traffic within a certain amount of time.
Currently, if client open connection without sending any message. The connection will keep open.
If there is no Connection: close header, the connection will keep open.
Unless there is "Connection: close" header, they always require client to close.
For Jetty, the default idleTimeOut for Connector is 30 seconds.
Even I can add an IdleHandler to NettyJaxrsServer.addChannelHandlers, it's better to build in NettyJaxrsServer and add setter and attribute for idleTimeOut.
"


ronsigal:
Added comment to https://issues.jboss.org/browse/RESTEASY-1325.

ronsigal:
Thank you for the unit test, Ruiqiang. This PR is ready to merge.

liweinan:
This is cool! Thanks :-)

PR #769, opened 2016-04-02
RESTEASY-1334 JUnit Assert.assertEquals(a,b) for floating point is deprecated

awhitford:
The [Assert.assertEquals(e,a)](http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(double, double)) for floating point variables is deprecated.  One must use the newer [Assert.assertEquals(e,a,d)](http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(double, double, double)).

In addition, the version of JUnit used throughout the project was inconsistent.  Some `pom.xml` was tweaked to make it more consistent and centralize the version at the parent pom.

### DIFF
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -17,7 +17,6 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.12</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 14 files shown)


ronsigal:
Thanks, Anthony. This PR is ready to merge.

liweinan:
Nice PR! Thanks!

PR #777, opened 2016-04-08
RESTEASY-1339 Arquillian Consistency and Upgrade

awhitford:
Consistent versions for `arquillian-bom` and `wildfly-arquillian-container-managed`, then upgraded to:
* `arquillian-bom-1.1.11.Final`
* `wildfly-arquillian-container-managed-8.2.1.Final`


liweinan:
Thanks!

PR #781, opened 2016-04-20
RESTEASY-1124

ronsigal:
Don't remove trailing '/' from path in ResteasyUriInfo.

### DIFF
--- a/jaxrs/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/nextgen/finegrain/TrailingSlashTest.java
+++ b/jaxrs/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/nextgen/finegrain/TrailingSlashTest.java
@@ -1,0 +1,147 @@
+package org.jboss.resteasy.test.nextgen.finegrain;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import javax.ws.rs.GET;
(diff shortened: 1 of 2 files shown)


jaxrs/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/nextgen/finegrain/TrailingSlashTest.java
```diff
@@ -0,0 +1,147 @@
...
+      ResteasyUriInfo ruri = new ResteasyUriInfo(s, "", "");
+      URI uri = new URI(s);
+      Assert.assertEquals(path, ruri.getPath());
+      Assert.assertEquals(path, ruri.getPath(true));
+      Assert.assertEquals(path, ruri.getPath(false));
+      Assert.assertEquals(uri, ruri.getAbsolutePath());
+      Assert.assertEquals(uri, ruri.getBaseUri().resolve(ruri.getPath(false))); 
+   }
+
+   //@Test
```
tbroyer: Why are those tests disabled?

ronsigal: Actually, I started to work on it, but they failed, and I didn't think it was worth spending any more time on.

PR #783, opened 2016-04-21
RESTEASY-1281

ronsigal:
Missing ResourceInfo injected by @Context returns null on
getResourceClass() and getResourceMethod().


ronsigal:
Superceded by PR 785.

PR #784, opened 2016-04-21
RESTEASY-833: Preserve manually added AtomLinks

alexrashed:
I wanted RESTEASY-833 to be fixed, so I just implemented the suggestion and added a unit test.
https://issues.jboss.org/browse/RESTEASY-833

### DIFF
--- a/jaxrs/resteasy-links/src/main/java/org/jboss/resteasy/links/RESTServiceDiscovery.java
+++ b/jaxrs/resteasy-links/src/main/java/org/jboss/resteasy/links/RESTServiceDiscovery.java
@@ -99,10 +99,67 @@ public class RESTServiceDiscovery extends ArrayList<RESTServiceDiscovery.AtomLin
 		public void setLength(String length) {
 			this.length = length;
 		}
+		@Override
+		public int hashCode() {
+			final int prime = 31;
+			int result = 1;
+			result = prime * result + ((href == null) ? 0 : href.hashCode());
+			result = prime * result + ((hreflang == null) ? 0 : hreflang.hashCode());
+			result = prime * result + ((length == null) ? 0 : length.hashCode());
+			result = prime * result + ((rel == null) ? 0 : rel.hashCode());
+			result = prime * result + ((title == null) ? 0 : title.hashCode());
+			result = prime * result + ((type == null) ? 0 : type.hashCode());
+			return result;
+		}
+
+		@Override
+		public boolean equals(Object obj) {
+			if (this == obj)
+				return true;
+			if (obj == null)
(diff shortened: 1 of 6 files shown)


ronsigal:
Hi Alexander,

Thank you for working on this. I just have a couple of points.

1. Why does it matter if a link is injected twice? Isn't it simpler and faster to just write over an existing link?

2. I don't think the "failed to inject" message makes sense here

		try {
			injectionField.setAccessible(true);
			ret = (RESTServiceDiscovery) injectionField.get(entity);
		} catch (Exception e) {
			LogMessages.LOGGER.error(Messages.MESSAGES.failedToInjectLinks(entity), e);
		}

because you then go on to create a new RESTServiceDiscovery.

-Ron

alexrashed:
Hi Ron,

1. What exactly do you mean with "write over an existing link"? The `RestServiceDiscovery` extends `ArrayList`, which allows duplicate entries. In my opinion it should probably extend `HashSet` or `LinkedHashSet` to avoid duplicate entries, but that change would probably be too big for a minor-release.
The duplicate check is necessary because `XMLSerializer#fireBeforeMarshalEvents` is called twice during the execution of `XMLSerializer#childAsRoot` for the same (root-) element. Once in the in the beginning of `#childAsRoot` (line #491) and once in the 
`#childAsSoleContent` (deeper in the stack but called during the execution of `childAsRoot` - line 494).
This can be debugged/observed by adding a conditional breakpoint to `RESTUtils#addDiscovery`.
That duplicate call makes it necessary that `RESTUtils#addDiscovery` is idempotent.
To be honest I didn't investigate any further (maybe it is not necessary that it is executed twice?).

2. I thought the method should be as robust as possible (that's why I go on with the discovery creation and the link adding), but on the other hand there's nothing worse than a suppressed error. ;)
Do you think it would be better if the links would just not be added?

-Alex

ronsigal:
Hey Alex,

re: "What exactly do you mean with "write over an existing link"?"

Probably nothing meaningful. ;-)  You're right.

re: " there's nothing worse than a suppressed error"

True enough, but he particular message about failing to inject links doesn't really work here, because you might still succeed with a new RESTServiceDiscovery. I think it should be a new message that says that the RESTServiceDiscovery constructed by the user cannot be used.

re: "Do you think it would be better if the links would just not be added?"

No, I think you're doing the right thing. 

-Ron

alexrashed:
Hi Ron.
Thanks for the feedback, I added a new error message "Failed to access/reuse user-created service discovery in %s". 
- Alex

ronsigal:
Thanks, Alex. Looks good.

PR #787, opened 2016-04-27
RESTEASY-1345

ronsigal:
Introduced LocalResteasyProviderFactory that doesn't reference parent after creation, to separate parent and child configuration in client framework.


ronsigal:
RESTEASY1325Test in resteasy-netty4 is failing. Strange ... it passes for me. Seems like some temporary problem. I'll try again tomorrow.

jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
```diff
@@ -335,7 +336,7 @@ protected ResteasyProviderFactory getProviderFactory()
       if (providerFactory == null)
       {
          // create a new one
-         providerFactory = new ResteasyProviderFactory();
```
NicoNes: This line can actually remain the same since ClientConfiguration constructor now instantiate a new LocalResteasyProviderFactory(). It will be consistent with ResteasyClientBuilder.withConfig() method where we also use ResteasyProviderFactory.  What do you think ?

jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/LocalResteasyProviderFactory.java
```diff
@@ -0,0 +1,84 @@
+package org.jboss.resteasy.client.jaxrs.internal;
```
NicoNes: Is it possible to also override getRuntimeType() method to return  RuntimeType.CLIENT instead of RuntimeType.SERVER? 
On ClientConfiguration this method is OK and return RuntimeType.CLIENT. But when registering any provider on it, it is the underlying ResteasyProviderFactory it wrap that is used. So for any client Feature or DynamicFeature when calling FeatureContext.getConfiguration().getRuntimeType() it will always return RuntimeType.SERVER.
Maybe I should open a separate issue for this ?

jaxrs/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
ronsigal: I thought that made sense, but then I overrided getRuntimeType() and realized that the ResteasyClientBuilder's config should be a LocalResteasyProviderFactory. In fact, I changed withConfig() for that reason.

ronsigal:
This is superceded by PR #798 .

PR #790, opened 2016-05-05
[RESTEASY-1240] Fixing TestSecureProcessingFeature test too

asoldano:


ronsigal:
Thanks, Alessio. I wouldn't be surprised if there were more of these.

asoldano:
Well, actually I stopped on the first issue and fixed it last time. This time I have the whole build passing, so we should be fine.

PR #791, opened 2016-05-06
[Resteasy-1197]: Performance Issue: org.jboss.resteasy.spi.ResteasyUriInfo instantiation takes 20% of all the CPU time

NicoNes:
The goal of this this pull request is to improve ResteasyUriBuilder performance by doing to things:

- Replace all StringBuffer instances by StringBuilder. Because of thread safety, StringBuffer instances are a bit slower than StringBuilder. The massive use of StringBuffer was enforced by the use of java.util.regex.Matcher.appendReplacement(...) and java.util.regex.Matcher.appendTail(...) methods.
The idea is to keep on using java.util.regex.Matcher but stop calling those methods. To do so lets replace this pattern:

> 		String templateString = ...;
	  Matcher matcher = Pattern.compile(...).matcher(templateString);
	  StringBuffer stringBuffer = new StringBuffer();
	  while (matcher.find()) {
		 String matchedValue = matcher.group(1);
		 String replacementString = ....;
		 matcher.appendReplacement(stringBuffer, replacementString);
	 }
	 matcher.appendTail(stringBuffer);
	 return stringBuffer.toString();

By this one:

> 	   String templateString = ...;
	 Matcher matcher = Pattern.compile(...).matcher(templateString);
	 StringBuilder stringBuilder = new StringBuilder();
	 int start = 0;
	 while (matcher.find()) {
		 stringBuilder.append(templateString, start, matcher.start());
		 String matchedValue = matcher.group(1);
		 String replacementString = ...;
		 stringBuilder.append(replacementString);
		 start = matcher.end();
	 }
	 stringBuilder.append(templateString, start, templateString.length());
	 return stringBuilder.toString();

- Reduce instantiation and use of expensive java.util.regex.Matcher when possible by performing simple string check  and a small refactoring to avoid doing same expensive calls twice.

### DIFF
--- a/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
+++ b/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
@@ -622,8 +622,6 @@ public class ResteasyUriBuilder extends UriBuilder
          {
             value = Encode.encodeNonCodes(value);
          }
-         // if there is a $ then we must backslash it or it will screw up regex group substitution
-         value = value.replace("$", "\\$");
 		 builder.append(value);
 		 start = matcher.end();
       }
@@ -674,7 +672,7 @@ public class ResteasyUriBuilder extends UriBuilder
                if (encodeSlash) stringValue = Encode.encodePathSegmentSaveEncodings(stringValue);
                else stringValue = Encode.encodePathSaveEncodings(stringValue);
             }
-			builder.append(Matcher.quoteReplacement(stringValue));
+			builder.append(stringValue);
 			start = matcher.end();
          }
          else
--- a/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
+++ b/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
@@ -24,6 +24,39 @@ import java.util.regex.Pattern;
  */
 public class ResteasyUriBuilder extends UriBuilder
 {
+	
+	private static final class URITemplateParametersMap extends HashMap<String, Object> {
+
+		private final Object[] parameterValues;
+		private int index;
+		
+		private URITemplateParametersMap(Object...parameterValues) {
+			this.parameterValues=parameterValues;
+		}
+
+		@Override
+		public Object get(Object key) {
+			Object object = null;
+			if ( !super.containsKey(key) && this.index != this.parameterValues.length) {
+				object = this.parameterValues[this.index++];
+				super.put((String) key, object);
+			}else{
+				object = super.get(key);
+			}
+			return object;
+		}
+
+		@Override
+		public boolean containsKey(Object key) {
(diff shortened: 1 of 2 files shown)


jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
```diff
@@ -723,24 +776,10 @@ public URI build(Object... values) throws IllegalArgumentException, UriBuilderEx
 
    protected URI buildFromValues(boolean encodeSlash, boolean encoded, Object... values)
    {
-      List<String> params = getPathParamNamesInDeclarationOrder();
```
NicoNes: Retrieving URI template parameters to check if each of them is bound to a value forced us to execute heavy java.util.regex.Matcher stuff two times while they can be done only once in buildString(...) method.
Actually in getPathParamNamesInDeclarationOrder() method "PathHelper.URI_PARAM_PATTERN.matcher(PathHelper.replaceEnclosedCurlyBraces(string)).find()" is called	 for each scheme, userInfo, host,path, query and fragment properties. 
Same calls are made in buildString(...) method for same properties. We can do it once for all in buildString(...) and perform checks to throw exceptions if needed with respect of the specification requirements.
By the way doing that will keep us consistent with all buildFromEncodedMap() and buildFromMap() methods where all checks are already performed by buildString(...) method.

jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
```diff
@@ -519,65 +552,66 @@ protected URI buildUriFromMap(Map<String, ? extends Object> paramMap, boolean fr
 
    private String buildString(Map<String, ? extends Object> paramMap, boolean fromEncodedMap, boolean isTemplate, boolean encodeSlash)
    {
-      for (Map.Entry<String, ? extends Object> entry : paramMap.entrySet())
```
NicoNes: Maybe useless check since paramMap key and value will be tested in replaceParameter(...) and replaceQueryStringParameter(...) methods.

asoldano:
Very interesting, I actually had todo notes on replacing StringBuffer with StringBuilder here and there, cool. The changes in the ResteasyUriBuilder seems ok too, but I would double check them by running a TCK test, so please give us some time.
Speaking of performances, have you been using any specific benchmark to collect data leading to the optimizations above?

asoldano:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/128435651/log.txt <-- 6 failures and 3 errors to be analyzed

NicoNes:
Hi Alessio,

Is this the latest test logs ? It's maybe not cause I saw these errors yesterday and I fixed them with my last commit (https://s3.amazonaws.com/archive.travis-ci.org/jobs/128446469/log.txt). It was because of string escaping. Tell me if I'm wrong.
About the benchmark used, I did the analysis with jprofiler running a simple standalone server. I attached my results in https://issues.jboss.org/browse/RESTEASY-1197.

asoldano:
This is superseded by https://github.com/resteasy/Resteasy/issues/807

PR #797, opened 2016-05-09
[RESTEASY-1351]:Fix test failures in ClientBuilderTest and DuplicateP…

jimma:
…roviderRegistrationTest;Redirect test output to a file instead of console to easily find the real test failure message

### DIFF
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -583,6 +583,7 @@
                     <configuration>
                         <forkMode>once</forkMode>
                         <argLine>-Xms512m -Xmx512m</argLine>
+                        <redirectTestOutputToFile>true</redirectTestOutputToFile>
                     </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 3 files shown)


ronsigal:
Hi Jim,

Thanks for fixing this.

1. It occurs to me that this is another case of a test that could fail if the locale language isn't English. Maybe the tests for the logging output should be based on, for example, LogMessages.LOGGER.providerInstanceAlreadyRegistered().

2. I don't see where the redirectTestOutputToFile parameter is getting used. What am I missing?

-Ron

jimma:
Hi Ron, 
There is no other i18n  properties in resteasy code base, so it won't print message with other language. But I refactor the test a bit to make sure it won't break if we add other i18n property file in the future. Please review this is OK now. 
About the redirectTestOutputToFile configuration, which version maven do you use ? It's added since surefire plugin 2.3.
Cheers,
Jim

jimma:
The failure seems nothing to do with my change.  Can someone please restart the travis build ?

ronsigal:
Hey Jim,

I see 

    <version.surefire.plugin>2.18.1</version.surefire.plugin>

in org.jboss:jboss-parent:19:pom, which is Resteasy's parent pom. I just don't see how redirectTestOutputToFile is used. But that's ok. The test looks good, and it's passing now.

PR #798, opened 2016-05-10
RESTEASY-1345

ronsigal:
Added a subclass of ResteasyProviderFactory which doesn't reference its
parent after it is created.

https://issues.jboss.org/browse/RESTEASY-1345


ronsigal:
Supercedes PR #787.

PR #807, opened 2016-05-20
[RESTEASY-1197] Performance Issue: org.jboss.resteasy.spi.ResteasyUriInfo instantiation takes 20% of all the CPU time

asoldano:
This is a fixed PR on top of https://github.com/resteasy/Resteasy/pull/791 to avoid TCK regressions.

### DIFF
--- a/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
+++ b/jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
@@ -656,7 +656,7 @@ public class ResteasyUriBuilder extends UriBuilder
 				start = matcher.end();
 				continue;
 			}
-            throw new IllegalArgumentException(Messages.MESSAGES.templateParameterNull(param));
+            throw new IllegalArgumentException(Messages.MESSAGES.pathParameterNotProvided(param));
 		 }
 		 Object value = paramMap.get(param);
 		 String stringValue = value != null ? value.toString() : null;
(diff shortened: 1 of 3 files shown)


jaxrs/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
```diff
@@ -656,7 +656,7 @@ protected StringBuilder replaceParameter(Map<String, ? extends Object> paramMap,
 				start = matcher.end();
 				continue;
 			}
-			throw new IllegalArgumentException(Messages.MESSAGES.pathParameterNotProvided(param));
+            throw new IllegalArgumentException(Messages.MESSAGES.templateParameterNull(param));
```
NicoNes: In ResteasyUriBuilderTest.testFromPathWithMissingValues the error message will be " NULL value for template parameter: w". But if we change test from 

> URI uri = ResteasyUriBuilder.fromPath("").path("{v}/{w}").build(new Object[] {"first"}, false);

to

> URI uri = ResteasyUriBuilder.fromPath("").path("{v}/{w}").build(new Object[] {"first",null}, false);

then the message will be "path param w has not been provided by the parameter map". Actually contrary to the previous case param w has been provided but with a null value so message should be " NULL value for template parameter: w" in this case and "path param w has not been provided by the parameter map" in the first case.

It's appears clearer when you change test to 

> Map<String, String> parameters=new HashMap<>(2);
	    	parameters.put("v", "first");
	    	parameters.put("w", null);
	    	URI uri = ResteasyUriBuilder.fromPath("").path("{v}/{w}").buildFromMap(parameters, false);

The error will be ""path param w has not been provided by the parameter map" it will have been better to have "NULL value for template parameter: w".

What do you think ?

asoldano: Thanks for the review. I've swapped the 2 error messages, erroneously inverted them when chasing the spec/tck problems.

asoldano:
Merged, thanks for the contribution @NicoNes

PR #809, opened 2016-05-25
[RESTEASY-1400] Replace use of deprecated junit.framework.Assert with…

asoldano:
… org.junit.Assert


asoldano:
Merged, it's a simple automated replacement ;)

PR #812, opened 2016-05-25
https://issues.jboss.org/browse/RESTEASY-1356

kurobako:
https://issues.jboss.org/browse/RESTEASY-1356
RESTEasy don't use correct logging on many places

Existing logging classes are used now instead of System.out/System.err. Logging in test code and main methods is left untouched.

### DIFF
--- a/resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/i18n/Messages.java
+++ b/resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/i18n/Messages.java
@@ -41,5 +41,6 @@ public interface Messages
    @Message(id = BASE + 35, value = "Serving %s")
    String servingPathInfo(String pathInfo);
    
-   
+   @Message(id = BASE + 36, value = "Error while processing WADL")
+   String cantProcessWadl();
 }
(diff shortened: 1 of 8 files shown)


tjws/src/main/java/Acme/Serve/Serve.java
```diff
@@ -109,6 +109,10 @@
 import java.util.Vector;
```
asoldano: There are a bunch of FINER level logs in the class here; those should be protected with a LOG.isLoggable(Level.FINER) check if any computation is happening to create the arguments for the logs. In particular, it would make sense to have a static DEBUG_ON member, similar to the STREAM_DEBUG being removed, that would take the value of LOG.isLoggable(Level.FINER)

resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/i18n/Messages.java
```diff
@@ -41,5 +41,6 @@
    @Message(id = BASE + 35, value = "Serving %s")
    String servingPathInfo(String pathInfo);
    
-   
+   @Message(id = BASE + 36, value = "Error whiel processing WADL")
```
asoldano: Minor typo

asoldano:
Thanks, please see my comments on the changes.

kurobako:
retest this please

kurobako:
@asoldano applied suggested changes

asoldano:
Thanks, I've merged #813 instead, with just minor additions on your proposal.

PR #815, opened 2016-05-26
RESTEASY-1405: Support InputPart form parameters for @MultipartForm.

rankinc:
Allow Resteasy to inject InputPart as well as InputStream, so that we also have access to the part's Content-Type header.
See https://issues.jboss.org/browse/RESTEASY-1405


liweinan:
@rankinc Could you please provide test cases to the PR? Thanks!

rankinc:
Ugh - I've created a new branch for this.

PR #816, opened 2016-05-27
[RESTEASY-1266]

ronsigal:
Fixed cookie handling.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.client.jaxrs.internal;
 
+import org.jboss.resteasy.plugins.delegates.ClientCookie;
 import org.jboss.resteasy.util.CaseInsensitiveMap;
 import org.jboss.resteasy.util.DateUtil;
 import org.jboss.resteasy.util.HeaderHelper;
@@ -132,9 +133,23 @@ public class ClientRequestHeaders
 
    public void cookie(Cookie cookie)
    {
-      headers.add(HttpHeaders.COOKIE, cookie);
+      if (!(cookie instanceof ClientCookie))
+      {
+         cookie = new ClientCookie(cookie);
+      }
+      String cookies = (String)headers.getFirst(HttpHeaders.COOKIE);
+      StringBuilder builder = buildCookieString(cookies, cookie);
+      headers.putSingle(HttpHeaders.COOKIE, builder.toString());
    }
 
+   private StringBuilder buildCookieString(String cookies, Cookie cookie)
+   {
+      StringBuilder builder = new StringBuilder();
+      if (cookies != null) builder.append(cookies).append("; ");
+      builder.append(configuration.toHeaderString(cookie));
+      return builder;
+   }
+   
    public void allow(String... methods)
    {
       HeaderHelper.setAllow(this.headers, methods);
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
@@ -308,9 +308,13 @@ public class ClientRequestHeaders
          }
          else
          {
-            String str = configuration.toHeaderString(obj);
-            Cookie cookie = Cookie.valueOf(str);
-            cookies.put(cookie.getName(), cookie);
+            String[] cs = ((String) obj).split(";");
+            for (String c : cs)
+            {
+               String str = configuration.toHeaderString(c);
+               Cookie cookie = Cookie.valueOf(str);
+               cookies.put(cookie.getName(), cookie);
+            }
          }
       }
       return cookies;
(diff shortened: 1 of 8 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
```diff
@@ -293,9 +308,13 @@ public MediaType getMediaType()
          }
          else
          {
-            String str = configuration.toHeaderString(obj);
-            Cookie cookie = Cookie.valueOf(str);
-            cookies.put(cookie.getName(), cookie);
+            String[] cs = ((String) obj).split(";");
+            for (String c : cs)
+            {
+               String str = configuration.toHeaderString(c);
```
tbroyer: Why use `toHeaderString` on a string? Should be used on the input instead of the cast to string IMO.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/ClientCookie.java
```diff
@@ -0,0 +1,22 @@
...
+
+import javax.ws.rs.core.Cookie;
+import javax.ws.rs.core.NewCookie;
+
+/**
+ * 
+ * @author <a href="mailto:ron.sigal@jboss.com">Ron Sigal</a>
+ * @date May 20, 2016
+ */
+public class ClientCookie extends NewCookie
```
tbroyer: I don't understand why you need to introduce that `ClientCookie` and can't just use `Cookie`.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
tbroyer: BTW,  why not use `CookieParser`?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/ClientCookieHeaderDelegate.java
```diff
@@ -0,0 +1,29 @@
...
+      if (value == null)
+      {
+         throw new IllegalArgumentException(Messages.MESSAGES.paramNull());
+      }
+      NewCookie cookie = (NewCookie) value;
+      StringBuilder b = new StringBuilder();
+      b.append(cookie.getName()).append('=');
+      if (cookie.getValue() != null)
+      {
+         quote(b, cookie.getValue());
```
tbroyer: This is serializing "version 0" cookies (which happens to match what the web uses, and has been spec'd in RFC 6265), but the JAX-RS spec relies on RFC 2965 (version 1 cookies, never actually deployed by anyone).
So while this works (except values aren't quoted, but their value range restricted), this breaks spec compliance.
I've filed a bug against the spec a while ago, left without any response.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/ClientCookie.java
ronsigal: You're quite right. What threw me was that org.jboss.resteasy.plugins.delegates.CookieHeaderDelegate.toString() was using ServerCookie.appendCookieValue(), so I created a ClientCookie with its own delegate. I should have just fixed CookieHeaderDelegate. Stay tuned.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java
ronsigal: re; "Why use toHeaderString on a string?"

You're right.  Useless.

ronsigal: re: "BTW, why not use CookieParser?"

Well, yeah. Cookie.valueOf() uses CookieHeaderDelegate, which I've fixed now. I think.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/ClientCookieHeaderDelegate.java
ronsigal: Hmmm. Well, that explains some of my confusion. I didn't know about this wrinkle. Actually, the javax.ws.rs.core.Cookie javadoc refers to RFC 2109.

tbroyer: Sorry, I actually meant RFC 2109, not RFC 2965 (which tried to replace everything with new headers, and obviously failed; I think Opera is the only browser that ever implemented it, and it no longer does now that it switched to Chromium); but it doesn't change the issue: RFC 2109 is not used in practice; everyone follows "the original Netscape spec", adding extensions over time (Secure, HttpOnly), that were never spec'd until RFC 6265 aimed at describing the current state of the web (similar to how HTML5 also describes the current state of browser parsing behavior, besides adding new features).

I have no problem with breaking JAX-RS-spec compliance by dropping RFC 2109, but I'm almost sure there are some TCK tests that would require it, and there may be some "Enterprisey" people relying on it (there must be reasons JRE `java.net.HttpCookie`, Servlets and JAX-RS support, or are based on, those RFCs).

BTW, FYI, the JAX-RS spec bug report: https://java.net/jira/browse/JAX_RS_SPEC-430

ronsigal: Hey Thomas,

Thank you for the clarification. Ok, I'm surrendering to RFC 2109. I've backed out some of my changes and updated CookieHeaderDelegate to print $Version, $Path, and $Domain, and it looks like that doesn't break anything. I'm going to close this pull request, as it is superceded by #827. And I'll mention JAX_RS_SPEC-430 to our representative to the JAX-RS expert group.

-Ron

ronsigal:
Superceded by #827.

PR #821, opened 2016-06-01
[RESTEASY-1410]

ronsigal:
Moved TJWS classes from resteasy-jaxrs to tjws, moved test classes from resteasy-jaxrs to resteasy-jaxrs-testsuite, and made tjws dependent on resteasy-jaxrs instead of the other way around.


asoldano:
I've merged your commits. Thanks.

PR #827, opened 2016-06-04
[RESTEASY-1266]

ronsigal:
Fixed client side representation of cookies.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocationBuilder.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocationBuilder.java
@@ -76,6 +76,10 @@ public class ClientInvocationBuilder implements Invocation.Builder
    @Override
    public Invocation.Builder cookie(Cookie cookie)
    {
+      if (!(Cookie.class.equals(cookie.getClass())))
+      {
+         cookie = new Cookie(cookie.getName(), cookie.getValue(), cookie.getPath(), cookie.getDomain(), cookie.getVersion());
+      }
       getHeaders().cookie(cookie);
       return this;
    }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
@@ -2,6 +2,8 @@ package org.jboss.resteasy.plugins.delegates;
 
 import javax.ws.rs.core.Cookie;
 import javax.ws.rs.ext.RuntimeDelegate;
+
+import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 import org.jboss.resteasy.util.CookieParser;
 
 /**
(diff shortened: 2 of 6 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
```diff
@@ -16,10 +18,40 @@ public Cookie fromString(String value) throws IllegalArgumentException
...
    {
-      StringBuffer buf = new StringBuffer();
-      ServerCookie.appendCookieValue(buf, 0, value.getName(), value.getValue(), value.getPath(), value.getDomain(), null, -1, false);
-      return buf.toString();
+      if (cookie == null)
+      {
+         throw new IllegalArgumentException(Messages.MESSAGES.paramNull());
+      }
+      StringBuilder b = new StringBuilder();
+      b.append("$Version=").append(cookie.getVersion()).append(";");
```
tbroyer: Should only be done if `getVersion` is 1; specifically shouldn't be done if it's 0, or you'll risk breaking the web.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocationBuilder.java
```diff
@@ -76,6 +76,10 @@ public ClientRequestHeaders getHeaders()
    @Override
    public Invocation.Builder cookie(Cookie cookie)
    {
+      if (!(Cookie.class.equals(cookie.getClass())))
```
tbroyer: Is it really needed here given it's done in `ClientRequestHeaders` too?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
ronsigal: Specifically, the "$Version" attribute?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocationBuilder.java
ronsigal: Belt and suspenders. ;-)

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
```diff
@@ -16,10 +18,40 @@ public Cookie fromString(String value) throws IllegalArgumentException
       return CookieParser.parseCookies(value).get(0);
    }
 
-   public String toString(Cookie value)
+   public String toString(Cookie cookie)
    {
-      StringBuffer buf = new StringBuffer();
-      ServerCookie.appendCookieValue(buf, 0, value.getName(), value.getValue(), value.getPath(), value.getDomain(), null, -1, false);
```
tbroyer: Ha ha, hadn't realized current version **always** uses "version 0" (aka Netscape spec, or now RFC 6265) cookies (but strangely includes path and domain).

If this passes the TCK (it's been that way for at least 8 years, so either the TCK doesn't run that code at all, or it doesn't really care about the serialization provided it's properly parsed on the server-side, and because Servlet implementations –which I believe the TCK requires– don't really support RFC 2109, the TCK probably doesn't check the path and domain –and version– attributes), then I see **no reason to change to support RFC 2109.** (but possibly change the code to omit the path and domain, as `ServerCookie.appendCookieValue` will unconditionally add them, irrespective of the version)

What was the reason you changed that code?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
```diff
@@ -16,10 +18,40 @@ public Cookie fromString(String value) throws IllegalArgumentException
...
+      }
+      if (cookie.getDomain() != null)
+      {
+         b.append(";").append("$Domain").append("=");
+         quote(b, cookie.getDomain());
+      }
+      return b.toString();
    }
+   
+   protected void quote(StringBuilder b, String value) {
```
tbroyer: If you keep that code, then use `ServerCookie.maybeQuote2` instead.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
ronsigal: JAX-RS seems to be a bit schizophrenic on the subject of cookies. We're talking about how to address this. Stay tuned.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/CookieHeaderDelegate.java
ronsigal: Ok, thanks.

ronsigal:
Hey Thomas,

Your comment on June 5 put everything in perspective, I think. In particular, I just don't want to change things that aren't clearly broken, at least as far as JAX-RS is concerned. So I've created a new pull request, https://github.com/resteasy/Resteasy/pull/935, which is similar to this one but omits the changes to CookieHeaderDelegate. I'll close this pull request.

Thank you for your help.

-Ron

PR #844, opened 2016-06-18
RESTEASY-1405: Support InputPart form parameters for @MultipartForm.

rankinc:
Allow Resteasy to inject InputPart as well as InputStream, so that we also have access to the part's Content-Type header.
See https://issues.jboss.org/browse/RESTEASY-1405


rankinc:
This pull request replaces #815.

asoldano:
Replaced by https://github.com/resteasy/Resteasy/pull/908

PR #845, opened 2016-06-20
Add cache and speed up the build

jimma:


asoldano:
@jimma do we actually need this? (IOW are you currently seeing failures due to time spent on downloading dependencies?) While I like the idea in theory, I'm a bit concerned about the risk of loosing control over what's in the local maven repo. AFAIU, the PR builds would use the cache from the master runs, right? Did you manage to find out how much time is saved for each build with this option? (the cache still needs to be downloaded...)

asoldano:
In any case, should we decide to go the cache approach, your change here needs https://github.com/resteasy/Resteasy/pull/848 (otherwise the new local repo will always be different from the cached one, even excluding snapshots)

jimma:
This will save at least 2 mins.  From travis doc,  the cache is similar to the .m2 folder in a physical machine, travis will mount this cache for each build. I don't think this will cause some issue by PR build. The snapshot should be updated and used for each build. If this is not the case, we need to fix the build as you did in https://github.com/resteasy/Resteasy/pull/848.

asoldano:
OK, let's give this a try

PR #846, opened 2016-06-23
[RESTEASY-1361]:Move all WF8 arquillian tests to RESTEASY-TEST-WF8 mo…

jimma:
…dule


asoldano:
Cool, a step forward in re-organizing the testsuite. Thanks

PR #847, opened 2016-06-27
[RESTEASY-1366]: fixed Jackson2Provider client deserialization…

askannon:
…regression that was introduced in version 3.0.14 that breaks the Jackson @JsonView feature


asoldano:
Cool, thanks for the patch. Any chance you can also add a testcase? Thanks

asoldano:
@askannon  (otherwise I'll prepare one later this week)

askannon:
I added tests as well as fixed the same bug for the writer in addition to the reader.

askannon:
Any chance to get this merged into WildFly 10.1? Without this patch we can't upgrade to WF10 at all because @JsonView is broken.

asoldano:
OK, I'm closing this PR (would have needed you to rebase against current master which changes quite a bit since you opened this), but I've merged your changes on master and added the same testcase you provided but in a different project module.
As for the changes being included in WildFly 10.1, it really depends on WildFly release date; I'm most likely merging this fix to RESTEasy 3.0.x branch, so either 3.0.19.Final will be included in WildFly 10.1.0.Final or you'll have to install it manually afterwards on top WildFly 10.1.0.Final (or current WildFly 10.0.0.Final).

asoldano:
Ah, and thanks a lot for the contribution ;)

PR #848, opened 2016-06-28
[RESTEASY-1361] skip artifact installation for book examples (especia…

asoldano:
…lly considering that they build into different and not-SNAPSHOT versions from the core build)


asoldano:
OK, I've merged this as it basically cleans up the build with no side effects and is required by #845

PR #851, opened 2016-06-30
RESTEASY-1402: Let user override the cache by defining a ContextResolver<JAXBContext

NicoNes:
The goal of this pull request is to review XmlJAXBContextFinder so that `JAXBContext findCachedContext(Class type, MediaType mediaType, Annotation[] parameterAnnotations)` method works as it is supposed to work according to its documentation and the way ContextResolver should be used in JaxRS.


asoldano:
Merged, thanks!

PR #853, opened 2016-07-01
RESTEASY-1423: Respect private & no-store cache control flags

DXTR66:
https://issues.jboss.org/browse/RESTEASY-1423

* add check to ServerCacheInterceptor for private & no-store flag
* add org.mockito for easier unit testing
* add unit test for ServerCacheInterceptor


asoldano:
Hi! Thanks for the contribution, I think the issue is valid and the fix is OK.
Would it be possible to have an integration test instead of a unit test relying on mocks? I'd really like to avoid introducing a dependency to mockito just for this and would prefer having checks in place on the actual messages / behavior.
While you're on it, could you please rebease on top of current master?
Thanks!

DXTR66:
Sure no problem - will do that :).

PR #860, opened 2016-07-08
[RESTEASY-1361] New integration testsuite

asoldano:


rsvoboda:
I see a lot of logger.info(...), git grep logger.info | grep "src/test" | wc -l returns 1004
It can be address later, it may or may not be a problem. In general too much IO operations can decrease speed of TS execution. But on the other hand we have more details when something fails :) Some logging could be eliminated.

There are empty comments like in https://github.com/resteasy/Resteasy/pull/860/commits/0895f1dce79cda5e94eb0f5e5aaa450ea4c97603#diff-342d44b9f369d28a3028bcb31d321a51R93 probably from some bulk cleanup

Overall +1000 to merge it :)

PR #861, opened 2016-07-08
RESTEASY-1439: ETag not quoted in ResponseBuilder.tag(String tag)

bedrin:
[RESTEASY-1439: ETag not quoted in ResponseBuilder.tag(String tag)](https://issues.jboss.org/browse/RESTEASY-1439)

`Response.status(412).entity(message).tag("123").build()` should generate `ETag: "123"` but instead does not quote the tag `ETag: 123.`
The [JAX-RS spec](http://docs.oracle.com/javaee/6/api/javax/ws/rs/core/Response.ResponseBuilder.html#tag(java.lang.String)) says:

> the ETag should be quoted by the runtime

Jersey adheres to the spec here, RESTEasy doesn't.
`ResponseBuilder.tag(EntityTag tag)` method works corrrectly but `ResponseBuilder.tag(String tag)` has this bug


asoldano:
Merged, thanks for the patch!

PR #869, opened 2016-07-15
Fix for nondescript RuntimeException during Multipart-reading

MarkusKull:
Small bugfix to prevent (IO)Exceptions being wrapped by unhandled RuntimeException, when reading posted Multiparts.

### DIFF
--- a/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/MultipartInputImpl.java
+++ b/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/MultipartInputImpl.java
@@ -130,21 +130,30 @@ public class MultipartInputImpl implements MultipartInput, ProvidersContextRetai
             decodedStream = is;
          }
 
+         BodyFactory factory;
          try
          {
-            BodyFactory factory = (BodyFactory)bodyFactoryField.get(this);
-            body = factory.binaryBody(decodedStream);
-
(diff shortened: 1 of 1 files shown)


asoldano:
Hi, just out of curiosity, is there a test scenario that led you to spot this?

Speaking of the patch, how about catching IOException before Exception in the same try-catch block, instead of splitting it?

MarkusKull:
There was an actual problem which showed this. Pretty much convoluted with an exception deep inside reading. I manually tested it, but dont have a concise unittest.

In actual, an RuntimeException was thrown, not an IOException. A special case for IOException would not be enough.

asoldano:
No problem for the test in this case.

Why won't a change like https://gist.github.com/asoldano/a06a67d9e849bdbcc1168319a4060f91 actually work?

MarkusKull:
The gist wont work if a RuntimeException is thrown (IllegalArgumentException, WebApplicationException, whatever). Catch(IOException) will not be called, instead catch(Exception) will be called, which wraps the original exception in an typeless RuntimeException

asoldano:
Ah, so the problem is really in not wrapping exceptions that already are RuntimeException, I thought it was with IOExceptions being wrapped while they're already in the list of exceptions declared for that method. How about the updated gist?

MarkusKull:
Latest gist is ok. But i would prefer try-blocks to be as small as possible to make the intent clearer.

PR #871, opened 2016-07-16
[RESTEASY-1443]: Fix partial registration of application properties

NicoNes:
This pull request is a fix proposition so that all the application properties can be retrieved using `featureContext.getConfiguration().getProperties()`.

### DIFF
--- a/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/nextgen/properties/GetPropertiesTest.java
+++ b/resteasy-jaxrs-testsuite/src/test/java/org/jboss/resteasy/test/nextgen/properties/GetPropertiesTest.java
@@ -1,24 +1,24 @@
 package org.jboss.resteasy.test.nextgen.properties;
 
-import com.sun.net.httpserver.HttpContext;
-import com.sun.net.httpserver.HttpServer;
+import java.net.InetSocketAddress;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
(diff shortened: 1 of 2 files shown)


asoldano:
@NicoNes , I've tried patching the GetPropertiesTest as you did and still the test passes without fixing ResteasyDeployment. I would expect the test to verify the fix, can you please have a look?

NicoNes:
Hi @asoldano,

I just had a look at my version of `GetPropertiesTest `and I was a little confused because as you said it still work well without fixing `ResteasyDeployment`. 
So I tried to remember how I found out the regression and it appeared to me that only **3.0.17.Final** and **3.0.18.Final** versions are affected.
According to me it is due to the regression introduced by https://issues.jboss.org/browse/RESTEASY-1284 in **3.0.17.Final**. It has been fixed by you in **3.0.19.Final** (https://issues.jboss.org/browse/RESTEASY-1444).

The short explaination of the regression (3.0.17.Final and 3.0.18.Final) is that each provider class were added into `providerClasses` variable so when trying to register the `Features` holding the properties, only the first one was registered and the other ones were ignored because of following code in RestEasyProviderFactory 

`      if (getClasses().contains(provider.getClass()))
      {
         LogMessages.LOGGER.providerClassAlreadyRegistered(provider.getClass().getName());
         return;
      }`

So thanks to your fix (https://issues.jboss.org/browse/RESTEASY-1444) this PR and it associated Jira can be closed. 
However although the fix I did in `ResteasyDeployment `was useless, maybe you can keep it to only add one feature instance in the `providers` Set variable, instead of as many as registered properties  in the application. What do you think ?

PR #873, opened 2016-07-18
[RESTEASY-1444] Fix regression introduced in RESTEASY-1284

asoldano:


NicoNes:
Hi,

Looks good for me. Thanks !

PR #876, opened 2016-07-22
[RESTEASY-1450] Testsuite - annotations, review failing tests, configurable resteasy version for unit tests

kanovotn:
* Add documentation annotations for various tests
* Check that tests expected to fail have appropriate jira assigned
* Add profile to unit-tests submodule to allow to run unit tests with different resteasy version than the current project version


asoldano:
The PR looks good. My only concern is with the actual need for the changes in XmlJAXBContextFinderTest, in particular moving all those small classes in dedicated java files. It's really unlikely that those classes will be needed anywhere else, having them in the test files improves code readability and keep the testsuite "simpler" (at least imho).
Do we really need to move those classes out of the testcase?

kanovotn:
Thank you for review. You are probably right:) I reverted the changes in XmlJAXBContextFinderTest.

asoldano:
Thanks!

PR #883, opened 2016-07-27
[RESTEASY-1453]:Remove tests which already checked by netty adaptor;c…

jimma:
…lean up imports


server-adapters/resteasy-vertx/src/test/java/org/jboss/resteasy/test/JaxrsAsyncTest.java
```diff
@@ -1,191 +0,0 @@
...
-{
-   static String BASE_URI = generateURL("");
-   static Client client;
-
-   static final int REQUEST_TIMEOUT = 1000;
-
-   @BeforeClass
-   public static void setupSuite() throws Exception
-   {
-      VertxContainer.start().getRegistry().addSingletonResource(new AsyncJaxrsResource());
```
asoldano: Well, Netty 4 adaptor might have a similar test, but this is using VertxContainer, so it would make sense to keep it.

server-adapters/resteasy-vertx/src/test/java/org/jboss/resteasy/test/RESTEASY1323Test.java
```diff
@@ -1,178 +0,0 @@
...
-public class RESTEASY1323Test
-{
-   static String BASE_URI = generateURL("");
-
-   static final int REQUEST_TIMEOUT = 4000;
-
-   @BeforeClass
-   public static void setupSuite() throws Exception
-   {
-      VertxContainer.start().getRegistry().addSingletonResource(new AsyncJaxrsResource());
```
asoldano: Same as above

asoldano:
Merged; I initially had some doubts, but the basic tests are still available.

PR #885, opened 2016-07-27
Refactoring

asoldano:
We have the TestPortProvider class in a test package of resteasy-client src/main. However that class is used in multiple server adapters (netty, undertow, vertx, ...) containers (not in tests only) and in some cases the containers are even in 'test' packages...
We should put some order here, refactoring the containers classes in different packages etc.


asoldano:
Closed by 3246ed3cc4c79a8c39b83ab1118fc254010274ca

PR #888, opened 2016-08-01
Resteasy 1463 bld test failures build43

rsearls:


asoldano:
I've cherry-picked b223008 commit only. So this can be closed, thanks.

PR #897, opened 2016-08-12
[RESTEASY-1407] [Client] Request fails when sending an empty ByteArra…

treblereel:
…yInputStream

JIRA: https://issues.jboss.org/browse/RESTEASY-1407


treblereel:
hmmm

org.junit.ComparisonFailure: expected:<html[2]> but was:<html[1]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.jboss.resteasy.test.cache.PersistentServerCacheTest.testAccepts(PersistentServerCacheTest.java:291)

treblereel:
retest this please

MarkusKull:
Uh, is this really good? Whenever the restclient-user sets a content-length-Header but the real data has a differing length, thats an error of the restclient-user. This change seems to *always* ignore content-length and enforces chunked-encoding. A fix only for the described problem would be better.

treblereel:
@MarkusKull done,i hope,it much better :) , plus test added

PR #900, opened 2016-08-19
minor doc tweak

ndrake:
Fix wording in the upgrade docs


asoldano:
Thanks for the PR. Actually, while checking your changes, I noticed that the sentences you modified still have problems "... Unzip this file while within the modules/x/y/z directory of the WildFly distribution. " Would you mind fixing the remaining problems so that I merge your PR? (otherwise I'll simply close this and go and fix the stuff directly).

PR #902, opened 2016-08-22
[RESTEASY-1474] Allow to override redirectTestOutputToFile attribute

marekkopecky:
Jira: https://issues.jboss.org/browse/RESTEASY-1474

----

Allow to override redirectTestOutputToFile attribute of surefire plugin in testsuite.

This attribute is false by default. If value of this attribute is true in pom, it can't be override by -Dmaven.test.redirectTestOutputToFile=<property>.

----

Use same way as WF use:
https://github.com/wildfly/wildfly-core/blob/master/testsuite/pom.xml#L127
https://github.com/wildfly/wildfly-core/blob/master/testsuite/pom.xml#L307


marekkopecky:
Better solution (better than WF solution) is to set maven.test.redirectTestOutputToFile in properties ...

PR #903, opened 2016-08-23
[RESTEASY-1481] UndertowJaxrsServer: Exposing the ability to add prefix paths to the server root path handler

gpjacobs:
Undertow allows adding a variety of paths, but the default behavior in the current UndertowJaxrsServer automatically registers a root handler and implicitly adds the JAX-RS defined paths to it, disallowing manually adding other paths. Adding this method allows the manual addition of other paths such as for serving static resources (e.g. web pages, API documentation) alongside the REST endpoints. 

Inspired by a discussion I found when trying to find a solution to this problem: http://lists.jboss.org/pipermail/undertow-dev/2014-February/000680.html


asoldano:
Thanks for the PR. Would you mind adding a testcase for the changes here and creating a RESTEASY jira? Thanks!

gpjacobs:
Will do, I'll add a test case and JIRA ticket.

gpjacobs:
JIRA ticket created and test case added.

asoldano:
Thanks a lot!

PR #905, opened 2016-08-28
[RESTEASY-1201]block close in part writer

ian4hu:
add a unit test to produce the RESTEASY-1201 and fix RESTEASY-1201


ian4hu:
previous pr is #715

asoldano:
Thanks for the new PR including tests. I've merged it, considering Ron agreed with the proposed solution in the former PR.

PR #909, opened 2016-08-31
[RESTEASY-1479] Clean up compiler warnings in the testsuite module

kanovotn:


testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/resource/CookieObjectTest.java
```diff
@@ -149,7 +152,7 @@ public void testNewCookie4() throws Exception {
         boolean pass = true;
         String NewCookie_toParse =
                 "Customer=WILE_E_COYOTE; Path=/acme; Domain=acme.com; Max-Age=150000000; " +
-                        "Expires=Thu, 03-May-2018 10:36:34 GMT; Secure; HttpOnly";
+                        "Expires=Thu, 03-May-2118 10:36:34 GMT; Secure; HttpOnly";
```
asoldano: Can you explain this change please? was there an error in the test?

kanovotn: You are right, it is not needed here, I reverted that.

asoldano:
Thanks

PR #911, opened 2016-08-31
Add execution options into testsuite readme

kanovotn:


asoldano:
Thanks @kanovotn ; I only have one concern, which is around server.home vs jboss.home option. I believe with the current master the user should be using server.home as that serves both the goal of specifying the location of the server to be used for testing and prevents the build from downloading a container. The user should not be dealing with jboss.home option.
Can you please verify / double check this and update the doc if there's need to?
Thanks

asoldano:
@kanovotn why have you applied changes to the pom? You're basically breaking what I did in https://github.com/resteasy/Resteasy/commit/769b6dd99522965e716394500dcfdca8cd08c491 .As a consequence, the testsuite is failing.

kanovotn:
@asoldano, I did it because when I specified server.home to maven, it wasn't propagated to jboss.home and arquillian here could'nt find jboss.home - https://github.com/kanovotn/Resteasy/blob/master/testsuite/integration-tests/src/test/resources/arquillian.xml.

I'm checking that fail, sorry for that.

asoldano:
@kanovotn mmh... perhaps I didn't test a specific usage scenario. Can you write tell me the exact mvn command that you were using when having the problem?

kanovotn:
For example this:
`cd testsuite`
`mvn clean verify -Dserver.home=/home/kanovotn/eap-bin/7.1.0.DR3/jboss-eap-7.1 -Dtest=BasicAuthTest -fn`

asoldano:
@kanovotn ok, let me try fixing this after lunch, thanks

asoldano:
@kanovotn , OK, I think I've fixed the bug, currently verifying at https://github.com/resteasy/Resteasy/pull/917 (I tried you mvn command with that fix and it works locally)

asoldano:
@kanovotn I've merged #917. I assume you can modify this pr to only change the readme.

kanovotn:
@asoldano Thank you, now only readme is modified in this pr.

asoldano:
@kanovotn cool, thanks for the bug report, I've merged the PR.

PR #913, opened 2016-09-01
Another try to address timeout failures on RESTEASY1323Test

asoldano:


asoldano:
I've had 9 testsuite runs with no failures in a row, so I'm merging this... let's see if it helps.

PR #697, opened 2015-11-11
Use non-deprecated API provided by HttpClient.

xiaodong-xie:
1. Replace the usage of deprecated API with non-deprecated ones.
2. Expose more configuration for `ResteasyClientBuilder`.

The second one could be controversy though.

PR #923, opened 2016-09-04
Use non-deprecated API provided by HttpClient.

xiaodong-xie:
1. Replace the usage of deprecated API with non-deprecated ones.
2. Expose more configuration for ResteasyClientBuilder.

This is the second trial of https://github.com/resteasy/Resteasy/pull/697, with the re-formatting needed...

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
@@ -1,51 +1,39 @@
 package org.jboss.resteasy.client.jaxrs;
 
 import org.apache.http.HttpHost;
-import org.apache.http.client.params.HttpClientParams;
-import org.apache.http.conn.ClientConnectionManager;
-import org.apache.http.conn.params.ConnRoutePNames;
-import org.apache.http.conn.scheme.PlainSocketFactory;
-import org.apache.http.conn.scheme.Scheme;
-import org.apache.http.conn.scheme.SchemeRegistry;
-import org.apache.http.conn.ssl.AllowAllHostnameVerifier;
-import org.apache.http.conn.ssl.BrowserCompatHostnameVerifier;
-import org.apache.http.conn.ssl.SSLSocketFactory;
-import org.apache.http.conn.ssl.StrictHostnameVerifier;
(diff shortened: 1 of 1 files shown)


asoldano:
@xiaodong-xie hi, I've had a look at your changes now (thanks for having cleaned up the PR) and noticed Travis CI fails on WildFly 9.0.2 run. The reason basically is that new httpclient api are not available on 9.0.2. Now, recently we actually dealt with this problem, the Resteasy client is able to use either the new or old httpclient apis, see https://github.com/resteasy/Resteasy/blob/350b44f144eef01b0afee8c6dd64b2e9d814c358/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java#L375-L426
So, I suspect this PR is actually out-of-date now, but please confirm. Thanks.

asoldano:
@xiaodong-xie in, particular, if it's not out-of-date, the new configuration stuff that you added should be made available on the proper / already available class(es) using the new apis.

xiaodong-xie:
Cool, thanks, @asoldano, I'll have another look, and try to make the changes compatible with `WildFly 9.0.2`. Thanks for your time... 👍

xiaodong-xie:
@asoldano, it was due to in `HttpClient 4.3.6`, the method in question named `setSslcontext`, while in `HttpClient 4.5.x`, the new method was named `setSSLContext`, while `setSslcontext` was deprecated... 

If we still need to support `4.3.6`, we have to use `setSslcontext`...

xiaodong-xie:
I've updated the code and force pushed, hope it will build this time...

asoldano:
@xiaodong-xie OK, I've scanned your PR again. It builds, sure, but you didn't get my point. The changes you applied are for avoiding using httpclient api that were deprecated in 4.3. We've added support for httpclient 4.3+ on master, see ResteasyClientBuilder::build() and HttpClientBuilder43. The code that you modified, instead, is reached from ResteasyClientBuilder::buildOld(), which is meant for older httpclient versions. So, if we're going to fix/improve anything related to httpclient 4.3 or greater usage, that have to go into HttpClientBuilder43 class.

xiaodong-xie:
Thanks for the information, @asoldano, will have a look...

xiaodong-xie:
Closing for now...

PR #930, opened 2016-09-07
Improvements of accept header and matrix parameter

MarvGilb:
Dear RESTeasy Team,

this pull request contains some improvements for the client and is basically the same as this [pull request](https://github.com/resteasy/Resteasy/pull/676), but does not include the first stated commit ("Support body in case of server side exception (http status >= 400)."). Also it is compatible with the current branch again and contains the requested jira tickets.

[RESTEASY-1493](https://issues.jboss.org/browse/RESTEASY-1493) -- [e4fd77a](https://github.com/resteasy/Resteasy/commit/e4fd77a43caaaf21f5508bda4690b1409695a745) Fixes the problem, that matrix parameter needed to be passed at the last sub resource. Now they can be passed at the corresponding resource.

[RESTEASY-1494](https://issues.jboss.org/browse/RESTEASY-1494) -- [d5ef8f5](https://github.com/resteasy/Resteasy/commit/d5ef8f59385a044b88420dc30a15e07db312ec6d) This makes the client more flexible and sends multiple accept header instead of just the first one.

### DIFF
--- a/jaxrs/resteasy-client/pom.xml
+++ b/jaxrs/resteasy-client/pom.xml
@@ -17,6 +17,12 @@
             <artifactId>resteasy-jaxrs</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>tjws</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
         
         <dependency>
             <groupId>org.jboss.logging</groupId>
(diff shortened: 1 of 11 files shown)


asoldano:
preliminary note, to other reviewers, please note the PR is against 3.0.x branch, so we'd need a corresponding fix for master

asoldano:
I've created a new pull request with the changes here but against master, see https://github.com/resteasy/Resteasy/pull/934 . I'm closing this PR because the changes here are basically new features, so we're not including them in 3.0.x (which is effectively bug-only/maintenance mode).
Thanks for the contribution!

PR #906, opened 2016-08-30
RESTEASY-1341 - #792 rebased against current master

asoldano:

PR #937, opened 2016-09-11
My resteasy 1341

NicoNes:
New pull request replacing https://github.com/resteasy/Resteasy/pull/906.

The goal of this pull request is to make few resteasy builders compliant with the Builder pattern (http://www.javaworld.com/article/2074938/core-java/too-many-parameters-in-java-methods-part-3-builder-pattern.html) and add consistency in the way Builder pattern is implemented in the project.
Actually almost all resteasy builders strictly follow the Builder pattern such as org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder but few others don't for unknown reason.

According to the pattern for a given builder instance, all "setters" methods (those which return a  builder) must return the updated builder instance not a new one:

>  public static class PersonBuilder
{
	....
	public PersonBuilder lastName(String newLastName)
	{
		this.nestedLastName = newLastName;
		**return this;**
	}
	public PersonBuilder firstName(String newFirstName)
	{
		this.nestedFirstName = newFirstName;
		**return this;**
	}
	...
}

And all "build()" methods (the one which return a built instance) must return a fresh new instance:

> public static class PersonBuilder
{
	....
	public Person createPerson()
	{
		 return **new Person**(
			nestedLastName, nestedFirstName, nestedMiddleName,
			nestedSalutation, nestedSuffix,
			nestedStreetAddress, nestedCity, nestedState,
			nestedIsFemale, nestedIsEmployed, nestedIsHomeOwner);
	}
}

I think it's the way we expect this pattern to work. It's not about immutability it can work whatever the type to build is mutable or not, it's clearly not my point. My point is about being consistent in how we implement the builders with respect to the Builder pattern. Once the object has been built from its builder, this object must not be altered by any builder method. If the built object is mutable it will be modified through its own methods (see ResteasyClient for example).

This pull request focuses on: 

- ClientInvocationBuilder
- LinkBuilderImpl
- ResteasyUriBuilder

This pull request allows to raise and fix another issue revealed by ClientInvocationBuilderTest.test_build_method_reset_entity().

### DIFF
--- a/jaxrs/resteasy-client/src/test/java/org/jboss/resteasy/test/internal/ClientInvocationBuilderTest.java
+++ b/jaxrs/resteasy-client/src/test/java/org/jboss/resteasy/test/internal/ClientInvocationBuilderTest.java
@@ -1,0 +1,101 @@
+package org.jboss.resteasy.test.internal;
+
+import java.io.ByteArrayOutputStream;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.Invocation.Builder;
+import javax.ws.rs.core.MediaType;
+
+import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
+import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
+import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
+import org.jboss.resteasy.core.Dispatcher;
+import org.jboss.resteasy.spi.ResteasyDeployment;
+import org.jboss.resteasy.test.EmbeddedContainer;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ClientInvocationBuilderTest {
+
+	protected static ResteasyDeployment deployment;
+	protected static Dispatcher dispatcher;
+	protected static ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+	@Path("/")
+	public static class TestResource {
+		@POST
+		@Produces("text/plain")
+		public String post(String s) {
+			System.out.println("server entity: " + s);
+			return s;
+		}
+
+		@GET
+		@Produces("text/plain")
+		public String get() {
+			String s = "default";
+			System.out.println("server entity: " + s);
+			return s;
+		}
+	}
+
+	@BeforeClass
+	public static void before() throws Exception {
+		deployment = EmbeddedContainer.start();
+		dispatcher = deployment.getDispatcher();
+		deployment.getRegistry().addPerRequestResource(TestResource.class);
+	}
+
+	@AfterClass
+	public static void after() throws Exception {
+		EmbeddedContainer.stop();
+		dispatcher = null;
+		deployment = null;
+	}
+
+	@Test
+	public void test_build_method_return_new_instance() {
+		ResteasyClient client = new ResteasyClientBuilder().build();
+		try {
+			ResteasyWebTarget webTarget = client.target("http://localhost:8081");
+			Builder invocationBuilder = webTarget.request();
+
+			// GET invocation
+			ClientInvocation getInvocation = (ClientInvocation) invocationBuilder.accept(MediaType.TEXT_PLAIN_TYPE)
+					.build("GET");
+			Assert.assertEquals("default", getInvocation.invoke(String.class));
+
+			// Alter invocationBuilder
+			invocationBuilder.accept(MediaType.APPLICATION_XML_TYPE);
+			invocationBuilder.property("property1", "property1Value");
+			// Previously built getInvocation must not have been altered.(Those
+			// tests are not about immutability of
+			// getInvocation instance but about Builder pattern behavior).
+			Assert.assertFalse(getInvocation.getHeaders().getAcceptableMediaTypes()
+					.contains(MediaType.APPLICATION_XML_TYPE));
+			Assert.assertFalse(getInvocation.getConfiguration().getProperties().containsKey("property1"));
+
+			// POST invocation
+			ClientInvocation postInvocation = (ClientInvocation) invocationBuilder.accept(MediaType.TEXT_PLAIN_TYPE)
+					.build("POST", Entity.text("test"));
+			// Previous lines must build a new postInvocation instance and not
+			// modify previously built getInvocation instance.
+			// (It's all about Builder pattern behavior not immutability since
+			// Invocation is a mutable object.)
+			Assert.assertNotSame(getInvocation, postInvocation);
+			Assert.assertEquals("default", getInvocation.invoke(String.class));
+			Assert.assertTrue(postInvocation.getConfiguration().getProperties().containsKey("property1"));
+			Assert.assertEquals("test", postInvocation.invoke(String.class));
+		} finally {
+			client.close();
+		}
+	}
+
+}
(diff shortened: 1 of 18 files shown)


jaxrs/resteasy-client/src/test/java/org/jboss/resteasy/test/internal/ClientInvocationBuilderTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
+import org.jboss.resteasy.core.Dispatcher;
+import org.jboss.resteasy.spi.ResteasyDeployment;
+import org.jboss.resteasy.test.EmbeddedContainer;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class ClientInvocationBuilderTest {
```
asoldano: I don't think this test is actually being run, see for example https://s3.amazonaws.com/archive.travis-ci.org/jobs/159184701/log.txt . In any case, it's duplicate of integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientInvocationBuilderTest.java , right?

asoldano:
@ronsigal , can you please have a look at this too? The changes are eventually passing the testsuite :-)

NicoNes:
Yes you're right I forgot to remove the old test, both are same.

jaxrs/resteasy-client/src/test/java/org/jboss/resteasy/test/internal/ClientInvocationBuilderTest.java
ronsigal: I don't see ClientInvocationBuilderTest ... . I would think that it's somewhat different, though, since Nicolas has changed the behavior of ClientInvocationBuilder. But this test should be moved from resteasy-client to be consistent with our current testsuite structure. The easiest thing might be to remove the call to invoke and make it a pure unit test in resteasy-unit-tests.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/AsynchronousInvoke.java
```diff
@@ -24,219 +24,189 @@ public AsynchronousInvoke(ClientInvocation invocation)
...
+      return method(HttpMethod.GET);
    }
 
    @Override
    public <T> Future<T> get(Class<T> responseType)
    {
-      invocation.setMethod(HttpMethod.GET);
-      return invocation.submit(responseType);
+      return method(HttpMethod.GET, responseType);
    }
```
ronsigal: These are all just simplifications, right? I mean, there's no substantive change, right?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientWebTarget.java
```diff
@@ -233,7 +233,7 @@ public ResteasyWebTarget matrixParam(String name, Object... values) throws NullP
       else
       {
          String[] stringValues = toStringValues(values);
-         copy = uriBuilder.clone().matrixParam(name, stringValues);
+         copy.matrixParam(name, stringValues);
```
ronsigal: Nice catch.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/AsynchronousInvoke.java
NicoNes: It's not just simplifications.
The main point is that each time `get`, `trace` etc is called `invocation.setEntity()` (see `method` function) is also called to set the entity to the correct value. It allows to avoid exception such as  `javax.ws.rs.ProcessingException: RESTEASY004565: A GET request cannot have a body.`

jaxrs/resteasy-client/src/test/java/org/jboss/resteasy/test/internal/ClientInvocationBuilderTest.java
ronsigal: Oh, you mean the ClientInvocationBuilderTest from commit bff6805214f146aef0795a9846d071cf85d30d38?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/AsynchronousInvoke.java
ronsigal: Ah, I was looking at the old versions of method().

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -117,26 +120,24 @@ protected ClientInvocation createRequest(Object[] args)
...
-         builder = (ClientInvocationBuilder)target.request();
-      }
-
       for (int i = 0; i < processors.length; i++)
       {
          if (processors != null && processors[i] instanceof InvocationProcessor)
          {
             InvocationProcessor processor = (InvocationProcessor)processors[i];
-            processor.process(builder, args[i]);
+            processor.process(clientInvocation, args[i]);
```
ronsigal: Ok, I can see that passing ClientInvocation rather than ClientInvocationBuilder is a little simpler, but I don't follow the statement "Further more, according to the way jaxrs Invocation.Builder is designed we can't built Invocation with entity in too steps contrary to what is done in ClientInvoker.createRequest(Object[]). If invocation need to be built with an entity it must be done in one step (see build(method,entity), method(method,entity)...)."

MessageBodyParameterProcessor is setting the entity in the invocation, right? What am I missing?

NicoNes: Want I wanted to say is that each `processor` is in charge of adding a "part" to the request so that at the end, after going through the whole `processors` list the request is fully built. 

So at first sight passing `ClientInvocationBuilder` can seems the right thing to do. Each processor will call a "setter" method on the builder and at the end the `build` method will be called to get a `ClientInvocation` ready to be invoked.

The problem is that there is no entity setter on the jaxrs `Invocation.Builder` because entity is not supposed to be "a property stored" in the builder (to my mind).   The only way to pass entity to the `Invocation.Builder` is at build phase, because entity is supposed to be a "property stored" in the built `Invocation` (to my mind) . 

So if we would have liked to properly use `ClientInvocationBuilder` in MessageBodyParameterProcessor we would have called the `build()` method. But as a common contract of `BuilderPattern.build()` it would have returned a new `ClientInvocation` and our current `ClientInvocationBuilder`  will not have been impacted since entity is not "stored" in it.

So passing `ClientInvocationBuilder` appears  not to be so good that why I guess `MessageBodyParameterProcessor` accessed `ClientInvocation` instead.

Hope it's a bit clearer.

The second and main reason why I used  `ClientInvocation` instead is that, with my changes in `ClientInvocationBuilder.build()' method, `MessageBodyParameterProcessor`  set entity in the invocation and after that, call to `ClientInvocationBuilder.build()'  reset it.

ronsigal: Hi Nicolas,

Sorry for the delay. Ok, I get it. It looks like everything is OK. Thanks.

-Ron

ronsigal: Just to be clear, I think this one can be merged.

asoldano: Thanks Ron

PR #939, opened 2016-09-13
UndertowTest - remove testing file index.html after the test execution

xstefank:


asoldano:
Thanks @xstefank , I was actually going to check what's leaving that file in the sources...
Now that we're at this, would it be possible to create the file in a temp location (/tmp or such) or in a directory that's going to be cleaned in any case during the build?
On top of that, the delete() should likely live in a finally block of a try-finally. Thanks!

xstefank:
Hello @asoldano, the PR is updated to create tmp file in the /target directory, which is automatically deleted with the exit of the test JVM.

PR #940, opened 2016-09-14
[RESTEASY-1483]

ronsigal:
Use SecureRandom in AsynchronousDispatcher to generate job ids.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AsynchronousDispatcher.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AsynchronousDispatcher.java
@@ -72,7 +72,7 @@ public class AsynchronousDispatcher extends SynchronousDispatcher
       private int maxUses = -1;
       private int uses = 0; // uses > maxUses so that context parameters will get checked upon first use.
       
-      public int nextInt()
+      synchronized public int nextInt()
       {
          if (++uses > maxUses)
          {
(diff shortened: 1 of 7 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AsynchronousDispatcher.java
```diff
@@ -62,20 +64,71 @@ public void setMaxSize(int maxSize)
...
+      }
+   }
 
    protected ExecutorService executor;
    private int threadPoolSize = 100;
    private Map<String, Future<MockHttpResponse>> jobs;
    private Cache cache;
    private String basePath = "/asynch/jobs";
-   private AtomicLong counter = new AtomicLong(0);
+   private volatile SecureRandomWrapper counter;
```
asoldano: OK, I see you replaced the atomic long with a volatile ref to the wrapper: is volatile really needed here (how about declaring counter as final, given it's set in the constructor and never changed afterwards). Besides that, do we perhaps need something to deal with concurrent access to the wrapper methods?

ronsigal: D'oh. Yes, of course. I started with a plain volatile SecureRandom, and then created SecureRandomWrapper to periodically reseed the SecureRandom. Thanks, Alessio.

asoldano: ok, thanks for the fix :-)

PR #945, opened 2016-09-18
Added ability to control slash encoding for ParamParam

garethahealy:
If a client uses @PathParam which contains a slash, its currently encoded which then causes the call to fail. With this change, the user can decide to not encode the slashes by adding @Encoded


asoldano:
Related thread on user list: http://lists.jboss.org/pipermail/resteasy/2016-September/000009.html

PR #946, opened 2016-09-20
[JBEAP-4899] : RESTEasy should log used Providers and Interceptors in DEBUG level.

panossot:
https://issues.jboss.org/browse/JBEAP-4899

### DIFF
--- a/jaxrs/eagledns/src/main/java/se/unlogic/eagledns/zoneproviders/db/DBZoneProvider.java
+++ b/jaxrs/eagledns/src/main/java/se/unlogic/eagledns/zoneproviders/db/DBZoneProvider.java
@@ -23,6 +23,10 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.Logger.Level;
+
 public class DBZoneProvider implements ZoneProvider {
 
 	private static final Field RECORD_RELATION = ReflectionUtils.getField(DBZone.class, "records");
@@ -43,6 +47,8 @@ public class DBZoneProvider implements ZoneProvider {
 	private QueryParameterFactory<DBZone, Integer> zoneIDQueryParameterFactory;
 	private QueryParameterFactory<DBRecord, DBZone> recordZoneQueryParameterFactory;
 
+        @LogMessage(level = Level.DEBUG)
+        @Message(value = "Provider : se.unlogic.eagledns.zoneproviders.db.DBZoneProvider , method call : init .")
 	public void init(String name) throws ClassNotFoundException {
 
 		this.name = name;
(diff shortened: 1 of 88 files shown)


panossot:
Failure is irrelevant. 

Retest this please.

asoldano:
Hi @panossot , I have two comments here:
1) I tried re-running the testsuite runs, but they failed again, so I suspect there's indeed something in the changes you applied that is causing them
2) the way you implemented the logs is not fine in terms of performances. You should not do something like 

LogMessages.LOGGER.debug(Messages.MESSAGES.interceptorStartMethodCall("org.jboss.resteasy.plugins.providers.jackson","JacksonJsonpInterceptor","aroundWriteTo"));

instead you should be adding @LogMessage annotated methods with proper level and use them directly. The way you did, the i18n messages will always be computed, even if the log configuration is such that they will not eventually be printed.

panossot:
Hello @asoldano. Thank you for your comments. I will do it by adding  @LogMessage annotated methods. It is not ready yet, as some tests are failing..

panossot:
@asoldano : There were some more providers to be annotated. Let's see if it builds.

marekkopecky:
Wouldn't be possible add logging only to one place instead of to all providers? Users could register custom provider to RESTEasy and this custom provider won't be logged.

rsvoboda:
It's too much work and code done in this PR.

I asked Jamez on Hipchat if there is any way how could JBoss logging help - e.g. set  default '@LogMessage(level = Level.DEBUG)' for class / package

I can think of aspects to log method invocation instead of adding @LogMessage and @Message to every method in those classes.

marekkopecky:
AFAIK, MessageBodyReader is selected here:
https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/interception/AbstractReaderInterceptorContext.java#L52

In my point of view, it would be enough to add proper logging here, example: (this works correctly, I test it with WF):
```
LogMessages.LOGGER.debugUsedMessageBodyReader(reader.getClass().getName().);
```
and to similar places for MessageBodyWriter and Interceptors.

This approach correctly logs also custom providers defined in deployments.

panossot:
According to https://github.com/jboss-logging/jboss-logging/blob/master/src/main/java/org/jboss/logging/LogMessage.java the target is method.

Also, if the proceed method of AbstractReaderInterceptorContext.java is overidden then there will be no logging.

I could limit the Logging Annotations only to the basic methods.

panossot:
On second thoughts maybe marekkopecky's idea would be the best solution. I will implement that one.

jamezp:
These annotations will be ignored unless they are on an interface annotated with `@MessageLogger` or `@MessageBundle`. Adding annotations to random messages won't do anything. The logging tooling generates implementations for logger or bundle interfaces. It's a compile time only tool.

There is also no requirement for debug messages to be internationalized. Therefore all you need to do is log a normal debug message. For example you just need to do something like the following:
```
LogMessages.LOGGER.debugf("Provider: %s", getClass().getName())
```

panossot:
Closing this one. Replaced via https://github.com/resteasy/Resteasy/pull/951.

PR #948, opened 2016-09-23
[RESTEASY-1507] Add tests for http response codes with spring mvc dis…

kanovotn:
…patcher


asoldano:
@kanovotn if we want to merge the PR here now, the tests should be (temporarly) excluded

kanovotn:
Thanks, for review, I forget to exclude it.

asoldano:
No problem

PR #951, opened 2016-09-26
[JBEAP-4899] : RESTEasy should log used Providers and Interceptors in DEBUG level.

panossot:
https://issues.jboss.org/browse/JBEAP-4899


rsvoboda:
I think you have lost method name in the log message comparing to the previous PR #946

PR #954, opened 2016-09-27
[RESTEASY-1415] : RESTEasy should log used Providers and Interceptors in DEBUG level.

panossot:
https://issues.jboss.org/browse/RESTEASY-1415


server-adapters/resteasy-undertow/index.html
```diff
@@ -0,0 +1 @@
+Hello static world!
```
asoldano: This file should not be added to the sources (its erroneously written in the sources by a test, there's a PR for fixing this which still needs to be processed). Please remove this file from the PR anyway.

panossot: Removed.

PR #957, opened 2016-09-28
Resteasy 1485 reader exception 3.1.0 snapshot

rsearls:
There are 2 new classes
2 service files required an update.
a handful of test needed to be adjusted.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,6 @@
         <version.com.sun.mail>1.5.5</version.com.sun.mail>
         <version.com.sun.xml.bind.jaxb-impl>2.2.7</version.com.sun.xml.bind.jaxb-impl>
         <version.commons-io.commons-io>2.4</version.commons-io.commons-io>
-        <version.commons-lang.commons-lang>2.6</version.commons-lang.commons-lang>
         <version.hamcrest>1.3</version.hamcrest>
         <version.io.netty.netty>3.10.6.Final</version.io.netty.netty>
         <version.io.netty.netty4>4.0.37.Final</version.io.netty.netty4>
@@ -388,12 +387,6 @@
                 <groupId>commons-io</groupId>
                 <artifactId>commons-io</artifactId>
                 <version>${version.commons-io.commons-io}</version>
-            </dependency>
-
-            <dependency>
-                <groupId>commons-lang</groupId>
-                <artifactId>commons-lang</artifactId>
-                <version>${version.commons-lang.commons-lang}</version>
             </dependency>
 
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,6 @@
         <relativePath/>
     </parent>
 
-
     <modelVersion>4.0.0</modelVersion>
 
     <name>RESTEasy JAX-RS</name>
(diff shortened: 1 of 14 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/encoding/MessageSanitizerContainerResponseFilter.java
```diff
@@ -0,0 +1,36 @@
...
+@Priority(Priorities.ENTITY_CODER)
+public class MessageSanitizerContainerResponseFilter implements ContainerResponseFilter {
+    @Override
+    public void filter(ContainerRequestContext requestContext,
+                       ContainerResponseContext responseContext) throws IOException {
+
+        if (HttpResponseCodes.SC_BAD_REQUEST == responseContext.getStatus()) {
+            Object entity = responseContext.getEntity();
+            if (entity != null && entity instanceof String) {
+                String tmpEscapedMsg = StringEscapeUtils.escapeXml((String) entity);
```
asoldano: I've a minor (performance) concern here. By looking at the StringEscapeUtils code, in case of xml escape, what ends up being done is a 5 chars replacement (double-quote, ampersand, less-than, greater-than, xml apostrophe). The entity string is scanned linearly. Then we use String#replaceAll for the slash replacement, which builds up a pattern, matches on it and replaces. How about we manually create the Pattern, provide a regexp matching all 6 chars and perform the replacement with that? Alternatively, how about we do the linear string scan in our code and replace the 6 chars in the same pass ?

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/UnrecognizedPropertyExceptionHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+import org.jboss.resteasy.util.HttpResponseCodes;
+
+/**
+ * Address concerns of a possible XSS attack by removing some details of the exception.
```
asoldano: I would add RESTEASY-1485 in the comment here for future reference :-)

rsearls:
I've made the changes requested.
1 of 3 travis CI  envs failed on the first run.  No code changes were made. travis was rerun and all envs passed.

pom.xml
```diff
@@ -390,6 +391,12 @@
             </dependency>
 
             <dependency>
+                <groupId>commons-lang</groupId>
+                <artifactId>commons-lang</artifactId>
```
asoldano: Is comomns-lang still needed now that the escape code is MessageSanitizerContainerResponseFilter ?

asoldano:
Thanks for the changes @rsearls . Please see my comment above on the commons-lang dependency and rebase the PR against master, as it's including stuff that's already merged.

rsearls:
Change and rebase done

asoldano:
Thanks, cherry picked the last 2 commits, merged/rebased them against master and finally squashed them.

PR #959, opened 2016-09-28
[RESTEASY-1513]

ronsigal:
Deprecate resteasy-jackson-provider.

### DIFF
--- a/docbook/reference/en/en-US/modules/Json.xml
+++ b/docbook/reference/en/en-US/modules/Json.xml
@@ -16,9 +16,9 @@
     </para>
     
     <sect1>
-        <title>Using Jackson 1.9.x Outside of Wildfly 10 or EAP 7</title>
-        <para>If you're deploying Resteasy outside of Wildfly 10 or EAP 7,
-        add the resteasy jackson provder to your WAR pom.xml build:</para>
+        <title>Using Jackson 1.9.x Outside of Wildfly</title>
+        <para>If you're deploying Resteasy outside of Wildfly,
+        add the Resteasy Jackson provder to your WAR pom.xml build:</para>
         <programlisting><![CDATA[
 <dependency>
    <groupId>org.jboss.resteasy</groupId>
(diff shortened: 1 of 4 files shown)


docbook/reference/en/en-US/modules/Json.xml
```diff
@@ -1,63 +1,59 @@
...
+    currently deprecated, and will be removed in a release subsequent to 3.1.0.Final. The resteasy-jackson2-provider
+    module is based on Jackson 2. 
     </para>
+    
     <sect1>
-        <title>Using Jackson 1.9.x Outside of JBoss AS7</title>
-        <para>If you're deploying Resteasy outside of JBoss AS7 add the resteasy jackson provder to your WAR
-        pom.xml build</para>
+        <title>Using Jackson 1.9.x Outside of Wildfly 10 or EAP 7</title>
+        <para>If you're deploying Resteasy outside of Wildfly 10 or EAP 7,
```
asoldano: Minor note: I would stay general here and say "WildFly or JBoss EAP", without versions.

ronsigal: Well, I thought of that, and I'm in favor of being as general as possible. I'm not sure if the sentence applies to, say, Wildfly 8 or EAP 6. Or EAP 5. Or Wildfly 17. ;-)

ronsigal: Actually, I'm thinking maybe we shouldn't even mention EAP in the community user guide.

asoldano: Yeah, that makes sense

ronsigal: re: "Yeah, that makes sense"

The versions, or the reference to EAP?

asoldano: I mean completly removing EAP references (and anyway mentioning WildFly without version if that's not strictly required)

ronsigal: ++1

ronsigal: Ok, I removed references to EAP. Also, I referred to Wildfly 8 or Wildfly 9 or above, as appropriate.

PR #960, opened 2016-09-28
Resteasy 1353 distributable

rsearls:
revered to file to use web-fragment_3_0.xsd


asoldano:
hi @rsearls , by looking at the overall changes resulting from the 13 commits above, I don't see anything related to the subject here. Can you please check / rebase against master?

PR #965, opened 2016-10-03
Update dependancies and build

ctomc:
only bit problematic part is resteasy-spring update to jetty 9.
that is why I put it in separate commit.


asoldano:
related jiras:
* https://issues.jboss.org/browse/RESTEASY-1523
* https://issues.jboss.org/browse/RESTEASY-1524
* https://issues.jboss.org/browse/RESTEASY-1526
* https://issues.jboss.org/browse/RESTEASY-1526

PR #969, opened 2016-10-05
[RESTEASY-1512]

ronsigal:
jboss-modules targets just wildfly (no longer AS7) and conforms to
wildfly 10.1.0.Final.


asoldano:
Cool, thanks Ron.

PR #972, opened 2016-10-10
[RESTEASY-1520]

ronsigal:
Updated Users Guide to refer to Wildfly.


docbook/reference/en/en-US/modules/Installation_Configuration.xml
```diff
@@ -579,12 +686,12 @@ public class MyApplication extends Application
 &lt;/servlet-mapping&gt;</programlisting>
         </para>
         <para>
-            Then the value of resteasy-servlet.mapping.prefix must be:
+            Then the value of Resteasy-servlet.mapping.prefix must be:
```
asoldano: Shouldn't this be resteasy.servlet.mapping.prefix instead of Resteasy-servlet.mapping.prefix?

ronsigal: Ah, damn.  "Replace all" mistake. Thought I fixed all of them. Great catch. Thanks.

asoldano:
Merged, thanks!

PR #977, opened 2016-10-27
[RESTEASY-1205] RestEasy-Netty4: ByteBuf leak in ResteasyHttpRequestDecoder

gustavonalle:
https://issues.jboss.org/browse/RESTEASY-1205

### DIFF
--- a/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/NettyHttpRequest.java
+++ b/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/NettyHttpRequest.java
@@ -1,5 +1,7 @@
 package org.jboss.resteasy.plugins.server.netty;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufInputStream;
 import io.netty.channel.ChannelHandlerContext;
 
 import org.jboss.resteasy.core.AbstractAsynchronousResponse;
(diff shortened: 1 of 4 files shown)


AlexKolpa:
This implementation holds onto the `ByteBuf` until the entire response has been closed. I'm pretty sure that once the content has been read, it can be closed. So that would be [right after this line](https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/interception/jaxrs/AbstractReaderInterceptorContext.java#L54).

Unfortunately, there's no easy way to get access to the `NettyHttpRequest` with the currently implementation. One option would be to pass it to the `AbstractReaderInterceptorContext` from `ServerReaderInterceptorContext`.

Preferably you'd want [this commit in netty](https://github.com/netty/netty/commit/c1932a8537b742aaf15a7cfacf9f76ad8239f3c7) to be available, so you'd just need to call `new ByteBufInputStream(buff, true)` in `RestEasyHttpRequestDecoder`, and then an `inputStream.close()` after having read the content in `AbstractReaderInterceptorContext`

gustavonalle:
@AlexKolpa that commit will surely help, thanks for pointing it out. It apparently will be on next Netty 4.1.x version, but in the meantime, holding the ByteBuf until after the response is closed is certainly better than never releasing the ByteBuf

crazyzh1984:
Too complex implemention.
I have a simpler implemention on pull request #998 
commit b8db6234f2b406b6eba3010ffea8e29806173a4a
@asoldano

asoldano:
@crazyzh1984 actually I found the impl here simpler, anyway, if you feel there's a need for and want to provide an additional improvement PR on top this, feel free to open that up. Thanks.

PR #590, opened 2014-10-11
providers/multipart: RESTEASY-754 Support apache-mime4j 0.7.2

crabbkw:
Apache mime4j 0.7.2 has been out for a while now. Update the
multipart provider to support it.

The multipart provider had extended some classes from 0.6 that are
nolonger available in 0.7.2. Therefore a new approach was taken of
copying a few bits of code from mime4j and rewrite small parts.

See Mime4JWorkaround.java, and Mime4jWorkaroundBinaryEntityBuilder
for all such code.

PR #981, opened 2016-11-02
mime4j 0.7.2 update

codylerum:
I've done a little more to try and get #590 up to date.

Unfortunately it looks like mine4j 0.6.0 is bundled in wildfly so I'm not sure how to get the integration tests to a place where they can run.

Right now getting no class def found errors for the newer mime4j classes


codylerum:
This is for https://issues.jboss.org/browse/RESTEASY-754

asoldano:
Superseded by #1890

PR #982, opened 2016-11-02
Fixes RESTEASY-1544

rsertelon:
Moves HttpObjectAggregator after HttpResponseEncoder as told here : https://issues.jboss.org/browse/RESTEASY-1544


rsertelon:
I don't know how to write a good test case, but I think it is more than required as this little modification may have unforeseen impacts...

Do you have any pointers please?

rsertelon:
Closing, not fixing the problem.

PR #983, opened 2016-11-02
Resteasy 1532 synchronous dispatcher exception mapper

rsearls:


providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java
```diff
@@ -175,49 +175,53 @@ public void flush() throws IOException {
...
+      }
+      value = endpoint.modifyBeforeWrite(value);
+      ObjectWriterModifier mod = ObjectWriterInjector.getAndClear();
+      if (mod != null) {
+         writer = mod.modify(endpoint, httpHeaders, value, writer, jg);
+      }
+
+      try {
          writer.writeValue(jg, value);
-      } finally {
```
asoldano: I know this is the point of the PR, leaving the JsonGenerator open for the ExceptionMapper, but is there any side effect in doing this? Shouldn't it be closed anyway, even if later?

ronsigal: I've looked at one of the implementations of JsonGenerator (com.fasterxml.jackson.core.json.UTF8JsonGenerator), and it looks like it has two resources to be released at the end.

1. OutputStream: this doesn't seem to be a problem, since it's created and closed by the servlet container. Well, I just verified that for TJWS, and I suppose it must be true for other servlet containers.

2. UTF8JsonGenerator maintains some buffers, in the form of byte arrays of a few thousand bytes. These are released in UTF8JsonGenerator.close(). Given that the JsonGenerator is a stack variable that passes out of scope when ResteasyJackson2Provider.writeTo() returns, we're probably OK there as well.

Make sense?

asoldano: Sorry for the late reply; my only concern with 2. would be that the close() method actually also forces a flush, which might possibly be required. This said, I've just added a comment on the jira on a possible idea for this issue.

asoldano: I've opened up https://github.com/resteasy/Resteasy/pull/1153

PR #986, opened 2016-11-04
Revert "RESTEASY-1539: Respect Vary headers"

ronsigal:
Reverts resteasy/Resteasy#978

SimpleServerCache$CacheEntry doesn't completely implement ServerCache$Entry: it's missing MultivaluedMap<String, String> getVaryHeaders();

PR #987, opened 2016-11-05
Revert "Revert "RESTEASY-1539: Respect Vary headers""

ronsigal:
Reverts resteasy/Resteasy#986: #978  was OK.


ronsigal:
It looks like SimpleServerCache was restore after #978 was submitted.

PR #990, opened 2016-11-09
[RESTEASY-1424] RESTEasy-client module should contains dependencies f…

xstefank:
…or important RESTEasy providers

JBEAP: https://issues.jboss.org/browse/JBEAP-5081
JIRA: https://issues.jboss.org/browse/RESTEASY-1424

As it is described in the comments of above issues the removal of the jackson2 provider is necessary because of the usage of com.fasterxml dependencies which are not supported by RESTEasy, so they cannot be supplied to the user without knowledge. After the deprecation and following privatization of the com.fasterxml APIs the jackson2 provider dependency can be added to the resteasy-client module.

### DIFF
--- a/profiling-tests/pom.xml
+++ b/profiling-tests/pom.xml
@@ -30,6 +30,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-jackson2-provider</artifactId>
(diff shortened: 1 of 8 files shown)


marekkopecky:
This PR looks good to me

xstefank:
After the discussion we have decided that it would be better to not change the behavior of the resteasy-client for now.

PR #994, opened 2016-11-14
Add wildfly-elytron dependency to make arquillian work with Remoting 5

kanovotn:


asoldano:
@kanovotn how have you reproduced failures / forced remoting 5 usage?

PR #997, opened 2016-11-16
[RESTEASY-1547] Profile to enable resteasy-legacy in container and as…

kanovotn:
… dependency


asoldano:
@kanovotn there seems to be a problem with this PR, all test jobs are failing consistently, can you have a look please?

kanovotn:
Yes, I need to work on this more, thanks.

PR #999, opened 2016-11-17
[RESTEASY-1553] add check for html content-type and reverted test str…

rsearls:
…ings as needed


rsearls:
Can't get all 3 travis-ci envs to complete successfully in a single run.  3 separate runs and 3 different envs fail.
None of the failures are do to my code changes.

PR #1000, opened 2016-11-21
Forward Compatibility profile excludes legacy integration tests

kanovotn:
* moves forward compatibility profile to the main testsuite pom
* excludes legacy integration tests module from the profile (that will be needed to added later When Wildfly/EAP with the resteasy 3.1 is released).

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -10,14 +10,6 @@
     <description/>
     <modelVersion>4.0.0</modelVersion>
     <packaging>pom</packaging>
-
-    <modules>
-        <module>arquillian-utils</module>
-        <module>unit-tests</module>
-        <module>legacy-integration-tests</module>
-        <module>integration-tests</module>
-        <module>integration-tests-spring</module>
-    </modules>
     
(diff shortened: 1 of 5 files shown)


testsuite/pom.xml
```diff
@@ -61,6 +53,41 @@
...
+            <id>forward.compatibility</id>
+            <activation>
+                <property>
+                    <name>forward.compatibility</name>
+                </property>
+            </activation>
+            <modules>
+                <module>arquillian-utils</module>
+                <module>unit-tests</module>
+                <!-- Uncomment module below when EAP/Widlfly with Resteasy 3.1 is released -->
```
kanovotn: When EAP/Wildfly is released with 3.1.0.Final, we will use this newly released container as "old" version to run against new versions of client and testsuite. Now the forward compatibility tests are run against EAP 7.0.0.GA which doesn't have resteasy-legacy module so the tests cannot pass. So, no, this PR is not required only until Resteasy Final. 
Please correct me, if it doesn't make sense to you.

PR #1023, opened 2016-12-14
RESTEASY-1566: Provide support for multi valued query param

NicoNes:
This pull request is about providing support for multivalued query param in addition to existant behaviour (singlevalued query param) using `javax.ws.rs.ext.ParamConverter` mechanism.


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/PathParamInjector.java
```diff
@@ -109,30 +109,21 @@ else if (pathSegmentList)
...
+	       return extractor.extractValues(null);
+	    }
+		return extractor.extractValue(null);
+	 }
+	 if (extractor.isInjectableAsCollectionOrArray())
+	 {
+	    return extractor.extractValues(list);
+	 }
+	 return extractor.extractValue(list.get(list.size() - 1));
    }
```
ronsigal: I don't see anything other than cosmetic changes here, and I don't think they constitute improvements. For example, taking the second half of the method out of the scope of an else, while currently equivalent semantically, invites the possibility of someone inserting code, some day, that doesn't leave the method and falls through to the former else block.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -65,7 +65,7 @@ public StringParameterInjector(Class type, Type genericType, String paramName, C
       initialize(type, genericType, paramName, paramType, defaultValue, target, annotations, factory);
    }
 
-   public boolean isCollectionOrArray()
+   public boolean isInjectableAsCollectionOrArray()
    {
```
ronsigal: Not a good idea to change the name of a public method. Even if it is, in principle, an implementation method, it is still visible outside of Resteasy.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -80,40 +80,39 @@ protected void initialize(Class type, Type genericType, String paramName, Class
...
+     
+      if (!baseType.isPrimitive())
       {
-         if (genericType != null && genericType instanceof ParameterizedType)
+         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
+         if (paramConverter != null) 
          {
-            ParameterizedType zType = (ParameterizedType) genericType;
-            baseType = Types.getRawType(zType.getActualTypeArguments()[0]);
-            baseGenericType = zType.getActualTypeArguments()[0];
```
ronsigal: This is interesting. Really, I don't understand why baseGenericType is getting modified here. Given that there may have been some justification that I don't see, I appreciate that you added the call on line 90 as an addition to, rather than a replacement of, the original call.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -80,40 +80,39 @@ protected void initialize(Class type, Type genericType, String paramName, Class
...
+	        	 return;
+	         }
          }
-      }
-      if (!baseType.isPrimitive())
-      {
-         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
-         if (paramConverter != null) return;
 
          unmarshaller = factory.createStringParameterUnmarshaller(baseType);
```
ronsigal: Maybe this call should get the same "double" treatment that factory.getParamConverter() gets.

ronsigal:
I think this is a nice and worthwhile change. I have a few coding quibbles - see comments in the code. Thank you.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/PathParamInjector.java
NicoNes: Right it's just cosmetic, no specific improvments here. I don't get your point about the second half of the method.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: Oups... bad idea right 👎

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: To my mind baseGenericType is modified from it's original value in case of we are dealing with a collection because in this case the `factory.getParamConverter(...)` must be in charge of finding a converter for the data type held in the collection and not for the collection type itself any longer since the call to this method on line 90 return null.
Whatever the reason as you said if prefer to add call on line 90 rather than changing everything. However I've just noticed that I did it wrong cause I've changed the original behavior...
 The call from line 88 to 94 should  have been
`     if (!baseType.isPrimitive())
      {
         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
         if (paramConverter != null) 
         {
        	 return;
}
}`
Then remove code from line 110 to 114 and put it in a `if ` block like following at line 116:
`         if (!baseType.isPrimitive())
         {
	         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
	         if (paramConverter != null) 
	         {
	        	 return;
	         }
     unmarshaller = factory.createStringParameterUnmarshaller(baseType);`

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: Which call exactly ?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: [Put > at beginning of line for a blockquote.]

re: "finding a converter for the data type held in the collection and not for the collection type itself any longer since the call to this method on line 90 return null."

Right. But before your changes there was no call like the one on line 90, which uses baseGenericType before it is modified.
re: "The call from line 88 to 94"

It looks identical to the original ...

re: "Then remove code from line 110 to 114"

I don't see the difference. It's all inside a "if(!baseType.isPrimitive()" test alreade

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/PathParamInjector.java
ronsigal: re: the second half of the method"

For example, consider changing

>   if (x == y)
>   {
>      return z;
>   }
>   else
>   {
>      z = 17;
>   }

to

>   if (x == y)
>   {
>      return z;
>   }
>   z = 17;

They're semantically equivalent, right? But suppose someone comes along and changes the second version to


>   if (x == y)
>   {
>      z = 19;
>   }
>   z = 17;

Now, control falls through to z = 17;, even if x == y. That wouldn't have happened before.

Not likely, but still ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: Believe me, I've learned the hard way. ;-)

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: Sorry if I wasn't clear. I meant, maybe 

>   unmarshaller = factory.createStringParameterUnmarshaller(baseType);

should get called both before and after baseType gets modified ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/PathParamInjector.java
NicoNes: Ok I get it. It's actually better to use the 'else' block to keep the control on flow. Thank you 👍

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: re: "Right. But before your changes there was no call like the one on line 90, which uses baseGenericType before it is modified."

Yes and that's why the user registered `javax.ws.rs.ext.ParamConverterProvider` was never called with the collection genericType since when it is a collection the `baseGenericType` is modified to the data type held by the collection.

re: "The call from line 88 to 94" and "Then remove code from line 110 to 114"

In the original code the call to `factory.getParamConverter(baseType, baseGenericType, annotations);` was only invoked if `!baseType.isPrimitive()`.
In my version the first call to `factory.getParamConverter(baseType, baseGenericType, annotations);` at line 90 is done only if `!baseType.isPrimitive()` but the second one at line 110 will be invoked even if 'baseType.isPrimitive()' which is different from the original behavior when dealing with a collection.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: Yes actually the same "double" treatment could be done for all calls from line 17 to the end. It will allow users to provide a way to convert `String` type to any type subclassing `List`, `Set ` or `SortedSet` before fallback on default implementation  `ArrayList`, `TreeSet ` and `HashSet ` and focusing on data held by the collection. 
The  "first" treatement will  end if one of the following is true :
> `paramConverter != null`
> `unmarshaller != null`
> `converter != null`
> `delegate != null` in case of `paramType.equals(HeaderParam.class)`
> `valueOf != null`

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: re: "but the second one at line 110"

Hmmm, I don't see that. I see

>      if (!baseType.isPrimitive())
>      {
>          paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
>          ...
>	         if (genericType != null && genericType instanceof ParameterizedType)
>            ...
>	         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
>            ...
>      }

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: re: "It will allow users to provide a way to convert String type to any type subclassing List, Set or SortedSet before fallback on default implementation ArrayList, TreeSet and HashSet"

So, use the actual type found in the resource method. Hmmm, I wonder why the type is getting "normalized" now ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: >    
      if (!baseType.isPrimitive())
      {
         //This call is ok baseType is not a primitive
         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
         ...
	         if (genericType != null && genericType instanceof ParameterizedType)
	         {
	            ParameterizedType zType = (ParameterizedType) baseGenericType;
                // baseType is modified here so it can become a primitive (seem's technicaly impossible to me in this case but we never know...)
	            baseType = Types.getRawType(zType.getActualTypeArguments()[0]);
	            baseGenericType = zType.getActualTypeArguments()[0];
	         }
          ...
             // If baseType has been modified to a primitive it's KO, we change the behavior of the original code
	         paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);
	      ...
       }

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
NicoNes: By "normalized" do you mean convert String to ArrayList, TreeSet or HashSet ?

If it is the case I think this conversion is done like that because of the spec about conversion between String and  type of any resource method argument:

> 1.Be a primitive type
2.Have a constructor that accepts a single String argument
3.Have a static method named valueOf or fromString that accepts a single String argument (see, for example, Integer.valueOf(String))
4.Have a registered implementation of javax.ws.rs.ext.ParamConverterProvider JAX-RS extension SPI that returns a javax.ws.rs.ext.ParamConverter instance capable of a "from string" conversion for the type.
5.Be List<T>, Set<T> or SortedSet<T>, where T satisfies 2, 3 or 4 above. The resulting collection is read-only.

But the problem with current code is that in case of a collection it directly jump to point 5, even if user provides a Collection (List Set or SortedSet) with either:

> 1.A constructor that accepts a single String argument
2. A static method named valueOf or fromString that accepts a single String argument (see, for example, Integer.valueOf(String))
3. A registered implementation of javax.ws.rs.ext.ParamConverterProvider JAX-RS extension SPI that returns a javax.ws.rs.ext.ParamConverter instance capable of a "from string" conversion for the type.

By the way about point 5 I'm not sure that resulting collection is read only in Resteasy impl (see `StringParameterInjector.extractValues(List<String> values)`)

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: So the second call

>paramConverter = factory.getParamConverter(baseType, baseGenericType, annotations);

on line 110 is in the scope of 

>if (!baseType.isPrimitive())

on line 88.

NicoNes: Yep but `baseType` is modified at line 102 (see my comment on this line). So `baseType` can be changed change to a primitive so we have to check it again before the line 110 .

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: Which version of the JAX-RS spec are you looking at? Version 2.0 Final Release says

1. Types for which a ParamConverter is available via a registered ParamConverterProvider . See
Javadoc for these classes for more information.
2. Primitive types.
3. Types that have a constructor that accepts a single String argument.
4. Types that have a static method named valueOf or fromString with a single String argument
that return an instance of the type. If both methods are present then valueOf MUST be used unless
the type is an enum in which case fromString MUST be used 1 .
5. List<T> , Set<T> , or SortedSet<T> , where T satisfies 3 or 4 above.

ronsigal: re: "But the problem with current code is that in case of a collection it directly jump to point 5, even if user provides a Collection (List Set or SortedSet) with either:"

Yes, I think you're right. It just looks like a mis-reading of the spec.

NicoNes: Sorry I said "spec" but I was looking at the Javadoc of any of the following annotations:
@MatrixParam, @QueryParam, @PathParam, @CookieParam, @HeaderParam

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
ronsigal: Hey Nicolas,

I've been digging a little deeper and re-thinking this issue. I think I understand now why StringParameterInjector.initialize() is written as it is. Basically, it expects the underlying URL parsing mechanisms to handle Collections, and then the various string translation mechanisms to handle the element type. For example, given the resource method

>		@GET
>		@Path("dates")
>		@Produces("text/plain")
>		public String dates(@QueryParam("q") List<Date> q)
>		{
>			for (Date d : q)
>			{
>				System.out.println(d);
>			}
>			return "ok";
>		}

running

>		String date1 = new Date(117, 1, 1).toString();
>		String date2 = new Date(117, 2, 2).toString();
>		ResteasyClient client = new ResteasyClientBuilder().build();
>		Invocation.Builder request = client.target("http://localhost:8081/dates?q=" + date1 + "&q=" + date2).request();
>		Response response = request.get();

will produce

Wed Feb 01 00:00:00 EST 2017
Thu Mar 02 00:00:00 EST 2017

That is, "...?q=a&q=b" can be turned into a list or a set, so maybe we don't really need another syntax.

-Ron

NicoNes: Hi Ron,

Yes this is it and that's the kind of tests done in `org.jboss.resteasy.test.resource.param.ParamConverterTest.testIt()` unit tests too.

I'm not saying that another synthax is really mandatory (for sure it's not). 
Just that it will be nice to provide support for it since it does not require a lot of code modification to handle it so that users can choose wheiter doing "...?q=a&q=b" or "...?q=a,b".

 The "...?q=a&q=b" is for sure the cleanest syntax and will be the one supported by default (just like now). But if users want to take advantage of the second one  by fulfilling JaxRS spec conditions listed previously why not ? 

More, according to the specs users are free to regsiter a `ParamConverter` to handle `List<Date>` for example. So they can use any syntax since they are able to parse it into `List<Date>`

ronsigal: Ok, yeah, I'm not against adding this change. You want to update with the changes we've discussed?

NicoNes: Yes, sure I'll do it 👍

NicoNes:
Hi Ron,

Sorry for the delay. I created a new PR https://github.com/resteasy/Resteasy/pull/1093 including changes we discussed about.
Hope I did everything well.

PR #1029, opened 2016-12-21
[RESTEASY-1559] Incorrectly matching resources "/list" -> PathParam "…

kanovotn:
…/{id}"


asoldano:
hi @kanovotn ; thanks for the test here, however in this case I'd rather wait for somebody to look at the issue and propose a potential fix before adding the testcase. I suspect it might not be that easy to achieve the behavior expected in the jira description and this could possibly even be covered by certification tests (TCK).

kanovotn:
@asoldano Thanks for feedback. So in this case I will close this PR and just link my branch with test into the jira, in case it will be needed, is it ok? Do you prefer this way in all cases? For example in PR#1030 I did the same:)

asoldano:
@kanovotn sure, fine, close for now and we'll see later. thanks

PR #1033, opened 2017-01-04
[RESTEASY-1574]: org.jboss.resteasy.spi.FactoryFinder does not close resources.

panossot:
[RESTEASY-1574] : https://issues.jboss.org/browse/RESTEASY-1574
[JBEAP-7988] : https://issues.jboss.org/browse/JBEAP-7988

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/FactoryFinder.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/FactoryFinder.java
@@ -3,6 +3,7 @@ package org.jboss.resteasy.spi;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.security.AccessController;
@@ -103,9 +104,10 @@ final class FactoryFinder {
        ClassLoader classLoader = getContextClassLoader();
 
        String serviceId = "META-INF/services/" + factoryId;
+       BufferedReader rd = null;
+       InputStream is = null;
        // try to find services in CLASSPATH
        try {
-           InputStream is;
            if (classLoader == null) {
                is = ClassLoader.getSystemResourceAsStream(serviceId);
            } else {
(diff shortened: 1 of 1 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/FactoryFinder.java
```diff
@@ -113,20 +115,26 @@ static Object find(final String factoryId, final String fallbackClassName) throw
...
            LOGGER.log(Level.FINER, "Failed to load service " + factoryId + " from " + serviceId, ex);
-       }
-
-
+       } finally {
+           try {
+	       if (rd != null)
+		   rd.close();
+	   } catch (IOException ex) {
+	       ex.printStackTrace();
```
asoldano: Can we please have something better than the ex.printStackTrace()? (use the logger instance that's available). Also please see similar additions below.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/FactoryFinder.java
```diff
@@ -157,24 +172,30 @@ static Object find(final String factoryId, final String fallbackClassName) throw
...
-         }
+         } finally {
+           try {
+	       if (rd != null)
+		   rd.close();
+	   } catch (IOException ex) {
+               LOGGER.log(Level.FINER, "Failed to close  BufferedReader.", ex);
+	   }
+	 }
        }
```
ronsigal: Same comment as above. I would check is in the finally clause.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/FactoryFinder.java
```diff
@@ -113,20 +115,26 @@ static Object find(final String factoryId, final String fallbackClassName) throw
...
-
+       } finally {
+           try {
+	       if (rd != null)
+		   rd.close();
+	   } catch (IOException ex) {
+	       LOGGER.log(Level.FINER, "Failed to close  BufferedReader.", ex);
+	   }
+	}
+
```
ronsigal: If new InputStreamReader(is, "UTF-8") fails, then rd would be null but is would still be open. I would check is as well as rd in the finally clause.

PR #1040, opened 2017-01-19
[RESTEASY-1559] False alarm ambiguity for /foo and /{id} -> long

treblereel:
JIRA: https://issues.jboss.org/browse/RESTEASY-1559

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
@@ -446,12 +446,20 @@ public class SegmentNode
       }
       Collections.sort(sortList);
       SortEntry sortEntry = sortList.get(0);
-      if (targetMethods != null && targetMethods.size() > 1)
+      if (targetMethods != null && targetMethods.size() > 1 && !checkExpressionsUnique(sortList))
       {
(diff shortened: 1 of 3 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
```diff
@@ -446,14 +446,22 @@ else if (!consumeMatch)
...
       request.setAttribute(RESTEASY_CHOSEN_ACCEPT, sortEntry.getAcceptType());
       return sortEntry.match;
    }
 
+   private boolean checkExpressionsUnique(List<SortEntry> sortList){
+      List<String> list = new ArrayList<>();
+      for(SortEntry elm: sortList){
+         list.add(elm.match.expression.getRegex());
+      }
+      return new HashSet<>(list).size() == list.size();
```
asoldano: @treblereel , do we really need the 'list' instance here? Can't we build up the HashSet directly and compare its size with sortList size? Or even better, at that point can't we simply return the first time that the HashSet#add method invocation returns false?

ronsigal:
I'm not really in favor of this issue. See comment on https://issues.jboss.org/browse/RESTEASY-1559.

treblereel:
retest this please

PR #1042, opened 2017-01-20
[RESTEASY-1342] RoleBasedSecurityFilter doesn't allow customizing res…

kanovotn:
…ponse


kanovotn:
Updated the PR, the tests pass now.

PR #1044, opened 2017-01-20
1586

ronsigal:


asoldano:
I've merged the first commit (i18n stuff) just before the 3.0.20.Final release. Merged the other two (after having squashed them together) just now. Thanks.

PR #1045, opened 2017-01-23
[RESTEASY-1591] Change catch Throwable to IOException in ClientInvocation

driev:
Currently every exception in .invoke() is wrapped in a ResponseProcessingException or it is mapped to specific HTTP exception in handleErrorStatus.

This change allows a Client Request/Repsonse Filters to throw custom exceptions.

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/interceptor/ResponseFilterCustomExceptionTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/interceptor/ResponseFilterCustomExceptionTest.java
@@ -1,0 +1,65 @@
+package org.jboss.resteasy.test.interceptor;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.resteasy.spi.ResteasyProviderFactory;
+import org.jboss.resteasy.test.interceptor.resource.CustomException;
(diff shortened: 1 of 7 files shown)


asoldano:
Hi, would you mind creating a JIRA issue for this, explaining why this is a problem with a code example and also provide a testcase for the fix?
Thanks

driev:
Sure!

driev:
I added it in Jira as an "enhancement", since it is not really a bug? -> https://issues.jboss.org/browse/RESTEASY-1591

Adding tests now.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/interceptor/ResponseFilterCustomExceptionTest.java
```diff
@@ -0,0 +1,65 @@
...
+     * @tpTestDetails Use ClientResponseFilter
+     * @tpSince RESTEasy 3.1.0
+     */
+    @Test(expected = CustomException.class)
+    public void testThrowCustomException() throws Exception {
+        ResteasyProviderFactory factory = ResteasyProviderFactory.getInstance();
+        factory.register(ThrowCustomExceptionResponseFilter.class);
+        client.register(ThrowCustomExceptionResponseFilter.class);
+        client.target(generateURL("/testCustomException")).request().post(Entity.text("testCustomException"));
+    }
```
driev: Before, the exception would be a RequestProcessingException, with the CustomException inside.

asoldano:
@driev ah, I forgot to tell that we require that all contributions be made under the terms of the ASL
Do you agree to these terms?

driev:
Yes, I agree.

asoldano:
Thanks

PR #1050, opened 2017-01-27
[RESTEASY-1590] Create a new SSLEngine for each connection

stiwarta:
Bug fix: Only the first client requests are successful, any other requests issued by a new connection timeout. 
Using wireshark, it appears that the handshake is performed only at the first connection. I checked the code and indeed resteasy keep using the same SSL session for all connections.
Fix: use a new SSLEngine for each new connection.

Netty documentation about using the SSLHandler and restarting a session: 
https://netty.io/4.0/api/io/netty/handler/ssl/SslHandler.html =

Jira issue: https://issues.jboss.org/browse/RESTEASY-1590


asoldano:
@stiwarta , we require that all contributions be made under the terms of the ASL
Do you agree to these terms?

stiwarta:
@asoldano  Yes I do agree :). Do I need to attach something to my PR or this  comment is enough ?

asoldano:
@stiwarta , it's fine like this, thanks. We'll test your PR later and merge.

PR #1052, opened 2017-01-30
Add test dependency for hibernate validation

kanovotn:


asoldano:
Hi @kanovotn , why is this change needed?

kanovotn:
Hi @asoldano , this is needed because when I run the testsuite with EAP dependencies, org.jboss.resteasy.test.client.proxy.ResponseObjectTest fails to create InMemoryClientExecutor instance without this dependency. Same dependency is present in https://github.com/resteasy/Resteasy/blob/master/testsuite/legacy-integration-tests/pom.xml#L339 .

PR #1056, opened 2017-01-31
[RESTEASY-1580] Prevent NPE

asoldano:


asoldano:
Use confirmed the patch fixes his regression, merging.

PR #1057, opened 2017-01-31
[RESTEASY-1534][RESTEASY-1577] Fix ClientWebTarget to always create R…

asoldano:
…ESTEasy impl of UriBuilder or convert to it when casting is not possible


asoldano:
Got positive feedback on https://issues.jboss.org/browse/RESTEASY-1577, merging.

PR #1058, opened 2017-02-01
[RESTEASY-1467] Workaround for deleteOnExit memory leak

asoldano:
Proposed fix to avoid the leak; unless a specific impl of StorageProvider is requested, use a modified version of the TempFileStorageProvider that does not flag files as to be deleted on exit. Files end up being delete when the message body is disposed.


ronsigal:
Why not just inherit from TempFileStorageProvider and override createStorageOutputStream()?

asoldano:
@ronsigal , AFAIR the problem in doing that would be that TempFileStorageOutputStream and TempFileStorage are not visible (they're private inner classes), so they'd still have to be duplicated... and that's pretty much the same as defining a whole new storage provider.

ronsigal:
@asoldano Ah, yeah, didn't notice that.

PR #1060, opened 2017-02-03
[RESTEASY-1563] Enabling test by kanovotn and adding charset specific…

asoldano:
…ation in RoleBasedSecurityFilter


asoldano:
Merging after having checked with @kanovotn

PR #1061, opened 2017-02-04
[RESTEASY-1507] Adding specific catch blocks in ResteasyHandlerMappin…

asoldano:
…g and fixing SpringMvcHttpResponseCodesTest to properly cover all scenarios

### DIFF
--- a/resteasy-spring/src/main/java/org/jboss/resteasy/plugins/spring/i18n/Messages.java
+++ b/resteasy-spring/src/main/java/org/jboss/resteasy/plugins/spring/i18n/Messages.java
@@ -48,6 +48,12 @@ public interface Messages
    @Message(id = BASE + 40, value = "Interrupted while starting up Jetty")
    String interruptedStartingUpJetty();
    
+   @Message(id = BASE + 42, value = "Not allowed")
+   String notAllowed();
+   
+   @Message(id = BASE + 43, value = "Not supported")
+   String notSupported();
+   
    @Message(id = BASE + 45, value = "%s is not initial request.  Its suspended and retried.  Aborting.")
    String pathNotInitialRequest(String path);
 
@@ -59,6 +65,9 @@ public interface Messages
 
    @Message(id = BASE + 60, value = "RESTeasy Registry is null, do you have the ResteasyBootstrap listener configured?")
    String registryIsNull();
+   
+   @Message(id = BASE + 62, value = "The requested media type is not acceptable.")
+   String requestedMediaNotAcceptable();
    
    @Message(id = BASE + 65, value = "ResourceFailure: %s")
    String resourceFailure(String message);
--- a/resteasy-spring/src/main/java/org/jboss/resteasy/springmvc/ResteasyHandlerMapping.java
+++ b/resteasy-spring/src/main/java/org/jboss/resteasy/springmvc/ResteasyHandlerMapping.java
@@ -99,15 +99,15 @@ public class ResteasyHandlerMapping implements HandlerMapping, Ordered, Initiali
          LogMessages.LOGGER.error(Messages.MESSAGES.resourceNotFound(e.getMessage()), e);
       }
       catch (NotAcceptableException na) {
-         requestWrapper.setError(HttpResponseCodes.SC_NOT_ACCEPTABLE, "The requested media type is not acceptable.");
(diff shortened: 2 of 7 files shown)


asoldano:
@kanovotn please have a look at this PR, as you prepared the testcase for this issue.

@kanovotn , @ronsigal : note I was not able to have the ResteasyDeployment configured with securityEnabled = true since the beginning to have the RoleBasedSecurityFeature installed and test HTTP 403 scenario using @RolesAllowed annotation.

ronsigal:
* re: RoleBasedSecurityFeature

What has changed so that "resteasy.role.based.security" doesn't work? I assume it used to work with the SpringMVC Dispatcher servlet, though I don't see how, that being one of the servlets that eliminates Resteasy configuration ...

* The new messages in ResteasyHandlerMapping should be i18n'd.

asoldano:
@ronsigal , as for the resteasy.role.based.security, I don't know. We should perhaps check if it was really working in the past. If that's the case, this could be related to one of the changes that we've "recently" applied on the spring integration.

Right, those messages have to be i18n'd.

ronsigal:
re: "We should perhaps check if it was really working in the past."

I'm running SpringMvcHttpResponseCodesTest on Branch_3_0, and what I see is that authentication is being handled in undertow; in particular, io.undertow.servlet.handlers.security.ServletAuthenticationCallHandler in io.undertow:undertow-servlet calls io.undertow.security.impl.SecurityContextImpl in io.undertow:undertow-core.

Hmmm. Now I'm wondering, if, given that undertow handles requests in Wildfly, maybe this is independent of spring and the whole "dueling dispatchers" issue. I.e., maybe "resteasy.role.based.security" is NEVER necessary in Wildfly. If so, then there's an argument for running at least some tests in Jetty or some such alternative servlet container.

ronsigal:
Well, I commented out "resteasy.role.based.security" in web.xml for org.jboss.resteasy.test.security.BasicAuthTest in resteasy-integration-tests, and got four "expected:<403> but was:<200>" failures, so there's something going on. To be continued.

ronsigal:
When BasicAuthTest runs, undertow goes directly to org.jboss.resteasy.plugins.server.servlet.HttpServlet30Dispatcher. 

@asoldano @kanovotn Do you understand this?

ronsigal:
re: "Right, those messages have to be i18n'd."

@asoldano I just pushed a commit to your branch RESTEASY-1507 with the i18n'd messages. I realized after the fact that I should have done it on a new branch. Sorry about that.

asoldano:
@ronsigal wow, np, I just didn't know you could push a commit to someone else branch O_o
As for the other comments above, I'm having a look soon.

ronsigal:
@asoldano Yeah, I didn't know that either. ;-) There must be a flag you can set prevent it.

ronsigal:
@asoldano: Settings -> Branches -> Protected Branches

asoldano:
Going back to the former discussion above, let me remind that authentication != authorization and the javax.annotation.security.* annotations are meant for controlling fine grained authorization that can't be achieved through web.xml. So that why this goes beyond Undertow.

re "When BasicAuthTest runs, undertow goes directly to org.jboss.resteasy.plugins.server.servlet.HttpServlet30Dispatcher."
That's because the web.xml for the deployment in BasicAuthTest does not define any servlet-class, so the WildFly integration code rewrites the web metadata in order to use RESTEasy HttpServlet30Dispatcher .

asoldano:
@ronsigal , can we say that this PR is fine? The problem with not being able to rely on resteasy.role.based.security option is really... a test problem. I propose creating an additional jira for tracking additional investigation of the problem and merging this PR. If you agree, please submit review approval here. Thanks.

ronsigal:
re: "the WildFly integration code rewrites the web metadata in order to use RESTEasy HttpServlet30Dispatcher"

Ah, I didn't know that. Thanks.

ronsigal:
re: "If you agree, please submit review approval here."

Sure, I don't think there's any problem with this PR itself. +1

asoldano:
@ronsigal you might want to go to the "File changed" tab above and then click on the green "Review Changes" button to formally approve the review (as I marked you as reviewer for this PR ;-) )

PR #1062, opened 2017-02-06
[RESTEASY-1225]

ronsigal:
Correcting ResteasyProviderFactory's handling of the context map stack.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
@@ -440,7 +440,12 @@ public class ResteasyProviderFactory extends RuntimeDelegate implements Provider
 
    public static void pushContextDataMap(Map<Class<?>, Object> map)
    {
-      contextualData.setLast(map);
+      contextualData.push(map);
+   }
+   
+   public static void popContextDataMap()
+   {
+      contextualData.pop();
    }
 
    public static Map<Class<?>, Object> getContextDataMap()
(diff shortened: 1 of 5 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -440,7 +440,12 @@ public void insertInterceptorPrecedenceBefore(String before, String newPrecedenc
 
    public static void pushContextDataMap(Map<Class<?>, Object> map)
    {
-      contextualData.setLast(map);
+      contextualData.push(map);
+   }
+   
+   public static void popContextDataMap()
```
asoldano: Isn't this the same of the existing removeContextDataLevel() method?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -440,7 +440,12 @@ public void insertInterceptorPrecedenceBefore(String before, String newPrecedenc
 
    public static void pushContextDataMap(Map<Class<?>, Object> map)
    {
-      contextualData.setLast(map);
+      contextualData.push(map);
```
asoldano: Just to be sure we're not missing anything here, is the scenario you fixed in SynchronousDispatcher the only one in which calling contextualData.push(..) instead of contextualData.setLast(..) makes any difference? Just want to be sure we're not introducing a bug in other scenarios (as one would probably have to call pop).

ronsigal: It looks like AbstractAsynchronousResponse is the only place that calls that method.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
ronsigal: Ah, yes, that's true.  I can fix that.

asoldano:
Superseded by https://github.com/resteasy/Resteasy/pull/1064

PR #1063, opened 2017-02-06
3.0 1225

ronsigal:
Correcting ResteasyProviderFactory's handling of the context map stack.


ronsigal:
It looks like I included some files from RESTEASY-1565. I'll let this PR run, but replace it.

asoldano:
@ronsigal, I assume I can simply pick the commit from master and apply it to branch 3.0 too, right?

asoldano:
Well, the testsuite is passing that way, so I'm picking the commit from master and closing this, thanks.

PR #1066, opened 2017-02-07
Try giving Travis CI more time for runs

asoldano:


asoldano:
I'm going to merge this: it does not prevent builds from failing because of timeouts (there has to be a reason for that, either in travis or in how some tests behave with travis), but at least each run will last 30 mins max, while before it would have taken the time to reach the test that eventually goes into a timeout issue + 10 mins, which is usually more than 30min.

PR #1078, opened 2017-02-18
[RESTEASY-1569] changed calculation to get proper path substring. Add…

rsearls:
…ed tests for regex capturing groups

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
@@ -101,7 +101,23 @@ public class SegmentNode
                }
                else
                {
-                  String substring = path.substring(0, length);
+                  // must find the end of the matched pattern
+                  // and get the substring from 1st char thru end
+                  // of matched chars
+                  Pattern p = expression.getPattern();
+                  Matcher m = p.matcher(path);
+                  m.region(start, path.length());
+                  String substring = path;
+                  while(m.find()) {
+                     String endText = m.group(m.groupCount());
+                     if (endText != null && !endText.isEmpty()) {
+                        int indx = path.indexOf(endText, length);
+                        if (indx > -1) {
+                           substring = path.substring(0, indx);
+                        }
+                     }
+                  }
+
                   uriInfo.pushMatchedPath(substring);
                   uriInfo.pushMatchedURI(substring);
                }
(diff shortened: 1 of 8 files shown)


liweinan:
I've played with the patch on my computer for two days and the PR looks good to me. @ronsigal, could you please help to double check? Because this is a change in `SegmentNode` and the different logic may have some affects that I missed.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
```diff
@@ -101,7 +101,23 @@ public ResourceInvoker match(HttpRequest request, int start)
...
                {
-                  String substring = path.substring(0, length);
+                  // must find the end of the matched pattern
+                  // and get the substring from 1st char thru end
+                  // of matched chars
+                  Pattern p = expression.getPattern();
+                  Matcher m = p.matcher(path);
+                  m.region(start, path.length());
+                  String substring = path;
+                  while(m.find()) {
```
ronsigal: Should this be a while? Wouldn't the final (/.+)? gobble up the rest of the path, so that the entire path gets matched and there can never be a second successful m.find()?  Well, maybe there's some strange regex such that that is not the case, but, even so, wouldn't you want to take the first such match?

liweinan: @ronsigal Good idea. I'm writing some more complex regexp expressions in my local machine, let me give my opinion soon.

asoldano: @liweinan any update here? This PR is quire relevant / critical.

liweinan: @asoldano @ronsigal @rsearls I've done all my local testings (include resteasy-js testings) and didn't find any problem with this patch. So I suggest we merge it.

asoldano: @liweinan thanks.
As for what @ronsigal is saying... I don't know, perhaps Rebecca can/want to add something. What I can say is that if we really expect multiple passes in the while loop, we should not keep on re-assigning the 'substring' variable to new String objects, we should instead store the latest indx value and eventually compute the substring using it at the end of the loop. Anyway, this can possibly be optimized later.

rsearls: Given I tested with relatively complex strings and @liwainan did so as well.  This fix seems to address the those cases. I suggest merging this fix and revisit the issue if a user has a use case that breaks this.

asoldano: OK, fine

PR #1082, opened 2017-02-20
Revisit profiles activation to avoid relying on Maven 'activeByDefaul…

asoldano:
…t' which causes problem when needing to have multiple profiles enabled at the same time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,9 +77,11 @@
 
     <profiles>
         <profile>
-            <id>travis-ci</id>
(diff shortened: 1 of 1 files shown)


asoldano:
@kanovotn please have a look (just in case, the profile for eap should still be working fine).

asoldano:
Btw, I had to apply this locally for starting building using JDK 9 (which requires having an additional profile to be enabled only when jdk 9 is on)

PR #1086, opened 2017-02-21
[RESTEASY-1616] added section 45.3, JAX-RS with Spring MVC

rsearls:


asoldano:
@rsearls do we have 'jaxrs' and 'jax-rs' elsewhere in the doc in place of of 'JAX-RS' ? If not, I'd use 'JAX-RS' as that's the correct way of referring to the technology.

rsearls:
Text changed to JAX-RS in all references

PR #1093, opened 2017-03-05
RESTEASY-1566 (2): Provide support for multi valued query param

NicoNes:
This pull request is about providing support for multivalued query param in addition to existant behaviour (singlevalued query param) using javax.ws.rs.ext.ParamConverter mechanism. End of https://github.com/resteasy/Resteasy/pull/1023


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -301,7 +301,14 @@ else if (isCollection)
          {
             collection.add(extractValue(str));
          }
-         return collection;
+     	if (ArrayList.class.equals(collectionType)) {
```
NicoNes: As stated into Java doc about String conversion into resource parameter : 

> 5.Be List, Set or SortedSet, where T satisfies 2, 3 or 4 above. The resulting collection is read-only.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -252,11 +252,11 @@ else if (valueOf == null)
    }
    
 	private Class<? extends Collection> convertParameterTypeToCollectionType() {
-		if (List.class.isAssignableFrom(type)) {
```
NicoNes: We should not use `isAssignableFrom(...)` method because if resource method is as follow:

> 	@GET
	public Response singleValuedQueryParam(@QueryParam("date") LinkedHashSet<Date> dates) {
		...
	}

According to the original code the collection type to instantiate will be `HashSet` and at runtime it will fail with error like: 

> 01:30:33,911 ERROR [org.jboss.resteasy.resteasy_jaxrs.i18n] (default task-3) RESTEASY002005: Failed executing GET /singleValuedQueryParam: org.jboss.resteasy.spi.InternalServerErrorException: RESTEASY003020: Bad arguments passed to public javax.ws.rs.core.Response org.jboss.resteasy.test.resource.param.resource.ParamConverterResource.singleValuedQueryParam(java.util.LinkedHashSet)  ( java.util.HashSet [Mon Dec 19 00:00:00 CET 2016, Sat Dec 17 00:00:00 CET 2016, Sun Dec 18 00:00:00 CET 2016] )
	at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:180)
	...
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:140)
	... 46 more

I think it's a mis understanding of the spec :

> 5.Be List, Set or SortedSet, where T satisfies 2, 3 or 4 above. The resulting collection is read-only. 

and not

> 5.Be **subtype** of List, Set or SortedSet, where T satisfies 2, 3 or 4 above. The resulting collection is read-only.

ronsigal: This worries me. I think item 5 is a **bit** ambigious, since ArrayList **is** a List, though I think your interpretation is more plausible. But, since StringParameterInjector.extractValues() returns a HashSet, 

> @GET
>public Response singleValuedQueryParam(@QueryParam("date") HashSet dates) {
>...
>}

should work, right? So we're excluding behavior that used to work. That's not good.

ronsigal:
Other than my comment above, I think that this pull request is much better and simpler than the previous one. Thanks!

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/resource/ParamConverterResource.java
```diff
@@ -20,4 +27,34 @@ public void put(@QueryParam("pojo") ParamConverterPOJO q, @PathParam("pojo") Par
...
+		return Response.ok(format(dates)).build();
+	}
+
+	// This method can handle multiple 'date' query parameters
+	@GET
+	@Path("singleValuedQueryParam")
+	public Response singleValuedQueryParam(@QueryParam("date") SortedSet<Date> dates) {
+		return Response.ok(format(dates)).build();
+	}
+
```
ronsigal: Don't you want a test for Set<T> also?

NicoNes:
> So we're excluding behavior that used to work. That's not good.

Oups sorry...you're right, I did not realize.

 I've just pushed a fix for that introducing:
 - `UnmodifiableArrayList`
- `UnmodifiableHashSet`
- `UnmodifiableTreeSet`

This way we are spec compliant (using read only collection) and behavior that used to work is still working (case with HashSet, ArrayList and TreeSet).
I also had  more tests to cover those cases.

asoldano:
@ronsigal what's the status of this? Is it OK and can be merged?

ronsigal:
Hey Nicolas,

Thanks for fixing that issue with collection types. But where do you see a javadoc requiring the collections to be unmodifiable? What am I forgetting? And how do your unmodifiable classes differ from those you can get by calling Collections.unmodifiableCollection()?

-Ron

NicoNes:
Hi Ron,

> But where do you see a javadoc requiring the collections to be unmodifiable?

In the javadoc of  `@QueryParam`, `@MatrixParam`, `@CookieParam`, `@HeaderParam` for example:
http://docs.oracle.com/javaee/7/api/javax/ws/rs/QueryParam.html
>  The type T of the annotated parameter, field or property must either:
    1.Be a primitive type
        ....
    5.Be List<T>, Set<T> or SortedSet<T>, where T satisfies 2, 3 or 4 above. **The resulting collection is read-only.**

> And how do your unmodifiable classes differ from those you can get by calling Collections.unmodifiableCollection()?

Lets consider the following resource:

> @GET
public Response singleValuedQueryParam(@QueryParam("date") HashSet dates) {
...
}

It will be bound to URI like **http://host/path?date=20170905&date=20170905** and  date query params will be parsed by `StringParameterInjector.extractValues(List<String> values)` and injected into the java resource.
By using Collections.UnmodifableSet() `StringParameterInjector.extractValues(List<String> values)` will return a `java.util.Collections.UnmodifiableSet` wich is not subtype of `java.util.HashSet`. So at runtime an exception like following will be thrown: 

> 01:30:33,911 ERROR [org.jboss.resteasy.resteasy_jaxrs.i18n] (default task-3) RESTEASY002005: Failed executing GET /singleValuedQueryParam: org.jboss.resteasy.spi.InternalServerErrorException: RESTEASY003020: Bad arguments passed to public javax.ws.rs.core.Response org.jboss.resteasy.test.resource.param.resource.ParamConverterResource.singleValuedQueryParam(java.util.HashSet) ( java.util.Collections.UnmodifiableSet [Mon Dec 19 00:00:00 CET 2016, Sat Dec 17 00:00:00 CET 2016, Sun Dec 18 00:00:00 CET 2016] )
at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:180)
...
Caused by: java.lang.IllegalArgumentException: argument type mismatch
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:140)
... 46 more

My unmodifiable collections are different cause each of them extends either ` java.util.HashSet`, ` java.util.ArrayList` or ` java.util.TreeSet`. Thanks to that code is compliant with the javadoc whitout breaking behavior that used to work like following:

> @GET
public Response singleValuedQueryParam(@QueryParam("date") ArrayList dates) {
...
}

> @GET
public Response singleValuedQueryParam(@QueryParam("date") HashSet dates) {
...
}

> @GET
public Response singleValuedQueryParam(@QueryParam("date") TreeSet dates) {
...
}

And basic cases like following will also work: 

> @GET
public Response singleValuedQueryParam(@QueryParam("date") List dates) {
...
}

> @GET
public Response singleValuedQueryParam(@QueryParam("date") Set dates) {
...
}

> @GET
public Response singleValuedQueryParam(@QueryParam("date") SortedSet dates) {
...
}

Hope it's clearer.

-Nicolas

ronsigal:
Hey Nicolas,

Thanks for fixing that issue with collection types. But where do you see a javadoc requiring the collections to be unmodifiable? What am I forgetting? And how do your unmodifiable classes differ from those you can get by calling Collections.unmodifiableCollection()?

-Ron

ronsigal:
re: "In the javadoc of @QueryParam, .."

Ah, yeah, that's probably the second or third time you've told me that.

re: "java.util.Collections.UnmodifiableSet wich is not subtype of java.util.HashSet"

I see. Thanks. Ok, I have two more points:

1. The tests should say "@tpSince RESTEasy 3.1.3.Final"
2. How about a note in Chapter 6. "@QueryParam" of the Resteasy User Guide demonstrating what you have made possible?

Other than that, it looks good.

-Ron

NicoNes:
Hi Ron,

I added the "@tpSince RESTEasy 3.1.3.Final" on tests, it should be OK now.
About writting a note in the User Guide, it would be great thank you. Just let me know how to do. By the way since `StringParameterInjector` is reponsible of injection for any  JaxRS `@XXXParam`, it will also work for any of them not just `@QueryParam`.

-Nicolas

ronsigal:
re: "By the way since StringParameterInjector is reponsible of injection for any JaxRS @XXXParam, it will also work for any of them not just @QueryParam."

Ah, interesting point. What would a multivalued path parameter look like? Something like @PathParam("{p}/x/{p}")? Now that you've brought it up, I think it would be useful and appropriate to add at least one test for each of the other kinds of parameters.  Or more, as you think appropriate.

As for the documentation, I can do that. I'd like to see the other parameter types before I start writing, though.

Unless you **want** to write documentation. If you're like me, you don't like to write documentation. Fortunately, not everyone is like me. ;-)

-Ron

NicoNes:
Hi Ron,

I've just added dedicated tests for multi valued XXX param. They are bit a redundant since we are testing almost the same behavior for each XXX param, but I think they are clear. Let me know if not.

So for each XXX param you will find : 

- one test for custom multivalued param conversion. I let you check the one on `@PathParam`.

- 6 tests for default multivalued param conversion since HTTP define the standard way to do so for almost all XXX param except `CookieParam` and `PathParam` wich can't be multivalued by default.

About the documentation, you're right. Just like you I don't really like to write it. (I prefer to read it ...when not to long :))

NicoNes:
The last build failed, but I do not see no error except 

> The command "travis_wait 30 mvn -B -fae -Dserver.version=$SERVER_VERSION install" exited with 137

Any idea why ? Timeout ?

ronsigal:
Hi Nicolas,

re: "The last build failed, "

Yeah, it happens sometimes. I restarted the tests.

re: "So for each XXX param you will find"

I like the very useful comments in the tests. Thank you.

I wanted to see what something like

      @GET
      @Path("paths/{path:.+}")
      public Response pathString(@PathParam("path") List<String> paths) {
    	  return Response.ok("ok").build();
      }

would look like where the "paths" parameter picks up multiple segments, and I don't think that appears in the tests. I noticed that when MultiValuedParamConverterProvider.getConverter() returns a DateParamConverter, only the first path segments gets processed.

But when I use

    public static class PathParamsConverterProvider implements ParamConverterProvider {

       @SuppressWarnings("unchecked")
       @Override
       public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations) {
           if (rawType.isAssignableFrom(ArrayList.class)) {
               return (ParamConverter<T>) new PathParamsConverter();
           } else {
              return null;
           }
       }
   }
   
     public static class PathParamsConverter implements ParamConverter<List<String>> {

       @Override
       public List<String> fromString(String value) {
           String[] ss = value.split("/");
           List<String> list = new ArrayList<String>();
               for (String s : ss) {
                   list.add(s);
              }
          return list;
       }

       @Override
       public String toString(List<String> value) {
          // TODO Auto-generated method stub
          return null;
       }
   }

and call

    @Test
    public void testStrings() throws Exception
    {
       ResteasyClient client = new ResteasyClientBuilder().build();
       Invocation.Builder request = client.target("http://localhost:8081/test/paths/20161217/20161218/20161219").request();
       Response response = request.get();
       String report = response.readEntity(String.class);
       System.out.println("response: " + report);
       Assert.assertEquals(200, response.getStatus());
       client.close();
    }

I get

>path segment: 20161217
>path segment: 20161218
>path segment: 20161219

Could you include a test like that?

Thanks,
Ron

NicoNes:
Hi Ron,

The test you're talking about is already included into:

`org.jboss.resteasy.test.resource.param.resource.MultiValuedParamResource.PathParamResource `
>
    @Path("pathParam")
    public static class PathParamResource {

	   // http://xxx/pathParam/customConversion_multiValuedPathParam/20161217/20161218/20161219
	  @GET
	  @Path("customConversion_multiValuedPathParam/{path: .+}")
	  public Response customConversion_multiValuedPathParam(@PathParam("path") MultiValuedPathParam<Date> paths) {
		return Response.ok(formatDates(paths)).build();
	  }
	...
    }

It uses a `MultiValuedPathParam` and its associated converter `MultiValuedPathParamConverter` to split and convert the segments. It works as you expected in your test.

Is it ok ?

-Nicolas

ronsigal:
Hi Nicolas,

re: "The test you're talking about is already included into:"

Ok, I was thinking MultivaluedPathParam didn't have a type variable and that that was an issue. I see that it does have a type variable.

I've been writing some documentation about parameter conversion, and I think I've run into something that isn't right. If the resource method is

		@GET
		@Path("pathsegment/{p}/{q}/{p}")
		public Response pathSegments(@PathParam("p") List<PathSegment> ps) {
			for (PathSegment p : ps) {
				System.out.println("p: " + p.getPath());
			}
			return Response.ok().build();
		}

and the test method is

    @Test
    public void testPathSegments() throws Exception
    {
	   ResteasyClient client = new ResteasyClientBuilder().build();
	   Invocation.Builder request = client.target("http://localhost:8081/test/pathsegment/a1/b2/a3").request();
	   Response response = request.get();
	   String report = response.readEntity(String.class);
	   System.out.println("response: " + report);
	   Assert.assertEquals(200, response.getStatus());
	   client.close();
    } 

then, based on the second item in the javadoc for @PathParam, I would expect to get output

    p: a1
    p: a3

but I'm getting

    p: a3

This behavior comes from org.jboss.resteasy.core.PathParamParamInjector:

    public Object inject(HttpRequest request, HttpResponse response)
    {
       if (extractor == null) // we are a PathSegment
       {
          ResteasyUriInfo uriInfo = (ResteasyUriInfo) request.getUri();
          List<PathSegment[]> list = null;
          if (encode)
          {
             list = uriInfo.getEncodedPathParameterPathSegments().get(paramName);
          }
          else
          {
             list = uriInfo.getPathParameterPathSegments().get(paramName);
          }
          if (list == null)
          {
             throw new InternalServerErrorException(Messages.MESSAGES.unknownPathParam(paramName, uriInfo.getPath()));
          }
          PathSegment[] segments = list.get(list.size() - 1);  // <<==== HERE
          ...

That doesn't look right. Am I missing something?

-Ron

NicoNes:
Hi Ron,

You're right I think that is not correct. If you try the same test using `List<String>` instead of `List<PathSegment>` it works as expected thanks to following line:

```
 public Object inject(HttpRequest request, HttpResponse response) 
{
  ...
   else
   {
      if (extractor.isCollectionOrArray())
       {
          return extractor.extractValues(list); <= HERE
        }
  ...
}
```

So it should be the same using `List<PathSegment>`. The special treatment done at the line you showed me should only be available in case we are injecting a simple `PathSement` and not a `List<PathSegment>`.
According to the `@PathParam` javadoc:  

> •Be javax.ws.rs.core.PathSegment, **the value will be the final segment of the matching part of the path**. 
•Be List<javax.ws.rs.core.PathSegment>, the value will be a list of PathSegment corresponding to the path segment(s) that matched the named template parameter

By the way the second point is interesting since  following line of the javadoc

> Be List<javax.ws.rs.core.PathSegment>

is translated as followed in the code:

```
 public Object inject(HttpRequest request, HttpResponse response) 
 {
   ...
   return List.class.equals(type)&& ...;
 }
```

And the same concept into `@QueryParam`, `@MatrixParam` etc javadoc:

> 5.Be List<T>, Set<T> or SortedSet<T>, where T satisfies 2, 3 or 4 above. The resulting collection is read-only

is translated as followed in `StringParameterInjector `class:

```
private Class<? extends Collection> convertParameterTypeToCollectionType() {
  if (List.class.equals(type) || ArrayList.class.equals(type)) {
    return ArrayList.class;
  } 
  ...
}
```
Maybe a good thing will be to make this consistent in `PathParamInjector.inject(HttpRequest request, HttpResponse response) ` like following:

```
 public Object inject(HttpRequest request, HttpResponse response) 
 {
   ...
   return (List.class.equals(type) || ArrayList.class.equals(type))&& ...;
 }
```

 What do you think ?

ronsigal:
Hi Nicolas,

I took care of the PathSegment issue in a separate JIRA (https://issues.jboss.org/browse/RESTEASY-1653). The changes have been committed.

re: "Maybe a good thing will be to make this consistent in PathParamInjector.inject(HttpRequest request,"

Good point, thank you. In fact, I changed the method to
```
    private boolean isPathSegmentList(Class type, Type genericType)
    {
       Class collectionBaseType = Types.getCollectionBaseType(type, genericType);
       return List.class.isAssignableFrom(type) && collectionBaseType != null && collectionBaseType.equals(PathSegment.class);
    }
```
That is, I'm testing `List.class.isAssignableFrom()` instead of `List.class.equals()`.  I know we talked about his at one point. If you remember, StringParameterInjector.initialize() currently looks like
```
      if (List.class.isAssignableFrom(type))
      {
         isCollection = true;
         collectionType = ArrayList.class;
      }
      else if (SortedSet.class.isAssignableFrom(type))
      {
         isCollection = true;
         collectionType = TreeSet.class;
      }
      else if (Set.class.isAssignableFrom(type))
      {
         isCollection = true;
         collectionType = HashSet.class;
      }
```
You turned that into
```
	private Class<? extends Collection> convertParameterTypeToCollectionType() {
		if (List.class.equals(type) || ArrayList.class.equals(type)) {
			return ArrayList.class;
		} else if (SortedSet.class.equals(type) || TreeSet.class.equals(type)) {
			return TreeSet.class;
		} else if (Set.class.equals(type) || HashSet.class.equals(type)) {
			return HashSet.class;
		}
		return null;
	}
```
and I think we should preserve the earlier, broader semantics. WDYT?

-Ron

NicoNes:
Hi Ron,

I actually got a problem with the previous code in `StringParameterInjector.initialize() `. 
This code is allowing something that will **obviously** failed at runtime. Let's consider the case of the `Set` injection in the following resource for example:

```
@GET
public Response singleValuedQueryParam(@QueryParam("date") LinkedHashSet dates) {
...
}
```

 Since `Set.class.isAssignableFrom(LinkedHashSet.class)` is `true` the code will inject the  `dates `argument with an 'HashSet' instance but at runtime it will fail with:

> 01:30:33,911 ERROR [org.jboss.resteasy.resteasy_jaxrs.i18n] (default task-3) RESTEASY002005: Failed executing GET /singleValuedQueryParam: org.jboss.resteasy.spi.InternalServerErrorException: RESTEASY003020: **Bad arguments passed to public javax.ws.rs.core.Response org.jboss.resteasy.test.resource.param.resource.ParamConverterResource.singleValuedQueryParam(java.util.LinkedHashSet) ( java.util.HashSet** [Mon Dec 19 00:00:00 CET 2016, Sat Dec 17 00:00:00 CET 2016, Sun Dec 18 00:00:00 CET 2016] )
> at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:180)
> ...
> Caused by: **java.lang.IllegalArgumentException: argument type mismatch**
> at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> at java.lang.reflect.Method.invoke(Method.java:498)
> at org.jboss.resteasy.core.MethodInjectorImpl.invoke(MethodInjectorImpl.java:140)
> ... 46 more

The only way for this code to work is if `dates` argument type is either equals to `Set.class` or `HashSet.class`. That's why i change the code to:
```
else if (Set.class.equals(type) || HashSet.class.equals(type)) {
     return HashSet.class;
}
```
That's the same for `List `and `TreeSet` and that's why I truly believe that when `@QueryParam`, `@MatrixParam` etc javadoc says:

  > 5.**Be List, Set or SortedSet**, where T satisfies ...

it means:

> 5. Be List.class not subtype, Set.class not subtype or SortedSet.class not subtype ....
Else the injection will obviously lead to runtime exception since user can use any subtype, standard like `LinkedHashSet` but custom ones too and we will never be able to inject this specific type.

To be fine with the spec the code would have been like:
```
else if (Set.class.equals(type)) {
     return HashSet.class;
}
```
but I also added `HashSet.class.equals(type)` to preserve old behavior.

About the code in `PathParamInjector.inject(HttpRequest request, HttpResponse response)` I think that the spec interpretation was good. I just added the `ArrayList.class.equals(type)` check to be consistent with code in `StringParameterInjector.initialize()`

ronsigal:
re: "This code is allowing something that will **obviously** failed at runtime"

Yes, you're right. I'll change PathParamInjector.isPathSegmentList() to
```
   private boolean isPathSegmentList(Class type, Type genericType)
   {
      Class collectionBaseType = Types.getCollectionBaseType(type, genericType);
      return (List.class.equals(type) || ArrayList.class.equals(type)) && collectionBaseType != null && collectionBaseType.equals(PathSegment.class);
   }
```
Thanks, Nicolas.

ronsigal:
Hey Nicolas,

I've created https://github.com/resteasy/Resteasy/pull/1251 to replace this pull request. It takes the same code and merges it into master branch. Also, it includes the documentation I wrote. Could you take a look at https://github.com/resteasy/Resteasy/pull/1251?

Thanks,
Ron

PR #1096, opened 2017-03-07
RESTEASY-1623:GZIPDecodingInterceptor may be registered twice when us…

jimma:
…ing JAX-RS and REsteasy client on the same application


jimma:
@asoldano  yes. I think this is better than add synchronize block to getDelegate() when the factory is a ThreadLocalResteasyProviderFactory in RegisterBuiltIn.register().  If we add this in RegisterBuiltin, we'll make each servelt request to acquire get ResteasyProviderFactory when calling this line :
if (factory.isBuiltinsRegistered() || !factory.isRegisterBuiltins()) return;
This is not necessary after the factory is already builtin registered.

jimma:
What  I concern is add synchronize block to  ResteasyProviderFactory with getDelete() when the pass in factory is an instance of ThreadLocalResteasyProviderFactory is different from add the synchronize block to ThreadLocalResteasyProviderFactory. This will affect this line code  :  if (factory.isBuiltinsRegistered() || !factory.isRegisterBuiltins()) return;

PR #1107, opened 2017-03-18
[RESTEASY-1628]

ronsigal:
Changed treatment of text media types without charset parameter to add
"charset=UTF-8" by default.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/MessageSanitizerContainerResponseFilter.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/MessageSanitizerContainerResponseFilter.java
@@ -81,9 +81,13 @@ public class MessageSanitizerContainerResponseFilter implements ContainerRespons
           Object o = it.next();
           if (o instanceof String) {
              String mediaType = (String) o;
-             String[] parts = mediaType.split("/");
-             if (parts.length >= 2 && parts[0].trim().equals("text") && parts[1].trim().startsWith("html")) {
-                return true;
+             String[] partsType = mediaType.split("/");
+             if (partsType.length >= 2) {
+                String[] partsSubtype = partsType[1].split(";");
+                if (partsType[0].trim().equalsIgnoreCase("text") && 
+                      partsSubtype[0].trim().toLowerCase().equals("html")) {
+                   return true;
+                }
              }
           }
        }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/MessageSanitizerContainerResponseFilter.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/MessageSanitizerContainerResponseFilter.java
@@ -3,6 +3,8 @@ package org.jboss.resteasy.plugins.interceptors;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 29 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/interceptors/MessageSanitizerContainerResponseFilter.java
```diff
@@ -74,5 +76,18 @@ private String escapeXml(String str) {
         return sb.toString();
     }
 
+    private boolean containsHtmlText(ArrayList<Object> list) {
+       for (Iterator<Object> it = list.iterator(); it.hasNext();  ) {
+          Object o = it.next();
+          if (o instanceof String) {
+             String mediaType = (String) o;
+             String[] parts = mediaType.split("/");
+             if (parts.length >= 2 && parts[0].trim().equals("text") && parts[1].trim().startsWith("html")) {
```
kanovotn: Just a question:
 * I haven't try it myself, but what if the content-type will be "TEXT/HTML", then containsHtmlText will return false and then the string in the entity (L36) will not be escaped? Other lines in this PR, who test content-type, use equalsIgnoreCase() method.

ronsigal: You're so right, Katka. Thanks. I'll fix that.

ronsigal: Changed to

          if (o instanceof String) {
             String mediaType = (String) o;
             String[] partsType = mediaType.split("/");
             if (partsType.length >= 2) {
                String[] partsSubtype = partsType[1].split(";");
                if (partsType[0].trim().equalsIgnoreCase("text") && 
                      partsSubtype[0].trim().toLowerCase().equals("html")) {
                   return true;
                }
             }
          }

kanovotn: :+1:

asoldano:
@rsvoboda , thanks for the review. I agree on the need for squashing the commits; we could move the containsHtmlText method to MediaTypeHelper (in that case it should probably be a more general-use one, like containsContentType(List<Object> contentTypes, String contentType), but that's not strictly needed. I agree on using "for (Object o : list)", while I think we could avoid the JDK8 syntax for now to have the same code in 3.0 branch. Finally, it's better not using MediaType.valueOf(o.toString()) as long as there's no real need of it, basically because building MediaType objects is quite heavy (they always come with a TreeMap in them and the valueOf method is quite complex).

ronsigal:
- I squashed the commits, but, since master had changed, I had to create a new branch and a new PR: https://github.com/resteasy/Resteasy/pull/1110 . Do we have a policy restricting PRs to a single commit?
- I changed to the "for (Object o : list)" syntax.
- While creating a general purpose containsContentType() might be useful, I would have to do more work to parse the parameters, and probably wouldn't even want to use it in this case. So i left that alone.

asoldano:
@ronsigal , thanks. No we don't have a policy for single commit PR in RESTEasy, but single commit changes for a given jira are usually better for future investigation of changes ;-) This said, it often happens to have many commits to fix a given jira..

ronsigal:
@asoldano I definitely agree that 80 commits is too many. ;-)

PR #1109, opened 2017-03-21
[RESTEASY-1628]

ronsigal:
Changed treatment of text media types without charset parameter to add
"charset=UTF-8" by default.


ronsigal:
Replaced by https://github.com/resteasy/Resteasy/pull/1111.

PR #1113, opened 2017-03-22
[RESTEASY-1629]

ronsigal:
UndertowJaxrsServer can handle context and initial parameters.


ronsigal:
Changes requested on master branch made here as well.

PR #1114, opened 2017-03-22
[RESTEASY-1627] Automatic HEAD requests breaks subsequent request(s)

asoldano:
This fix the issue by avoiding the body contents to be returned in HEAD method responses, as a consequence of RESTEASY-1365 changes.
The bits on the wire are still slightly different from what we had before RESTEASY-1365: before that, resteasy-jaxrs was not handling control over to the resteasy-netty4 server adapter for dealing with responses to HEAD requests, basically only writing http status plus few headers (Connection and Content-Type headers). Now, resteasy-netty4 is still processing the response, but the contents of the ChunkOutputStream are actually not flushed to the wire, resulting in an empty chunk being written only. So an exchange with a GET request followed by a HEAD request to the same resource ends up being as follows (capture with Wireshark):

GET /test HTTP/1.1
Host: localhost:8081
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.2 (Java/1.8.0_77)

HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/plain
transfer-encoding: chunked

b
hello world
0

HEAD /test HTTP/1.1
Content-Length: 0
Host: localhost:8081
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.2 (Java/1.8.0_77)

HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/plain
transfer-encoding: chunked

0

### DIFF
--- a/server-adapters/resteasy-netty/src/test/java/org/jboss/resteasy/test/NettyTest.java
+++ b/server-adapters/resteasy-netty/src/test/java/org/jboss/resteasy/test/NettyTest.java
@@ -83,6 +83,8 @@ public class NettyTest
       String val = ClientInvocation.extractResult(new GenericType<String>(String.class), getResponse, null);
       Assert.assertEquals("hello world", val);
       Response headResponse = target.request().build(HttpMethod.HEAD).invoke();
+      Assert.assertFalse(headResponse.hasEntity());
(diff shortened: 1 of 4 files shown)


asoldano:
This is replaced by https://github.com/resteasy/Resteasy/pull/1115

PR #1115, opened 2017-03-22
[RESTEASY-1627] Improving test for resteasy-netty and fixing resteasy…

asoldano:
…-netty4 to avoid sending any body and Content-Length header in response to HEAD requests

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
@@ -78,7 +78,8 @@ public class ServerResponseWriter
       
       executeFilters(jaxrsResponse, request, response, providerFactory, method);
 
-      if (jaxrsResponse.getEntity() == null)
+      //[RESTEASY-1627] check on response.getOutputStream() to avoid resteasy-netty4 trying building a chunked response body for HEAD requests 
+      if (jaxrsResponse.getEntity() == null || response.getOutputStream() == null)
       {
          response.setStatus(jaxrsResponse.getStatus());
          commitHeaders(jaxrsResponse, response);
--- a/server-adapters/resteasy-jdk-http/src/test/java/org/jboss/resteasy/test/HeadContentLengthTest.java
+++ b/server-adapters/resteasy-jdk-http/src/test/java/org/jboss/resteasy/test/HeadContentLengthTest.java
@@ -1,0 +1,82 @@
+package org.jboss.resteasy.test;
+
+import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
+import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
+import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
+import org.jboss.resteasy.plugins.server.sun.http.HttpContextBuilder;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import com.sun.net.httpserver.HttpServer;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.HttpMethod;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.GenericType;
+import javax.ws.rs.core.Response;
+
+import static org.jboss.resteasy.test.TestPortProvider.generateURL;
+
+import java.net.InetSocketAddress;
+
+public class HeadContentLengthTest
+{
+   @Path("/")
+   public static class Resource
+   {
+      @GET
+      @Path("/test")
+      @Produces("text/plain")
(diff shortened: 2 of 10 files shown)


asoldano:
@ronsigal , @liweinan , can any of you have a quick look at this? I'd like this issue to be fixed in next release, which should be tagged in 1-2 days.

asoldano:
Some additional explanations at https://issues.jboss.org/browse/RESTEASY-1627?focusedCommentId=13382654&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13382654

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
```diff
@@ -78,7 +78,8 @@ public static void writeNomapResponse(BuiltResponse jaxrsResponse, final HttpReq
       
       executeFilters(jaxrsResponse, request, response, providerFactory, method);
 
-      if (jaxrsResponse.getEntity() == null)
+      //[RESTEASY-1627] check on response.getOutputStream() to avoid resteasy-netty4 trying building a chunked response body for HEAD requests 
+      if (jaxrsResponse.getEntity() == null || response.getOutputStream() == null)
```
ronsigal: Wouldn't it be easier, and more general, to check for request.getHttpMethod().equalsIgnoreCase("HEAD") here?

asoldano: That's what we had before RESTEASY-1365 was implemented. However, the problem was that causing Content-Length to always be set to 0 for responses to HEAD requests, which goes against rfc7230/7231. Basically, either the Content-Length is not specified at all, or it has to have the same value that would returned for a GET request to the same resource. In order to compute such a value, the response processing has to be performed and the the headers set after that, hence the removal of the check on the httpmethod name there.

The idea with the fix here is to allow resteasy-netty4 to reply to HEAD request without setting the Content-Length header, while leaving the fix for other servers as it was implemented in RESTEASY-1365 (so a valid Content-Length is computed and set).

ronsigal: I see, @asoldano. Thanks for the explanation. Ok, how about this: If verb is HEAD, commitHeaders() removes the Content-Length header. Let me say:

- I don't see anything wrong with the solution you've presented.
- I'm not trying to be difficult. I'm just wondering if there's a general solution that would apply to other server adaptors.

asoldano: @ronsigal , sure, I'd be fine with a general solution applying to all adapters (even if that would change the behavior implemented in RESTEASY-1365). Can commitHeaders() actually remove the Content-Length there (or perhaps that header is added later?) anyway?
I should likely check...

ronsigal: re: "perhaps that header is added later?"

I'll look into that.

asoldano: In case of resteasy-netty, that's added later by RestEasyHttpResponseEncoder::encode, for resteasy-netty4 that's added later by NettyHttpResponse::getEmptyHttpResponse .
I don't know about wfly/eap and the other servers, but looks like we should fix them all in the relevant places (actually, I just figured out the resteasy-vertx is probably affected by this issue too, for example).

asoldano: I've been debugging and thinking a bit about this. We can't remove headers from HttpServletResponse instances, so an attempt of removing the Content-Length header from the response would fail. This said, when commitHeader is invoked, the Content-Length header is not set yet in the case of resteasy-undertow and wildfly, I assume because Undertow sets it later on. So I decided to at least complete the proposed fix here, by covering all the adapters & providing tests; this means either preventing the Content-Length header from being emitted if that's something we can control (resteasy-netty4, resteasy-vertx, resteasy-jdk-http) or by having that computed properly instead of being set to 0 (resteasy-netty, resteasy-undertow, wildfly/eap).
My proposal would be to go with this solution for now, given it basically keeps the current resteasy-netty/resteasy-undertow/resteasy-jdk-http/wildfly/eap behaviors while fixing broken head responses in resteasy-netty4 and resteasy-vertx. We can possibly create a new jira for further investigating unifying all server adapters' behavior when processing HEAD requests.

ronsigal: Maybe I was being overly optimistic about being able to solve this problem in resteasy-jaxrs. It **feels** like it should be possible, but that and 10,000 lire will get you an uber ride. ;-)

So, +1.

asoldano: OK, thanks @ronsigal . I've created https://issues.jboss.org/browse/RESTEASY-1632 and I'm going to merge this.

PR #1119, opened 2017-03-31
[RESTEASY-1422]

ronsigal:
Handle response headers with View entity in resteasy-html.

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/html/HeadersInViewResponseTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/html/HeadersInViewResponseTest.java
@@ -1,0 +1,93 @@
+package org.jboss.resteasy.test.providers.html;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+
+import javax.ws.rs.client.Invocation;
+import javax.ws.rs.core.NewCookie;
+import javax.ws.rs.core.Response;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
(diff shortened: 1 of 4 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/html/HeadersInViewResponseTest.java
```diff
@@ -0,0 +1,93 @@
...
+@RunWith(Arquillian.class)
+@RunAsClient
+public class HeadersInViewResponseTest {
+
+    static ResteasyClient client;
+
+    @Deployment()
+    public static Archive<?> deploy() {
+        WebArchive war = TestUtil.prepareArchive(HeadersInViewResponseTest.class.getSimpleName());
+        war.addAsLibrary(getResteasyHtmlJar());
```
asoldano: Just asking, why not using the same same approach as in org.jboss.resteasy.utils.TestUtilSpring, basically reading the maven artifact from the Maven Repository?

ronsigal: Ah, well, I didn't know about it. Do you want me to use it?

asoldano: Not sure to be honest, I was wondering if you considered that and if there's a reason for preferring one of the two approaches.

ronsigal: Well, I can't think of a compelling reason one way or the other. MavenUtil is way more complicated and, presumably, more general. You could argue that using the maven repository is immune to Resteasy restructuring, but how often are we going to move resteasy-html?

asoldano: ok, let's keep this as is then, at least as long as no actual problem is found. Thanks.

PR #1127, opened 2017-04-10
RESTEASY-1636 Part2

liweinan:
- Remove documents on skeleton-key and OAuth modules
- Add security modules deprecation document
- Move `login-module` to `security-legacy` directory


docbook/reference/en/en-US/modules/Deprecated_security_modules.xml
```diff
@@ -0,0 +1,18 @@
+<chapter id="Deprecated_security_modules">
+    <title>Deprecated Security Modules</title>
+
+    <para>
+        These security modules are deprecated since 3.1.3.Final and users should use Keycloak provided by JBoss community instead:
```
asoldano: How about the following line instead? I'd rather not mentioning the version this is being deprecated in (it's basically a mess when you make releases)

"The following security related modules are deprecated and meant to be removed in the future:"

I would also move the reference to Keycloak after the list, in place of the paragraph on the directory structure, which is not that relevant after all.

docbook/reference/en/en-US/modules/Deprecated_security_modules.xml
```diff
@@ -0,0 +1,18 @@
...
+
+    <itemizedlist>
+        <listitem>keystone</listitem>
+        <listitem>login-module-authenticator</listitem>
+        <listitem>resteasy-oauth</listitem>
+        <listitem>skeleton-key-idm</listitem>
+    </itemizedlist>
+
+    <para>
+        The above modules are moved from 'security' directory to 'security-legacy' directory.
```
asoldano: How about something like the following line instead?

"Please refer to http://www.keycloak.org/ for a modern solution for the usecases previously covered by the deprecated modules".

liweinan:
@asoldano Thanks for the advice, it's much better! I've made the modifications for you to check.

PR #1129, opened 2017-04-10
[RESTEASY-1635]

ronsigal:
Update web.xml examples in User Guide to omit unnecessary elements like
<servlet> and <servlet-mapping>.

### DIFF
--- a/docbook/reference/en/en-US/modules/Installation_Configuration.xml
+++ b/docbook/reference/en/en-US/modules/Installation_Configuration.xml
@@ -260,10 +260,9 @@
             You can use it with EJB and CDI and you can rely completely on WildFly to scan for and deploy your JAX-RS services and providers.
             All you have to provide is your JAX-RS service and provider classes packaged
             within a WAR either as POJOs, CDI beans, or EJBs.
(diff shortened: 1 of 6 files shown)


docbook/reference/en/en-US/modules/Installation_Configuration.xml
```diff
@@ -278,8 +277,8 @@
         </para>
         <para>
             Since we're not using a &lt;servlet-mapping&gt; element, we must define an
-            <classname>Application</classname> class (see <xref linkend="javax.ws.rs.core.Application"/>) that is annotated with
-            the <classname>@ApplicationPath</classname> annotation.  If you return any empty set for classes and singletons,
+            <classname>javax.ws.rs.core.Application</classname> class (see <xref linkend="javax.ws.rs.core.Application"/>) that is annotated with
```
asoldano: I'm not a native English speaker, but should it be "a javax.ws.rs.core.Application class" instead of "an javax.ws.rs.core.Application class" ?

ronsigal: D'oh. Busted. ;-)

PR #1132, opened 2017-04-12
[JBEAP-9423] Base64 class uses printStackTrace method

xstefank:
JBEAP: https://issues.jboss.org/browse/JBEAP-9423


resteasy-jaxrs/src/main/java/org/jboss/resteasy/util/Base64.java
```diff
@@ -1388,7 +1388,7 @@ else if (len < 4)
             }   // end try
             catch (java.io.IOException e)
             {
-               e.printStackTrace();
+               Messages.MESSAGES.unableToDecodeGZIPBase64();
```
asoldano: The i18n message here should be printed in the log, shouldn't it?
Moreover, I'd also like the IOException to be logged too, to simplify troubleshooting.

xstefank:
@asoldano sorry, I missed the logging class. It should be alright now. Thank you for the review.

PR #1135, opened 2017-04-14
[RESTEASY-1464]

ronsigal:
On server side, FileProvider adds temporary File to Cleanables so that
it will be deleted at end of invocation.


ronsigal:
I've 
1. removed the call to File.deleteOnExit();
2. added a DEBUG level log message about temporary files on the client side;
3. added a note about temporary files to the User Guide.

PR #1136, opened 2017-04-14
[RESTEASY-1464]

ronsigal:
On server side, FileProvider adds temporary File to Cleanables so that
it will be deleted at end of invocation.


ronsigal:
Same response as in https://github.com/resteasy/Resteasy/pull/1135.

PR #1138, opened 2017-04-19
Jbeap 10058 branch 3.0

rsearls:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/CustomExceptionMapperTest.java
```diff
@@ -64,6 +65,12 @@
                         CustomExceptionMapperReport.class)
                 .addAsLibrary(file, "validation-versioning.jar")
                 .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+            new RuntimePermission("getClassLoader"),
```
kanovotn: Only `new RuntimePermission("accessDeclaredMembers")` is required to make the test pass, no getClassloader permission is required.

The stacktrace of the AccessControll exception (attached in https://issues.jboss.org/browse/JBEAP-10058) shows that problem is in the class: `org.jboss.resteasy.plugins.validation.ConstraintTypeUtil11.getMethod(ConstraintTypeUtil11.java:95)`.

Why is it better in this case to give the 'accessDeclaredMembers' permission to the application, instead of creating AccessController.doPrivileged() block in the resteasy code?

 For example when application itself calls some code which reguires permission directly:
`Thread.currentThread().getContextClassLoader()`, it is obvious it must have granted the permission directly.
If the above getContextClassLoader() is called somewhere in the resteasy classes then probably it should be handled in the resteasy code.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/ValidationThroughRestTest.java
```diff
@@ -42,6 +43,12 @@
     public static Archive<?> createTestArchive() {
         WebArchive war = TestUtil.prepareArchive(ValidationThroughRestTest.class.getSimpleName())
                 .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
```
kanovotn: Same as my previous comment - https://github.com/resteasy/Resteasy/pull/1138/files#diff-9a07efadc51ffa6fbff43e1a8b715eacR70

rsearls:
Revisions made

kanovotn:
I don't see changes to review, in comparison to my previous comments.

rsearls:
Please elaborate.  I reviewed the touched files and don't see any changes to be made.

kanovotn:
Discussed over email with @rsearls and the changes looks good to me.

PR #1143, opened 2017-04-20
[RESTEASY-1597] add logging of file not deleted

rsearls:


rsearls:
Will not add requested community member change.  Will wait for jdk fix

asoldano:
@rsearls just to be sure we didn't misunderstand each other, with the comment above I didn't imply we should leave the code untouched. I meant that we don't simply need to write a log about the problem, instead we need to avoid the leak or make it possible to avoid it. In the RESTEASY-1597 jira comments, I proposed an approach for dealing with the problem and a community user also proposed an alternative solution.

asoldano:
Btw, I'm not counting on this JDK "issue" being addressed any time soon.

PR #1149, opened 2017-04-28
[RESTEASY-1630] new tests for Servletinitializer

rsearls:


arquillian/RESTEASY-1630-jetty-resteasy-servlet-initializer/src/test/java/org/jboss/resteasy/test/resteasy1630/ServletInitializerTwoTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import static org.junit.Assert.assertEquals;
+
+/**
```
asoldano: Please add some javadoc here to explain the test. Basically, the reason why we are adding this test is to show how the deployment based on the resteasy-servlet-initializer approach works. So, those looking at this test need to either find the resteasy-servlet-initializer explicitly added to the deployment created using ShrinkWrap, or read a note explaining that Arquillian is automatically adding the resteasy-servlet-initializer dependency to the deployment as a consequence of that dependency having been specified in the pom.xml for this test project. Without that, most readers will simply miss the point. Thanks.

rsearls:
done

asoldano:
Thanks, I see you explicitly added the jar to the shrinkwrap created deployments, fine.

PR #1150, opened 2017-05-10
[RESTEASY-1646] update section, discussed BOM files.

rsearls:


docbook/reference/en/en-US/modules/Maven_and_RESTEasy.xml
```diff
@@ -6,100 +6,87 @@ JBoss's Maven Repository is at:
...
+       </dependency>
+        ....
+   </dependencies>
+  ]]>
+  </programlisting>
+   <para>
+     Using the BOM file is not required.  Individual RESTEasy artifacts can be
+     declared in the dependencies section of the POM file like this.
+   </para>
+  <programlisting>
```
asoldano: We should not be promoting this approach, right? I'd probably skip this part, it's implicit in the sections above imho. WDYT?

rsearls:
File edited.

PR #1156, opened 2017-05-15
[RESTEASY-1650] Make HttpResponse closeable and always close its outp…

asoldano:
…utstream, while still avoiding flushing/closing on the lower level servlet outputstream

This solves the issue mentioned at https://issues.jboss.org/browse/RESTEASY-1650, please reads the comments too to understand the problem.
The possible downside of this patch is that it possibly breaks user applications that were explicitly calling close() or flush() on the response outputstream and expecting that to actually flush/close the underlying servlet outputstream (despite that being bad and against JAX-RS recommendations).

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -22,7 +22,7 @@ public class HttpServletResponseWrapper implements HttpResponse
    protected OutputStream outputStream = new DeferredOutputStream();
 
    /**
-    * RESTEASY-684 wants to defer access to outputstream until a write/flush/close happens
+    * RESTEASY-684 wants to defer access to outputstream until a write happens
     *
     */
    protected class DeferredOutputStream extends OutputStream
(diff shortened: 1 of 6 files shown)


ronsigal:
Hi @asoldano,

I was playing with ContainerResponseFilters, and that works as well. ContainerResponseContextImpl.setEntityStream() is

    public void setEntityStream(OutputStream entityStream)
    {
       httpResponse.setOutputStream(entityStream);
    }

So, I think this pull request is OK.

-Ron

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -48,13 +48,13 @@ public void write(byte[] bytes, int i, int i1) throws IOException
       @Override
       public void flush() throws IOException
       {
-         response.getOutputStream().flush();
+         //NOOP (RESTEASY-1650)
       }
```
jimma: @asoldano , SseEventOutputImpl retrieves response with ResteasyProviderFactory.getContextData(HttpServletResponse.class) to flush sse event. So this won't effect SseEventOutputImpl.writeEvent(), right ?

jimma: @asoldano  Then nothing I need to worry.

PR #1159, opened 2017-05-16
[RESTEASY-1138] keep a reference to the ClientResponse object as long…

asoldano:
… as there's a live inputstream derived from it

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
@@ -21,6 +21,7 @@ import javax.ws.rs.ext.Providers;
 import javax.ws.rs.ext.ReaderInterceptor;
 
 import java.io.ByteArrayInputStream;
+import java.io.FilterInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
@@ -129,7 +130,24 @@ public abstract class ClientResponse extends BuiltResponse
    {
       if (bufferedEntity != null) return new ByteArrayInputStream(bufferedEntity);
       if (isClosed()) throw new ProcessingException(Messages.MESSAGES.streamIsClosed());
-      return getInputStream();
+      InputStream is = getInputStream();
+      return is != null ? new InputStreamWrapper(is, this) : null;
+   }
+   
+   private static class InputStreamWrapper extends FilterInputStream {
+      
+      private ClientResponse response;
+      
+      protected InputStreamWrapper(InputStream in, ClientResponse response) {
+         super(in);
+         this.response = response;
+      }
+      
+      @Override
+      public void close() throws IOException {
+         super.close();
+         this.response.close();
+      }
    }
 
    protected abstract void setInputStream(InputStream is);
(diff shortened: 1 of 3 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -129,7 +130,24 @@ protected InputStream getEntityStream()
...
+      
+      protected InputStreamWrapper(InputStream in, ClientResponse response) {
+         super(in);
+         this.response = response;
+      }
+      
+      @Override
+      public void close() throws IOException {
+         super.close();
+         this.response.close();
```
asoldano: I could have simply set this.response = null here, but that would have been possibly "optimized" (removed) by an aggressive JIT compiler figuring out this.response wouldn't be used. So I'm actually explicitly closing the response if the inputstream is explicitly closed.

jimma: This PR looks good to me. One of the minor things: is the super.close() is enough here ?  response.close() will close the underlying inputstream and other stuff too.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/response/ResponseStreamPrematurelyClosedTest.java
```diff
@@ -0,0 +1,89 @@
...
+   }
+
+   @Test
+   public void testStream() throws Exception
+   {
+      Builder builder = client.target(generateURL("/test/document/abc/content")).request();
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+      //builder.get().readEntity explicitly on the same line below and not saved in any temp variable
+      //to let the JVM try finalizing the ClientResponse object
+      IOUtils.copy(builder.get().readEntity(InputStream.class), baos);
```
jimma: Another minor thing :  we can call a system.gc() (although it only suggest jvm to do gc) and wait for  gc notification by calling javax.management api. After gc happens , we do IOUtils.copy and check the target stream. Is it better ?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
asoldano: For the records, I've been discussing this a bit with Jim and we eventually agreed that it should be fine to leave the fix as is.

asoldano: For the records, I've been discussing this a bit with Jim and we eventually agreed that it should be fine to leave the fix as is.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/response/ResponseStreamPrematurelyClosedTest.java
asoldano: Discussed this too, Jim might propose an additional test later.

coxon:
hello, i update resteasy to 3.0.23.Final, but i still get this exception. if you need, i can post exception.

asoldano:
@coxon , sure, please share as much details as you can.
Thanks

coxon:
@asoldano  ok, i try to describe what i faced clearly.  
i used openstack4j to download objects, and openstack4j send http-post via reateasy. ApacheHttpClient4Executor-> ClientResponse execute(ClientRequest request) and i get Stream closed exception throws out.
it's happend occasionally,  so next time when exception  throws , i will post it.

PR #1161, opened 2017-05-17
[RESTEASY-1596] deleteOnExit on temp file only when the Cleaneable me…

asoldano:
...chanism eventually fails in deleting it

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/DataSourceProvider.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/DataSourceProvider.java
@@ -9,8 +9,6 @@ import org.jboss.resteasy.resteasy_jaxrs.i18n.*;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.util.NoContent;
 
-import javax.xml.transform.stream.StreamSource;
-
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileInputStream;
@@ -114,7 +112,6 @@ public class DataSourceProvider extends AbstractEntityProvider<DataSource>
          int count = in.read(buffer, 0, buffer.length);
          if (count > -1) {
              tempFile = File.createTempFile("resteasy-provider-datasource", null);
-             tempFile.deleteOnExit();
              FileOutputStream fos = new FileOutputStream(tempFile);
              fos.write(buffer, 0, count);
              try
@@ -250,7 +247,10 @@ public class DataSourceProvider extends AbstractEntityProvider<DataSource>
 	   {
 		   if(tempFile.exists())
 		   {
-			   tempFile.delete();
+			   if (!tempFile.delete()) //set delete on exit only if the file can't be deleted now
+			   {
+			      tempFile.deleteOnExit();
+			   }
 		   }
 	   }
    }


rsearls:
We have a test in Resteasy for this, EntityBufferingInFileTest.  Currently it is not deploying because it relies on
org.jboss.resteasy.resteasy-legacy which is not present on the server.   
     Caused by: org.jboss.modules.ModuleNotFoundException: org.jboss.resteasy.resteasy-legacy 

working on getting this to run.

asoldano:
@rsearls can you expain better what you mean? I see two EntityBufferingInFileTest files in the sources (one in testsuite/integration-tests, one in testsuite/legacy-integration-tests) and both are part of the build, hence passing. Is there anything to be enabled in those tests, or do they have to be enriched to cover this ?

rsearls:
@ asoldano  There are 2 tests with identical names, 
./testsuite/legacy-integration-tests/src/test/java/org/jboss/resteasy/test/client/EntityBufferingInFileTest.java
./testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/EntityBufferingInFileTest.java

The one in legacy-integration-tests is encountered first.  It fails because of the legacy package.
Moving this test a side.  The other one runs.  
I'm still try to confirm the desired code is being executed.

asoldano:
@rsearls ok, given the tests passed here though, I'm wondering if you have local changes on them or what.

rsearls:
@ asoldano
531 ApacheHttpClient4Engine line 531 always succeeds in deleting the
file so the warning message is never written.

I ran the test with this

mvn clean verify \
  -Dserver.home=/home/rsearls/j1/wfly10/wildfly/build/target/wildfly-11.0.0.Beta1-SNAPSHOT \
  -fn -Dtest=EntityBufferingInFileTest \
  -Dmaven.test.redirectTestOutputToFile=false \
  -Djava.io.tmpdir=/tmp/foofoo  \
  -Dmaven.surefire.debug

asoldano:
@rsearls I've manually tested the fix works with the following procedure:
* set a debugger breakpoint on line 531 of ApacheHttpClient4Engine
* delete the file using a console
* step forward with the debugger and let the code fail in deleting the file (as it's not there)
* re-create the file that was previously there, using a console
* disconnect the debugger and let the test finish
* verify the file is not there anymore (as the deleteOnExit causes it to be removed then the JVM shuts down).

As for the legacy test failing, my bet is that's because the server instance you're explicitly telling the testsuite to use is not properly patched to enable the legacy stuff. In this case, to make it easy, simply avoid setting the server.home here and let the testsuite download, patch and run the server he wants.

kanovotn:
@asoldano The issue here is that file.deleteOnExit() keeps the reference even when File.delete() succeeds is it correct?

Maybe another option could be to use try-with-resources with with custom file handler implementing Autoclosable (desribed here - http://stackoverflow.com/a/34050507). With this we could avoif the  usage of deleteOnExit() at all. Just a thought.

asoldano:
@kanovotn, yes the issue being fixed here is that deleteOnExit() is always called, as soon as the file is created. So, despite the file being later deleted (unless something wrong happens on the disk), we still have consumed memory that won't be freed till the JVM goes down (there's basically a static LinkedHashSet<String> instance storing the path of the files marked as to be deleted on exit). This is basically a leak.

This said, we already have a mechanism to figure out when to delete those files (see the CleanableSequenceInputStream), so i believe there's no need for additional AutoCloseable abstraction.
I could have simply just removed the deleteOnExit(), actually that's what I did initially; then I thought it would have been better to keep the deleteOnExit() as a safety net, to be used just in case when we go and delete the file, we fail in deleting it, so we give the system a chance to try performing the cleanup just before shutting down. But that's an exceptional scenario.

kanovotn:
@asoldano  ok, thank you for explanation.

PR #1167, opened 2017-05-18
[RESTEASY-1655] Allow RESTEasy Spring Boot starter to provide a custom async dispatcher

fabiocarvalho777:
Allowing ResteasyDeployment to use previously set async dispatcher
This is necessary for RESTEasy Spring Boot starter
See https://issues.jboss.org/browse/RESTEASY-1655
See paypal/resteasy-spring-boot#61


asoldano:
Hi, I've merged your PR, thanks. Actually, despite the changes being minor and considering we didn't do this for your previous PR either, it would be better if you could formally agree here that your contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0.
Do you agree to these terms?

fabiocarvalho777:
I agree to those terms. Thanks.

PR #1168, opened 2017-05-19
[RESTEASY-1238]

ronsigal:
Avoid ERROR level stacktraces when response is commited before Exception
can be written.


asoldano:
I'm replacing this PR with https://github.com/resteasy/Resteasy/pull/1170, where I basically excluded your new test on WildFly 10.0.0.Final. I believe the web layer behaves differently/erroneously there.

PR #1176, opened 2017-05-25
[RESTEASY-1599] refactor code to allow user to subclass generation of…

rsearls:
… ClientHttpEngine

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/HTTPClientVersionCheck.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/HTTPClientVersionCheck.java
@@ -1,0 +1,34 @@
+package org.jboss.resteasy.client.jaxrs;
+
+/**
+ * This is a workaound.  For backward compatibility Apache client APIs pre-4.3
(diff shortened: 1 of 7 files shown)


asoldano:
@rsearls thanks for the PR. However, I've had a deeper look and I think we can do better than this, so I've prepared another PR (https://github.com/resteasy/Resteasy/pull/1182) that builds on top of yours. Please review it, its changes are:
* refactored the new builder interface to be a real builder; to be honest we actually don't even need this interface, but it makes things a little bit clearer
* removed the setHttpClientBuilder method from the interface, especially as that does not make sense for the old builder implementation
* factored out the few lines of code that create the HttpClient into a _protected ClientHttpEngine createEngine(...)_ method that the user can override, instead of having to override a 150+ lines method
* renamed HttpClientEngineBuilder* into ClientHttpEngineBuilder* (as they are builders of ClientHttpEngine classes)
* changed the VerifierWrapper class from package protected to public, otherwise we could have left it where it was before (the user is not going to use our own package)

rsearls:
PR (#1182) changes look good.  I'll with draw this PR

PR #1178, opened 2017-06-01
Resteasy 1662

NicoNes:
@ronsigal Another pull request about String marshalling for String based @*Param.

This pull request is about enabling String marshalling for String based @*Param (@QueryParam, @MatrixParam...) into char primitive type or Character type as already done  for other primitive/wrapper types.


asoldano:
Hi @NicoNes ,
I've reviewed your PR and it looks fine, thanks. I'm only going to update a bunch of "@tpSince RESTEasy 3.0.16" with proper versions, after having merged the PR.
Btw, I think we forgot to deal with this previously, sorry, but we require that contributions are made under the terms of the ASL2.0 License: http://www.apache.org/licenses/LICENSE-2.0. Do you agree to these terms? (please explicitly comment below). Thanks.

NicoNes:
Hi @asoldano,

Yes I agree to the  terms of the licence 👍.

PR #1179, opened 2017-06-02
[RESTEASY-1659] changed the algorithm for HEADER putAll and added tes…

rsearls:
…t for change

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseHeaders.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseHeaders.java
@@ -150,11 +150,32 @@ public class HttpServletResponseHeaders implements MultivaluedMap<String, Object
       for (Map.Entry<? extends String, ? extends List<Object>> entry : map.entrySet())
       {
          List<Object> objs = entry.getValue();
-         for (Object obj : objs)
-         {
-            add(entry.getKey(), obj);
+         // When the header key does not exist, undertow creates it and saves this value.
+         // When the header key exists undertow clears the existing values and saves this value.
+         // All subsequent values must be added so they will be appended properly.
+         if (!objs.isEmpty()) {
+            putResponseHeader(entry.getKey(), objs.get(0));
+            for (int i = 1; i < objs.size(); i++) {
+               add(entry.getKey(), objs.get(i));
+            }
          }
       }
+   }
+
+   protected void putResponseHeader(String key, Object value)
+   {
+      RuntimeDelegate.HeaderDelegate delegate = factory.getHeaderDelegate(value.getClass());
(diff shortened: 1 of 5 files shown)


rsearls:
see detailed comments about this proposed fix here, https://issues.jboss.org/browse/RESTEASY-1659

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseHeaders.java
```diff
@@ -150,13 +150,32 @@ public void putAll(Map<? extends String, ? extends List<Object>> map)
       for (Map.Entry<? extends String, ? extends List<Object>> entry : map.entrySet())
       {
          List<Object> objs = entry.getValue();
-         for (Object obj : objs)
-         {
-            add(entry.getKey(), obj);
+         // When the header key does not exist, undertow creates it and saves this value.
+         // When the header key exists undertow clears the existing values and saves this value.
+         // All subsequent values must be added so they will be appended properly.
+         putResponseHeader(entry.getKey(), objs.get(0)); // rls test
```
asoldano: This is going to fail in case we get an empty list (yes, an edge case, but still... a check on the size would probably be safe). Or perhaps an empty list has a special meaning?

asoldano: @ronsigal ^ FYI

rsearls: I tried making this adjustment.  This change breaks the return headers count.  Each object in the header input list is registered as an unique array element in undertow.  Submitting our list as a single
string breaks the expected header count upon return of headers.  Parsing the string back into a list on
our side does not seen like a good idea to me.
@ronsigal @asoldano

asoldano: @rsearls ok, I fine, thanks for the verification and explanation. Then please just protect against the case of empty list. Thanks.

rsearls: @asoldano Complete. check for empty list.

dmak:
Thanks for the fix @rsearls.
How is it possible to see comments in RESTEASY-1659? Even being logged-in I don't see any comments, however I receive them by email.

> Does anyone know if a distinction is being made?

As you have correctly noted, if one needs to append he should call `addAll()`.

rsearls:
Hi @dmak.   We have a way of tagging comments for employees only.  I've untagged my description of the
analysis of the cause of the bug.  Check the jira and you should see it.

PR #1181, opened 2017-06-05
[RESTEASY-1663] Reduce memory allocation for better scalability

asoldano:
https://issues.jboss.org/browse/RESTEASY-1663


asoldano:
Tests on the bench environment show improved performance, merging.

PR #1183, opened 2017-06-06
Update Interceptors.xml

wattdave:
The custom attribute used for annotating filters and interceptors must have @Retention(RententionPolicy.RUNTIME) in order to actually be seen by the RestEasy runtime. This commit makes that clear in the documentation.


asoldano:
Thanks!

PR #1193, opened 2017-06-27
[RESTEASY-1672]:Select highest priority provider if there are more su…

jimma:
…itable MessageBodyReaders/MessageBodyWriters


asoldano:
Thanks @jimma .
I've merged the changes in https://github.com/asoldano/Resteasy/tree/jsr370 for now; that will eventually be merged into master in the next future.

PR #1189, opened 2017-06-20
Allow sub-resouce locators to return class

jimma:

PR #1188, opened 2017-06-16
RESTEASY-1668 Allow sub-resource locators to return classes as well.

liweinan:
https://issues.jboss.org/browse/RESTEASY-1668

PR #1195, opened 2017-06-28
[RESTEASY-1668]

asoldano:
This merges https://github.com/resteasy/Resteasy/pull/1188 and https://github.com/resteasy/Resteasy/pull/1189 , with the aim of keeping both PR tests.
I had to modify the netty4 test from PR 1188, so please have a look and confirm that's fine (basically the test was not valid) or we should add a special case impl.


asoldano:
I've merged the changes in https://github.com/asoldano/Resteasy/tree/jsr370 for now; that will eventually be merged into master in the next future.

PR #1201, opened 2017-06-30
[RESTEASY-1669] Deprecating ApacheHttClient*Engine constructors allow…

asoldano:
…ing to specify a HttpContext instance and explaining potential issues with that approach in the javadoc; added a new HttpContextProvider abstraction to allow users controlling the HttpContext lifecycle.


rsearls:
The changes look fine to me.

asoldano:
After having reasoned on the jira comments and talked with the user, I'm merging. Thanks for the review.

PR #1207, opened 2017-07-14
[RESTEASY-1686] MediaType negotiation issues

asoldano:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
```diff
@@ -403,49 +418,61 @@ public int compareTo(SortableMediaType o)
...
+      Class<?> wt1 = m1 instanceof SortableMediaType ? ((SortableMediaType)m1).writerType : null;
+      Class<?> wt2 = m2 instanceof SortableMediaType ? ((SortableMediaType)m2).writerType : null;
+      return mostSpecific(m1, wt1, m2, wt2);
+   }
+   
    /**
     * m1, m2 are compatible
     */
-   private static SortableMediaType mostSpecific(MediaType m1, MediaType m2)
+   private static SortableMediaType mostSpecific(MediaType m1, Class<?> wt1, MediaType m2, Class<?> wt2)
```
NicoNes: Should this method not be the S fucntion defined by spec section 3.7.2 ?

- if m1="*/xml" and m2="application/*" => the method wil return "application/*"  instead  of "application/xml;d=2"
- if m1="text/*;q=0.5;" and m2="text/html;qs=0.8" => the method will return "text/html;qs=0.8" instead of "text/html;q=0.5;qs=0.8;d=1"

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
```diff
@@ -336,12 +339,24 @@ public static void commitHeaders(BuiltResponse jaxrsResponse, HttpResponse respo
    {
       double q = -1;
```
NicoNes: This SortableMediaType looks like the result of the S() function of spec section 3.7.2. If so both "q" and "qs" default value should be set to 1 :

> q and qs are assumed to be 1:0 if absent.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
```diff
@@ -336,12 +339,24 @@ public static void commitHeaders(BuiltResponse jaxrsResponse, HttpResponse respo
...
+      
+      public SortableMediaType(String type, String subtype, Map<String, String> parameters, Class<?> writerType)
+      {
+         this(type, subtype, parameters);
+         this.writerType = writerType;
+      }
+      
       public SortableMediaType(String type, String subtype, Map<String, String> parameters)
       {
          super(type, subtype, parameters);
```
NicoNes: "qs" should be evaluated whatever "q" value

NicoNes: About the `compareTo()` method:
- Lets consider media types "application/xml" and "text/*". According to the spec "application/xml" > "text/*" since n/m > n/* > */* . Even if media types are not compatible the comparaison should not be based on "qs" nor "q" since n/m > n/* > */*.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -1887,29 +1886,40 @@ public MediaType getConcreteMediaTypeFromMessageBodyWriters(Class type, Type gen
...
+               mbwc = mbwc.getSuperclass();
+            }
+            Class writerType = Types.getTemplateParameterOfInterface(mbwc, MessageBodyWriter.class);
+            if (writerType == null || !writerType.isAssignableFrom(type)) continue;
             Produces produces = mbw.getClass().getAnnotation(Produces.class);
             if (produces == null) continue;
             for (String produce : produces.value())
             {
                MediaType mt = MediaType.valueOf(produce);
                if (mt.isWildcardType() || mt.isWildcardSubtype()) continue;
```
NicoNes: Why are we filtering `MessageBodyWriter ` with produces media types `isWildcardType() || isWildcardSubtype())` ?

Suppose that the client is asking for media type "x/y" and that the only `MessageBodyWriter` able to handle it is the one annotated with `@Produces("x/*")` it will be filtered.

To my mind it should be accepted and then when computing for most specific media types between "x/y" and "x/*" it will be "x/y".

asoldano:
@NicoNes , thanks for the feedack, will reply to your questions/comments later.
In any case, I'm currently working on a revisited implementation of the Secton 3.8 of the spec, you can check the WIP at https://github.com/asoldano/Resteasy/commits/RESTEASY-1686-revisit .

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
asoldano: This check on wildcards is gone in the new proposal, thanks.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
asoldano: I think I have to take my (well, ours) words back with respect to the S function in 3.7.2 being the same as the one mentioned in 3.8...or at least I'm not fully convinced anymore. The behavior that previously seemed to be explained by the S functions being the same was actually caused by a bug that's fixed in the new proposal branch, so I'm leaning towards not changing the SortableMediaType because of this.

asoldano:
This is replaced by https://github.com/resteasy/Resteasy/pull/1211

PR #1217, opened 2017-07-20
[RESTEASY-1686] Adding proper processing of q / qs parameters in S fu…

asoldano:
…nction

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
@@ -400,24 +400,14 @@ public class ServerResponseWriter
    
    private static class SortableMediaType extends MediaType implements Comparable<SortableMediaType>
    {
-      double q = -1;
-      double qs = -1;
+      double q = 1;
+      double qs = 1;
       Class<?> writerType = null;
-      
-      public SortableMediaType(MediaType m, Class<?> writerType)
-      {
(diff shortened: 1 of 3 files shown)


rkadle:
@asoldano - 

We are seeing media types mismatch (reported in RESTEASY-1686, RESTEASY-1227) in 3.0.24.Final versions as well.

We have lot of existing APIs with following @Produces declarations:

`@Produces({ "application/xml", "application/json" })`

We started receiving ` application/xml ` responses for a variety of following Accept headers. From additional testing it appears that, response type is always the first one declared in ` @Produces ` annotation. If I change order in ` @Produces ` annotation above, I am seeing ` application/json ` response format where earlier versions retuned ` application/xml ` responses.

- application/json, application/xml, text/json, text/x-json, text/javascript, text/xml
- application/json, application/xml, text/json, text/x-json, text/javascript, text/xml, application/json, text/json, text/x-json, text/html

Reaching out to see if there is any proposed fix for this in upcoming 3.0.x.Final version.

asoldano:
@rkadle , you might want to create a jira with the details of the defect you're seeing and the reason why you believe it's really a defect / bug (basically the change in behavior, even if unpleasant, might possibly also be due to a bug that's been fixed). 
Generally speaking, a lot of thinking has been done before coming to the fixes that you mentioned, so we really need to see the reasoning behind the claim. Thanks.

rkadle:
@asoldano - Thanks for getting back on this.

I created following JIRA to track this behavior change: https://issues.jboss.org/browse/RESTEASY-1888

Please review and let me know if additional info needed.

PR #1221, opened 2017-07-25
[RESTEASY-1672]:Select highest priority provider if there are more su…

jimma:
…itable MessageBodyReaders/MessageBodyWriters


asoldano:
@jimma isn't this already included through https://github.com/resteasy/Resteasy/commit/fb5e2032aab2f12433b7eb86cd89914e8d817147 ?

PR #1228, opened 2017-07-27
Resteasy async client

liweinan:


liweinan:
Still not finished yet, put it here to record the progress and for reference.

liweinan:
SpecialResourceTest.test583() failed because of httpcomponents minor version update (It uses stricter mime type checking, so the test need to be rewritten. Currently it is ignored)

liweinan:
AsyncTimeoutExceptionsTest.futureTimeoutAndMoreRequestsTest() won't timeout when async client is used.

liweinan:
AsyncTimeoutExceptionsTest .invocationCallbackTimeoutAndMoreRequestsTest() won't timeout when async client is enabled.

liweinan:
Current progress:

- Migrated Markus PR
- Made async-client fully optional (When async-client is in classpath, it will be used, or else it will fallback to the existing asyncExecutor)
- legacy-tests test with out async-client
- non legacy integration tests are tested with async-client now

ronsigal:
This is pretty cool stuff. I've written a test that uses the NIO client, and it seems to be kinda long:
```
   @Test
   public void testNIO() throws Exception {
      HttpAsyncClientBuilder asyncClientBuilder = HttpAsyncClientBuilder.create();
      CloseableHttpAsyncClient httpAsyncClient = asyncClientBuilder.build();
      ClientHttpEngine engine = new ApacheHttpAsyncClient4Engine(httpAsyncClient, true);
      ResteasyClientBuilder clientBuilder = new ResteasyClientBuilder();
      clientBuilder.httpEngine(engine);
      Client client = clientBuilder.build();
      WebTarget target = client.target("http://localhost:8081/test");
      Builder builder = target.request();
      AsyncInvoker asyncInvoker = builder.async();
      Future<Response> future = asyncInvoker.get();
      Response response = future.get();
      String report = response.readEntity(String.class);
      System.out.println("response: " + report);
      client.close();
   }
```
Obviously, I'm not taking advantage of the fluent API, which was just for my own understanding, but building the Client takes a lot of work. Is there an easier way?

-Ron

ronsigal:
How about something like ResteasyClientBuilder.buildAsync() that creates an ApacheHttpAsyncClient4Engine?

asoldano:
I'm working on this at https://github.com/resteasy/Resteasy/pull/1249 ...

PR #1230, opened 2017-07-28
[RESTEASY-1671] checked for null inputStream

rsearls:


rsearls:
1. I was never able to create a test case to generate the exception.  I tried to create a testcase that uses the default wfly DB.  I was not able to find a way to programmatically  turn off the ref to the DB.  I will look further 
into creating some type of test for this.

2. I have rebased the PR to the current master.

rsearls:
A test has been added

PR #1239, opened 2017-08-07
Lazy HTTP 200 response from @Produces(MediaType.SERVER_SENT_EVENTS) a…

asoldano:
…nnotated resource methods

This is to be improved before merging; it's using the Cleanable abstraction as a way to run code just after the invoker is done, we most likely need a better solution.
The fix to the state machine in SseEventSourceImpl is not strictly required, but still desirable imho.


asoldano:
OK, considering the review, I'm merging this for now, to further stabilize test runs, but we should consider improving the implementation. Using the Cleanable for this issue, despite working, still looks really a misuse to me ;-)

PR #1242, opened 2017-08-08
Resteasy 1699: Fix potential NPE

NicoNes:


NicoNes:
Hi @asoldano,

Sorry for the 15 commits I'll reset them. About the change you suggested, I don't understand your concern since according to me both `getClassz().hashCode()` and `classz.hashCode()` return the same hash code (if the weak reference has not be cleared of course). Am I wrong ?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
```diff
@@ -272,7 +272,7 @@ private CachedMediaTypeAndClass(Class clazz, MediaType mediaType)
       {
          this.clazz = new WeakReference(clazz);
          this.mediaType = mediaType;
-         int result = getClazz().hashCode();
+         int result = clazz.hashCode();
```
NicoNes: In `clazz.hashCode()` `clazz `refers to the 1st argument of type `Class` and not to `this.clazz` of type `WeakReference<Class<?>>`. Your concern was maybe because of this ?

asoldano: Yes. The hash code of the reference object is not the same of the Class that's referenced within the reference. CachedMediaTypeAndClass seems to be designed to that its hash code is the same of the Class that's passed when building it. And the reason for that is most likely related to fact that CachedMediaTypeAndClass instances are being added to a hashmap. By using the hash code of every new WeakReference that's built when creating a new CachedMediaTypeAndClass, you would basically alter the behavior.

NicoNes: Totally agree but what I am saying is that the `clazz` I'm using in `int result = clazz.hashCode();` refers to `Class clazz` argument and not `WeakReference<Class<?>> clazz` property. In other words:
`int result = clazz.hashCode()` is not the same as `int result = **this**.clazz.hashCode()` where `this.clazz` refers to `WeakReference<Class<?>> clazz` in this case.

I should have renamed the variable to be clearer like this:

```
...
      // we need a weak reference because of possible hot deployment
      // Although, these reference should get cleared up with any add() invocation
private WeakReference<Class<?>> clazzReference;

private CachedMediaTypeAndClass(Class clazz, MediaType mediaType){
         this.clazzReference= new WeakReference(clazz);
         this.mediaType = mediaType;
         int result = clazz.hashCode();
     ...
}
```

asoldano: Ah, sorry, I missed that the parameter for the constructor has the same name. So, at this point, I'm fine. Thanks.

asoldano: Ah, sorry, I missed that the parameter for the constructor has the same name. So, at this point, I'm fine. Thanks.

PR #1244, opened 2017-08-09
Resteasy 1700

NicoNes:


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -14,6 +14,7 @@
 import org.jboss.resteasy.util.InputStreamToByteArray;
 import org.jboss.resteasy.util.ReadFromStream;
 import org.jboss.resteasy.util.Types;
+import org.omg.IOP.IOR;
```
asoldano: This is not needed, is it?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -70,12 +71,38 @@ public void setClientConfiguration(ClientConfiguration configuration)
...
+	@Override
+	public Object getEntity() {
+		abortIfClosed();
+		Object entity = super.getEntity();
+		if (entity != null) {
+			return entity;
+		}
+		//Check if the entity was previously fully consumed
+		try {
+			InputStream inputStream = getEntityStream();
```
asoldano: The call to getEntityStream() here is possibly a problem. Basically, we need to make sure that method is not called while a ClientResponse object is garbage collected as a consequence of finalize method having been aggressively called. So far, we didn't bother about this problem because the only invoker to that protected method is already synchronized (readFrom(...) ). In any case, it would probably be safe to simply turn getEntityStream() into a synchronized method.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
NicoNes: Oups... useless you're right

NicoNes:
Yes, this PR addresses what's written in javadoc of  `javax.ws.rs.core.Response.getEntity()`

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
NicoNes: Oh thanks, I did not pay attention to that.
However how can it be possible for the garbage collector to call `finalize` while another thread is calling `readFrom()` on the object being finalized ? I mean, `finalize` should be invoked on the object only if there is no reference on this object anymore. Am I missing something ?

asoldano: Unfortunately, the implementation of the finalizer mechanism is somehow bugged. I should go and verify this specific scenario and to tell the exact condition that would cause an error (that's why I've said "possibly a problem" ;-) ), but we've been suffering from similar issues, some of which led to making those other methods synchronized there. If you want to look at an interesting video on this topic, check https://www.youtube.com/watch?v=UrGP6pfb0H8

NicoNes: Interesting, I 'll have a look at it thanks 👍 .
So ok I'll  turn the `getEntity()` into a synchronized method.

asoldano: Makes sense, thanks

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -70,12 +71,38 @@ public void setClientConfiguration(ClientConfiguration configuration)
...
+	@Override
+	public synchronized Object getEntity() {
+		abortIfClosed();
+		Object entity = super.getEntity();
+		if (entity != null) {
+			return entity;
+		}
+		//Check if the entity was previously fully consumed
+		try {
+			InputStream inputStream = getEntityStream();
```
ronsigal: To be honest, I've never really understood the purpose or meaning of getEntity(), but this does seem to be correct.

ronsigal:
I remember why I used to think Response.getEntity() **might** be correct. The javadoc for Response.readEntity() says,

>A message instance returned from this method will be cached for
>subsequent retrievals via {@link #getEntity()}.

I interpreted that to **imply** that readEntity() had to be called before getEntity(). 

I don't know ...

asoldano:
@ronsigal @NicoNes , I've cleaned up this PR into https://github.com/resteasy/Resteasy/pull/1283 and I'm leaning towards merging that. Any comment / concern?

asoldano:
OK, I've merged #1283 .

PR #1247, opened 2017-08-18
Add a cache of MediaType objects

asoldano:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
```diff
@@ -7,17 +7,35 @@
...
+
+      if (map == null) return parse(type);
+      MediaType result = map.get(type);
+      if (result == null)
+      {
+         result = parse(type);
+         final int size = map.size();
+         if (size >= MAX_MT_CACHE_SIZE)
+         {
+            map.clear();
```
rsvoboda: Is full map clear really necessary ?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
```diff
@@ -7,17 +7,35 @@
...
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
  * @version $Revision: 1 $
  */
 public class MediaTypeHeaderDelegate implements RuntimeDelegate.HeaderDelegate
 {
+   private static final int MAX_MT_CACHE_SIZE = Integer.getInteger("org.jboss.resteasy.plugins.delegates.max_mediatype_cache_size", 200);
+   private static final Map<String, MediaType> map = MAX_MT_CACHE_SIZE > 0 ? new ConcurrentHashMap<String, MediaType>() : null;
```
rsvoboda: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html#ConcurrentHashMap(int)

initial size could be set to the limit

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
rsvoboda: could be something like https://stackoverflow.com/a/36255034/6266477 used ?

asoldano:
@rsvoboda I know full eviction of the cache is unfair and looks simplistic; in facts it's really just the most straightforward safety belt for avoiding explosion of the cache. A LRU approach would be better, assuming it's efficiently implemented. Let's see the benefits of the current solution first and then evaluate improvements, ok?

rsvoboda:
+1 for simple solution now + evaluate improvements

asoldano:
Closing for now as it looks like the improvements are not as great as expected. Might re-evaluate in future.

PR #1248, opened 2017-08-18
[RESTEASY-1680]:test sseEventSource with try-with-resources

jimma:


jimma:
I'd like to only change try-with-resources in testAddMessage() and leave others to test explicitly close SseEventSource.

asoldano:
OK, fine

PR #1249, opened 2017-08-18
Async client

asoldano:
Followup from https://github.com/resteasy/Resteasy/pull/1228.
PR 1228 description pretended the new async/NIO engine is enabled by default when the libs are available, however that was not the case, given the new engine was not actually being constructed anywhere. I tried defaulting to the new engine, but that clearly does not fit most of the common usage scenario, as for example built-in providers can ask for blocking reads in a thread handling IO. So I think it's better to let the use decide to build a new engine and set it in the client when it makes sense.

### DIFF
--- a/resteasy-client/pom.xml
+++ b/resteasy-client/pom.xml
@@ -28,11 +28,11 @@
             <artifactId>resteasy-jaxrs-services</artifactId>
             <version>${project.version}</version>
         </dependency>
-        
+
         <dependency>
             <groupId>org.jboss.logging</groupId>
             <artifactId>jboss-logging</artifactId>
-       </dependency>
+        </dependency>
         <dependency>
             <groupId>org.jboss.logging</groupId>
             <artifactId>jboss-logging-annotations</artifactId>
@@ -46,12 +46,16 @@
             <artifactId>httpclient</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpasyncclient</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.spec.javax.ws.rs</groupId>
             <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
         </dependency>
     </dependencies>
-
-   <profiles>
+    <profiles>
         <profile>
             <id>i18n</id>
             <activation>
@@ -64,7 +68,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-resources-plugin</artifactId>
-                       <executions>
+                        <executions>
                             <execution>
                                 <id>copy-resources</id>
                                 <phase>initialize</phase>
@@ -72,7 +76,9 @@
                                     <goal>copy-resources</goal>
                                 </goals>
                                 <configuration>
-                                    <outputDirectory>${basedir}/src/main/resources/org/jboss/resteasy/client/jaxrs/i18n</outputDirectory>
+                                    <outputDirectory>
+                                        ${basedir}/src/main/resources/org/jboss/resteasy/client/jaxrs/i18n
+                                    </outputDirectory>
                                     <resources>
                                         <resource>
                                             <directory>${basedir}/src/test/resources/i18n</directory>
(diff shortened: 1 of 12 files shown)


asoldano:
@ronsigal , I've been thinking about your comment at https://github.com/resteasy/Resteasy/pull/1228#issuecomment-322926434 (and the following one). The problem I see is configuration of the asyn engine and perhaps picking sensible defaults for them. I thought I could add an AsyncClientHttpEngineBuilder to play the same role of ClientHttpEngineBuilder4(3) but still I don't know which of the engine parameters are really relevant enough to justify explicit addition to the RESTEasy public api. WDYT? With the fluent API we'd have something like this...

CloseableHttpAsyncClient asyncClient = HttpAsyncClientBuilder.create().setMaxConnTotal(5).build();
return new ResteasyClientBuilder().httpEngine(new ApacheHttpAsyncClient4Engine(asyncClient, true)).build();

ronsigal:
@asoldano , your code looks like mine only fluent.

I'd say, why not do both? For example, there's org.apache.http.impl.nio.client.HttpAsyncClients.createDefault():

```
    /**
     * Creates {@link CloseableHttpAsyncClient} instance with default
     * configuration.
     */
    public static CloseableHttpAsyncClient createDefault() {
        return HttpAsyncClientBuilder.create().build();
    }
```
for people (like me) that don't know so much about HttpClient.

asoldano:
@ronsigal ok, how about https://github.com/resteasy/Resteasy/pull/1249/commits/e93b2c6b2ff6e5020c6f2687e22337e7bed66d8a#diff-e8190e0a45f9a8246df450575a55385cR64 ?

asoldano:
@ronsigal sure, quite a lot of doc needs to be written for this feature, especially on when to use the new engine and why. Yes, the createSystem looked like an easy way for letting users configure the client without doing to much in RESTEasy ;-)

ronsigal:
@asoldano 

re: "sure, quite a lot"

D'oh. Excuse my totally useless comment. ;-)

asoldano:
@ronsigal nothing to excuse :-) Actually it looks like the change is causing test failures here on Travis, investigating that soon...

jimma:
@asoldano , I was thinking how can we configure the thread pool/ExecutorService to HttpAsyncClient,  and it looks the executor created in ResteasyClientBuilder can't be used to configure HttpAsyncClientEngine and we can only configure the IOReactor's thread count,  so we don't need to create executor if the engine is HttpAsyncClientEngine: https://github.com/resteasy/Resteasy/blob/master/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java#L394 . WDYT ?

asoldano:
@jimma interesting and tempting optimization, however at the moment I see that executor is also used for serving the rx(..) methods in ClientInvocationBuilder... right?

asoldano:
@jimma, thanks for thinking about possible enhancement here, I appreciate this. So what you're proposing is to leverage the new async engine for improving the current implementation of rx(), right? That looks interesting to me, even if I'm not that sure about possible side effects of this. How about we move this specific part of the discussion to a resteasy dev list thread? You might want to start it as soon as we have the initial async engine PR here merged (so that people can look at the code more easily). I'm currently trying to nail down a test regression here, so this PR is a bit of a mess atm, I'll clean the latest commits later.

jimma:
OK. I'll start the discussion in the dev list after this is merged.

asoldano:
I've added an excluded test for comparing performances of the old and new client engine. Feel free to enable it locally, give it a run and let me know what you think. In particular, I'd like confirmation on it actually performing the comparison in correct conditions (same number of connections allowed).

asoldano:
OK, I've merged the PR. Will work on doc before marking jira as solved. Thanks everybody for the feedback so far, keep testing / playing with this stuff ;-)
@MarkusKull , @liweinan FYI

asoldano:
> @jimma, thanks for thinking about possible enhancement here, I appreciate this. So what you're proposing is to leverage the new async engine for improving the current implementation of rx(), right? That looks interesting to me, even if I'm not that sure about possible side effects of this. How about we move this specific part of the discussion to a resteasy dev list thread? You might want to start it as soon as we have the initial async engine PR here merged (so that people can look at the code more easily). I'm currently trying to nail down a test regression here, so this PR is a bit of a mess atm, I'll clean the latest commits later.

Many months later... RESTEASY-2196 was created and I've just finished working with the reporter to fix it. It's related to the discussion here, so here is the link: #1990

PR #1253, opened 2017-08-21
[RESTEASY-1680]:Add test for sseEventSink with try-with-resources

jimma:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/AnotherSseResource.java
```diff
@@ -25,24 +25,33 @@
...
-	   }
-	   sseBroadcaster.register(sink);	  
+      synchronized (outputLock)
+      {
+         if (sink == null)
+         {
+            throw new IllegalStateException("No client connected.");
+         }
+      }
+      eventSink = sink;
```
asoldano: So, the line here above (eventSink = sink;) is really a bug fix, as it stores a reference to the sink in the resource, allowing the close() method to actually do something. Nice catch.

asoldano: This said, I think the "synchronized (outputLock)" is not needed, considering you're checking the sink parameter in there.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/AnotherSseResource.java
```diff
@@ -25,24 +25,33 @@
...
   
    @DELETE
-   public void close() throws IOException {
+   @Produces(MediaType.TEXT_PLAIN)
+   public boolean close() throws IOException {
        synchronized (outputLock) {
            if (eventSink != null) {
-               eventSink.close();
-               eventSink = null;
+              try (SseEventSink sink = eventSink) {
```
asoldano: While this is good and it's one of the reasons for this PR, I think you should add a comment saying something like "we could have simply invoked eventSink.close(), verifying auto-closeable functionalities though". This would help users understand the test more easily (the test here is becoming complex).

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
```diff
@@ -135,13 +135,16 @@ public void testBroadcast() throws Exception
...
        Assert.assertTrue("Waiting for broadcast event to be delivered has timed out.", latch.await(20, TimeUnit.SECONDS));
+       
+       Client closeClient = new ResteasyClientBuilder().build();
+       WebTarget closeTarget = closeClient.target(generateURL("/service/sse"));
+       Assert.assertTrue("Subscribed eventsink is not closed", closeTarget.request().delete().readEntity(Boolean.class));
+       
        eventSource.close();
        eventSource2.close();
-       client.close();
-       client2.close();
```
asoldano: can you reassure me on why we don't need these client.close() anymore? Thanks

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/AnotherSseResource.java
jimma: Ok. will add this.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/AnotherSseResource.java
jimma: +1

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
asoldano: Ah, right, thanks.

PR #1260, opened 2017-08-25
[RESTEASY-1706] Add support for jax-rs annotation inheritance with generics.

Syquel:
Adjust the discovery process of overridden resource methods to resolve generic type parameters in method signatures by the actual type at runtime.
* Replace implementation of annotation inheritance discovery in ResourceMethodRegistry by the usage of the implementation in ResourceBuilder.
* Enhance the implementation of annotation inheritance discovery in ResourceBuilder.
* Move general generics logics to and enhance the utility class Types.
JIRA Issue: https://issues.jboss.org/projects/RESTEASY/issues/RESTEASY-1706


Syquel:
@rsearls @asoldano @ronsigal Could someone review the pull request and provide feedback for enhancements, please?

asoldano:
@Syquel thanks for the contribution. Besides from @rsearls note above (which can easily be solved in the IDE), I actually have 2 requests:
1) would you mind adding a testcase for the scenario being fixed here?
2) we require that all contributions be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 . Do you agree to these terms?

Thanks

Syquel:
@rsearls I fixed my import statements and amended my commit.

@asoldano 
1. I tried to implement an Arquillian Integration Test, but got hung up on an issue in the JsonCollectionProvider / ResteasyJackson2Provider which prevents me from successfully calling Invocation.invoke(GenericEntity<Collection<T>>).
Would a plain JUnit Test be sufficient?
2. Yes, I agree.

asoldano:
@Syquel , thanks. Yes, a plain JUnit Test is better than nothing.

Syquel:
@asoldano I added an Arquillian integration test. (I just had to switch from Json to XML MediaType...)
Do you know if you will backport it to Resteasy 3? There should be no conflicts.

asoldano:
@Syquel thanks. As for backporting to RESTEasy 3, it's not decide yet, but for now let's say it's likely.

PR #1263, opened 2017-08-29
[RESTEAY-1665]:Add @Priority to JsonBindingProvider to really enable it in JsonBindingTest

jimma:


testsuite/integration-tests/pom.xml
```diff
@@ -348,6 +348,37 @@
...
+                            <configuration>
+                                <includes>
+                                    <include>**/Test*.java</include>
+                                </includes>
+                                <excludes>
+                                     <!-- Tests requires excluding JsonBindingProvider-->
+                                     <exclude>**/JsonBindingTest.java</exclude>
+                                </excludes>
+                                <classpathDependencyExcludes>
+                                    <classpathDependencyExcludes>org.jboss.resteasy:resteasy-json-binding-provider</classpathDependencyExcludes>
```
asoldano: mmh, I'm not really sure about this. Does it mean that the new json-b provider is going to break the behavior of most apps out there when they upgrade to latest resteasy? If that's really the case (is this a consequence of the @Priority you added?), we should evaluate different solutions (like not having direct and transitive dependencies to the new module by default and adding it explicitly in the json-b test). Do we really have such a non backward compatibility?

jimma:
On Tuesday, August 29, 2017, Alessio Soldano <notifications@github.com>
wrote:

> *@asoldano* requested changes on this pull request.
> ------------------------------
>
> In testsuite/integration-tests/pom.xml
> <https://github.com/resteasy/Resteasy/pull/1263#discussion_r135751384>:
>
> > +        <plugins>
> +            <plugin>
> +                <artifactId>maven-surefire-plugin</artifactId>
> +                    <executions>
> +                        <execution>
> +                            <id>default-test</id>
> +                            <configuration>
> +                                <includes>
> +                                    <include>**/Test*.java</include>
> +                                </includes>
> +                                <excludes>
> +                                     
> +                                     <exclude>**/JsonBindingTest.java</exclude>
> +                                </excludes>
> +                                <classpathDependencyExcludes>
> +                                    <classpathDependencyExcludes>org.jboss.resteasy:resteasy-json-binding-provider</classpathDependencyExcludes>
>
> mmh, I'm not really sure about this. Does it mean that the new json-b
> provider is going to break the behavior of most apps out there when they
> upgrade to latest resteasy?
>
If JsonBindingProvider is in the classpath , it will be picked up and take
precedence. I don't think this will break something if json-b is not
required and not included in classpath.

>  If that's really the case (is this a consequence of the @priority
> <https://github.com/priority> you added?), we should evaluate different
> solutions (like not having direct and transitive dependencies to the new
> module by default and adding it explicitly in the json-b test). Do we
> really have such a non backward compatibility?
>
No. Adding @priority is required to enable jsonb and make jsonb take
precedence if it is in classpath.If you print more info in the JsonBingTest
before my change, you'll find it actually reads and writes with jackson
provider instead of jsonb. I don't think this is something ugly dependency
configration because maven surefire already have this excluding dependency
feature for specific tests.We don't need to creat another maven module with
explicitly decalre jsonb dependency only for one JsonBingTest. WDYT?

asoldano:
@jimma ok, thanks for having brought this to my attention, I missed this before.
So, at the moment I think we should not have the dependency to the new json-b provider module in the existing testsuites, as long as most of the tests in them actually enforce the jackson binding behavior. IMHO, it does not make sense to add the maven dependency in they testsuite module and them exclude it in the surefire configuration for all tests except one. That one (the JsonBindingTest) is currently the "exception" and needs a special treatment (for example a custom surefire execution with additional dependency and some comments in the test referring to that).
At a given point considering json-b has higher priority than jackson (btw, can you link here the reference to the jsr370 spec enforcing this, for future reference) we might want to revisit the tests that are relying on jackson impl by chance (not on purpose, to verify jackson provider details).
Finally, regarding the addition you put in JsonBindingTest ... I see you're expecting the attributes order to be "color, domesticated, name, sort", but I would have expected "color, sort, name, domesticated" judging from the annotation on Cat. What am I missing?

jimma:
> @jimma ok, thanks for having brought this to my attention, I missed this before.
So, at the moment I think we should not have the dependency to the new json-b provider module in the existing testsuites, as long as most of the tests in them actually enforce the jackson binding behavior. IMHO, it does not make sense to add the maven dependency in they testsuite module and then exclude it in the surefire configuration for all tests except one. That one (the JsonBindingTest) is currently the "exception" and needs a special treatment (for example a custom surefire execution with additional dependency and some comments in the test referring to that).

Adding extra dependency for this test can with:
 1) exclude dependency for specific test like what I did now 
 2) create a profile 
 3) create another maven module jsonbinding-integration-tests to include this dependency  

I thinks neither 2) or 3) is good to create only for this test.  Regarding 1) , I don't see any harmful thing we use this maven surefire plugin configuration . Including dependency for this test is another kind of special treatment too if we don't create another standalone maven module for this test. 

> At a given point considering json-b has higher priority than jackson (btw, can you link here the reference to the jsr370 spec enforcing this, for future reference) we might want to revisit the tests that are relying on jackson impl by chance (not on purpose, to verify jackson provider details).

Spec 11.2.7 Java API for JSON Binding under  about the integration with JEE technology says : 
In a product that supports the Java API for JSON Binding (JSON-B) [19], implementations MUST support entity providers for all Java types supported by JSON-B in combination with the following media types: application/json , text/json as well as any other media types matching */json or */*+json .
Note that if JSON-B and JSON-P are both supported in the same environment, entity providers for JSON-B take precedence over those for JSON-P for all types except JsonValue and its sub-types. "

There is nothing else except this about json binding, and there is no item explicitly says about jsonb provider has higher priority than jackson either. If we want to enable json binding when it is included in class path, annotated a @Priority is the easy solution to select json binding provider first for  application/json media type.

> Finally, regarding the addition you put in JsonBindingTest ... I see you're expecting the attributes order to be "color, domesticated, name, sort", but I would have expected "color, sort, name, domesticated" judging from the annotation on Cat. What am I missing?

Nice findings. There is another issue we should export jsonb api in the jsonbinding provider jboss module. I've fixed this.

asoldano:
hi @jimma ,
I'm still not 100% convinced on the way we should be dealing with the json-b provider selection. I think I'll merge your PR as is and open up a JIRA for re-evaluating this later (more details coming on the jira). Thanks.

asoldano:
https://issues.jboss.org/browse/RESTEASY-1714

PR #1265, opened 2017-08-29
[RESTEASY-1703] added subsection in upgrade guide

rsearls:


resteasy-upgrade-guide/reference/en/en-US/modules/Miscellaneous.xml
```diff
@@ -121,4 +121,49 @@ ResteasyProviderFactory.getInstance().registerProvider(POJOConverterProvider.cla
...
+      <para>
+         Improvements and corrections have been made in Resteasy 3 to the implementation of
+         the matching rules as defined in the JAX-RS 2.0 specification.  In particular
+         a change for handling an ambiguous uri on a sub-resource method and
+         sub-resource locator was made.  In Resteasy 2 it was possible for a
+         sub-resource locator to execute successfully even when there was also a sub-resource
+         with the same uri.  This was incorrect behavior according to the specification.
+         In Resteasy 3 when there is an ambiguous uri for a sub-resource and a
+         sub-resource locator.  Calling the sub-resource will be successful.  Calling
+         the sub-resource locator will result in an HTTP, 405 Method Not Allowed error.
```
asoldano: @rsearls, thanks for the PR. I'm probably too picky, but can you fix the punctuation here (the 3 parts of the sentence here should probably be together) ?

rsearls:
Done.

asoldano:
@rsearls thanks, merging :-)

resteasy-upgrade-guide/reference/en/en-US/modules/Miscellaneous.xml
```diff
@@ -121,4 +121,49 @@ ResteasyProviderFactory.getInstance().registerProvider(POJOConverterProvider.cla
...
+      <para>
+         Improvements and corrections have been made in Resteasy 3 to the implementation of
+         the matching rules as defined in the JAX-RS 2.0 specification.  In particular
+         a change for handling an ambiguous uri on a sub-resource method and
+         sub-resource locator was made.  In Resteasy 2 it was possible for a
+         sub-resource locator to execute successfully even when there was also a sub-resource
+         with the same uri.  This was incorrect behavior according to the specification.
+         In Resteasy 3 when there is an ambiguous uri for a sub-resource and a
+         sub-resource locator, calling the sub-resource will be successful, but calling
+         the sub-resource locator will result in an HTTP, 405 Method Not Allowed error.
```
ronsigal: @rsearls , I'm uncomfortable with the phrase "calling the sub-resource locator". How does one call a sub-resource locator? You have to send a URI that matches the sub-resource locator but which does not match a sub-resource method (which would "mask" the sub-resource locator). So, when a sub-resource method and sub-resource locator have the same URI, it isn't possible to call the sub-resource locator. So, I can't make sense of "calling the sub-resource locator will result in an HTTP, 405 Method Not Allowed error". I would say that it is the calling of the sub-resource method with the wrong verb is what leads to the 405 status. WDYT?

resteasy-upgrade-guide/reference/en/en-US/modules/Miscellaneous.xml
```diff
@@ -121,4 +121,49 @@ ResteasyProviderFactory.getInstance().registerProvider(POJOConverterProvider.cla
...
+
+         @POST
+         @Path("items")
+         public Response create(@QueryParam("msg") final String msg) {
+         return Response.ok().build();
+         }
+
+         @Path("items")
+         public List&lt;String&gt; splitPattern(@QueryParam("pattern") final String pattern) {
+         String [] split = pattern.split
```
ronsigal: @rsearls , shouldn't splitPattern() return a resource?

rsearls:
ok.  I'll rewrite the whole thing tomorrow.

rsearls:
@ronsigal, Would you review this text?

PR #1266, opened 2017-08-30
[RESTEASY-1680]:Fix EventHandler.reconnect() and add a reconnect test

jimma:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseResource.java
```diff
@@ -68,6 +68,28 @@ public void addMessage(final String message) throws IOException {
        eventSink.send(event);
    }
    
+   @POST
+   @Path("/addMessageAndDisconnect")
+   public void addMessageAndDisconnect(final String message) throws IOException, InterruptedException {
+        //clear events store first
+        eventsStore.clear();
```
asoldano: This should go into a syncronized (eventStore) block, or we need to use a thread safe collection for the eventStore.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
```diff
@@ -127,15 +129,14 @@ public void testSseEvent() throws Exception
...
-       {})[5].indexOf("Done") > -1);
+       Assert.assertTrue("Expect the last event is Done event, but it is :" + results.toArray(new String[]{})[5], 
+               results.toArray(new String[] {})[5].indexOf("Done") > -1);
        eventSource.close();
        client.close();
     }
     
     
-    @Test
+    @Ignore
```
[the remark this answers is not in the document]
jimma: I ignored it in debug mode and forgot to add it back. I've enabled it.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseResource.java
jimma: This could not be necessary for the test purpose and here are the reasons: 
1. This is to force empty the evens store. 
2. evensStore is only read after eventSink is closed and eventSource is reconnected with a Last-Event-ID header.

asoldano: OK, makes sense

PR #1268, opened 2017-08-30
[RESTEASY-1712] added check and warning msg for ambiguous paths.

rsearls:


rsearls:
OK for review and further consideration.  This is something the customer has requested.  I am ambivalent
about providing it.
@Produces/@consumes are not involved in this part of path matching.

ronsigal:
For the record, one warning about multiple matching resources was due to a stupid coding mistake I made...

asoldano:
@ronsigal , I'm not blaming anybody (I probably even reviewed and merged that PR ;-) ), just saying that experience is suggesting me to carefully think if we could be missing something here, given a false alarm would likely be worse than the really specific issue being detected here. So what do you think about the check impl here?

rsearls:
@ronsigal, just for futher ref here is the initial description of the issue. https://issues.jboss.org/browse/RESTEASY-1703.  There is a subsection in the Upgrade guide about this now.   I don't think this is a
scenario that happens very often, so providing a warning may not be worth the effort of maintaining it. Also
as Alessio said it might report a false positive which would not be good.

ronsigal:
* I'm looking at the algorithm from Section 3.7 of JAX-RS specs 1.0 and 2.1, and I don't see any difference with respect to this issue. I think that, if the behavior described in RESTEASY-1703 is real (I haven't tested it), then Resteasy 2.x has a bug and Resteasy 3.x is correct. But I wouldn't say that "RestEasy 3.x does not allow same path ...". I don't see anything illegal about the example in RESTEASY-1703. It's just that the sublocator gets "masked".

* I think it's a little odd to single out one of the many bugs we fixed in going from Resteasy 2 to Resteasy 3 and print a message saying, "Look, it's fixed!". It seems to me that a note in the upgrade guide is sufficient. 

* On the other hand, I don't think it would be a terrible thing to log a warning for EVERY bug we fixed that leads to a behavior change. It could be something that could get turned on when the user is migrating and turned off when the migration is done. On the other, other hand, it could be rather time consuming to do. But if it sounds worthwhile, we could start now for Resteasy 4.0, if there are any such changes.

* @rsearls , I've added a couple of comments to https://github.com/resteasy/Resteasy/pull/1265.

* Next step: I haven't actually studied the changes in this PR. I will.

-Ron

ronsigal:
re: "Look, it's fixed!"

Well, another way to think about it is that, if we were disposed to issue warnings, this is a place that we could do a warning independent of the fact that there used to be a related bug. We would be warning the developer that the resource is ill-formed because a sub-resource locator is being shadowed by a sub-resource method.

jimma:
I agree with @ronsigal the migration/upgrade guide about this could be enough. 
IMHO, technically speaking this is doable if resteasy matches request with HTTP method. But jaxrs spec defines request matching setps , which is matching the resource method first exclude the resource locator then check if http method is equals, we have to implement like this.

rsearls:
@ronsigal , your thoughts on this rewrite.

ronsigal:
Hi @rsearls , I can't find the new version ...

rsearls:
@ronsigal, sorry.  Wrong PR.  I intended it to be for 1703.  I just sent a review request from that one.

ronsigal:
@rsearls, @jimma, I'm a little bit leaning towards logging the warning, just because it seems user friendly. Suppose someone made the "shadowing" mistake in Resteasy 3.x without having used Resteasy 2.x. Then they wouldn't be reading the upgrade guide.

But, the warning isn't quite complete yet. It turns out that a method in one class can shadow a locator in another class. For example,
```
	@Path("test")
	public static class TestResource1 {

		@GET
		@Path("x")
		@Produces("text/plain")
		public Response method() {
			return Response.ok("ok").build();
		}
	}
	
	@Path("test")
	public static class TestResource2 {

		@Path("x")
		@Produces("text/plain")
		public SubResource locator() {
			return new SubResource();
		}
	}

	public static class SubResource {

		@POST
		@Path("")
		public String sub(String s) {
			return "sub(" + s + ")";
		}
	}
```
But the test here looks at one class at a time, so this one doesn't get caught. But, if I add
```
	@Path("other")
	public static class TestResource3 {

		@GET
		@Path("y")
		public String s() {
			return "ok";
		}

		@Path("y")
		public SubResource t() {
			return new SubResource();
		}
	}
```
and register it
```
	@BeforeClass
	public static void before() throws Exception
	{
		server = new UndertowJaxrsServer().start();
		ResteasyDeployment deployment = new ResteasyDeployment();
		deployment.setDeploymentSensitiveFactoryEnabled(true);
		deployment.setApplication(new TestApp());
		deployment.start();
		deployment.getRegistry().addPerRequestResource(TestResource3.class); // <<===
		server.deploy(deployment);
	}
```
I get 
```
20:37:20.640 WARN  [org.jboss.resteasy.resteasy_jaxrs.i18n] (main) RESTEASY002195: The following sub-resource methods and sub-resource locators have the same path, [other/y].  The paths should be unique.  [org.jboss.resteasy.experiment.undertow.Test1703b$TestResource3.s, org.jboss.resteasy.experiment.undertow.Test1703b$TestResource3.t]
20:37:20.641 WARN  [org.jboss.resteasy.resteasy_jaxrs.i18n] (main) RESTEASY002195: The following sub-resource methods and sub-resource locators have the same path, [test/x].  The paths should be unique.  [org.jboss.resteasy.experiment.undertow.Test1703b$TestResource1.method, org.jboss.resteasy.experiment.undertow.Test1703b$TestResource2.locator]
```
It works because
1. the shadowing in a single class triggers the test, and
2. TestResource1 and TestResource2 are already registered, and the test sees the shadowing in root.getBounded().entrySet().

ronsigal:
@rsearls, github is confusing me. When I go to https://github.com/resteasy/Resteasy/pull/1265 and click on View Changes, I see the old version. Maybe it won't pull in your new version because the PR is closed?

rsearls:
@ronsigal, I revised the code and added a test.   Would you review again?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyDeployment.java
```diff
@@ -475,6 +475,7 @@ public void registration()
       {
          registry.addPerRequestResource(actualResourceClass);
       }
+      registry.checkAmbiguousUri();
```
ronsigal: @rsearls, is there a reason you put the test before processing the ResourceFactories?

ronsigal:
@rsearls, that's really cool that you found a way to test the logging output.

rsearls:
On initial look, it did not seem to be needed.   I've traced that code.  Your are right it the call should be moved after that look.  I have made this change.

ronsigal:
@rsearls, Ok, thanks for that change. I had to re-read the registry / node code again, and it looks like ResourceMethodRegistry.root.getBounded().entrySet() is just what we need in ResourceMethodRegistry.checkAmbiguousUri(), so I approve this PR. Nice!!

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
```diff
@@ -279,6 +280,45 @@ protected void register(ResourceFactory rf, String base, ResourceClass resourceC
...
+                  methodCnt++;
+               }
+            }
+            if (methodCnt > 0 && locatorCnt > 0)
+            {
+               StringBuilder sb = new StringBuilder();
+               for (ResourceInvoker exp : values) {
+                  sb.append(exp.getMethod().getDeclaringClass().getName() + "."
+                          + exp.getMethod().getName()+ ", ");
+               }
```
asoldano: The StringBuilder instance should be used better here. You should be doing sb.append(exp.getMethod().getDeclaringClass().getName()).append(".").append(exp.getMethod().getName()).append(", ); to avoid creating too much useless String instances.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
```diff
@@ -279,6 +280,45 @@ protected void register(ResourceFactory rf, String base, ResourceClass resourceC
...
+
+               LogMessages.LOGGER.uriAmbiguity(entry.getKey(),
+                       sb.toString().substring(0, sb.toString().length()-2));
+            }
```
asoldano: This is also improvable; instead of creating yet a new String to crop the last 2 chars, we should only add the ", " before if not dealing with the last ResourceInvoker value. So please refactor the loop above instead. (btw StringBuilder also have a length() method, which does not requiring building the string)

rsearls:
All changes made.  TI runs successfully.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/Registry.java
```diff
@@ -105,4 +105,6 @@
    void addJndiResource(String jndiName, ResourceClass resourceClass);
 
    void addJndiResource(String jndiName, ResourceClass resourceClass, String basePath);
+
+   void checkAmbiguousUri();
```
marekkopecky: @asoldano @rsearls Whouldn't be better to use default interface method? It would be backward compatible, Registry is public interface. This default method could be just empty, like VertxRegistry.java has.

This commit is present in 3.6 and master ...

PR #1274, opened 2017-09-04
[RESTEASY-1680]:If the data buffer's last character is a U+000A LINE FEED (LF) character, then remove the last character from the data buffer

jimma:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/InboundSseEventImpl.java
```diff
@@ -97,9 +97,17 @@ public Builder write(byte[] data)
...
       public InboundSseEvent build()
       {
+         //from https://html.spec.whatwg.org/multipage/server-sent-events.html#processField
+         //If the data buffer's last character is a U+000A LINE FEED (LF) character, 
+         //then remove the last character from the data buffer
+         byte[] result = dataStream.toByteArray();
+         if (result.length > 0 && '\n' == result[result.length-1]) 
+         {
+            result = Arrays.copyOfRange(result, 0, result.length-1);
+         }
```
ronsigal: Maybe it's not worth it, but a subclass of ByteArrayOutputStream could look at ByteArrayOutputStream.buf and decide to copy all of buf or all but the last byte, saving one copy.

ronsigal:
Hey @jimma,

Looking at the code, I have another comment that's not directly relevant to these two classes, but I'm not sure where to put it.

SseEventInputImpl and SseEventOutputImpl both use "\r\n\r\n".getBytes(), but they could be running on different machines. Should you use String.getBytes(Charset charset)?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/InboundSseEventImpl.java
ronsigal: Oh, well, you already have EventByteArrayOutputStream ...

jimma: @ronsigal  Thanks for the review and comments. You are right. We can use EventByteArrayOutputStream here too to avoid copy the array for 2 times.  I've fixed this . Please help review again.

jimma:
@ronsigal While I read the spec again yesteardy , I found this is another thing I need to fix/improve later. Spec says the event stream should be encode with "UTF-8"  and the end of line is :

> end-of-line = ( cr lf / cr / lf )
>lf            = %x000A ; U+000A LINE FEED (LF)
>cr            = %x000D ; U+000D CARRIAGE RETURN (CR)

Now "\r\n\r\n".getBytes() event boundary match is clearly too simple to handle this.

PR #1275, opened 2017-09-04
Async stream provider

FroMage:
This adds support for async stream providers, via the Reactive Streams API, as discussed in https://issues.jboss.org/browse/RESTEASY-1701.

- Added dep to reactive-streams 1.0.1 (tiny jar)
- Added AsyncStreamProvider to convert async streams to a common type
- By default collect every element of the stream into a List and serialise that at the end in one go
- If using the new `@Stream` annotation, stream every element of the stream to the output (hopefully in chunks)
- If using `@Produces(SSE)` on the method, stream every element of the stream into an SSE event
- Added tests for that, with rx2, and for converting an rx2 Single via AsyncResponseProvider
- Fix SSE deserialising to not append a LF at the end (and test)
- Fix SSE serialising to split data on any accepted line terminator (and test)

I'll add further remarks to the JIRA issue.


FroMage:
Added a fix that removes a servlet dependency in the current code for `CompletionStage` (via `HttpServletRequest`) which makes Resteasy not embeddable in Vert.x anymore.

FroMage:
Can I get a review please?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/response/resource/PublisherResponseMessageBodyWriter.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyWriter;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+@Produces("abc/xyz")
+public class PublisherResponseMessageBodyWriter implements MessageBodyWriter<PublisherResponseTestClass> {
```
asoldano: Sorry, I'm not getting what this is used for / what is tested through this (together with the PublisherResponseTestClass). What am I missing?

jimma:
@FroMage  Just try to understand the new async stream provider here.  The new AsyncStreamResponseConsumer is responsible for consume each element in publisher and flush it to servlet response, right ? 
And another question about the @Stream annotation to enable AsyncStreamResponseConsumer, do you think we can create AsyncStreamResponseConsumer for the Publisher return type method without @Stream annotated too ?  Does Publisher already imply it's a stream?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/response/resource/PublisherResponseMessageBodyWriter.java
FroMage: Totally a mistake, well spotted I'll remove it.

FroMage:
> I like this PR very much, thanks. Would you mind solving the conflict with the changes Jim already did in SSE? Thanks.

I will remove the conflict now that his fix has been merged.

FroMage:
Note that I have now tested this with a real application and noticed that both async response types (`CompletionStage` and `Publisher`) fail to call the async response callbacks on completion. I assume this isn't per spec, right? If not I'll fix it, because I need this to dispose CDI contexts on async requests.

asoldano:
> Note that I have now tested this with a real application and noticed that both async response types (CompletionStage and Publisher) fail to call the async response callbacks on completion. I assume this isn't per spec, right?

Right, at least AFAIR.

> If not I'll fix it, because I need this to dispose CDI contexts on async requests.

Yes please. Btw, we've just got access to the JSR 370 TCK and will run it soon. We'll let you know privately if something strange is found while running it.

asoldano:
@FroMage , feel free to add a testcase for this callback on completition issue if it does not take too much to create it ;-) Thanks!

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/response/resource/PublisherResponseMessageBodyWriter.java
asoldano: Thanks!

FroMage:
> @FroMage , feel free to add a testcase for this callback on completition issue if it does not take too much to create it ;-) Thanks!

Working on it. And almost done except one question: in the `CompletionStageResponseTest` class you test the resource throwing a `RuntimeException` which is not user-mapped, and so calling `writeException` throws `UnhandledException` all the way back to the container, which handles the writing of the exception.

It's unclear to me if I should call the completion listener, since I can only call them _before_ I send the exception to the container, and so, technically _before_ the response has been sent, which the listener says should happen _after_ it has been sent. On the other hand, if I don't call the listeners we can have leaks if people rely on them to be called for cleanup (such as me with CDI).

WDYT?

FroMage:
I see in `AbstractAsynchronousResponse.internalResume(Throwable exc)` you do call the callbacks in a `finally` block which means they get called even before the exception is thrown back at the user. I suppose that counts as a "yes we should call the callbacks even before responding if the exception is not mapped".

FroMage:
OK, added the fix, the tests, and verified that the fixes to SSE worked. Should be good to merge, even though I still have to add docs.

asoldano:
@FroMage github still complains that there're conflicts, but I can probably deal with that manually. The problem I see is that there are few test failures when running on Travis CI, which can't be reproduced locally here btw. It has happened to me other times, it's somewhat related to the different architecture and performances of the testing environment. Can you see the Travis CI logs? do they ring any bell there? If you want to use the Travis CI for testing and figuring out where the problem is, you might want to do something like https://github.com/resteasy/Resteasy/pull/1284/commits/2cddab1d595665f93d47b1f289f6ad3e97bd837c in another temporary PR.

FroMage:
Apparently it's ` PublisherResponseTest.testSse:174 expected:<[two]> but was:<[one]>`, which suggests I got items out of order. That's going to take some head scratching to puzzle out…

FroMage:
I can reproduce locally if I run the test several times in a row. At least that should help me figure it out.

FroMage:
Well, adding print statements makes the issue go away, great :(

asoldano:
@FroMage, I'm make some progress in investigating this issue (still using the other temp PR for testing on Travis). There's definitely an ordering "issue", most likely related to the scheduling of the threads that actually send events from the server; not really a problem IMHO anyway, the test simply needs to be fixed to avoid assuming a given order (I've done that in the other PR).
Besides that, it happens sometimes to get "one" followed by "one" on client, which is way more problematic. I'm investigating and I think that's related to exception handling being broken somewhere in SSE impl. Will update here with my findinds...

FroMage:
> most likely related to the scheduling of the threads that actually send events from the server

Mmm, I was pretty sure there was only one thread on the server sending both events, and only a single client at a time, but perhaps I'm wrong.

FroMage:
I do get some exceptions in the logs, but they appear to be client exceptions, no?

```
javax.ws.rs.ProcessingException: RESTEASY004655: Unable to invoke request
	at org.jboss.resteasy.client.jaxrs.engines.ApacheHttpClient4Engine.invoke(ApacheHttpClient4Engine.java:317)
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.invoke(ClientInvocation.java:453)
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocation.invoke(ClientInvocation.java:466)
	at org.jboss.resteasy.client.jaxrs.internal.ClientInvocationBuilder.get(ClientInvocationBuilder.java:169)
	at org.jboss.resteasy.plugins.providers.sse.client.SseEventSourceImpl$EventHandler.run(SseEventSourceImpl.java:268)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.IllegalStateException: Connection is still allocated
	at org.apache.http.util.Asserts.check(Asserts.java:34)
	at org.apache.http.impl.conn.BasicHttpClientConnectionManager.getConnection(BasicHttpClientConnectionManager.java:266)
	at org.apache.http.impl.conn.BasicHttpClientConnectionManager$1.get(BasicHttpClientConnectionManager.java:217)
	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:190)
	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:184)
	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:88)
	at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:184)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:82)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:55)
	at org.jboss.resteasy.client.jaxrs.engines.ApacheHttpClient4Engine.invoke(ApacheHttpClient4Engine.java:313)
	... 11 more
```

FroMage:
I think the client dies, and reconnects and gets two times the first element `one`, no?

FroMage:
I thought perhaps I have to call `source.close` in the client, but that makes the next run fail with `IllegalState RESTEASY004510` which isn't really clear.

FroMage:
Well, apparently, closing the event source closes the `ResteasyClient`, that's something I didn't expect. Is it normal?

FroMage:
So what I think is happening is that indeed my test failed to close the event source. I can't reproduce the error once that is fixed. Let me push and check what travis says.

FroMage:
So it turns out the test passes locally on its own, but not when run with all tests with `mvn clean install`, even locally. Even once I close the event source. Once I add print statements to debug it, it passes in both cases. Sorta depressed here.

asoldano:
@FroMage , it's true that the source should be closed on client side and it's expected to also close the client. I've done that in the other PR, but that didn't prevent the failures from still happening.
I've started adding some debug logs in the SseEventOutputImpl and noticed that when the test fails, the writeEvent method is not fully run, basically mean an exception is thrown somewhere and is then swallowed (which happens because the future that's returned by sink.send() is not used anywhere). Now, I've added a printstacktrace statement and here are the logs: https://paste.fedoraproject.org/paste/lKwFkXCejT2r5Guv9NodXQ/raw . It looks the server keeps on trying writing to a stream that is closed, till it eventually succeeds...

FroMage:
> the source should be closed on client side and it's expected to also close the client

Really? I would have thought closing the stream would only close this request, not the entire client.

> I've started adding some debug logs in the SseEventOutputImpl and noticed that when the test fails, the writeEvent method is not fully run, basically mean an exception is thrown somewhere and is then swallowed (which happens because the future that's returned by sink.send() is not used anywhere).

OK brilliant work, man. Apparently I forgot to check on that. I wonder if I should wait on it. I guess not, in theory, but at least I should have some guarantee that sent messages are sent in order, even if they're all sent by a different future.

As to who the hell closed that connection, I've no idea. Why would this even happen?

FroMage:
I will wait for the sending to be done before requesting the next element in the stream, that should be easy.

asoldano:
> > the source should be closed on client side and it's expected to also close the client
> Really? I would have thought closing the stream would only close this request, not the entire client.

let's put this aside for now, let me check later, I remember the team discussed about this in the future, need to go back and verify first.

> OK brilliant work, man. Apparently I forgot to check on that. I wonder if I should wait on it. I guess not, in theory, but at least I should have some guarantee that sent messages are sent in order, even if they're all sent by a different future.

mmh.. at first I thought a wait on that future was missing, so implemented it locally...but eventually I trashed it because I also thought that this is an asynch, so it should basically be fire and forget. But I might be wrong.
As for the ordering guarantees, I'm not that convinced, to be honest. It probably depends on how Flow is defined (not an expert here), however I guess that to ensure the ordering you'd actually have to wait on that future...

> As to who the hell closed that connection, I've no idea. Why would this even happen?
The timestamps on the logs are suspicious. Almost always 500 ms between tries, so I think that's because the client is automatically closing and re-connecting the source (there's a configurable delay for that, whose default is 500ms). I'm going and checking that part of the code now...

FroMage:
OK I believe I've fixed it by waiting for the event to be sent before I ask for the next element. But I'm pretty sure lots of people will be confused by the random sending of elements, due to `send` being done in a thread pool with no queue. I'm pretty sure this is wrong and there should be a FIFO behaviour, even if async, no?

FroMage:
Damn, looks like it still fails on Travis :( I'll try to increase the reconnect timeout, perhaps the machine is too slow to send events...

asoldano:
Thanks for the change to deal with the ordering problem. Adding a FIFO queue is likely nice future feature request.
As for the failures, they're indeed caused by the connection being closed when the server tries writing on the ouput stream. What happens is that the client eventually figures out the connection is broken so it re-connects. That causes can lead to receiving 2 equivalent data, which makes the test fail. Increasing the re-connect timeout is not going to help, the problem is really understanding what is closing the connection and if that's really required.

FroMage:
OK I got it I'm pretty sure. Thought about it this weekend, the only thing that would make sense is if RS tells us the stream is complete before we've (async) sent the event. I can't reproduce it locally even with a huge `sleep` in the async sending, but I see these logs:

```
[31m11:47:48,561 ERROR [stderr] (default task-1) Sending SSE data async: one
[0m[31m11:47:53,570 ERROR [stderr] (ForkJoinPool.commonPool-worker-15) DONE Sending SSE data async: one
[0m[31m11:47:53,571 ERROR [stderr] (ForkJoinPool.commonPool-worker-15) Sending SSE data async: two
[0m[31m11:47:53,574 ERROR [stderr] (ForkJoinPool.commonPool-worker-15) On complete
[0m[31m11:47:58,572 ERROR [stderr] (ForkJoinPool.commonPool-worker-4) DONE Sending SSE data async: two
```

Which confirms that we get the `onComplete` as soon as the `onNextElement` receives the second element, and sends it async. So what MUST be happening is that `onComplete` closes the channel _before_ we've sent it. I'll handle this and pray that Travis likes it, because it's the only logical explanation of what's happening.

FroMage:
OK, it works on two builds, but one of them fails with:

```
Running org.jboss.resteasy.test.providers.sse.SseTest
Tests run: 8, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 43.768 sec <<< FAILURE! - in org.jboss.resteasy.test.providers.sse.SseTest
testAddMessage(org.jboss.resteasy.test.providers.sse.SseTest)  Time elapsed: 0.161 sec  <<< FAILURE!
java.lang.AssertionError: 5 messages not equal expected:<[message 0, message 1, message 2, message 3, message 4]> but was:<[message 2, message 0, message 1, message 3, message 4]>
	at org.jboss.resteasy.test.providers.sse.SseTest.testAddMessage(SseTest.java:91)
```

And err… I don't know.

asoldano:
@FroMage , looks like I was not the only one getting crazy with this during the weekend, nice. I managed to reproduce the problem locally (75-80% of times), but I suspected problems in the lower level SSE implementation, so spent a lot of time debugging that (and setting up Byteman ;-)). Your explanation makes sense indeed, I've tried with my reproducer and it fixes the failures 👍 , cool!
Now, on the last Travis CI runs, we got a different failures in SseTest (last one seems to be related to event ordering btw), but I'm confident those are can be fixed and possibly are note related to your changes. I'll take over from here and eventually merge the PR, will get back to you if I think there's something to be changed in this PR. Thanks again!

FroMage:
Yeah, these concurrency issues are horrible. Sorry to put you through this :(

You think the ordering issues could come from the client side? I can't see how they would come from the server. Unless there's still some hidden exception and reconnect going on, but that does not seem possible to me, because we're much more likely to get things like `message 0, message 0…` rather than start at `message 2`, because _as it is_ I can't think the server can send them out of order. I strongly suspect the client, at this point.

asoldano:
I've never reproduced the ordering issue locally (while I can reproduce another failure that I could see here from time to time in your new testEscapedMessage), but I believe this could happen in a slow machine with a different scheduler, considering the sink sends events in new threads (https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java#L112 )and the event send is basically atomic (https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java#L120). I might be wrong, but as you previously said we should probably consider adding a queue here.
I've rebased this PR in https://github.com/resteasy/Resteasy/pull/1288 and I'm improving the tests and cleaning up few minor glitches (imports, etc). Will merge that after a couple of green runs on Travis ;-)

asoldano:
@FroMage I've merged the other PR, in which I pulled all your commits here. Thanks a lot for now, we'll eventually have to write some doc... ;-)

FroMage:
Yes, I will work on the docs ASAP. Did you find the last remaining error, then?

asoldano:
https://github.com/resteasy/Resteasy/pull/1288/commits/c668768801eaae5f0e20927cf30f963018ad5205#diff-aff9561ead9cd3cae236201c0240adc9R30 was a problem because the sink was being (auto)closed while events were sent.

https://github.com/resteasy/Resteasy/pull/1288/commits/e879cd091257f5a71cb42264b40773d1a724c86a was also a bug (that we already had), even if I think it was not causing problems here.

FroMage:
Ooo, nice catches!

murphye:
Curious when resteasy-rxjava2 will be available on https://repository.jboss.org/nexus/content/repositories/snapshots

I have tried it locally and works great!

asoldano:
Thanks for the feedback @murphye !
I've just mvn deployed a snapshot, it should be there now (anyway please note the web pages with the list of artifacts are sometime not in sync with the actual repo contents).

murphye:
Seems there is a permissions issue?

https://repository.jboss.org/nexus/content/repositories/snapshots/org/jboss/resteasy/resteasy-rxjava2/

> Request is marked as local-only, remote access not allowed from M2Repository(id=snapshots)

asoldano:
I can try investigating / asking someone from infra. Still the only thing that matters is whether a maven project having a dependency on that snapshot artifact can actually fetch the jar or not. The full url of the pom is https://repository.jboss.org/nexus/content/repositories/snapshots/org/jboss/resteasy/resteasy-rxjava2/4.0.0-SNAPSHOT/resteasy-rxjava2-4.0.0-20171106.131840-1.pom

murphye:
Ok I see it now. So no problem at this point. Thanks!

PR #1276, opened 2017-09-05
[RESTEASY-1680]:Enable create SseEvent with multiple data fields

jimma:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventProvider.java
```diff
@@ -105,9 +107,34 @@ public void writeTo(OutboundSseEvent event, Class<?> type, Type genericType, Ann
...
+                           if (isNewLine)
+                           {
+                              entityStream.write(SseConstants.DATA_LEAD);
+                           }
+                           entityStream.write(b);
+                           isNewLine = false;
+                        }
+                     }
+
+                  });
```
ronsigal: @jimma, I'm not sure about this. I get that write() is the only abstract method in OutputStream, but what about entityStream.flush() and entityStream.close()?

jimma: @ronsigal  Because the outputStream is passed to MessageBodyWriter to serialize message to underlying entityStream.  flush() or close() on this ByteArrayOutputStream is not allowed before the buffer is read.

jimma:
@rsearls  copy/paste error. I fixed in the next change. Thanks for the review.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventProvider.java
ronsigal: @jimma I'm not sure why "flush() or close() on this ByteArrayOutputStream is not allowed before the buffer is read". Maybe it doesn't make sense to call close(), but is it illegal? What am I missing?

jimma: @ronsigal  sorry, I didn't explain this well.  Yes. it doesn't make sense to close and it should do nothing when close() is called for ByteArrayOutputStream . There is no need to close io or clean up something here.

ronsigal: @jimma, I'm not trying to be difficult ;-) but isn't it true that an event can have data of any type, which means that any MessageBodyWriter could be called, even one supplied by the user. But in that case we can't predict what that MBW will be doing ...

jimma: @ronsigal  We leave the flush() and close() to no operation/effect which OutputStream does. MBW only takes care of writing byte buffer is enough, user defined MBW call flush() or close() will take no effect.

ronsigal: @jimma re: "We leave the flush() and close() to no operation/effect which OutputStream does."

There's
```
   @Override
   public void flush() throws IOException
   {
      commit();
      delegate.flush();
   }



   @Override
   public void close() throws IOException
   {
      commit();
      delegate.close();
   }
```
in CommitHeaderOutputStream ...

jimma: I meant the Anonymous OutputStream class which is passed to the the selected MBW to write :
```
                  new OutputStream()
                 {
                    boolean isNewLine = false;
 
                    @Override
                    public void write(int b) throws IOException
                    {
                       if (b == '\n' || b == '\r')
                       {
                          if (!isNewLine) {
                             entityStream.write(SseConstants.EOL);
                          }
                          isNewLine = true;
                       }
                       else
                       {
                          if (isNewLine)
                          {
                             entityStream.write(SseConstants.DATA_LEAD);
                          }
                          entityStream.write(b);
                          isNewLine = false;
                       }
                    }
```

asoldano: @jimma , I think @ronsigal is right here, we should probably creating a full wrapper/delegate of OutputStream here. IOW, the new OutputStream should also override the other methods of the class, including close and flush. The implementation for those methods should simply delegate to the entityStream object.

jimma:
I added this two methods. Do you think it's better now ?

asoldano:
@jimma yes, thanks. Merging now.

PR #1277, opened 2017-09-07
[RESTEAY-1680]:Reads event with different eol; clean up charset constant

jimma:
Fix couple of things : 

-  each event is now ended with '\n' insteand of '\r\n' 
-  eventInputImpl can read the event with delimiter '\r', '\n' or '\r\n' (it's only '\r\n' before this change)
-  all event fields are encoded with "UTF-8" and value is encoded with mediaType's charset  if it exists.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
@@ -18,13 +18,4 @@ public class EventByteArrayOutputStream extends ByteArrayOutputStream
       }
       return Arrays.copyOf(buf, count);
    }
-   
-   public synchronized byte[] getEventData()
-   {
-      if (count >=1 && buf[count-1] == '\n')
-      {
-         return Arrays.copyOf(buf, count - 1);
-      }
-      return Arrays.copyOf(buf, count);
-   }
 }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
@@ -5,23 +5,23 @@ import java.util.Arrays;
 
 public class EventByteArrayOutputStream extends ByteArrayOutputStream
 {
-   private void removeBlankLine()
-   {
-      if (this.count > 4)
-      {
-         count = count - 4;
-      }
-   }
-
    public synchronized byte[] getEventPayLoad()
    {
-      removeBlankLine();
+      // delimiter is \r or \n
+      if (count >=2 && this.buf[count-2] == this.buf[count-1])
+      {
+         count = count -1;
+      }
+      //delimiter is 
+      if (count >= 1 && buf[count-2] == '\r' && buf[count-1] == '\n') {
+         count = count -2;
+      }
       return Arrays.copyOf(buf, count);
    }
    
    public synchronized byte[] getEventData()
    {
-      if (buf[count] == '\n')
+      if (buf[count-1] == '\n')
       {
          return Arrays.copyOf(buf, count - 1);
       }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
@@ -14,7 +14,7 @@ public class EventByteArrayOutputStream extends ByteArrayOutputStream
       }
       //delimiter is \r\n
       if (count >= 2 && buf[count-2] == '\r' && buf[count-1] == '\n') {
-         Arrays.copyOf(buf, count-2);
+         return Arrays.copyOf(buf, count-2);
       }
       return Arrays.copyOf(buf, count);
    }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
@@ -10,18 +10,18 @@ public class EventByteArrayOutputStream extends ByteArrayOutputStream
       // delimiter is \r or \n
       if (count >=2 && this.buf[count-2] == this.buf[count-1])
       {
-         count = count -1;
+         return Arrays.copyOf(buf, count -1);
       }
       //delimiter is 
-      if (count >= 1 && buf[count-2] == '\r' && buf[count-1] == '\n') {
-         count = count -2;
+      if (count >= 2 && buf[count-2] == '\r' && buf[count-1] == '\n') {
+         Arrays.copyOf(buf, count-2);
       }
       return Arrays.copyOf(buf, count);
    }
    
    public synchronized byte[] getEventData()
    {
-      if (buf[count-1] == '\n')
+      if (count >=1 && buf[count-1] == '\n')
       {
          return Arrays.copyOf(buf, count - 1);
       }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
@@ -12,7 +12,7 @@ public class EventByteArrayOutputStream extends ByteArrayOutputStream
       {
          return Arrays.copyOf(buf, count -1);
       }
-      //delimiter is 
+      //delimiter is \r\n
       if (count >= 2 && buf[count-2] == '\r' && buf[count-1] == '\n') {
          Arrays.copyOf(buf, count-2);
       }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
@@ -25,7 +25,7 @@ public class SseEventInputImpl implements EventInput, Closeable
    private InputStream inputStream;
    private volatile boolean isClosed = false;
    private boolean lastFieldWasData;
-   private final String DELIMIETER = new String(SseConstants.EVENT_DELIMITER, StandardCharsets.UTF_8);
+   private final String DELIMITER = new String(SseConstants.EVENT_DELIMITER, StandardCharsets.UTF_8);
    public SseEventInputImpl(Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, String> httpHeaders,
          InputStream inputStream)
    {
@@ -230,7 +230,7 @@ public class SseEventInputImpl implements EventInput, Closeable
             eolBuffer[pos] = b;
             //if it meets \r\r , \n\n , \r\n\r\n or \n\r\n\r\n
             if ((pos > 0 && eolBuffer[pos] == eolBuffer[pos - 1])
-                  || (pos >= 3 && new String(eolBuffer, 0, pos, StandardCharsets.UTF_8).contains(DELIMIETER)))
+                  || (pos >= 3 && new String(eolBuffer, 0, pos, StandardCharsets.UTF_8).contains(DELIMITER)))
             {
                boundary = true;
             }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
@@ -8,6 +8,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.annotation.Annotation;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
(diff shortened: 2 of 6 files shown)


ronsigal:
Hi @jimma, I have some questions about the treatment of end of lines.

I don't understand EventByteArrayOutputStream.getEventPayLoad(). The first if construction will remove a blank line if and only if the last two characters are '\r\r' or '\n\n'. Regardless of what the first if does, the second if will remove '\r\n'.  So, 1) you may or may not remove the last blank line, and 2) well, I'm not sure what the second if is doing.

I think the goal of getEventPayload() is to return the entire event minus the last blank like. Now, It looks to me like there are eight variations on the end of a file that contains a non-blank line followed by a blank line: "\r\n\r\n", "\r\n\r", '\r\n\n", "\r\r\n", "\r\r", "\n\r\n", "\n\r", and "\n\n". So I think that's what it should be looking for.

Similarly, SseEventInputImpl.readEvent() should be looking for those sequences.

ronsigal:
re: "all event fields are encoded with "UTF-8" and value is encoded with mediaType's charset if it exists."

Section 9.2.4 "Parsing an event stream" of https://html.spec.whatwg.org/multipage/server-sent-events.html just says "Event streams in this format must always be encoded as UTF-8." Are you sure it's OK to encode the data field differently?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -22,9 +23,8 @@
    private MediaType mediaType;
    private MultivaluedMap<String, String> httpHeaders;
    private InputStream inputStream;
-   private final byte[] EventEND = "\r\n\r\n".getBytes();
    private volatile boolean isClosed = false;
-
+   private final String DELIMIETER = new String(SseConstants.EVENT_DELIMITER, StandardCharsets.UTF_8);
    public SseEventInputImpl(Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, String> httpHeaders,
```
ronsigal: DELIMIETER -> DELIMITER  ;-)

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -118,11 +120,11 @@ public InboundSseEvent read() throws IllegalStateException
...
                temSave.write(b);
                b = readLine(entityStream, ':', temSave);
-               String fieldName = temSave.toString(charset.toString());
+               String fieldName = temSave.toString(StandardCharsets.UTF_8.name());
                temSave.reset();
                if (b == ':')
                {
-                  //space after the colon is ignored
+                  //spec says there is space after colon
                   do
```
ronsigal: Actually, it says, "If value starts with a U+0020 SPACE character, remove it from value."  That is, it seems to be saying to remove only a single space.

ronsigal:
One other thing is bothering me. Section 9.2.5 "Interpreting an event stream" of https://html.spec.whatwg.org/multipage/server-sent-events.html says "If the data buffer's last character is a U+000A LINE FEED (LF) character, then remove the last character from the data buffer." This seems to be specific to HTML. In JAX-RS, we have MessageBodyReaders parsing the data, so maybe it's not relevant. I don't know. What do you think?

jimma:
> EventByteArrayOutputStream at line 16
if (count >= 1 && buf[count-2] == '\r' && buf[count-1] == '\n')
Isn't there a chance that count=1 in which case count-2=-1 which
would throw an array out of bounds exception.

This is a obviously mistake. I fixed it .Thanks.

jimma:
> EventByteArrayOutputStream at line 24, 
can count ever have a value of
zero here?
I need to add this check.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: Thanks.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: @ronsigal  I looked at this in spec. I think more blank spaces after the filed name is useless.  The field value should be started with non blank spaces content. I can't think of an event case which the field value should be started with blank space.

jimma:
> One other thing is bothering me. Section 9.2.5 "Interpreting an event stream" of https://html.spec.whatwg.org/multipage/server-sent-events.html says "If the data buffer's last character is a U+000A LINE FEED (LF) character, then remove the last character from the data buffer." This seems to be specific to HTML. In JAX-RS, we have MessageBodyReaders parsing the data, so maybe it's not relevant. I don't know. What do you think?

I am not sure about this either. When I looked at this , I think this could be different for text/* media type.  If the spec dosen't count the last LF in the data payload, we need to process the same for text type data value.

jimma:
> Hi @jimma, I have some questions about the treatment of end of lines.
I don't understand EventByteArrayOutputStream.getEventPayLoad(). The first if construction will remove a blank line if and only if the last two characters are '\r\r' or '\n\n'. Regardless of what the first if does, the second if will remove '\r\n'. So, 1) you may or may not remove the last blank line, and 2) well, I'm not sure what the second if is doing.

I need to fix this.  For the first check , it should return and not go into another \r\n check. 

> I think the goal of getEventPayload() is to return the entire event minus the last blank like. Now, It looks to me like there are eight variations on the end of a file that contains a non-blank line followed by a blank line: "\r\n\r\n", "\r\n\r", '\r\n\n", "\r\r\n", "\r\r", "\n\r\n", "\n\r", and "\n\n". So I think that's what it should be looking for.

@ronsigal  From https://en.wikipedia.org/wiki/Newline,  '\n', '\r" or '\r\n" is OS dependent. "\r\n\r"  or "\r\n\n" the mixed eol format usage isn't we should look for ?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
ronsigal: @jimma I don't disagree with you, I'm just repeating from the spec.

ronsigal:
@jimma re: " If the spec dosen't count the last LF in the data payload, we need to process the same for text type data value."

The spec seems a little strange to me. It says to add a LF after each line, in effect normalizing the EOL character in the data buffer. But then you're supposed to remove the last LF. I don't get it. But for us, StringTextStar, for example, would write an entire string, with no line breaks, and then we would add a LF at the end, so I guess it makes sense to then remove that LF. So, I think I agree.

ronsigal:
@jimma re: ""\r\n\r" or "\r\n\n" the mixed eol format usage isn't we should look for ?"

I agree that  "\r\n\r", '\r\n\n", "\r\r\n", "\n\r\n", and "\n\r" don't make any sense. But, technically, they're legal. Some weird MBW could use them, I guess. But I won't argue if you want to ignore them.

asoldano:
@ronsigal , @jimma, have all the concerns been worked here?

jimma:
@ronsigal, Can you please have a look ? Anything else I need to improve, please let me know.

asoldano:
@jimma also note the PR has conflicts and needs to be rebased

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
```diff
@@ -5,23 +5,23 @@
...
+   public synchronized byte[] getEventPayLoad()
    {
-      if (this.count > 4)
+      // delimiter is \r or \n
+      if (count >=2 && this.buf[count-2] == this.buf[count-1])
       {
-         count = count - 4;
+         return Arrays.copyOf(buf, count -1);
+      }
+      //delimiter is
```
ronsigal: // delimiter is \r\n

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
```diff
@@ -5,23 +5,23 @@
...
-      if (this.count > 4)
+      // delimiter is \r or \n
+      if (count >=2 && this.buf[count-2] == this.buf[count-1])
       {
-         count = count - 4;
+         return Arrays.copyOf(buf, count -1);
+      }
+      //delimiter is 
+      if (count >= 2 && buf[count-2] == '\r' && buf[count-1] == '\n') {
+         Arrays.copyOf(buf, count-2);
```
ronsigal: return Arrays.copyOf(buf, count-2) ?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
```diff
@@ -5,23 +5,23 @@
...
-      if (this.count > 4)
+      // delimiter is \r or \n
+      if (count >=2 && this.buf[count-2] == this.buf[count-1])
       {
-         count = count - 4;
+         return Arrays.copyOf(buf, count -1);
+      }
+      //delimiter is \r\n
+      if (count >= 2 && buf[count-2] == '\r' && buf[count-1] == '\n') {
+         Arrays.copyOf(buf, count-2);
```
asoldano: @jimma , as Ron pointed out, isn't a return statement still missing here?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -118,11 +120,11 @@ public InboundSseEvent read() throws IllegalStateException
             {
                temSave.write(b);
                b = readLine(entityStream, ':', temSave);
-               String fieldName = temSave.toString(charset.toString());
+               String fieldName = temSave.toString(StandardCharsets.UTF_8.name());
```
ronsigal: @jimma, did you mean to change charset.toString() to charset.name() here? You don't want to hard code UTF-8, do you?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -212,25 +214,40 @@ else if ("retry".equals(name))
       EventByteArrayOutputStream buffer = new EventByteArrayOutputStream();
       int data;
       int pos = 0;
+      boolean boundary = false;
+      byte[] eolBuffer = new byte[5];
       while ((data = in.read()) != -1)
```
ronsigal: @jimma, at the end of the loop, you return null. What if a line is terminated by EOF? The BNF says
```
end-of-line   = ( cr lf / cr / lf / eof )
```

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
ronsigal: Oh, wait, the spec says
```
Event streams in this format must always be encoded as UTF-8.
```
Should we just ignore charsets?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/EventByteArrayOutputStream.java
jimma: Ah. What a mistake ! Fixed. Thanks @ronsigal and @asoldano

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: What do you mean by ignore charsets ?  temSave.toString() will use platform default character set.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: @ronsigal From the latest spec https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events, I only see end-of-line =  ( cr lf / cr / lf )

ronsigal: @jimma, ah, I was looking at https://www.w3.org/TR/2009/WD-eventsource-20091029/#parsing-an-event-stream.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventProvider.java
```diff
@@ -47,7 +48,7 @@ public void writeTo(OutboundSseEvent event, Class<?> type, Type genericType, Ann
          MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream)
          throws IOException, WebApplicationException
    {
-      Charset charset = SseConstants.UTF8;
+      Charset charset = StandardCharsets.UTF_8;
       if (mediaType != null && mediaType.getParameters().get(MediaType.CHARSET_PARAMETER) != null)
       {
          charset = Charset.forName(mediaType.getParameters().get(MediaType.CHARSET_PARAMETER));
```
ronsigal: @jimma, I'm not sure how to think about this. If the spec says that "Event streams in this format must always be encoded as UTF-8.", then shouldn't creating an event with a media type that has a charset (other than UTF-8) be considered an error?

asoldano:
@jimma , sorry but this needs to be rebased against master again, as a consequence of me having merged the other big asynch stream PR. Thanks!

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventProvider.java
jimma: @ronsigal  I back to look at this again and it looks we can do this : always encode as "UTF-8" and we simply ignore the other charset.  User should always use "UTF-8" charset as event spec says .  I removed other charset support.

jimma:
@ronsigal @asoldano   Looks the big mistakes and issues are resolved after your review. Thanks for the comments. I am going to merge this. If you have more comments or concerns on this change, please let me know. I'll create another PR to fix.

PR #1278, opened 2017-09-07
Async request filters

FroMage:
Hi,

This is a proto for supporting async request filters, which I need in another project.

This is not done by far, but given as a PR for discussion (do not merge it).

I only support post-match requests going async, and for backwards compat the default behaviour is to be synchronous in post-match filters. To go async from within a filter, you need to cast the context argument and invoke `suspend()` on it. After that you can resume the filtering with `abortWith` (existing method) or `resume()` to move on to the next filter or the resource method itself.

It's possible to have a filter that goes async, and a non-async filter or a non-async resource method after it, and we do handle this case. In the case of non-async resource methods, I assume that if the method is not declared as returning `void`, it's a non-async method and we should use its return value. Perhaps this is wrong and we can make it better.

I'm not too happy with how I implemented this in the context because the implementation is confusing by requiring a mix of async/sync filters, but I'm not sure how to make this easier to maintain, short of having a context instance per filter.

The biggest question I have is: what sort of API can we have that is backwards compatible. The current approach to keep the existing filter and context API and require the filter to typecast the passed context into a more capable subclass is not ideal, but perhaps you have other ideas?

If we can find ways to fix those issues and we can agree to add this, I'll go ahead and add async support for pre-match request and response filters too.

WDYT?

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
@@ -391,9 +391,9 @@ public class ResourceMethodInvoker implements ResourceInvoker, JaxrsInterceptorR
       }
       if (request.getAsyncContext().isSuspended())
       {
-         if(method.getReturnType().equals(void.class))
+         if(method.isAsynchronous())
             return null;
-         // resume an async request that returns something
+         // resume a sync request that got turned async by filters
          request.getAsyncContext().getAsyncResponse().resume(rtn);
          return null;
       }
@@ -575,5 +575,10 @@ public class ResourceMethodInvoker implements ResourceInvoker, JaxrsInterceptorR
    public MediaType[] getConsumes()
    {
       return method.getConsumes();
+   }
+
+   public void markMethodAsAsync()
+   {
+      method.markAsynchronous();
    }
 }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
@@ -294,13 +294,8 @@ public class ResourceMethodInvoker implements ResourceInvoker, JaxrsInterceptorR
 
       PostMatchContainerRequestContext requestContext = new PostMatchContainerRequestContext(request, this, requestFilters, 
             () -> invokeOnTargetAfterFilter(request, response, target));
-      BuiltResponse abortResponse = requestContext.filter();
-      if(abortResponse != null)
-         return abortResponse;
-      if(!requestContext.isSuspended())
-         return invokeOnTargetAfterFilter(request, response, target);
-      // don't return anything: the request has been suspended
-      return null;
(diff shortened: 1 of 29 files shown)


FroMage:
BTW, I couldn't find anywhere in the spec what to do with the return value of a method that has a `@Suspended` parameter. Do you know? Is it just ignored?

FroMage:
Added support for async filters in pre-match and response.

Can I get a review please?

asoldano:
> BTW, I couldn't find anywhere in the spec what to do with the return value of a method that has a @Suspended parameter. Do you know? Is it just ignored?

http://docs.oracle.com/javaee/7/api/javax/ws/rs/container/Suspended.html

asoldano:
> I think I can make the filters less complex by always leaving the continuation handling to them, rather than allow the possibility that filters were synchronous.

oh, well, I think the current proposal would be good, but if you want to propose an alternative, feel free.

> Also apparently I need another method resume(Throwable) because abortWith(Response) does not let me easily map that Throwable to a Response. Unless there's an API that does that for me?

mmh... I can't think of something that would do that here; @ronsigal ?

FroMage:
>> BTW, I couldn't find anywhere in the spec what to do with the return value of a method that has a @suspended parameter. Do you know? Is it just ignored?
> http://docs.oracle.com/javaee/7/api/javax/ws/rs/container/Suspended.html

OK thanks!

FroMage:
> oh, well, I think the current proposal would be good, but if you want to propose an alternative, feel free.

I'll try and see if that makes the code cleaner.

FroMage:
> Same as above, I would accept the current solution if we can't think of anything better. Adding Ron as reviewer, to let him also provide his opinion.

I suppose I can create a new interface just for those operations, that extends the JAX-RS interfaces, so that when we get there we can push the interface on to the next JAX-RS. Sort-of cleaner as it decouples it from the implementation.

FroMage:
I have made an interface, and tried to simplify the async filter handling, though in one case I'm not sure the simplification is really one.
WDYT?

asoldano:
@FroMage , I'm having a look at this soon. Would you mind rebasing against master in the mean time?

FroMage:
Sure, you want a rebase or a merge?

asoldano:
@FroMage a rebase would be better, thanks

asoldano:
Just FYI, should you start getting new failures after the rebase, that could be a consequence of part of the testsuite being erroneously disable at the time you initially submitted this PR. I've figured out that part of the testsuite was disabled only few days ago, when I got back from PTO.

FroMage:
OK, just rebased and forced push. The tests still all pass, locally.

asoldano:
Thanks. The failures in the latest Travis CI are known ones (to be fixed in another PR under review), so I've re-started the test runs.

FroMage:
Which failures? Travis passed all the tests :)

asoldano:
@FroMage , I re-run the tests before you could notice the failures.
Anyway, I think the changes being proposed here are good, are you ok with us merging them? (considering what you wrote in the initial message)

FroMage:
I'm OK to merge this, because I've tested it and use it in another project. I'll also add docs for it ASAP.

FroMage:
Thanks!

PR #1282, opened 2017-09-19
Restore integration-test execution

asoldano:


asoldano:
@jimma FYI, just figured out one of your PR (the one you added a new surefire execution for the json b test) accidentally disabled the whole integration-testsuite.

asoldano:
Merged immediately to prevent regressions. We can improve the solution in other commits, if needed.

jimma:
@asoldano Can you explain why that change disabled the whole integration test suite ?

asoldano:
@jimma I'm not 100% sure, it has to do with mixed configuration in plugin and plugin execution, plus one of the execution being the default one. I just noticed that the build was taking way less time than before and then noticed that no test was run in that module. So I applied the same "setup" that we've been using successfully for years in JBWS (disable default execution and explicitly enable 2 new ones, avoiding mixing includes and excludes in the same execution config) and it solved the problem.

PR #1286, opened 2017-09-25
[RESTEASY-1719]: NewCookie(s) are not commited in server response

NicoNes:


NicoNes:
Hi @rsearls,

It seem's that my issue is linked to another one you fixed in 3.0.23 (https://issues.jboss.org/browse/RESTEASY-1659).

So my PR is obviously not correct and there is many ways to fix it in my opinion involving changes to `org.jboss.resteasy.spi.HttpResponse` contract/implementations.

 Is the `org.jboss.resteasy.spi.HttpResponse.addNewCookie()` method really necessary since `NewCookie` can be added to  `org.jboss.resteasy.spi.HttpResponse.getOutputHeaders()`  map ?

By the way, another thing: I think that the `org.jboss.resteasy.plugins.server.servlet.put(String s, List<Object> objs)` method does not follow the Map contract since it adds headers instead of replacing them.

rsearls:
Your code changes in ServerResponseWriter does not change the fundamental
function of the method.  There is no difference between calling
jaxrsResponse.getHeaders(); and jaxrsResponse.getMetadata().  The call
to jaxrsResponse.getHeaders() is in fact calling jaxrsResponse.getMetadata().
If you look deeper into the code you will see that BuiltResponse (jaxrsResponse)
is a subclass of javax.ws.rs.core.Response.  javax.ws.rs.core.Response is
an abstract class.  It defines getMetadata and getHeaders() as follows.
   public abstract MultivaluedMap<String, Object> getMetadata();

   public MultivaluedMap<String, Object> getHeaders() {
      return this.getMetadata();
   }

BuiltResponse implements getMetadata(), so the use of getHeaders is calling
getMetadata() already.  This code change is inefficient because it is calling
a method which is calling the method we desire rather than calling the method
directly.

Similarly your use of outputHeaders.addAll is unneeded and inefficient. Method 
addAll requires a key/value pair input param and thus needs the for loop but 
method pullAll takes a single Map<? extends K, ? extends V> input param.

I'm rejecting this code change as unneeded.

NicoNes:
Hi @rsearls 

Sorry I forgot to mention that my PR is about this JIRA https://issues.jboss.org/browse/RESTEASY-1700.
The unit test is linked to the JIRA and to this PR too. It will help you to spot the issue I'm talking about.

So you're right my changes about using getHeaders instead of `Response.getMetadata()` is not really important I did it just because of the method java doc says that:

> See getHeaders(). This method is considered deprecated. Users are encouraged to switch their code to use the getHeaders() method instead. The method may be annotated as @Deprecated in a future release of JAX-RS API.

My PR is not about this little change but about calling `addAll()` instead of `putAll` so that both `javax.ws.rs.core.Cookie` and `javax.ws.rs.core.NewCookie` instances can be commited into server response.

As I said before:

> So my PR is obviously not correct and there is many ways to fix it in my opinion involving changes to org.jboss.resteasy.spi.HttpResponse contract/implementations.

Also I think that my previous point is still valid:

> By the way, another thing: I think that the org.jboss.resteasy.plugins.server.servlet.put(String s, List<Object> objs) method does not follow the Map contract since it adds headers instead of replacing them.

rsearls:
This is not the proper solution.  I am discussing with the
team how best to address this issue.

asoldano:
@rsearls does it mean this can be closed as superseded by #1295 ?

PR #1290, opened 2017-09-27
[RESTEASY-1717] added proposed code and new testcase

rsearls:
What I find disconcerting about this code is that @Path does not support the @Inherit annotation like @Produces and @Consumes does.   This requires each interface subclass must have the same @Path value in order to access the endpoint.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
@@ -4,7 +4,6 @@ import org.jboss.resteasy.client.jaxrs.ProxyConfig;
 import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
 import org.jboss.resteasy.client.jaxrs.internal.ClientConfiguration;
 import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
-import org.jboss.resteasy.client.jaxrs.internal.ClientInvocationBuilder;
 import org.jboss.resteasy.client.jaxrs.internal.ClientRequestHeaders;
 import org.jboss.resteasy.client.jaxrs.internal.ClientResponse;
 import org.jboss.resteasy.client.jaxrs.internal.proxy.extractors.ClientContext;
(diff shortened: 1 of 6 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -77,11 +82,87 @@ public Method getResourceMethod()
...
+            for (Class<?> next = interfaces.poll(); next != null; next = interfaces.poll())
+            {
+               Consumes annotation = (Consumes) next.getAnnotation(Consumes.class);
+               if (annotation != null) {
+                  String[] values = annotation.value();
+                  for (String value : values)
+                  {
+                     WeightedMediaType candidateWeightedMediaType = WeightedMediaType.valueOf(value);
+                     if (weightedMediaType == null || candidateWeightedMediaType.compareTo(weightedMediaType) < 0) {
+                        weightedMediaType = candidateWeightedMediaType;
```
asoldano: Can you explain me the reasoning behind this please?
And isn't the new checkConsumesMediaTypes(..) method here changing the behavior in case of defaultConsumes being null?

rsearls: When there is no @Consumes annotation on the current class, resteasy should
check the super-class for the @Consumes annotation.  Because @Consumes has 
annotation @Inherited, every sub-class should inherit the super-class's @Consumes 
annotation when the sub-class does not have a @Consumes annotation declared.
However, when the current sub-class has a @Consumes annotation it overrides 
the parent annotation.

rsearls:
I did not foresee the backward compatibility issues with this change, but yes that is another reason to dislike this change.

asoldano:
mmh... considering the "issues" mentioned above and what the user said at https://issues.jboss.org/browse/RESTEASY-1717?focusedCommentId=13466931&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13466931 , I'm not really convinced here. Unless you can provide a more convincing solution, I'm leaning towards rejecting this and closing the issue as won't fix.

rsearls:
My vote is to reject it.

asoldano:
OK, closed this, please deal with the jira.

PR #1294, opened 2017-10-02
[RESTEASY-1718] added utility to check for IPv6 format

rsearls:
@ronsigal @asoldano Please review.  All the variants of an IPv6 addresses can't
easily be determined by a Pattern object.  I've appropriated IPv parsing code
from URI and created utility, IPv6FormatParser.  In addition Resteasy supports 
uri "templating" (see UriBuilderTest line 87) which allows left and right brackets.
IPv format considers {} "excluded characters" because gateways and other transport
agents are known to sometimes modify such characters, so a Pattern for template
text was created and checked first so we can identify these before the IPv
address format is checked.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
@@ -89,6 +89,7 @@ public class ResteasyUriBuilder extends UriBuilder
    public static final Pattern opaqueUri = Pattern.compile("^([^:/?#]+):([^/].*)");
    public static final Pattern hierarchicalUri = Pattern.compile("^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?");
    private static final Pattern hostPortPattern = Pattern.compile("([^/:]+):(\\d+)");
+   private static final Pattern squareHostBrackets = Pattern.compile( "(\\[(([0-9A-Fa-f]{0,4}:){2,7})([0-9A-Fa-f]{0,4})\\]):(\\d+)" );
 
    public static boolean compare(String s1, String s2)
    {
@@ -189,20 +190,37 @@ public class ResteasyUriBuilder extends UriBuilder
             host = host.substring(at + 1);
             this. userInfo = user;
          }
+
          Matcher hostPortMatch = hostPortPattern.matcher(host);
          if (hostPortMatch.matches())
          {
             this.host = hostPortMatch.group(1);
             int val = 0;
-            try {
+            try
+            {
                this.port = Integer.parseInt(hostPortMatch.group(2));
-            }
-            catch (NumberFormatException e) {
+            } catch (NumberFormatException e)
+            {
                throw new IllegalArgumentException(Messages.MESSAGES.illegalUriTemplate(uriTemplate), e);
             }
-         }
-         else
+         } else
          {
+            if (host.startsWith("["))
+            {
+               // Must support an IPv6 hostname of format "[::1]" or [0:0:0:0:0:0:0:0]
+               Matcher bracketsMatch = squareHostBrackets.matcher(host);
+               if (bracketsMatch.matches())
+               {
+                  host = bracketsMatch.group(1);
+                  try
+                  {
+                     this.port = Integer.parseInt(bracketsMatch.group(5));
+                  } catch (NumberFormatException e)
+                  {
+                     throw new IllegalArgumentException(Messages.MESSAGES.illegalUriTemplate(uriTemplate), e);
+                  }
+               }
+            }
             this.host = host;
          }
       }
--- a/testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/util/UriBuilderTest.java
+++ b/testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/util/UriBuilderTest.java
@@ -1033,6 +1033,20 @@ public class UriBuilderTest {
                 builder.build();
             }
         }
+
+        // RESTEASY-1718 checks
+        {
+            Assert.assertEquals("http://foo", UriBuilder.fromUri("http://foo").build().toString());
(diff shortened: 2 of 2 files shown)


rsearls:
- isn't the problem here "simply" that the patterns are using ":" only as the 
  character separating the hostname from the port and not expecting it to be 
  part of an IPv6 address?
    That may me true.  The text that came through was [::1]:8080.  My understanding 
    of the pattern is that this is what it intended.  I read the pattern
    to be ... one or more characters that are not / and : followed by : 
    followed by one or more numbers only.  Is that how you interpret the pattern?
    
- to resolve the column issue, URIs/URLs including an IPv6 address are required 
  to use square parenthesis around the host/IP, can't we fix the patterns to use 
  this information? or perhaps we need to check for the existence of square 
  parenthesis and use a different set of patterns?
    I will work on such a pattern.

- I'm failing to understand why templating enters the game here and why it's 
  a problem for IPv6 only.
    In the URI code that I took {} are illegal characters for the host text.
    
- as for identifying IPv6 addresses using pattern matching, please check 
  https://github.com/jbossws/jbossws-common/blob/jbossws-common-3.1.5.Final/src/main/java/org/jboss/ws/common/utils/AddressUtils.java 
  which is what we use in JBossWS
- the proposed solution so far looks a bit too complicated; I'm also concerned 
  by the fact that you had to duplicate code in the JDK (can't you delegate to it instead?)
- creating a new parser and then directly accessing one of its fields to get 
  the result of a computation is not really good
- the changes to the LinkTest kind of suggest me we could be breaking backward 
  compatibility, is that true?
    Because LinkTest still fails with an exception I did not see it as a 
    backward compatibilty issue, but I suppose it could be.

rsearls:
@asoldano  Please review this change in the handling of the host.  I switched to using Patterns only.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/proxy/UriUsingIPv6Test.java
```diff
@@ -0,0 +1,108 @@
...
+
+
+/**
+ * @tpSubChapter Resteasy-client
+ * @tpChapter Client tests
+ * @tpSince RESTEasy 4.0.0
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class UriUsingIPv6Test {
```
asoldano: Integration tests should be used when actual network communication is needed for the test, when a mix of client and server behavior is tested, when there's a user reproducer to be verified, etc.
In the case here, it might make sense to write an integration test, but that should really perform the invocation using the IPv6 address, if IPv6 is enabled. Judging from the test code, I think this test wouldn't pass if IPv6 was on, would it?

rsearls:
@asoldano   Revisions made.  Please review

asoldano:
@rsearls OK, this seems to be processing addresses correctly. However, I'm wondering if we can make the implementation simpler and probably faster, adding just a single regexp instead of two. Can't we simply check if the host starts with a "[" and then either use a new regexp for ipv6 addresses only or the previous hostPortPattern regexp? Am I missing edge cases with such approach?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
```diff
@@ -89,6 +89,11 @@ public UriBuilder clone()
    public static final Pattern opaqueUri = Pattern.compile("^([^:/?#]+):([^/].*)");
    public static final Pattern hierarchicalUri = Pattern.compile("^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?");
    private static final Pattern hostPortPattern = Pattern.compile("([^/:]+):(\\d+)");
+   //private static final Pattern curelyOrSquareHostBrackets = Pattern.compile("((\\[.*\\]|\\{.*\\}):?.*)");
+   //private static final Pattern basicHostnamePattern = Pattern.compile("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$", Pattern.CASE_INSENSITIVE);
+   private static final Pattern curelyOrSquareHostBrackets = Pattern.compile( "(\\[(([0-9A-Fa-f]{0,4}:){2,7})([0-9A-Fa-f]{0,4})\\]|\\{.*\\}):?(\\d*)" );
```
asoldano: s/curely/curly/

resteasy-jaxrs/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilder.java
```diff
@@ -89,6 +89,11 @@ public UriBuilder clone()
    public static final Pattern opaqueUri = Pattern.compile("^([^:/?#]+):([^/].*)");
    public static final Pattern hierarchicalUri = Pattern.compile("^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?");
    private static final Pattern hostPortPattern = Pattern.compile("([^/:]+):(\\d+)");
+   //private static final Pattern curelyOrSquareHostBrackets = Pattern.compile("((\\[.*\\]|\\{.*\\}):?.*)");
+   //private static final Pattern basicHostnamePattern = Pattern.compile("^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])$", Pattern.CASE_INSENSITIVE);
```
asoldano: Please remove commented out lines here

rsearls:
@asoldano changes made.  Please review.

rsearls:
@asoldano   Changed

asoldano:
Merged, thanks

PR #1295, opened 2017-10-02
[RESTEASY-1719] changed processing for a Cookie as a String in a resp…

rsearls:
…onse object

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
@@ -21,6 +21,7 @@ import javax.ws.rs.container.ContainerResponseFilter;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.NewCookie;
+import javax.ws.rs.core.Cookie;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.MessageBodyWriter;
 import javax.ws.rs.ext.WriterInterceptor;
@@ -408,9 +409,14 @@ public class ServerResponseWriter
                   response.addNewCookie(cookie);
                   it.remove();
                }
+               else
+               {
+                  response.getOutputHeaders().add(javax.ws.rs.core.HttpHeaders.SET_COOKIE, next);
+                  it.remove();
+               }
             }
-            if (cookies.size() < 1)
-               jaxrsResponse.getMetadata().remove(HttpHeaderNames.SET_COOKIE);
+
+            jaxrsResponse.getMetadata().remove(HttpHeaderNames.SET_COOKIE);
          }
       }
       if (jaxrsResponse.getMetadata() != null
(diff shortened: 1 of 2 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
```diff
@@ -408,9 +409,16 @@ public static void commitHeaders(BuiltResponse jaxrsResponse, HttpResponse respo
                   response.addNewCookie(cookie);
                   it.remove();
                }
+               else if (next instanceof Object)
```
NicoNes: `next instance Object` will always be true right ?

rsearls: Sadly yes.  I am testing a change for this.

rsearls:
Code change to check for char[] or String pushed

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
```diff
@@ -402,15 +403,22 @@ public static void commitHeaders(BuiltResponse jaxrsResponse, HttpResponse respo
...
             {
                Object next = it.next();
+               String typeName = next.getClass().getTypeName();
                if (next instanceof NewCookie)
                {
                   NewCookie cookie = (NewCookie) next;
                   response.addNewCookie(cookie);
                   it.remove();
                }
+               else if ("java.lang.String".equals(typeName) || "char[]".equals(typeName))
```
NicoNes: Actually I think that it is not a good idea to only allow `NewCookie `, `String `or `char[] `as valid Set-Cookies header values.
Any object  should be considered as valid as long as there is an instance of `javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate` able to convert it to a String or using it `toString() ` method.
It seems that such classes already exist in the project such as `org.jboss.resteasy.plugins.delegates.CookieHeaderDelegate` wich handle `javax.ws.rs.core.Cookie` or `org.jboss.resteasy.plugins.delegates.NewCookieHeaderDelegate` wich handle `javax.ws.rs.core.NewCookie`.

The java doc about `javax.ws.rs.core.Response.ResponseBuilder.header(String, Object)` is:

> 
>            Add an arbitrary header.
>           
>            @param name  the name of the header
>            @param value the value of the header, the header will be serialized
>                         using a {@link javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate} if
>                         one is available via {@link javax.ws.rs.ext.RuntimeDelegate#createHeaderDelegate(java.lang.Class)}
>                         for the class of {@code value} or using its {@code toString} method
>                         if a header delegate is not available. If {@code value} is {@code null}
>                         then all current headers of the same name will be removed.
>            @return the updated response builder.

rsearls: javax.ws.rs.ext.RuntimeDelegate.HeaderDelegate  should not be a concern.
The call to response.getOutputHeaders().add in actuality calls
HttpServletResponseHeaders.add which in turn calls
HttpServletResponseHeaders.addResponseHeaders line 93
This method gets the proper RuntimeDelegate.HeaderDelegate and executes it.

rsearls:
I've made the changes and added test cases

NicoNes:
Hi @rsearls 

Thanks 👍

PR #1297, opened 2017-10-05
Added rxjava, rxjava2 providers

FroMage:
Here are the rxjava and rxjava2 plugins to support both.


asoldano:
Thanks a lot! There's a failure on JDK9 and few very minor things which I've already fixed locally, I'm going to merge this PR and push another commits just after that.

PR #1299, opened 2017-10-06
[RESTEASY-1728] Look for resteasy.allowGzip system property in Regist…

asoldano:
…erBuiltin

This is a proposal for a "fix" applying to both client and server side. Basically every time a new ResteasyProviderFactory instance is created,we would go and check the system property.
This said, we can probably do better... please provide feedback.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/RegisterBuiltin.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/RegisterBuiltin.java
@@ -1,20 +1,25 @@
 package org.jboss.resteasy.plugins.providers;
 
-import org.jboss.resteasy.core.ThreadLocalResteasyProviderFactory;
-import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
-import org.jboss.resteasy.spi.ResteasyProviderFactory;
-
-import javax.ws.rs.ext.Providers;
-
 import java.io.BufferedReader;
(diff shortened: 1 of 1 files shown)


ronsigal:
This is related to the issue that got me thinking about a general purpose facility for checking system properties. Something like
```
         ResteasyConfiguration context = ResteasyProviderFactory.getContextData(ResteasyConfiguration.class);
         if (context != null)
         {
            String s = context.getParameter("resteasy.allowGzip");
            if (s == null)
            {
                 s = AccessController.doPrivileged(new PrivilegedAction<String>() {
                        @Override
                        public String run() {
                            return System.getProperty("resteasy.allowGzip");
                        }
            }
            if (s != null)
            {
               allowGzip = Boolean.parseBoolean(s);
            }
         }
```

iweiss:
@ronsigal Take a look at Elytron's security manager facility, https://github.com/wildfly-security/wildfly-elytron/blob/master/src/main/java/org/wildfly/security/manager/WildFlySecurityManager.java#L1114.

asoldano:
@ronsigal , sure, it's related, and this is probably yet another confirm that we should be implementing your proposal (in that case we'd refactor this fix a bit, should be possible in backward compatible way, right?).
@iweiss , what do you actually want to suggest with the link above?

TomasHofman:
Do you guys think this can be used in 2.3.x branch as it is, so that it's not holding EAP release any longer? If there's going to be some refactoring later on, this branch could perhaps live without it, if it doesn't change a substance of the fix.

PR #1303, opened 2017-10-10
[JBEAP-11180, RESTEASY-1484] Disable GZIP by default; impose maximum …

spyrkob:
…file size.

JIRA: https://issues.jboss.org/browse/JBEAP-11180

Backport of #947 for 3.0.19.SPx branch.


spyrkob:
added a couple test fixes

spyrkob:
@asoldano it looks like the failures on jdk7 are due to env issues - can they be safely ignored?

asoldano:
@spyrkob most likely yes, but I've re-started the run just in case.

PR #1304, opened 2017-10-11
[RESTEASY-1680]:Set eventsink to closed state after eventsource closes the connection

jimma:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseResource.java
```diff
@@ -157,6 +157,33 @@ public void run()
...
+                       break;
+                   }
+               }
+           }
+       }.start();
+   }
+   @GET
+   @Path("/isopen")
+   public boolean isOpen() {
+       return this.eventSink.isClosed();
```
asoldano: The method name and @Path value are misleading, considering your return eventSink.isClosed(). How about adding a "!" and reversing the checks in the client?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
```diff
@@ -338,6 +339,51 @@ public void testEscapedMessage() throws Exception
...
+        Client isOpenClient =  ClientBuilder.newClient();
+        Invocation.Builder isOpenRequest = isOpenClient.target(generateURL("/service/server-sent-events/isopen")).request();
+        javax.ws.rs.core.Response response = isOpenRequest.get();
+        Assert.assertFalse("EventSink isClosed=false is expected ", response.readEntity(Boolean.class));
+        
+        eventSource.close();
+        Assert.assertFalse("Closed eventSource state is expceted", eventSource.isOpen());
+        
+        boolean isClosed = false;
+        for (int i = 0 ; i < 5; i ++)
```
asoldano: why do we need to make this 5 times? It looks suspicious to me, if this is only timing issue, it's likely to break / fail intermittently depending on the test environment.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseResource.java
jimma: Another stupid mistake and it is fixed now.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
jimma: Server needs some time to know the connection is already closed by write another event to outputstream and get an IOException. I tried to refactor the test and remove this loop, it looks works.

asoldano:
@jimma the test is still intermittently failing. I believe you need to wait (sleep) on the client a bit before checking if the connection is closed. Also note that a slight sleep (few millisecs ) in the loop where events are sent might make the thread scheduling more fair and hence allow for more predictable test execution.

PR #1305, opened 2017-10-11
[RESTEASY-1075] HTTP redirect support in ApacheHttpClient engine

asoldano:


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ApacheHttpClient4Engine.java
```diff
@@ -434,7 +435,7 @@ public String getMethod()
 
    protected boolean isRedirectRequired(final ClientInvocation request, final HttpRequestBase httpMethod)
    {
-      if (httpMethod instanceof HttpGet && false) // todo  && request.followRedirects())
+      if (httpMethod instanceof HttpGet && isFollowRedirects())
```
ronsigal: @asoldano, according to the HttpClient documentation
```
5.3. Automatic redirect handling

HttpClient handles all types of redirects automatically, except those explicitly prohibited by the HTTP specification as requiring user intervention. See Other (status code 303) redirects on POST and PUT requests are converted to GET requests as required by the HTTP specification.
```
So maybe we shouldn't restrict redirection to GETS?

ronsigal:
@asoldano, it looks fine. I usually test "true", "false", and "default", but ...  ;-)

asoldano:
Thanks Ron, merging

PR #1318, opened 2017-10-18
Fix early connection completion on async response filters

FroMage:
Now that response filters can be completed asynchronously, we can no longer rely on `finally` blocks to complete the connection and call callbacks: we need to register listeners to call them when the response is sent.


asoldano:
@FroMage first of all thanks for the PR. I still have to look at this carefully, however wanted to start by giving an initial feedback. There are many methods whose signature is being changed to add a "Consumer<Throwable> onComplete", still in many cases a lambda that does nothing is passed on that parameter (t -> {}); how about, when possible, keeping methods with the old signature too and delegating to the new ones by passing that "empty" lambda as parameter? That would grant a better backward compatibility and probably make the changeset easier.

FroMage:
I can. But I fear it will distract people from using the `onComplete` and assume things can be done with `finally`. But I can.

asoldano:
Well, the fact is that I'd like to backport you stream/reactive changes to a 3.x minor in the next future, to make sure they land in WFLY soon. That's why I'm looking a bit at backward compatibility. But I think I can probably do the changes I suggested on the branch only, and mark the old methods as deprecated ;-)

FroMage:
That's a possibility, just let me know if you want me to add the methods. I can always add them and mark the others as deprecated on master too. That would serve enough as warning IMO.

asoldano:
@FroMage thanks and sorry for the late answer. To me the PR here is fine; to merge it, it has to be rebased, so I've created https://github.com/resteasy/Resteasy/pull/1336, in which I've also added an additional commit for the mentioned backward compatibility methods. Unless you see anything wrong, I'd merge #1336 after a successful Travis CI run.
Thanks again.

FroMage:
+1On 27 Oct 2017 01:52, Alessio Soldano <notifications@github.com> wrote:@FroMage thanks and sorry for the late answer. To me the PR here is fine; to merge it, it has to be rebased, so I've created #1336, in which I've also added an additional commit for the mentioned backward compatibility methods. Unless you see anything wrong, I'd merge #1336 after a successful Travis CI run.
Thanks again.

PR #1320, opened 2017-10-18
[RESTEASY-1738]

ronsigal:
If an entity is sent with media type application/x-www-form-urlencoded
entity, reconstruct body from form parameters if the InputStream has
been read.

[RESTEASY-1738]

Before constructing form entity from form parameters, check that form
parameters were already read.


asoldano:
Thanks!

PR #1325, opened 2017-10-21
[RESTEASY-1741] Dependencies update

rsvoboda:
https://issues.jboss.org/browse/RESTEASY-1741

Dependencies update to catch up with upstream and WF dependencies.
Some dependencies updates are causing problems, will be solved case by case.

### DIFF
--- a/arquillian/RESTEASY-1056-jetty-bv11/pom.xml
+++ b/arquillian/RESTEASY-1056-jetty-bv11/pom.xml
@@ -54,7 +54,7 @@
     </dependency>
     <dependency>
         <groupId>com.sun.xml.bind</groupId>
-        <artifactId>jaxb-impl</artifactId>
+        <artifactId>jaxb-core</artifactId>
     </dependency>
     <dependency>
         <groupId>org.jboss.resteasy</groupId>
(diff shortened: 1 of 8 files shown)


rsvoboda:
Java 9 fails :( The rest is OK.

rsvoboda:
I get PASS on my local machine when ddb8f528fcdc7fdd7eba2303b2eee90f128353e5 commit is combined with
```diff
-            <modular.jdk.args>--add-modules=java.se</modular.jdk.args>
+            <modular.jdk.args>--add-modules=java.se,java.xml.bind</modular.jdk.args>
```

@asoldano is it ok for you to enhance `add-modules` by `java.xml.bind` ?

rsvoboda:
jaxb shift from 2.2.7 to 2.2.11.jbossorg-1
and there is split in jaxb packages - core + impl etc.

but do not know details what could trigger the need for java.xml.bind in add-modules tbh

asoldano:
@rsvoboda FYI, this https://github.com/asoldano/jbossws-cxf/commit/50007e8e7baff47bd84ae39820d8f0730716d780 is what has just allowed me to stay with java.se module dependency only in jbossws and consume jaxb-runtime from maven dependencies instead.

rsvoboda:
Hi @asoldano, I tried to add javax.activation:activation to the first failing module (providers/jackson) but see the same exceptions.

```
java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBException

```
Feels like I need jaxb-api dependency ... either via add-module or javax.xml.bind:jaxb-api:2.3.0

asoldano:
This is superseded by https://github.com/resteasy/Resteasy/pull/1332

PR #1329, opened 2017-10-24
[RESTEASY-1025] async ClientHttpEngine documentation.

rsearls:


asoldano:
Thanks Rebecca, merging. I'm actually moving the doc added here to the RESTEasy Client API chapter, as it's not really something strictly related to JAX-RS 2.1

PR #1251, opened 2017-08-20
[RESTEASY-1566]

ronsigal:
I've merged the code from https://github.com/resteasy/Resteasy/pull/1093
into master branch, along with documentation in StringConverter.xml in
the User Guide.

PR #1331, opened 2017-10-25
[RESTEASY-1746] Provide support for multi valued resource method array type arguments

NicoNes:
This PR is about providing support for custom conversion between multi valued param (queryParam, MatrixPartam, headerParam...) and array type arguments as it has been done in https://github.com/resteasy/Resteasy/pull/1251


asoldano:
This looks ok to me. @ronsigal , do you want to have a look too, given you reviewed RESTEASY-1566 previously?

ronsigal:
@asoldano, it looks OK to me.

@NicoNes, I've attached a version of Chapter 25 "String marshalling for String based @*Param" of the Resteasy User Guide with an array example. Could you take a look, correct it if necessary, and include it with your pull request? Thanks.

PR #1336, opened 2017-10-26
[RESTEASY-1701] Fix early connection completion on async response filters

asoldano:
rebased and more backward compatible version of https://github.com/resteasy/Resteasy/pull/1318


asoldano:
Considering @FroMage 's +1 on #1318, I've merged this. Thanks!

PR #1337, opened 2017-10-27
[RESTEASY-1739]

ronsigal:
ResteasyProviderFactory handles ExceptionMappers and
ParamConverterProviders according to @Priority.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
@@ -75,6 +75,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -106,21 +107,21 @@ public class ResteasyProviderFactory extends RuntimeDelegate implements Provider
       
       public int priority = Priorities.USER;
 
-      private SortedKey(Class<?> intf, T reader, Class<?> readerClass, int priority, boolean isBuiltin)
+      protected SortedKey(Class<?> intf, T reader, Class<?> readerClass, int priority, boolean isBuiltin)
       {
          this(intf, reader, readerClass);
          this.priority = priority;
          this.isBuiltin = isBuiltin;
       }
 
-      private SortedKey(Class<?> intf, T reader, Class<?> readerClass, boolean isBuiltin)
+      protected SortedKey(Class<?> intf, T reader, Class<?> readerClass, boolean isBuiltin)
       {
          this(intf, reader, readerClass);
          this.isBuiltin = isBuiltin;
       }
 
 
-      private SortedKey(Class<?> intf, T reader, Class<?> readerClass)
+      protected SortedKey(Class<?> intf, T reader, Class<?> readerClass)
       {
          this.readerClass = readerClass;
          this.obj = reader;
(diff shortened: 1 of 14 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -1428,6 +1535,8 @@ protected int getPriority(Integer override, Map<Class<?>, Integer> contracts, Cl
          Integer p = contracts.get(type);
          if (p != null) return p;
       }
+      // Check for weld proxy.
+      component = component.isSynthetic() ? component.getSuperclass() : component;
```
sermojohn: I agree with this approach, so the check applies to any Provider priority retrieval.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -156,8 +157,45 @@ public int compareTo(SortedKey<T> tMessageBodyKey)
       {
          return template;
       }
+
+       public T getObj() {
+           return obj;
+       }
    }
+   
+   protected static class ExtSortedKey<T> extends SortedKey<T>
```
sermojohn: I do not see any reason for extending the SortedKey comparator. It seems that it provides no extra comparison logic as equal object comparison is already performed in the first step of SortedKey#compareTo.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -383,13 +438,28 @@ protected void initializeRegistriesAndFilters()
       if (stringConverters == null && parent != null) return parent.getStringConverters();
       return stringConverters;
    }
-   protected List<ParamConverterProvider> getParamConverterProviders()
+   
+   public List<ParamConverterProvider> getParamConverterProviders()
```
sermojohn: This method and state it maintains (paramConverterProviders) look obsolete.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
ronsigal: @sermojohn, given instances of
```
   @Provider
   @Priority(20)
   public static class ExceptionMapper1 implements ExceptionMapper<TestException> 
```
and
```
   @Provider
   @Priority(20)
   public static class ExceptionMapper2 implements ExceptionMapper<TestException> 
```
SortedKey.compareTo() will return 0 and ExtSortedKey.compareTo() will return -1. The latter will result in an insertion into sortedParamConverterProviders, and the former will not (because an "equal" element already exists). I.e., a ParamConverterProvider will get lost with SortedKey.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
ronsigal: @sermojohn, that's one of the methods I included to maintain API compatibility.  I changed your method
```
protected Set<SortedKey<ParamConverterProvider>> getParamConverterProviders()
```
to
```
protected Set<SortedKey<ParamConverterProvider>> getSortedParamConverterProviders()
```
and restored the original
```
   public List<ParamConverterProvider> getParamConverterProviders()
```

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
sermojohn: You are totally right, I did not spot this!

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
sermojohn: @ronsigal I do not get the meaning of being backwards compatible on public methods that are not part of the interfaces that the class implements. I would assume that these methods (like getExceptionMappers() and getParamConverterProviders()) should never be called from another class, if the "programming to an interface" principle. What do you think?

ronsigal: @sermojohn, well, maybe "API compatibility" isn't quite the right term. But, technically, public and even protected methods are accessible to Resteasy users. However inadvisable, someone COULD be calling getParamConverterProviders(). All things being equal, I would like to prevent breaking someone's code unless there's a good reason.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
ronsigal: And that's what I spent yesterday doing. ;-)

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
asoldano: @sermojohn while I agree in principle with you, what @ronsigal is doing is unfortunately needed at least till we go and rework the way RESTEasy is split into components and integrated in WildFly. That's https://issues.jboss.org/browse/RESTEASY-1533 btw, but I'm not very confident we'll be able to work on that in RESTEasy 4.

PR #1339, opened 2017-10-29
Added PERMANENT_REDIRECT to HttpResponseCodes

mithunsasidharan:
<a href="https://tools.ietf.org/html/rfc7538">RFC 7538</a> describes the behavior of the HTTP status code <strong>308</strong>. While this RFC is quite new it solves an actual problem regarding POST redirections.

And it completes quite well the 307 status code found in the <a href="https://tools.ietf.org/html/rfc7231">HTTP 1.1 RFC</a> (previously described in the obsolete <a href="https://tools.ietf.org/html/rfc2616">RFC 2616</a>).

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/util/HttpResponseCodes.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/util/HttpResponseCodes.java
@@ -141,6 +141,15 @@ public interface HttpResponseCodes
    public static final int SC_TEMPORARY_REDIRECT = 307;
 
    /**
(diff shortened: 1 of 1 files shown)


asoldano:
@mithunsasidharan , thanks for your contribution. The change is straightforward, however, before I merge it, I'd need to you to explicitly agree to the statement below.
We require that all contributions be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0
Do you agree to these terms?

mithunsasidharan:
@asoldano : I agree to the terms. Thank you.

PR #1346, opened 2017-11-13
Resteasy 1753: Bean validation handler should not swallow ConstraintDefinitionException, ConstraintDeclarationException nor GroupDefinitionException

NicoNes:


ronsigal:
Ok, I now believe that there is no down side to this change. There is no detectable change, except in the case that someone is using an implementation of ExceptionMapper<ResteasyViolationException>.

ronsigal:
I've created pull request https://github.com/resteasy/Resteasy/pull/1368 which has an addition to the User Guide discussing the change I just mentioned.

PR #1349, opened 2017-11-20
Do not stringify FormData

dumitrubogdanmihai:
**resteasy-jsapi/resteasy-client.js** Do not stringify FormData.

Developers may send the form content inside FormData to be properly handled by XMLHttpRequest.

### DIFF
--- a/resteasy-jsapi/src/main/resources/resteasy-client.js
+++ b/resteasy-jsapi/src/main/resources/resteasy-client.js
@@ -108,7 +108,7 @@ REST.Request.prototype = {
 				}else if(this.entity instanceof Document){
 					if(!contentTypeSet || REST._isXMLMIME(contentTypeSet))
 						data = this.entity;
-				}else if(this.entity instanceof Object){
+				}else if(this.entity instanceof Object && !(this.entity instanceof FormData)){
 					if(!contentTypeSet || REST._isJSONMIME(contentTypeSet))
 						data = JSON.stringify(this.entity);
 				}


asoldano:
hi @dumitrubogdanmihai ,
thanks for reaching us. Would it be possible for you to provide some additional info on the defect here or perhaps (even better) a testcase / example ?

dumitrubogdanmihai:
Here more details about my usecase. It was hard for me to create a complete test for this because I had to do another hack to be able to achieve sending a file with its filename like this:
```
var request = new XMLHttpRequest();
request.open("POST", "http://blablabla");
request.send(new FormData(formElement));
```

```
So we had a rest endpoint like:
  @POST
  @Path("/upload")
  @Consumes(MediaType.MULTIPART_FORM_DATA)
  public Response uploadFile(MultipartFormDataInput uploadedFile) {
  ....
```
  
that generated the following javascript function:

```
var RESTFileService = {};
// POST /upload
RESTFileService.uploadFile = function(_params){
 var params = _params ? _params : {};
 var request = new REST.Request();
 request.setMethod('POST');
 var uri = params.$apiURL ? params.$apiURL : REST.apiURL;
 uri += '/upload';
 if(params.$entity)
  request.setEntity(params.$entity);
 request.setURI(uri);
 if(params.$username && params.$password)
  request.setCredentials(params.$username, params.$password);
 if(params.$accepts)
  request.setAccepts(params.$accepts);
 else
  request.setAccepts('text/plain');
if (REST.antiBrowserCache == true) {
  request.addQueryParameter('resteasy_jsapi_anti_cache', (new Date().getTime()));
    var cached_obj = REST._get_cache_signature(REST._generate_cache_signature(uri));
    if (cached_obj != null) { request.addHeader('If-Modified-Since', cached_obj[1]['Last-Modified']); request.addHeader('If-None-Match', cached_obj[1]['Etag']);}
}
 if(params.$contentType)
  request.setContentType(params.$contentType);
 else
  request.setContentType('multipart/form-data');
 if(params.$callback){
  request.execute(params.$callback);
 }else{
  var returnValue;
  request.setAsync(false);
  var callback = function(httpCode, xmlHttpRequest, value){ returnValue = value;};
  request.execute(callback);
  return returnValue;
 }
```
In order to be able to send the content of a form wrapped inside FormData like: {$entity: new FormData(this.uploadForm)} I had to do another hack to let the content type to be 'undefined' so that the browser would set the boundary properly,
so I overridden the execute function and checked if the given content type was equal to a inner convention like "NO_CONTENT_TYPE" we set it to 'undefined' and let the request pass.
but REST.Request.execute would stringified the entity if the content type wasn't set:
```
if(this.entity){
	if(this.entity instanceof Element){
		if(!contentTypeSet || REST._isXMLMIME(contentTypeSet))
			data = REST.serialiseXML(this.entity);
	}else if(this.entity instanceof Document){
		if(!contentTypeSet || REST._isXMLMIME(contentTypeSet))
			data = this.entity;
	}else if(this.entity instanceof Object){
		if(!contentTypeSet || REST._isJSONMIME(contentTypeSet))
			data = JSON.stringify(this.entity);
	}
}
```

asoldano:
@rsearls , would you mind dealing with this? Btw, a JIRA describing the "task" should be created and inked here.

rsearls:
Created https://issues.jboss.org/browse/RESTEASY-1770
code review pending.

rsearls:
Proposed change reviewed.  I agree with this change.

asoldano:
@dumitrubogdanmihai we're ready for merging the changes. As last step before that, we require that all contributions be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0.
Do you agree to those terms?

dumitrubogdanmihai:
I agree those terms.
Bogdan

PR #1350, opened 2017-11-27
limit marklimit to max value of int

juddmei:
if marklimit expands to a number greater than max value of an int, the while loop in read() method would become an infinite loop because marklimit is negative.


juddmei:
Cool, thanks! I am fine with the changes. I'll close this PR then.

asoldano:
Thanks!

PR #1351, opened 2017-11-28
[RESTEASY-1757]: fix NPE when input stream is null

JohannesRudolph:
See the attached JIRA issue for details: https://issues.jboss.org/browse/RESTEASY-1757


asoldano:
@JohannesRudolph , thanks for having reported the issue and the fix. We require that all contributions be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0
Do you agree to these terms?

JohannesRudolph:
@asoldano sure, I agree to the Apache License. Glad to help :-)

PR #1352, opened 2017-11-30
[RESTEASY-1761] Limit size of SelfExpandingBufferedInputStream

asoldano:


asoldano:
This is a followup of https://github.com/resteasy/Resteasy/pull/1350

juddmei:
BTW, I think it is better to add some comments to doc so as to reminder the developers this class only support a stream that has size smaller than 1GB. What do you think?

asoldano:
@juddmei , it would make sense. Where (which section) in the doc would you add this piece of information?

juddmei:
How about add it to the Java Doc of the class?

PR #1358, opened 2017-12-09
[RESTEASY-1765] Remove q-qs parameters from header Content-type in responses

gaol:
Jira: https://issues.jboss.org/browse/RESTEASY-1765

EAP 7.1 Jira: https://issues.jboss.org/browse/JBEAP-13928

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
@@ -359,6 +359,17 @@ public class ServerResponseWriter
          map.remove(SegmentNode.RESTEASY_SERVER_HAS_PRODUCES.toLowerCase());
          chosen = new MediaType(chosen.getType(), chosen.getSubtype(), map);
       }
+      if (chosen.getParameters().containsKey("q")) {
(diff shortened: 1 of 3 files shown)


ronsigal:
I've added a note to RESTEASY-1765].

ronsigal:
Hi @gaol,

I'm having second thoughts. I see now that Resteasy is removing "q" and "qs" in every case except when there is no @Produces on the resource class or method (as you mentioned on RESTEASY-1765). The other case was ignored prior to RESTEASY-1227 "MediaType set incorrectly in SegmentNode when multiple media types are offered." In response to RESTEASY-1227, I had Resteasy check for @Produces on relevent MessageBodyWriters, which can be used to supply a MediaType. I guess I didn't notice at the time that "q" and "qs" were being removed in the other cases, and I didn't remove them. Now, I think that, for consistency, they should be removed by default in this case also. But then, we're changing an existing behavior, and I don't think leaving them in is a bug. On the other hand, this new behavior didn't exist before last July, when RESTEASY-1227 was closed. Also, RESTEASY-1227 was open for almost two years before I worked on it, so that case must not be too common. I'm inclined to make the change without requiring a flag. @asoldano, what do you think?

-Ron

gaol:
@ronsigal I agree with you too. +1

ronsigal:
Tests pass.

PR #1361, opened 2017-12-10
RESTEASY-1766: Closing SseEventSink should not call JAX-RS filters

NicoNes:
Closing the SseEventSink instance once an event has been sent should not return any response and thus not pass through the JAX-RS response pipeline.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
@@ -72,9 +72,6 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
       {
          if (asyncContext.isSuspended())
          {
-            //resume(null) will call into AbstractAsynchronousResponse.internalResume(Throwable exc)
-            //The null is valid reference for Throwable:http://stackoverflow.com/questions/17576922/why-can-i-throw-null-in-java
-            //Response header will be set with original one
             asyncContext.getAsyncResponse().complete();
          }
       }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
@@ -131,9 +131,9 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
       {
          throw new IllegalStateException(Messages.MESSAGES.sseEventSinkIsClosed());
       }
-      flushResponseToClient();
       try
       {
+    	 flushResponseToClient();
          writeEvent(event);
 
       }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
@@ -75,7 +75,7 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
             //resume(null) will call into AbstractAsynchronousResponse.internalResume(Throwable exc)
             //The null is valid reference for Throwable:http://stackoverflow.com/questions/17576922/why-can-i-throw-null-in-java
             //Response header will be set with original one
-            asyncContext.getAsyncResponse().resume(Response.noContent().build());
+            asyncContext.getAsyncResponse().complete();
          }
       }
 
@@ -85,8 +85,17 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
    {
       if (!responseFlushed)
       {
-         //set back to client 200 OK to implies the SseEventOutput is ready
-         BuiltResponse jaxrsResponse = (BuiltResponse) Response.ok().type(MediaType.SERVER_SENT_EVENTS).build();
+         BuiltResponse jaxrsResponse = null;
+		 if(this.closed)
+		 {
+			 jaxrsResponse = (BuiltResponse) Response.noContent().build();
+		 }
+		 else
+		 {
+			//set back to client 200 OK to implies the SseEventOutput is ready
+			 jaxrsResponse = (BuiltResponse) Response.ok().type(MediaType.SERVER_SENT_EVENTS).build();
+		 }
+		 
          try
          {
             ServerResponseWriter.writeNomapResponse(jaxrsResponse, request, response,
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
@@ -1,0 +1,108 @@
+package org.jboss.resteasy.test.providers.sse;
+
+import java.util.Arrays;
+
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.logging.Logger;
+import org.jboss.resteasy.utils.PortProviderUtil;
+import org.jboss.resteasy.utils.TestUtil;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(Arquillian.class)
+@RunAsClient
+public class SseEventSinkClosingTest {
+
+	private final static Logger logger = Logger.getLogger(SseEventSinkClosingTest.class);
+
+	@Deployment
+	public static Archive<?> deploy() {
+		WebArchive war = TestUtil.prepareArchive(SseEventSinkClosingTest.class.getSimpleName());
+		war.addClass(SseEventSinkClosingTestResource.class);
+		war.addClass(SseEventSinkClosingTestResource.ContainerFilter.class);
+		return TestUtil.finishContainerPrepare(war, null, Arrays.asList(SseEventSinkClosingTestResource.class),
+				SseEventSinkClosingTestResource.ContainerFilter.class);
+	}
+
+	private String generateURL() {
+		return PortProviderUtil.generateBaseUrl(SseEventSinkClosingTest.class.getSimpleName());
+	}
+
+	@After
+	public void reset() throws Exception {
+		Client client = ClientBuilder.newClient();
+		try {
+			client.target(generateURL()).path(SseEventSinkClosingTestResource.BASE_PATH)
+					.path(SseEventSinkClosingTestResource.RESET_RESPONSE_FILTER_INVOCATION_COUNT_PATH).request()
+					.delete();
+		} finally {
+			client.close();
+		}
+	}
+
+	@Test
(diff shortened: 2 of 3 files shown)


asoldano:
@jimma , please review this.

jimma:
@asoldano  OK. I'll review this today or tomorrow.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
```diff
@@ -0,0 +1,108 @@
...
+				Assert.assertEquals(Integer.valueOf(1), response.readEntity(Integer.class));
+			}
+
+		} finally {
+			client.close();
+		}
+
+	}
+
+}
```
jimma: @NicoNes  SseEventSink is internally constructed with a servlet response.  Should it all invoke response filter when close SseEventSink, no matter the sse event is sent or not ?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTestResource.java
```diff
@@ -0,0 +1,93 @@
...
+			@Context SseEventSink sseEventSink) {
+		requestContext.setProperty(RESPONSE_FILTER_INVOCATION_COUNTER_PROPERTY, this.responseFilterCountInvocation);
+		sseEventSink.send(sse.newEvent("message"));
+		sseEventSink.close();
+	}
+
+	@GET
+	@Path(CLOSE_WITHOUT_SENDING_PATH)
+	@Produces(MediaType.SERVER_SENT_EVENTS)
+	public void closedWithourSending(@Context ContainerRequestContext requestContext, @Context Sse sse,
```
jimma: closedWithoutSending?

NicoNes: Hi @jimma I added this test because it is a possible use case since nothing is preventing us from doing this.
So the SseEventSink should be handle to handle it correctly by sending the appropriate 204 response and thus going through the JAX-RS standard processing pipeline (filters) since no event nor 200 response has already been sen't.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
NicoNes: Based on my understanding of the spec **9.5 Processing Pipeline**, the filters must only been called when sending the initial response, the one with the headers:

> For compatibility purposes, implementations MUST initiate processing of an SSE response when either the
> first message is sent or when the resource method returns, whichever happens first. The initial SSE response,
> which may only include the HTTP headers, is processed using the standard JAX-RS pipeline as described
> in Appendix C. Each subsequent SSE event may include a different payload and thus require the use of a
> specific message body writer. Note that since this use case differs slightly from the normal JAX-RS pipeline,
> implementations SHOULD NOT call entity interceptors on each individual event 1.

So when the first 200 response is sent, either when resource method returns or when a sse event is sent whichever happens first the filters must be called and the headers will be committed.
Then if closing the SseEventSink after that, the filters should not be called and by the way  sending any new response header such as 204 will be useless since headers has already been commited, it will never be received by client.
So to be coherent with the spirit of the specs if headers has already been committed it is uselless to called filters again which can be confusing since headers will never be received by client.

However when closing the SseEventSink if no sse event has been sent, the 204 response should be sent to the client as the initial response and the filters must be called

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
```diff
@@ -85,8 +85,17 @@ protected synchronized void flushResponseToClient()
...
+		 if(this.closed)
+		 {
+			 jaxrsResponse = (BuiltResponse) Response.noContent().build();
+		 }
+		 else
+		 {
+			//set back to client 200 OK to implies the SseEventOutput is ready
+			 jaxrsResponse = (BuiltResponse) Response.ok().type(MediaType.SERVER_SENT_EVENTS).build();
+		 }
+
```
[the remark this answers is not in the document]
NicoNes: With the change I did in close() method (https://github.com/resteasy/Resteasy/pull/1361/files/2edd2389102cbfc199d881de0dadff73707b3071#diff-97c04cdfc0891675b26b1a572dba739cR78) calling `SseEventSink.close()` will only complete the asynchContext. Method flushResponseToClient() will be invoked when resource method returns since it was added to `Cleanables` by SseEventSinkInterceptor (https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventSinkInterceptor.java#L58)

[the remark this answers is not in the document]
jimma: Thanks. This looks good.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
jimma: Just to understand the change for this issue.The cleanable is called after resource method returns, so closedWithoutSending() will call SseEventOuputImpl#close() without process pipeline. The pipeline is processed at these lines  ?

> 	 if(this.closed)
> 	 {
> 		 jaxrsResponse = (BuiltResponse) Response.noContent().build();
> 	 }

NicoNes: Yes calling SseEventOuputImpl#close() will just complete the asynch response without processing the pipeline.
The processing will be done at line 101 (https://github.com/NicoNes/Resteasy/blob/2edd2389102cbfc199d881de0dadff73707b3071/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java#L101)
>ServerResponseWriter.writeNomapResponse(jaxrsResponse, request, response,
> ResteasyProviderFactory.getInstance(), t -> {
>}, true);

jimma: Thanks @NicoNes . Do you think these are better test names in SseEventSinkClosingTest?
~~Should_NotInvokeResponseFilter_When_ClosingSseEventSinkOnceMessageSent ()~~  to testFilterForEventSent() 
~~Should_InvokeResponseFilterOnlyOnce_When_ClosingSseEventSinkWithoutSending()~~ to testFilterForMethodReturn()

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTestResource.java
jimma: I meant this is typo. The correct one is closedWithoutSending(), right ?

NicoNes: Oh sorry, your'e right 👍

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
NicoNes: Personaly I think that the convention I used is clearer. It allows a quick and better understanding of what is being tested , under what cicumstances and what is the expected result whithout having to read the test code. (https://dzone.com/articles/7-popular-unit-test-naming)

But it's personnal and if it's not the resteasy convention let me know and I will update the names.

jimma: We now name most of resteasy tests with the name convention #3 or something similar in the linked doc : 
>test[Feature being tested]: This one makes it easy to read the test as the feature to be tested is written as part of test name. Although, there are arguments that the “test” prefix is redundant. However, some sections of developer love to use this technique

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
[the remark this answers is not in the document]
NicoNes: Hey @jimma two more points about this class:

- invocation of **SseEventOutputImpl#flushResponseToClient()** at line 128 (https://github.com/resteasy/Resteasy/pull/1361/files#diff-97c04cdfc0891675b26b1a572dba739cL128) should be done in the try/catch block in order to follow the doc requirement about sending event:

> If there is a problem during sending of an event, completion stage will be completed exceptionally.

- **SseEventOutputImpl#send(OutboundSseEvent event, BiConsumer<SseEventSink, Throwable> errorConsumer)** should be marked as **synchronized** to prevent following race condition:

Thread 1 calls **SseEventOutputImpl#send(OutboundSseEvent event, BiConsumer<SseEventSink, Throwable> errorConsumer)** and since it is not closed  goes to line 131 (https://github.com/resteasy/Resteasy/pull/1361/files#diff-97c04cdfc0891675b26b1a572dba739cR137) without entering the method.
Thread 2 calls **SseEventOutputImpl#close()** so the outputstream is being closed. Then thread 1 continues entering **SseEventOutputImpl#writeEvent(OutboundSseEvent event)** and since the outputstream is closed it will fail with a IOException being thrown and **SseEventOutputImpl#close()** being called for nothing a second time.

Actually it's not really bad since object internal state is still consitent but just a kind of incoherent. Making the method **synchronized** will prevent this behavior without really adding more blocking since both **flushResponseToClient()** and **writeEvent(...)** are already **synchronized**.

What do you think ?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseEventSinkClosingTest.java
NicoNes: OK names switched to the one you proposed.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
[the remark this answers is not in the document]
jimma: >invocation of SseEventOutputImpl#flushResponseToClient() at line 128 (https://github.com/resteasy/Resteasy/pull/1361/files#diff-97c04cdfc0891675b26b1a572dba739cL128) should be done in the try/catch block in order to follow the doc requirement about sending event:
+1.  This requires a try/catch block.
> SseEventOutputImpl#send(OutboundSseEvent event, BiConsumer<SseEventSink, Throwable> errorConsumer) should be marked as synchronized to prevent following race condition
I think close() is the priority operation and it shouldn't wait the send event is done then execute. close() can interrupt a send event operation at any time.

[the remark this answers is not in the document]
NicoNes: Ok so if the close() is the priority let's remove the **synchronized** on it so that it truly does not wait for the send event to be done.
However let's replace the **boolean closed** by an AtomicBoolean to follow spec and be sure that close() will only be invoked once:

>Subsequent calls have no effect and are ignored.

so close() will be like:

	@Override
	public void close() {
		if (!closed.compareAndSet(false, true)) {
			return;
		}
		if (asyncContext.isSuspended() && asyncContext.getAsyncResponse() != null) {
			if (asyncContext.isSuspended()) {
				asyncContext.getAsyncResponse().complete();
			}
		}
	}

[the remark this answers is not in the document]
NicoNes: Hi @jimma,

Forget about the last comment about  thread safety for now please. I will open a new PR to talk about this.
So I've just committed the last fix about the  try/catch block.

asoldano:
Thanks @NicoNes and @jimma , I assumed the discussion was over and the changes look ok to me, so I've merged the PR.

PR #1368, opened 2017-12-16
[RESTEASY-1753]

ronsigal:
Adding documentation for the change in RESTEASY-1753.

### DIFF
--- a/docbook/reference/en/en-US/modules/Migration_from_older_versions.xml
+++ b/docbook/reference/en/en-US/modules/Migration_from_older_versions.xml
@@ -13,6 +13,41 @@
     </para>
 </sect1>
 <sect1>
+    <title>Migrating to 4.0.0</title>
+    <para>
+       Version 4.0.0 of Resteasy is the first release to support JAX-RS 2.1 (see
+        <ulink url="https://jcp.org/en/jsr/detail?id=370">https://jcp.org/en/jsr/detail?id=370</ulink>),
+        which is discussed in <xref linkend="JAX-RS_2.1_additions"/>. Changes not related to the new spec
+        include the following.
+    </para>
+    <itemizedlist>
+        <listitem>
+            <para>
+                Prior to release 4.0.0, when bean validation (<xref linkend="Validation"/>) threw instances of exceptions
+            </para>
+            <itemizedlist>
+                <listitem><classname>javax.validation.ConstraintDefinitionException</classname>,</listitem>
+                <listitem><classname>javax.validation.ConstraintDeclarationException</classname>, or</listitem>
+                <listitem><classname>javax.validation.GroupDefinitionException</classname>,</listitem>
+            </itemizedlist>
+            <para>
+                they were wrapped in a
+                <classname>org.jboss.resteasy.api.validation.Resteasy.ResteasyViolationException</classname>,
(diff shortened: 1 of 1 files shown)


ronsigal:
This pull request adds some documentation for https://github.com/resteasy/Resteasy/pull/1346.

docbook/reference/en/en-US/modules/Migration_from_older_versions.xml
```diff
@@ -12,6 +12,41 @@
         <ulink url="http://docs.jboss.org/resteasy/docs/resteasy-upgrade-guide-en-US.pdf">http://docs.jboss.org/resteasy/docs/resteasy-upgrade-guide-en-US.pdf</ulink>
     </para>
 </sect1>
+<sect1>
+    <title>Migrating to 4.0.0</title>
+    <para>
+       Version 4.0.0 of Resteasy is the first release to support JAX-RS 2.1 (see
```
asoldano: It would be probably better to avoid the statement on 4.0.0 being the "first release to support JAX-RS 2.1", as we will eventually backport JAX-RS 2.1 to the RESTEasy 3 series and at that point people might be confused with ordering.

docbook/reference/en/en-US/modules/Migration_from_older_versions.xml
```diff
@@ -12,6 +12,41 @@
         <ulink url="http://docs.jboss.org/resteasy/docs/resteasy-upgrade-guide-en-US.pdf">http://docs.jboss.org/resteasy/docs/resteasy-upgrade-guide-en-US.pdf</ulink>
     </para>
 </sect1>
+<sect1>
+    <title>Migrating to 4.0.0</title>
+    <para>
+       Version 4.0.0 of Resteasy is the first release to support JAX-RS 2.1 (see
+        <ulink url="https://jcp.org/en/jsr/detail?id=370">https://jcp.org/en/jsr/detail?id=370</ulink>),
+        which is discussed in <xref linkend="JAX-RS_2.1_additions"/>. Changes not related to the new spec
+        include the following.
```
asoldano: I'm probably showing up as very picky here, sorry, but afaiu we're going to only list topics that affect migration here, which is not directly related to them being about the jsr370 or not, right? IOW, in the release other non jaxrs21 changes are coming, but they are not going to be listed here if they do not affect migration, right?

docbook/reference/en/en-US/modules/Migration_from_older_versions.xml
ronsigal: Ok, I changed the section title to "Migrating from 3.0.x to 4.0.0" and opened the section with "In releases 3.0.x". That same text should work for 3.5.

docbook/reference/en/en-US/modules/Migration_from_older_versions.xml
ronsigal: On the contrary, you're exactly right. I've removed the whole opening paragraph that referenced JAX-RS 2.1.

ronsigal:
@asoldano, as for whether this was a bug ... . I think of a bug as a failure to adhere to some kind of specification. In this case, the JAX-RS spec doesn't say much about how to report validation violations, so I don't think it was a bug. I would call it a MISTAKE. ;-)

PR #1369, opened 2017-12-18
[RESTEASY-1754] add getter for JAXBContext and add testcase

rsearls:
I have run CI several times.  On each run, one of the env fails on test org.jboss.resteasy.test.providers.sse.SseTest.
The env that fails appears to be random.  I reviewed SseTest.  I don't see that this code change is causing this randon test failure.


asoldano:
@rsearls you PR has a conflict with master at the moment, you should start by resolving it. Thanks

rsearls:
Conflicts resolved.

PR #1370, opened 2017-12-18
Resteasy 1772:  Use kind of PostResourceMethodInvoker instead of Cleanable to achieve sse spec requirement

NicoNes:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/PostResourceMethodInvokers.java
```diff
@@ -0,0 +1,32 @@
...
+ * 
+ * @author Nicolas NESMON
+ *
+ */
+public class PostResourceMethodInvokers {
+
+	private final Set<PostResourceMethodInvoker> invokers;
+
+	public PostResourceMethodInvokers() {
+		this.invokers = new LinkedHashSet<>();
```
asoldano: Do we really need a Set internally relying on a Map here? Can't we simply have an ArrayList ? I don't think we really care about duplicates and ordering (it's really an internal abstraction), do we?
Basically, I'd rather not affecting the memory footprint, considering this object is created for each invocation.

asoldano:
@NicoNes btw, as a consequence of having merged your other PRs, this needs to be rebased against master to resolve conflicts. Can you do that too, please? Thanks

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/PostResourceMethodInvokers.java
NicoNes: Hi @asoldano,

I can't explain why I made this stange choice of using a Set sorry. But you're right there is no need to use such a Collection here and we don't really care about duplicates for this treatment.

Thanks

PR #1374, opened 2017-12-22
New parameter annotations

FroMage:
As discussed recently, here's support for param annotations that take their names from (depending on where they're placed):

- parameter name, via reflection (requires use of `-parameters` `javac` flag, otherwise params are named `arg0..X`)
- setter property name
- field name

The new parameter names are unoriginal and should probably change. I'd call them like the originals, myself, but this may be found to be confusing by users.

Perhaps we should add an optional `String value() default ""` argument, in case you want to override the name, rather than use the old annotation?

I've added tests, but I have a feeling I only tested configuration via the `ResourceBuilder`, not via injectors. Can you confirm/deny? If I have to test the injectors, how would I go about it?

Also, I'm not sure what strategy we should have if parameter names are absent from reflection. As it is, it will default to `arg0`..`argX` names. Should we throw an error? Warning? Nothing? We currently get no warning if you pass a query param `param` to a resource expecting `@PathParam("parm") String param` (note the typo), so I'm torn between treating this as a silent user mistake, and a one-time warning ("Hey, we noticed that your parameter is named `arg0` which indicates you've very likely forgotten to compile with `-parameters`"). WDYT?

### DIFF
--- a/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/CdiInjectorFactory.java
+++ b/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/CdiInjectorFactory.java
@@ -113,15 +113,15 @@ public class CdiInjectorFactory implements InjectorFactory
       return new CdiPropertyInjector(delegate.createPropertyInjector(resourceClass, factory), resourceClass, sessionBeanInterface, manager);
    }
 
-   public ValueInjector createParameterExtractor(Class injectTargetClass, AccessibleObject injectTarget, Class type, Type genericType, Annotation[] annotations, ResteasyProviderFactory factory)
+   public ValueInjector createParameterExtractor(Class injectTargetClass, AccessibleObject injectTarget, String defaultName, Class type, Type genericType, Annotation[] annotations, ResteasyProviderFactory factory)
    {
-      return delegate.createParameterExtractor(injectTargetClass, injectTarget, type, genericType, annotations, factory);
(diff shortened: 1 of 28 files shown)


asoldano:
@FroMage , I've started looking at this. First of all, unfortunately this would need a rebase as it conflicts with some other changes (namely https://github.com/resteasy/Resteasy/commit/fae873c8b2ae407c20b6c9b2c9685fe4d5e13bf7 ): ResourceConstructor and ResourceLocator are now interfaces with the same methods they had before, the actual implementations are in DefaultResourceConstructor and DefaultResourceLocator. I've rebased in https://github.com/resteasy/Resteasy/pull/1467 , should that help.
In the #1467 PR there's also a commit for refactoring the new annotations so that they have the same name of the original ones, yet a different package; I've also added the String value() default "".
As for the strategy in case of missing names from reflection, I'm fine with silently defaulting to arg0..argX as long as the behavior is documented.
I'll think about testing the injectors.
Finally, we should try to ensure backward compatibility, so that we can easily include the changes in WildFly / EAP at a given point (and include the changes in a release soon). So we should add new methods with additional parameter (e.g. in org.jboss.resteasy.spi.InjectorFactory) instead of changing public method signatures (old ones can be marked as deprecated if needed).


https://github.com/resteasy/Resteasy/pull/1467

FroMage:
Looks great to me, I will close this PR then.

PR #1376, opened 2017-12-28
Resteasy 1776: Sse feature should only be enabled when resource method only produces "text/event-stream" and is injected with @Context SseEventSink

NicoNes:
This pull request introduce the concept of **SSE resource method** as described in the specification:

> 9.3 Server API
> ...
> A resource method that injects an SseEventSink and produces the media type text/event-stream is an
> SSE resource method.

It helps to fixe issues listed in https://issues.jboss.org/browse/RESTEASY-1776

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AsyncResponseConsumer.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AsyncResponseConsumer.java
@@ -60,23 +60,15 @@ public abstract class AsyncResponseConsumer
    }
 
    public static AsyncResponseConsumer makeAsyncResponseConsumer(ResourceMethodInvoker method, AsyncStreamProvider<?> asyncStreamProvider) {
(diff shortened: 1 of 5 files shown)


NicoNes:
Thanks to the tests failing I found out that there was a resteasy specific kind of SSE resource method like that:
```
   @GET
   @Path("sse")
   @Produces(MediaType.SERVER_SENT_EVENTS)
   public Publisher<String> sse() {
	   return Flowable.fromArray("one", "two");
   }
```
So I updated the **ResourceMethodInvoker.isSseResourceMethod(ResourceMethod resourceMethod)** to take it into account.

So now conditions to be a SSE method are :

- Produce only text/event-stream

- Be  injected with SseEventSink parameter or a having a return type for which a **org.jboss.resteasy.spi.AsyncStreamProvider** is regsitered in the given application and not being annotated with any **org.jboss.resteasy.annotations.Stream**.

@asoldano let me know if it is acceptable.

PR #1380, opened 2018-01-11
Introduce a JettyClientEngine to use jetty-client with resteasy JAX-RS Client

stevenschlansker:
Jetty has a mature, efficient, and very nice client that is very handy if you are already using Jetty as your server and would prefer a single stack solution over introducing AsyncHttpClient.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.client.jaxrs.engines;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
@@ -24,7 +25,7 @@ import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.client.api.Request;
 import org.eclipse.jetty.client.api.Response;
 import org.eclipse.jetty.client.api.Result;
-import org.eclipse.jetty.client.util.DeferredContentProvider;
+import org.eclipse.jetty.client.util.BytesContentProvider;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.jboss.resteasy.client.jaxrs.AsyncClientHttpEngine;
@@ -106,12 +107,17 @@ public class JettyClientEngine implements AsyncClientHttpEngine {
             request.followRedirects(false);
         }
 
-        final DeferredContentProvider content;
+//        final DeferredContentProvider content;
         if (invocation.getEntity() != null) {
-            content = new DeferredContentProvider();
-            request.content(content);
-        } else {
-            content = null;
+            final ByteArrayOutputStream os = new ByteArrayOutputStream();
+            try {
+                invocation.writeRequestBody(os);
+            } catch (IOException e) {
+                future.completeExceptionally(e);
+                callback.failed(e);
+                return future;
+            }
+            request.content(new BytesContentProvider(os.toByteArray()));
         }
 
         request.send(new Response.Listener.Adapter() {
@@ -165,9 +171,9 @@ public class JettyClientEngine implements AsyncClientHttpEngine {
 
             @Override
             public void onComplete(Result result) {
-                if (content != null) {
-                    content.close();
-                }
+//                if (content != null) {
+//                    content.close();
+//                }
                 try {
                     if (extractor != null) {
                         stream.close();
@@ -189,22 +195,28 @@ public class JettyClientEngine implements AsyncClientHttpEngine {
             }
         });
 
-        if (content != null) {
-            client.getExecutor().execute(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        invocation.writeRequestBody(new JettyContentStream(client.getByteBufferPool(), content));
-                    } catch (IOException e) {
-                        request.abort(e);
-                        future.completeExceptionally(e);
-                        callback.failed(e);
-                    } finally {
-                        content.close();
-                    }
-                }
-            });
-        }
+//        if (content != null) {
+//            client.getExecutor().execute(new Runnable() {
+//                @Override
+//                public void run() {
+//                    final JettyContentStream stream = new JettyContentStream(client.getByteBufferPool(), content);
+//                    try {
+//                        invocation.writeRequestBody(stream);
+//                    } catch (IOException e) {
+//                        request.abort(e);
+//                        future.completeExceptionally(e);
+//                        callback.failed(e);
+//                    } finally {
+//                        try {
+//                            stream.close();
+//                        } catch (IOException e) {
+//                            // ignore
+//                        }
+//                        content.close();
+//                    }
+//                }
+//            });
+//        }
         return future;
     }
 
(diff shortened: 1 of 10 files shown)


stevenschlansker:
Note that I've added a little cleanup around the executor handling -- right now we require `ExecutorService` but never really use it.  In Jetty's case we already have an Executor we can reuse (but unfortunately it is *not* an ExecutorService).  This is not strictly necessary for this feature but it led to some nice cleanup that could opportunistically make 4.0 (and otherwise probably can't come until 5)

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+    };
+
+    private final HttpClient client;
+
+    public JettyClientEngine(HttpClient client) {
+        if (!client.isStarted()) {
+            try {
+                client.start();
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
pennello: Is this how this library prefers to handle the construction of these engines?  Seems unfortunate to depend on using an object that throw a checked exception, but then not have handling for checked exceptions for the construction of these engines.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    public SSLContext getSslContext() {
+        return client.getSslContextFactory().getSslContext();
+    }
+
+    @Override
+    public HostnameVerifier getHostnameVerifier() {
+        throw new UnsupportedOperationException();
```
pennello: Does the lack of an implementation of this method merit documentation?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    public ClientResponse invoke(ClientInvocation invocation) {
+        Future<ClientResponse> future = submit(invocation, true, NOP, null);
+        try {
+            return future.get(client.getIdleTimeout(), TimeUnit.MILLISECONDS);
+        } catch (InterruptedException e) {
+            future.cancel(true);
```
pennello: And reset the interrupt flag?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+            throw clientException(e, null);
+        } catch (ExecutionException | TimeoutException e) {
+            throw clientException(e.getCause(), null); // NOPMD
```
pennello: What was PMD complaining about here?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+    }
+
+    /**
+     * Implementation note: due to lack of asynchronous message decoders the request must either be buffered,
+     * or it must have a {@code null} extractor and type parameter {@code <T>} must be {@link ClientResponse},
+     * which will read the data through its stream.  It is not possible to use the synchronous JAX-RS message
+     * decoding infrastructure without buffering or spinning up auxiliary threads (arguably more expensive than buffering).
```
pennello: The argument would depend on the size of the payload.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+
+            @Override
+            public void onHeaders(Response response) {
+                cr = new JettyClientResponse(invocation.getClientConfiguration(), stream, () -> future.cancel(true));
+                cr.setProperties(invocation.getMutableProperties());
+                cr.setStatus(response.getStatus());
+                cr.setHeaders(extract(response.getHeaders()));
+                if (!buffered) {
+                    if (extractor != null) {
+                        throw new IllegalStateException("TODO: ResultExtractor is synchronous and may not be used without buffering.");
```
pennello: This is neat.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,223 @@
...
+
+            @Override
+            public void onFailure(Response response, Throwable failure) {
+                future.completeExceptionally(failure);
+                callback.failed(failure);
+            }
+
+            @Override
+            public void onSuccess(Response response) {
+                if (buffered) {
```
pennello: In the case that we're unbuffered, we'll have a `null` extractor, so is that why it's okay to do nothing here in that case?

stevenschlansker:
Thanks for the feedback.  I'm working on the test failures and will have a fix for that and the feedback soon.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
stevenschlansker: Yup, exactly.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
stevenschlansker: Yeah, I agree -- we may end up adding the threaded option later, but it's not needed to ship the first version.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
stevenschlansker: Not sure what the consequence of this is.  I mirrored the Apache implementation.

testsuite/arquillian-utils/pom.xml
```diff
@@ -191,7 +191,7 @@
 
         <dependency>
             <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-nop</artifactId>
+            <artifactId>slf4j-simple</artifactId>
```
stevenschlansker: I'm not sure why the test utils are pulling in `slf4j-nop` -- that just leads to no log output, making debugging difficult.

stevenschlansker:
Hi @ronsigal @asoldano sorry for the direct ping, but I am wondering -- is this a contribution RESTEasy could accept?  I'm excited to get it into 4.0 as early as possible!

asoldano:
@stevenschlansker np at all for the direct ping. Yes, this is a contribution we can accept, thanks. It probably just needs to be cleaned up a little bit and enriched with some documentation in the docbook, but in any case it looks interesting :-)
While we're here, this would be your first contribution since I'm leading the project; we started requiring that all contributions be made under the terms of the ASL2.0 License: http://www.apache.org/licenses/LICENSE-2.0 . Do you agree to these terms?

stevenschlansker:
Yes, the code is original and I release it as Apache2.

stevenschlansker:
hi @asoldano we've been running this in production for a while now and haven't found additional issues, I introduced a brief documentation blurb, and rebased it against 4.0.0.Beta2

What else could I do to move this forward?  Thanks!

asoldano:
@stevenschlansker thanks, I've created https://issues.jboss.org/browse/RESTEASY-1862 , will review this again and eventually merge in the next weeks.

asoldano:
@stevenschlansker , I'm closing this and replacing it with https://github.com/resteasy/Resteasy/pull/1495 , which is a rebased version of this PR (to solve the conflict); it has the same commits as this PR, plus a minor fix to the documentation. If the testsuite passes on #1495, I'll merge.

PR #1388, opened 2018-01-19
DynamicFeature should warn user when registering wrong provider type

NicoNes:
Signed-off-by: NicoNes <nicolas.nesmon@gmail.com>

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
@@ -23,6 +23,7 @@ import org.jboss.resteasy.spi.metadata.Parameter;
 import org.jboss.resteasy.spi.metadata.ResourceMethod;
 import org.jboss.resteasy.spi.validation.GeneralValidator;
 import org.jboss.resteasy.spi.validation.GeneralValidatorCDI;
+import org.jboss.resteasy.util.DynamicFeatureContextDelegate;
 import org.jboss.resteasy.util.FeatureContextDelegate;
 
 import javax.ws.rs.ProcessingException;
@@ -103,7 +104,7 @@ public class ResourceMethodInvoker implements ResourceInvoker, JaxrsInterceptorR
       this.resourceMethodProviderFactory = new ResteasyProviderFactory(providerFactory);
       for (DynamicFeature feature : providerFactory.getServerDynamicFeatures())
       {
-         feature.configure(resourceInfo, new FeatureContextDelegate(resourceMethodProviderFactory));
+         feature.configure(resourceInfo, new DynamicFeatureContextDelegate(resourceMethodProviderFactory));
       }
 
       this.methodInjector = injector.createMethodInjector(method, resourceMethodProviderFactory);
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
@@ -172,6 +172,11 @@ public interface LogMessages extends BasicLogger
    @LogMessage(level = Level.WARN)
    @Message(id = BASE + 195, value = "The following sub-resource methods and sub-resource locators have the same path, [%s].  The paths should be unique.  [%s]")
    void uriAmbiguity(String path, String methodList);
+   
+   @LogMessage(level = Level.WARN)
+   @Message(id = BASE + 196, value = "Component of type %s can't be dynamically bound to a resource method as a %s provider.")
+   void providerCantBeDynamicallyBoundToMethod(Class<?> componentClass, Class<?> providerType);
+
 
    ///////////////////////////////////////////////////////////////////////////////////////////////////////////
    //                                                  INFO                                                 //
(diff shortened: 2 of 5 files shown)


jimma:
Can we simply check this in ResourceMethodInvoker ?  If there is dyanmicFeature registered and the wrong providers comes from resourceMethodProviderFactory,  we can log the warning message.

NicoNes:
Can you please show me a simple example of what you mean?

Thanks

NicoNes:
Can you please show me a simple example of what you mean?

Thanks

jimma:
@NicoNes   My idea is add some check after 
https://github.com/resteasy/Resteasy/blob/master/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java#L118 like:
```    
  if (!providerFactory.getServerDynamicFeatures().isEmpty()) {
         if (resourceMethodProviderFactory.getParamConverterProviders().equals(parentProviderFactory.getParamConverterProviders())) {
            //log warning
         }
      }
```

NicoNes:
Ok I get you point. You would like something like that:

```
if (!providerFactory.getServerDynamicFeatures().isEmpty()) {
	if (!resourceMethodProviderFactory.getContexResolvers().equals(
			parentProviderFactory.getContexResolvers())) {
		// log warning
	}
	if (!resourceMethodProviderFactory.getExceptionMappers().equals(
			parentProviderFactory.getExceptionMappers())) {
		// log warning
	}
	if (!resourceMethodProviderFactory.getMessageBodyWriters().equals(
			parentProviderFactory.getMessageBodyWriters())) {
		// log warning
	}
	if (!resourceMethodProviderFactory.getMessageBodyReaders().equals(
			parentProviderFactory.getMessageBodyReaders())) {
		// log warning
	}
	if (!resourceMethodProviderFactory.getDynamicFeatures().equals(
			parentProviderFactory.getDynamicFeatures())) {
		// log warning
	}
}
```
Actually to my opinion, unless I misunterstood your point, there is a major drawback fixing the issue like that.

Suppose that a user is registering following `DynamicFeature`:

```
private static final class CustomDynamicFeature implements DynamicFeature{

	@Override
	public void configure(ResourceInfo resourceInfo, FeatureContext featureContext) {
		featureContext.register(new ContextResolverImpl<Object>());
		featureContext.register(new ExceptionMapperImpl<Throwable>());
		featureContext.register(new MessageBodyWriter<Object>());
		featureContext.register(new MessageBodyReader<Object>());
		featureContext.register(new OtherDynamicFeature());
	}
	
}
```

With your approach all of those providers will be registered in the `resourceMethodProviderFactory` and once done, we will compare the providers registered on this `resourceMethodProviderFactory` and the original one `parentProviderFactory` to log warnings when it is necessary.

Sorry but IMO it's a kind of weird of registering something that we know for sure will not be used, just in order to print warnings. 
A more usual way to do this, is to print warning as soon as the `register(...)` method is invoked and also do not register things we are warning the user about as being forbidden.

jimma:
ResoureceMethodProviderFactory is only intended to resolve the post-matching providers and I don't think this is wired or not acceptable when the providers are registered internally to it and we print some message to warn these providers will be ignored.

NicoNes:
Sorry I don't know what to say more about this approach. 

As a general rule I was expected us to agree about the fact that we should better not register something that we know for sure will never be used and warn user about it instead of register it and warn user about it.
Why do we want to fill up memory  by registering useless provider when we can do the same without registering nothing ?

For example in JAXRS we cannot register two providers of a same type and we have to warn user about the fact the second registration will not be taken into account.
So to achieve this, what is done in `ResteasyProviderFactory.registerProviderInstance()` (https://github.com/NicoNes/Resteasy/blob/5e18d02170f03fec51ee78b05b6ca999f5549842/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java#L1875) is that we first make the check test and if KO the warning is printed but without addding the useless provider since it is useless and forbidden.

jimma:
> Why do we want to fill up memory by registering useless provider when we can do the same without registering nothing ?

Log warning message with new added extra classes increase memory consumption too.

> For example in JAXRS we cannot register two providers of a same type and we have to warn user about the fact the second registration will not be taken into account.
So to achieve this, what is done in ResteasyProviderFactory.registerProviderInstance() (https://github.com/NicoNes/Resteasy/blob/5e18d02170f03fec51ee78b05b6ca999f5549842/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java#L1875) is that we first make the check test and if KO the warning is printed but without addding the useless provider since it is useless and forbidden.

Parent resteasyProviderFactory is the class to get all provider things for all resources in the application and add these wrong providers leads the error processing.   ResourceMethodProviderFactory is created to handle the method providers and it's the method scope thing. It seems adding these providers and simply ignore them doesn't break anything.  

What I suggested is only my idea and I am not trying to ask you to take this approach.  What I want to know is do we need introduce another class to only print warn message and  Is there simple approach to do this job ?

NicoNes:
> What I suggested is only my idea and I am not trying to ask you to take this approach. What I want to know is do we need introduce another class to only print warn message and Is there simple approach to do this job ?

I can understand your point.

Let me know what you decided, I'll let you know if I have another idea.

Thanks

NicoNes:
@asoldano  :+1:

PR #1391, opened 2018-01-23
RESTEASY-1789

jesuino:
RESTEASY-1789: RESTeasy still request for GZIP even with resteasy.allowGzip=false when using the 2.3.x client


asoldano:
@jesuino the changes are causing testsuite regressions.

rsearls:
Why not patch with the code from resteasy-4.0.0-SNAPSHOT org.jboss.resteasy.plugins.providers.RegisterBuiltin
lines 144-155

asoldano:
Looks like there's no more interest in moving this PR forward, closing it.

PR #1395, opened 2018-02-01
Resteasy 1793: SseBroadcasterImpl does not follow the javadoc requirements

NicoNes:
The current SseBroadcasterImpl break following java doc requirements:

Requirement from *javax.ws.rs.sse.SseBroadcaster.close()* doc:

>     Once the SseBroadcaster is closed, subsequent calls have no effect and are ignored.
>     Once the SseBroadcaster is closed, invoking any other method on the
>     broadcaster instance would result in an IllegalStateException being thrown

Requirement from *javax.ws.rs.sse.SseBroadcaster.onClose(Consumer<SseEventSink> onClose*) doc:

>     We have to notify close listeners if the SSE event output has been
>     closed (either by client closing the connection (IOException) or
>     by calling SseEventSink.close() (IllegalStateException)
>     on the server side.

Potential resource leak due to SseEventSink not closed when SseBroadCaster.close() is invoked:

Actually most of the time when a **SseEventSink** is registered to a **SseBroadcaster**, user is expected its termination to be handled by the **SseBroadcaster** itself. So user will never call **SseEventSink.close()** on each **SseEventSink** he registered but instead he will just call **SseBroadcaster.close()**.
So **SseBroadcasterImpl** must be safe enough to ensure that calling **SseBroadcasterImpl.close()** will actually close all regsitered **SseEventSink** and to prevent any concurrent access to **SseBroadcasterImpl.register()** that will add any new **SseEventSink** which will never be closed by the **SseBroadcasterImpl**.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
@@ -149,7 +149,7 @@ public class SseBroadcasterImpl implements SseBroadcaster
       //return event immediately and doesn't block anything
       return CompletableFuture.runAsync(() -> {
          outputQueue.forEach(eventSink -> {
-            SseEventOutputImpl outputImpl = (SseEventOutputImpl) eventSink;
+            SseEventSink outputImpl = eventSink;
             try
 			{
 				outputImpl.send(event).whenComplete((object, err) -> {
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.plugins.providers.sse;
 
+import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -38,13 +39,10 @@ public class SseBroadcasterImpl implements SseBroadcaster
 	   }
       //Javadoc says close the broadcaster and all subscribed {@link SseEventSink} instances.
       //is it necessay to close the subsribed SseEventSink ?
-      outputQueue.forEach(evenSink -> {
-         evenSink.close();
-         closeConsumers.forEach(consumer -> {
-            consumer.accept(evenSink);
-         });
+      outputQueue.forEach(eventSink -> {
+         eventSink.close();
+         notifyOnCloseListeners(eventSink);
       });
-      outputQueue.clear();
    }
    
    private void checkClosed()
@@ -54,7 +52,34 @@ public class SseBroadcasterImpl implements SseBroadcaster
 		   throw new IllegalStateException(Messages.MESSAGES.sseBroadcasterIsClosed());
 	   }
    }
-
+   
+   private void notifyOnCloseListeners(SseEventSink eventSink)
+   {
+	   // First remove the eventSink from the outputQueue to ensure that
+	   // concurrent calls to this method will notify listeners only once for a
+	   // given eventSink instance.
+	   if (outputQueue.remove(eventSink))
+	   {
+			closeConsumers.forEach(consumer -> {
+				consumer.accept(eventSink);
+			});
+	   }
+   }
+   
+   private void notifyOnErrorListeners(SseEventSink eventSink, Throwable throwable)
+   {
+		// We have to notify close listeners if the SSE event output has been
+		// closed (either by client closing the connection (IOException) or by
+		// calling SseEventSink.close() (IllegalStateException) on the server
+		// side).
+		if (throwable instanceof IOException || throwable instanceof IllegalStateException) {
+			notifyOnCloseListeners(eventSink);
+		}
+		onErrorConsumers.forEach(consumer -> {
+			consumer.accept(eventSink, throwable);
+		});
+   }
+   
    @Override
    public void onError(BiConsumer<SseEventSink, Throwable> onError)
    {
@@ -84,26 +109,20 @@ public class SseBroadcasterImpl implements SseBroadcaster
       return CompletableFuture.runAsync(() -> {
          outputQueue.forEach(eventSink -> {
             SseEventOutputImpl outputImpl = (SseEventOutputImpl) eventSink;
-            if (!outputImpl.isClosed())
+            try
+			{
+				outputImpl.send(event).whenComplete((object, err) -> {
+					if (err != null) {
+						notifyOnErrorListeners(eventSink, err);
+					}
+				});
+			}
+            catch (IllegalStateException e)
             {
-               outputImpl.send(event, callAllErrConsumers());
-            }
-            else
-            {
-               outputQueue.remove(eventSink);
-            }
+				notifyOnErrorListeners(eventSink, e);
+			}
          });
       });
-   }
-
-   BiConsumer<SseEventSink, Throwable> callAllErrConsumers()
-   {
-      return (eventSink, err) -> {
-         onErrorConsumers.forEach(consumer -> {
-            consumer.accept(eventSink, err);
-         });
-      };
-
    }
 
 }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
@@ -5,6 +5,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
@@ -12,22 +13,29 @@ import javax.ws.rs.sse.OutboundSseEvent;
 import javax.ws.rs.sse.SseBroadcaster;
 import javax.ws.rs.sse.SseEventSink;
 
+import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
+
 public class SseBroadcasterImpl implements SseBroadcaster
 {
-   private ConcurrentLinkedQueue<SseEventSink> outputQueue = new ConcurrentLinkedQueue<SseEventSink>();
+   private ConcurrentLinkedQueue<SseEventSink> outputQueue = new ConcurrentLinkedQueue<>();
 
    private final List<BiConsumer<SseEventSink, Throwable>> onErrorConsumers = new CopyOnWriteArrayList<>();
 
    private final List<Consumer<SseEventSink>> closeConsumers = new CopyOnWriteArrayList<>();
+   
+   private final AtomicBoolean closed = new AtomicBoolean();
 
    public SseBroadcasterImpl()
    {
-
    }
 
    @Override
    public void close()
    {
+	   if (!closed.compareAndSet(false, true))
+	   {
+		   return;
+	   }
       //Javadoc says close the broadcaster and all subscribed {@link SseEventSink} instances.
       //is it necessay to close the subsribed SseEventSink ?
       outputQueue.forEach(evenSink -> {
@@ -38,29 +46,40 @@ public class SseBroadcasterImpl implements SseBroadcaster
       });
       outputQueue.clear();
    }
+   
+   private void checkClosed()
+   {
+	   if (closed.get())
+	   {
+		   throw new IllegalStateException(Messages.MESSAGES.sseBroadcasterIsClosed());
+	   }
+   }
 
    @Override
    public void onError(BiConsumer<SseEventSink, Throwable> onError)
    {
+	  checkClosed();
       onErrorConsumers.add(onError);
    }
 
    @Override
    public void onClose(Consumer<SseEventSink> onClose)
    {
+	  checkClosed();
       closeConsumers.add(onClose);
    }
 
    @Override
    public void register(SseEventSink sseEventSink)
    {
+	  checkClosed();
       outputQueue.add(sseEventSink);
-
    }
 
    @Override
    public CompletionStage<?> broadcast(OutboundSseEvent event)
    {
+	  checkClosed();
       //return event immediately and doesn't block anything
       return CompletableFuture.runAsync(() -> {
          outputQueue.forEach(eventSink -> {
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java
@@ -7,6 +7,9 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 
@@ -25,9 +28,24 @@ public class SseBroadcasterImpl implements SseBroadcaster
    private final List<Consumer<SseEventSink>> closeConsumers = new CopyOnWriteArrayList<>();
    
    private final AtomicBoolean closed = new AtomicBoolean();
+   
+   // Used to perform a mutual exclusion between register and close operations
+   // since every registered SseEventSink needs to be closed when
+   // SseBroadcaster.close() is invoked to prevent leaks due to SseEventSink
+   // never closed.
+   // Actually most of the time when a SseEventSink is registered to a
+   // SseBroadcaster, user is expected its termination to be handled by the
+   // SseBroadcaster itself. So user will never call SseEventSink.close() on
+   // each SseEventSink he registers but instead he will just call
+   // SseBroadcaster.close().
+   private final Lock readLock;
+   private final Lock writeLock;
 
    public SseBroadcasterImpl()
    {
+	   	ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
+	  	this.readLock = readWriteLock.readLock();
+	  	this.writeLock = readWriteLock.writeLock();
    }
 
    @Override
@@ -37,12 +55,20 @@ public class SseBroadcasterImpl implements SseBroadcaster
 	   {
 		   return;
 	   }
-      //Javadoc says close the broadcaster and all subscribed {@link SseEventSink} instances.
-      //is it necessay to close the subsribed SseEventSink ?
-      outputQueue.forEach(eventSink -> {
-         eventSink.close();
-         notifyOnCloseListeners(eventSink);
-      });
+	   writeLock.lock();
+	   try
+	   {
+	      //Javadoc says close the broadcaster and all subscribed {@link SseEventSink} instances.
+	      //is it necessay to close the subsribed SseEventSink ?
+	      outputQueue.forEach(eventSink -> {
+	         eventSink.close();
+	         notifyOnCloseListeners(eventSink);
+	      });
+	   }
+	   finally
+	   {
+		   writeLock.unlock();
+	   }
    }
    
    private void checkClosed()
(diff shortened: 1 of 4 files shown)


jimma:
> the method SseEventOutputImpl#send(OutboundSseEvent event, BiConsumer<SseEventSink, Throwable> errorConsumer) really necessary ?
The errorConsumer registration could be done on the CompletionStage<?> returned by SseEventOutputImpl#send(OutboundSseEvent event) as I did in SseBroadcasterImpl#broadcast() (https://github.com/resteasy/Resteasy/pull/1395/files#diff-1363d4b305d7f39bc9a10a727872a498R155)

If I remembered this correctly, the CompletableFuture is backed with low priority daemon thread and changed what you proposed can't trigger the error processing immediately and sometimes error consumer even isn't called before the tests execution is finished.

NicoNes:
Hi @jimma 

About the closed state check, it actually would have been simpler to do it using a volatile closed state and a simple lock like that:

```
public class SseBroadcasterImpl implements SseBroadcaster {

	private final Object lock = new Object();
	private volatile boolean closed;

	@Override
	public void close()
	{
		if (this.closed) {
			return;
		}
		synchronized (this.lock) {
			if (this.closed) {
				return;
			}
			...
			this.closed=true;
		}
	}

	@Override
	public void register(SseEventSink sseEventSink)
	{
		if (this.closed) {
			return;
		}
		synchronized (lock) {
			if (this.closed) {
				return;
			}
			...
		}
	}

}
```

But I did it using a `ReadWriteLock` to prevent excessive/useless blocking.
With simple lock, multiple threads calling `register(...)` on an open `SseBroadcasterImpl` instance will have to wait for the lock to be available since it is exclusive.
With `ReadWriteLock` in the same situtation (on an open `SseBroadcasterImpl`) no need to wait since read lock can be held by multiple threads at the same time.

NicoNes:
> If I remembered this correctly, the CompletableFuture is backed with low priority daemon thread and changed what you proposed can't trigger the error processing immediately and sometimes error consumer even isn't called before the tests execution is finished.

Actually since I'm calling `whenComplete()` instead of `whenCompleteAsynch()` to trigger the error it will be called immediately by the same thread that completes the given CompletableFuture

- CompletableFuture java doc says:

> •Actions supplied for dependent completions of non-async methods may be performed by the thread that completes the current CompletableFuture, or by any other caller of a completion method.

By the way since `SseEventOutputImpl.send(...)` returns a `CompletableFuture.completedFuture(...)` no new thread will be neither created nor used. Both send action and error notification will be done in the daemon thread used by `broadcast(...)`  (CompletableFuture.runAsync(...))

Am I wrong ?

jimma:
@NicoNes   The SseEventSinks to broadcast already in a ConcurrentLinkedQueue, so we can remove this lock object and set the close state once enter the close method:
```
        @Override
	public void close()
	{
		this.closed = true;
                outputQueue.forEach(evenSink -> {
                evenSink.close();
               closeConsumers.forEach(consumer -> {
               consumer.accept(evenSink);
               });
             });
          outputQueue.clear();
	}
```
Does this help ?

jimma:
>By the way since SseEventOutputImpl.send(...) returns a CompletableFuture.completedFuture(...) no new thread will be neither created nor used. Both send action and error processing will be done in the same daemon thread used by broadcast(...) (CompletableFuture.runAsync(...))
> Am I wrong ?

What I meant the daemon thread thing is I tried to implement SseOutputImpl.send with async methods and trigger the error with  whenComplete(), but you don't know when this will be executed,  hence the error consumer can't be triggered immediately.
You can change this to throws exception instead of return CompletableFuture.completedFuture(ex), then it can call the withComplete(errorConsumer).So  SseEventOutputImpl should be refactored too.

NicoNes:
Hi @jimma 

About the closed check: Removing the lock object just because the queue is thread safe does not help at all:

**1.** First point: the java doc of the `SseBroadcaster.close()` says that 

> Once the SseBroadcaster is closed, subsequent calls have no effect and are ignored.

- So with the code you shown me without no lock and no check, if two threads are calling the `SseBroadcaster.close()`  at the same time: 
T1 enters the method  and **pause for thread scheduling purpose before line `outputQueue.clear();`**
Then T2 enters the method till the end and exits.
Following cases will then happen:

   -  'closed' status will be set to true twice => useless but not really bad. **OK**
   

   -  `SseEventSink.close()` will be called twice on each eventSink from the queue => useless but not really bad since `SseEventSink.close()` is safe enough according to the  `SseEventSink.close()` doc that guaranty that subsequent calls to  `SseEventSink.close()` have no effect and are ignored. In other world only the first call to `SseEventSink.close()`  will be taken into account. **OK**

    - `consumer.accept(eventSink)` will be called twice on each eventSink from the queue =>being "closed notified" twice for a same `SseEventSink` means that this given `SseEventSink` has been closed twice. It's a kind of non sense. Since a `SseEventSink` can only be closed once, a given close notifier bounded to it should only be notified once . **KO**

So a lock object is needed an minimal acceptable code will be:

```
	@Override
	public void close() {
		this.closed = true;
		synchronized (this.lock)
		{
			outputQueue.forEach(evenSink -> {
				evenSink.close();
				closeConsumers.forEach(consumer -> {
					consumer.accept(evenSink);
				});
			});
			outputQueue.clear();
		}
	}
```
In order to prevent taking the monitor on the lock object each time a better version of the code will be:

	@Override
	public void close() {
		if(this.closed){
			return;
		}
		this.closed = true;
		synchronized (this.lock)
		{
			outputQueue.forEach(evenSink -> {
				evenSink.close();
				closeConsumers.forEach(consumer -> {
					consumer.accept(evenSink);
				});
			});
			outputQueue.clear();
		}
	}

And in order not to use lock anymore and to keep code simple the best option is to use an Atomic boolean:
```
	@Override
	public void close() {
		if(!this.closed.compareAndSet(false,true)){
			return;
		}
		outputQueue.forEach(evenSink -> {
			evenSink.close();
			closeConsumers.forEach(consumer -> {
				consumer.accept(evenSink);
			});
		});
		outputQueue.clear();
	}
```
So hope we are agree on this first point ?

**2.** Second point: A lock is needed to ensure mutual exclusion between  `SseBroadcaster.close()` and `SseBroadcaster.register()` in order to prevent the case I was talking about in my first comment:

>  Potential resource leak due to **SseEventSink** not closed when **SseBroadCaster.close()** is invoked:
>  
>  Actually most of the time when a **SseEventSink** is registered to a **SseBroadcaster**, user is expected its termination to be handled by the **SseBroadcaster** itself. So user will never call **SseEventSink.close()** on each **SseEventSink** he registered but instead he will just call **SseBroadcaster.close()**.
>  So **SseBroadcasterImpl** must be safe enough to ensure that calling **SseBroadcasterImpl.close()** will actually close all regsitered **SseEventSink** and to prevent any concurrent access to **SseBroadcasterImpl.register()** that will add any new **SseEventSink** which will never be closed by the **SseBroadcasterImpl**.

So let's imagine the case without any lock.
Suppose that T1 thread enters the register method pass the closed status check and then pause for thread scheduling purpose before adding the `SseEventSink` to the queue.
Suppose that T2 thread executes the `SseBroadcaster.close()` method and exists.
Then T1 will resume and the `SseEventSink` will be added to the queue => Since the `SseBroadcaster` is closed the newly registered `SseEventSink` will never be used and will never be closed by this `SseBroadcaster`. **KO**

So to fix this a minimal acceptable code will be:

```
 	@Override
 	public void close() {
 		if (!this.closed.compareAndSet(false, true)) {
 		        return;
 		}
 		synchronized (this.lock) {
 			outputQueue.forEach(evenSink -> {
 				evenSink.close();
 				closeConsumers.forEach(consumer -> {
 					consumer.accept(evenSink);
 				});
 			});
 			outputQueue.clear();
 		}
 	}
 
 	@Override
 	public void register(SseEventSink sseEventSink) {
 		synchronized (this.lock) {
 			if (this.closed.get()) {
 		                throw new IllegalStateException(Messages.MESSAGES.
                                 sseBroadcasterIsClosed());
 			}
 			outputQueue.add(sseEventSink);
 		}
 	}
```

In order to prevent taking the monitor on the lock object each time in the `register()` method a better version of the code will be:

```
	@Override
	public void register(SseEventSink sseEventSink) {
		if (this.closed.get()) {
		       throw new IllegalStateException(Messages.MESSAGES.
                                 sseBroadcasterIsClosed());
		}
		synchronized (this.lock) {
			if (this.closed.get()) {
		               throw new IllegalStateException(Messages.MESSAGES.
                                 sseBroadcasterIsClosed());
			}
			outputQueue.add(sseEventSink);
		}
	}
```

And since there is no need of mutual exclusion between concurrent calls of `SseBroadcaster.register()` (this method can be invoked by multiple threads safely since we are using a `ConcurrentLinkedQueue`) and that the mutual exclusion is only needed between `SseBroadcaster.register()` and `SseBroadcaster.close()` lets use a readWriteLock to better achieve this. So the improved version of the code will be:

```
	@Override
	public void close() {
		if (!this.closed.compareAndSet(false, true)) {
			return;
		}
		this.writeLock.lock();
		try {
			outputQueue.forEach(evenSink -> {
				evenSink.close();
				closeConsumers.forEach(consumer -> {
					consumer.accept(evenSink);
				});
			});
			outputQueue.clear();
		} finally {
			this.writeLock.unlock();
		}
	}

	@Override
	public void register(SseEventSink sseEventSink) {
		if (this.closed.get()) {
                     throw new IllegalStateException(Messages.MESSAGES.
                                 sseBroadcasterIsClosed());
		}
		if (this.readLock.tryAcquire()) {
			try {
				if (this.closed.get()) {
		                     throw new IllegalStateException(Messages.MESSAGES.
                                     sseBroadcasterIsClosed());
				}
				outputQueue.add(sseEventSink);
			} finally {
				this.readLock.unlock();
			}
		}else{
			throw new IllegalStateException(Messages.MESSAGES.sseBroadcasterIsClosed());
		}
	}
```

Sorry for this long explainaition but I hope it's clearer.

So If you're agree lets first finish talking about this point before talking again about the CompletableFuture one.

Thanks

-Nicolas

jimma:
>1. First point: 
 And in order not to use lock anymore and to keep code simple the best option is to use an Atomic  boolean
> So hope we are agree on this first point ?

OK. I agree. AtomicBoolean is better. 

>2. Second point 
> So let's imagine the case without any lock.
> Suppose that T1 thread enters the register method pass the closed status check and then pause for > thread scheduling purpose before adding the SseEventSink to the queue.
> Suppose that T2 thread executes the SseBroadcaster.close() method and exists.
> Then T1 will resume and the SseEventSink will be added to the queue => Since the SseBroadcaster >is closed the newly registered SseEventSink will never be used and will never be closed by this > SseBroadcaster. KO

I think this would not a problem.  SseEventSink backed with a http connection. http connection will timeout and be closed in the end.   As you probably notice the comment in SseBroadcaster.close(), I actually think close subscribed SseEventSink in SseBroadcaster.close() is  arguable IMO. When there is a SseEventSink which is both registered to broadcaster and used to send normal sse even message, close this SseEventSink in broadcaster is actually too much . 

> And since there is no need of mutual exclusion between concurrent calls of SseBroadcaster.register() (this method can be invoked by multiple threads safely since we are using a ConcurrentLinkedQueue) and that the mutual exclusion is only needed between SseBroadcaster.register() and SseBroadcaster.close() lets use a readWriteLock to better achieve this. So the improved version of the code will be:

If we have to use ReentrantReadWriteLock, the fair ReentrantReadWriteLock should be used here. Then the waiting writer can take the control instead of another waiting reader.  Under the write/read lock, volatile boolean looks enough and we don't need another lock in AtomicBoolean. WDYT ?
```
   private ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
   private volatile boolean closed;
   @Override
   public void close()
   {
      lock.writeLock().lock();
      try
      {
         if (closed)
         {
            return;
         }
         closed = true;
         outputQueue.forEach(evenSink -> {
            evenSink.close();
            closeConsumers.forEach(consumer -> {
               consumer.accept(evenSink);
            });
         });
         outputQueue.clear();
        
      }
      finally
      {
         lock.writeLock().unlock();
      }
   }
   @Override
   public void register(SseEventSink sseEventSink)
   {

      lock.readLock().lock();
      try
      {
         if (closed)
         {
            throw new IllegalStateException(Messages.MESSAGES.
                                 sseBroadcasterIsClosed());
         }
         outputQueue.add(sseEventSink);
      }
      finally
      {
         lock.readLock().unlock();
      }
   }
```

NicoNes:
> I think this would not a problem. SseEventSink backed with a http connection. http connection will timeout and be closed in the end

Yes for sure it will close at the end but depending on the connection configuration it can take long and long before being closed. And during that time we are spending useless extra resources on server side to keep this `SseEventSink` alive/open instead of closing it since we already know that it will never be used again by the broadcaster. So to my opinion it's a problem that can probably leads to resource leak.
And more, this approach fully respect the spirit of the spec that states that `SseBroadcaster.close()` is responsible of closing all registered `SseEventSink`. In orther words, once closed  the `SseBroadcaster` is not supposed to hold any open `SseEventSink`.

> As you probably notice the comment in SseBroadcaster.close(), I actually think close subscribed SseEventSink in SseBroadcaster.close() is arguable IMO. When there is a SseEventSink which is both registered to broadcaster and used to send normal sse even message, close this SseEventSink in broadcaster is actually too much .

Yes I saw you're comment. Actually you're right, if the SseEventSink is used for both broadcasting and send normal message, closing the SseEventSink  when closing the SseBroadcaster can be too much.
But in the other case if the SseEventSink is just used for broadcasting, we have to close it when the SseBroadcaster  is closed. We can't let this SseEventSink open this way and wait for it to be closed later (thanks to timeout or anything else) => resource leak I was talking above.

Anyway, we can talk about this for long but the spec mades the decision for us and personaly I'm OK with it.

> If we have to use ReentrantReadWriteLock, the fair ReentrantReadWriteLock should be used here. Then the waiting writer can take the control instead of another waiting reader. 

As you may have noticed in `register()` method, I used `this.readLock.tryAcquire()` instead of `lock.readLock().lock();` in order not to block and to fail fast if the writeLock is already held by a thread closing the SseBroadcaster. And with this `this.readLock.tryAcquire()` the fairness policy has no effect.
But I think you're right let's set the fairness policy to true and use `lock.readLock().lock();`. I think it is acceptable fo this case to privilege slow lactency for a lock request instead of throughput. **OK** :+1: 

> Under the write/read lock, volatile boolean looks enough and we don't need another lock in AtomicBoolean. WDYT ?

Whatever the `closed` state is  an AtomicBoolean or just a boolean we still have to check its state before trying to take any lock, in order to prevent excessive and useless locking (in case the SseBroadcaster is already closed for example). So the code will be:

```
	 private ReentrantReadWriteLock lock = new ReentrantReadWriteLock(true);
	 private volatile boolean closed;

	@Override
	public void close() {
		if (closed) {
			return;
		}
		lock.writeLock().lock();
		try {
			if (closed) {
				return;
			}
			closed = true;
			outputQueue.forEach(evenSink -> {
				evenSink.close();
				closeConsumers.forEach(consumer -> {
					consumer.accept(evenSink);
				});
			});
			outputQueue.clear();

		} finally {
			lock.writeLock().unlock();
		}
	}

	@Override
	public void register(SseEventSink sseEventSink) {
		if (closed) {
			throw new IllegalStateException(Messages.MESSAGES
                         .sseBroadcasterIsClosed());
		}
		lock.readLock().lock();
		try {
			if (closed) {
				throw new IllegalStateException(Messages.MESSAGES
                                 .sseBroadcasterIsClosed());
			}
			outputQueue.add(sseEventSink);
		} finally {
			lock.readLock().unlock();
		}
	}
```
If T1 and T2 are calling `close()` method at the same time and if both pass the first `closed` check they will both take the write lock one after the other, but the second one will take it for nothing.

So the AtomicBoolean just prevent this case in the close method() (in a non blocking fashion by the way).

In the register() method it does not change nothing since the AtomicBoolean.get() method is just a non blocking volatile access to a boolean field (the extra cost of invoking method is not really significant IMO).

So it's up to you, but I think that we should better keep it as a way to prevent useless excesivve locking in close() method. And since it is a non blocking structure contrary to read/write lock the throughput will be better, and the only drawback is the little memory cost of creating an AtomicBoolean object instead of just using a boolean. Not significant IMO.

Let me know what you decide.

Thanks

jimma:
> So it's up to you, but I think that we should better keep it as a way to prevent useless excesivve locking in close() method. And since it is a non blocking structure contrary to read/write lock the throughput will be better, and the only drawback is the little memory cost of creating an AtomicBoolean object instead of just using a boolean. Not significant IMO.

This is the minor thing and AtomicBoolean works too.

NicoNes:
Ok thanks. So I just made the change we discussed about fairness policy.

So about the second point: **using completableFuture.whenComplete() in SseBroadcaster to launch error processing.** (https://github.com/NicoNes/Resteasy/blob/690b55fd3de1b9e3785807a5f4248948fb08d83f/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java#L149)

```
   @Override
   public CompletionStage<?> broadcast(OutboundSseEvent event)
   {
	....
			{
				outputImpl.send(event).whenComplete((object, err) -> {
					if (err != null) {
						notifyOnErrorListeners(eventSink, err);
					}
				});
			}
         ...
}
```

First of all to be clear what is done in `SseEventOutputImpl.send()` is OK for me. I was not asking to add any async processing here. And it is not a spec requirement neither, so we are fine.
Returning `CompletableFuture.completedFuture(ex)`/ `CompletableFuture.completedFuture(event)` is also fine and OK for me.

What I was suggesting is to simplify code by removing the `SseEventOutputImpl.send(OutboundSseEvent event, BiConsumer<SseEventSink, Throwable> errorConsumer)` so that we only have the `SseEventOutputImpl.send(OutboundSseEvent event)`.

This way to trigger an error processing immediately right after the send action fails we just have to do:

`SseEventOutputImpl.send(OutboundSseEvent event).whenComplete(BiConsumer<SseEventSink, Throwable> errorConsumer)`

As I was saying the other day `CompletableFuture` guaranty us that in our current case the `BiConsumer<SseEventSink, Throwable> errorConsumer` error processing will be invoked right after the invocation of `CompletableFuture.completedFuture(...)` and in the exact same thread.

If for any reason `SseEventOutputImpl.send()` becomes asynchronous in the future, we will still have the same guaranty. 

- If the `whenComplete(...)` is called when the given `CompletableFuture` is not complete, that is to say if the send action is not finished. Then the `BiConsumer<SseEventSink, Throwable> errorConsumer` will be invoked right away and in the exact same daemon/asynch thread that is computing and that will complete the send action.

- If the `whenComplete(...)` is called when the given `CompletableFuture` has already been completed, that is to say if the send action is already finished. Then the `BiConsumer<SseEventSink, Throwable> errorConsumer` will be invoked right away and in the exact same thread that is invoking the `whenComplete(...)` method.

If user does not want error processing to be done immediately he will be free to do:

`SseEventOutputImpl.send(OutboundSseEvent event).whenAsync(BiConsumer<SseEventSink, Throwable> errorConsumer)`

Do you agree with that ? Is there any problem here ?

jimma:
> Do you agree with that ? Is there any problem here ?
The comment what I added just to explain why the SseEventOutputImpl.send(OutboundSseEvent event, BiConsumer<SseEventSink, Throwable> errorConsumer) is there. And I do agree we can remove this and only keep SseEventOutputImpl.send(OutboundSseEvent event), but you need to do something like completableFuture.completeExceptionally(ex) when there is exception thrown in SseEventoutputImpl.send(),  right ?

NicoNes:
> And I do agree we can remove this and only keep SseEventOutputImpl.send(OutboundSseEvent event), but you need to do something like completableFuture.completeExceptionally(ex) when there is exception thrown in SseEventoutputImpl.send(), right ?

Yep right, the send method will look like this:

```
   public CompletionStage<?> send(OutboundSseEvent event)
   {
	  synchronized (lock)
      {
	      if (closed)
	      {
	         throw new IllegalStateException(Messages.MESSAGES.sseEventSinkIsClosed());
	      }
	      try
	      {
	    	 internalFlushResponseToClient(true);
	         writeEvent(event);

	      }
	      catch (Exception ex)
	      {
	         return CompletableFuture.completedFuture(ex);
	      }
	      return CompletableFuture.completedFuture(event);  
	  }
      }
```
Ok so if we both agree about that I will make the change.

Thanks

jimma:
@NicoNes  CompletableFuture.completedFuture(ex) won't trigger the exception consumer in broadcast method : 
```
if (err != null) {
  notifyOnErrorListeners(eventSink, err);
}
```
You need to complete this stage with exception : completableFuture.completeExceptionally(ex) , right?
BTW, Can you please explain why we need internalFlushResponseToClient(true) ?

jimma:
> The internalFlushResponseToClient(true) is needed in order to throw the original IOException instead of a ProcessingException one, since it is the IOException that will help us to spot a closed connection problem (see SseBroadcasterImpl.notifyOnErrorListeners()).

So the flushResponseToClient() is an unreference method now ? Can you please fix this ? 
And also rename test case name with what Resteasy's approach ? Thanks.

NicoNes:
`flushResponseToClient()` is refrenced here https://github.com/NicoNes/Resteasy/blob/65593301fdaf1ae1c67128e86a90c403936d1115/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventSinkInterceptor.java#L32.
Sure, I'll change test names.

NicoNes:
Test cases renamed :+1:

asoldano:
@NicoNes thanks for the contribution. Just a note for the future, can you please double check that you have the resteasy codestyle/format setup in your IDE? Thanks.

NicoNes:
Hi @asoldano ,

You're right I did not set the codestyle/format files in my IDE.
The eclipse files to use are there https://github.com/resteasy/Resteasy/tree/36f6700202e0819025d0823023db819c2a31e2f4/config/eclipse right?

asoldano:
@NicoNes right, thanks

PR #1397, opened 2018-02-05
[RESTEASY-1794] JAX-RS 2.0 backward compatibility mode

asoldano:
A mechanism for building resteasy-jaxrs and resteasy-client with jaxrs 2.0 dependency instead of jaxrs 2.1. This is off by default ('-Djaxrs20' flag required) and when enabled produces alternative maven artifacts with '-jaxrs20' classifier.


rsvoboda:
@psakar @Skullman EE7 vs EE8 stuff, please look at the changes to be able to build and use '-jaxrs20' artifacts too

psakar:
@asoldano Can you please document on related JIRA ? Do you suppose build will produce only single set of artifacts (jaxrs 2.0 compatible ()) or both at the same time (eg. other consumers (products) may require jaxrs 2.1) ?

asoldano:
@psakar the build will create new (jaxrs21) artifacts by default. Given this is a temporary measure, for a temporary wfly need, the expectation is that whoever will build resteasy from sources, he will manually re-build resteasy-jaxrs and resteasy-client modules (only!) with -Djaxrs20 option enabled.
I'm adding this to the jira too.

rsvoboda:
I noticed that commit has bad id in the message - RESTEASY-1784 instead of RESTEASY-1794

PR #1402, opened 2018-02-07
Upgrade jaxb version to 2.3

jimma:


asoldano:
@jimma btw, we need a RESTEASY jira for tracking this

asoldano:
@jimma actually, I've just created it: https://issues.jboss.org/browse/RESTEASY-1796

PR #1409, opened 2018-02-14
Exclude unstable tests

kanovotn:


kanovotn:
Closed, as the tests failures were addressed before the tag.

PR #1410, opened 2018-02-15
[RESTEASY-1805] Extract interfaces from resource metadata classes

chkal:
This pull request is the first step to implement [RESTEASY-1805](https://issues.jboss.org/browse/RESTEASY-1805) as discussed here:

http://lists.jboss.org/pipermail/resteasy-dev/2017-December/000513.html

A few notes regarding this pull request:
* I extracted interfaces from `ResourceClass`, `ResourceMethod` and `ResourceLocator`.
* The new interfaces are named like the classes are named today. The default implementation got the prefix `Default`. So for example the old class `ResourceClass` was refactored to an interface `ResourceClass` with a default implementation `DefaultResourceClass`.
* There are a few other classes for which this refactoring would make sense. Especially the metadata classes describing parameters. However, I will create a separate pull request for this later.
* Each "extract method" refactoring is a separate commit. I hope this helps during the review.

Please let me know if these changes are fine or if I should change anything.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/LocatorRegistry.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/LocatorRegistry.java
@@ -26,17 +26,18 @@ public class LocatorRegistry
    public LocatorRegistry(Class<?> clazz, ResteasyProviderFactory providerFactory)
    {
       this.providerFactory = providerFactory;
+      ResourceBuilder resourceBuilder = providerFactory.getResourceBuilder();
       if (Proxy.isProxyClass(clazz))
       {
          for (Class<?> intf : clazz.getInterfaces())
          {
-            ResourceClass resourceClass = ResourceBuilder.locatorFromAnnotations(intf);
+            ResourceClass resourceClass = resourceBuilder.locatorFromAnnotations(intf);
             register(resourceClass);
          }
       }
       else
       {
-         ResourceClass resourceClass = ResourceBuilder.locatorFromAnnotations(clazz);
+         ResourceClass resourceClass = resourceBuilder.locatorFromAnnotations(clazz);
          register(resourceClass);
       }
    }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
@@ -48,11 +48,13 @@ public class ResourceMethodRegistry implements Registry
    protected RootClassNode root = new RootClassNode();
    protected boolean widerMatching;
    protected RootNode rootNode = new RootNode();
+   protected ResourceBuilder resourceBuilder;
 
 
    public ResourceMethodRegistry(ResteasyProviderFactory providerFactory)
    {
       this.providerFactory = providerFactory;
+      this.resourceBuilder = providerFactory.getResourceBuilder();
    }
 
    public boolean isWiderMatching()
@@ -67,7 +69,7 @@ public class ResourceMethodRegistry implements Registry
 
    public void addPerRequestResource(Class clazz, String basePath)
    {
-      addResourceFactory(new POJOResourceFactory(clazz), basePath);
+      addResourceFactory(new POJOResourceFactory(resourceBuilder, clazz), basePath);
 
    }
 
@@ -78,13 +80,13 @@ public class ResourceMethodRegistry implements Registry
     */
    public void addPerRequestResource(Class clazz)
    {
-      addResourceFactory(new POJOResourceFactory(clazz));
+      addResourceFactory(new POJOResourceFactory(resourceBuilder, clazz));
    }
 
    @Override
    public void addPerRequestResource(ResourceClass clazz)
    {
-      POJOResourceFactory resourceFactory = new POJOResourceFactory(clazz);
+      POJOResourceFactory resourceFactory = new POJOResourceFactory(resourceBuilder, clazz);
       register(resourceFactory, null, clazz);
       resourceFactory.registered(providerFactory);
    }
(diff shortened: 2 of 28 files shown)


chkal:
Is there anything I can do to help to get this one reviewed? I would love to continue working on this feature, but it would be great to get some first feedback on this part before doing so.

asoldano:
@chkal sorry for the delay here; we've been busy on the 3.5.0 release, for which we had a very tight schedule. I'll review and comment here soon.

chkal:
@asoldano: Thanks a lot. Please let me know if there is anything I should change.

jimma:
I think this is good to merge. @chkal,  will the ResourceClassProcessors interface be added and put in ResteasyProviderFactory in the next PR ? If you already have it, it's no problem to add to this PR. If there is a test to demonstrate the usage of this feature, that will be great.

chkal:
@jimma My idea was to provide a total of thee pull requests representing the three steps which I described in [my mail](http://lists.jboss.org/pipermail/resteasy-dev/2017-December/000513.html). I hoped that this will simplify the reviewing of the pull request. However, I you prefer one big pull request, I could also continue my work and add more commits into this one. This could also work if I slice my work into reasonable sized commits. Just let me know which way you prefer. ;-)

jimma:
@chkal  Put everything in this PR is OK. Thanks .

chkal:
Hi everyone! I just finished the implementation of [RESTEASY-1805](https://issues.jboss.org/browse/RESTEASY-1805). As requested, I added the commits to this pull request and rebased it against master.

I tried to group the changes into individual commits. I hope that this will help during the review. Please let me know if you have any questions. But the overall steps it took me to implement were:

* Extract interfaces for the `Resource[Class/Method/...]` classes
* Refactored `ResourceBuilder` so that callers don't call static methods on it anymore, but instead retrieve a single instance from `ResteasyProviderFactory` and use this one. This is a rather huge change, as it affects many classes, but the individual changes are rather small.
* Added the `ResourceClassProcessor` interface and modified `ResteasyProviderFactory` to add discovered processors to the `ResourceBuilder` instance.

Let me know if you have any questions or ideas for improvement.

I also created a first version of a `ResourceClassProcessor` for the JSR371 reference implementation Ozark. This processor basically adds `@Produces("text/html")` to all MVC controllers which don't specify a different media type. You can have a look at the code [here](https://github.com/chkal/ozark/tree/resteasy-produces/resteasy/src/main/java/org/mvcspec/ozark/resteasy/metadata). Just to prove that this SPI is actually useful. ;-)

chkal:
Sorry, I forget to mention one thing. I just noticed one of the four Travis jobs failed for some reason. But the Travis build on my fork was successful. See the build for the corresponding branch [here](https://travis-ci.org/chkal/resteasy/builds/349696583). So perhaps it is just a matter of restarting the job!?

asoldano:
Ah, and yes that Travis CI failure was most likely a travis env issue.

chkal:
Hi @asoldano! Thanks a lot for your comments.

First of all: I agree that this PR is rather huge. And actually I hoped that I can implement this feature without touching so many files. But unfortunately refactoring `ResourceBuilder` affected many many other classes.

I've two question:
* You mentioned backwards compatibility issues. Could you elaborate on this? AFAIK I didn't touch any existing public APIs. Although "public API" may be a bit of a vague term.
* Could you share some more details about the current roadmap? What is the plan for RESTEasy 4.0? Will the next Wildfly version use RESTEasy 3.5.0?

To sum up: I fully understand your concerns about backporting this to the 3.5.0 branch. Please let me know if there is anything I could to to minimize your concerns about backwards compatibility.

asoldano:
@chkal regarding backward compatibility, right the problem is exactly that unfortunately the RESTEasy project doesn't have a clear distinction between a public API and internal classes. So unless something is fully backward compatible (no changes of public method signatures on / removal of any classes), we need to evaluate the chances that the not backward compatible changes are actually affecting users or not.
As for the roadmap, 3.5.0.Final is already included in WildFly, starting from 12.0.0.Final (see https://github.com/wildfly/wildfly/blob/12.0.0.Final/pom.xml#L195) . RESTEasy 4 is being treated more as a feature boxed release, so we don't have an actual planned release date for 4.0.0.Final yet.
Btw, isolating API from impl (or at least stating clearly what is what) is one of the goals of 4...

chkal:
Hi @asoldano,

> regarding backward compatibility, right the problem is exactly that unfortunately the RESTEasy project doesn't have a clear distinction between a public API and internal classes. So unless something is fully backward compatible (no changes of public method signatures on / removal of any classes), we need to evaluate the chances that the not backward compatible changes are actually affecting users or not.

Ok, understood. If there is anything I can do to help here, just let me know.

> As for the roadmap, 3.5.0.Final is already included in WildFly, starting from 12.0.0.Final (see https://github.com/wildfly/wildfly/blob/12.0.0.Final/pom.xml#L195) . RESTEasy 4 is being treated more as a feature boxed release, so we don't have an actual planned release date for 4.0.0.Final yet.

Thanks for sharing these details. I wasn't aware that there is no concrete plan how to proceed with 4.0 yet.

asoldano:
@chkal so, I've looked at this a little bit more and I think something we could do to be more backward compatible would be keeping the old static methods in ResourceBuilder and old constructors in POJOResourceFactory too. With those restored, the remaining not backward compatible changes are only the one in SingletonResource (which I believe is not a problem) and the ResourceClass/ResourceConstructor/ResourceLocator/ResourceMethod having turned into interfaces. The latter would only affect those having built something really custom (hacky?) directly on resteasy, common users won't notice that for sure.
I've applied the changes to the PR https://github.com/resteasy/Resteasy/pull/1449 , which is built on top of yours, perhaps you can have a look?
In the mean time I'm assessing the risk of backporting that to branch 3.5 with the internal QA team here.

asoldano:
I've merged #1449

PR #1423, opened 2018-02-20
Restore excluded tests

asoldano:
Try restoring excluded tests (RESTEASY-1446, RESTEASY-1778), after having fixed RESTEASY-1800.


asoldano:
12/12 passing runs, I'm merging

PR #1429, opened 2018-02-26
[RESTEASY-1810] Usage of Map.entrySet instead of keySet in tests

rsvoboda:
Use Map.entrySet instead of iterating keySet and getting value later

https://issues.jboss.org/browse/RESTEASY-1810

~~Please **do not merged yet**, we need to discuss few items:~~


rsvoboda:
@asoldano PR updated with legacy-integration-tests

"debugging" blocks commented-out (instead of removal) as discussed on HipChat

PR #1433, opened 2018-03-01
[RESTEASY-1815] Sse tests part2

kanovotn:


kanovotn:
@asoldano Same comment as https://github.com/resteasy/Resteasy/pull/1434

PR #1434, opened 2018-03-01
[RESTEASY-1815 + RESTEASY-1819] Sse tests

kanovotn:


kanovotn:
@asoldano, The failing one test was trying to simulate call of onError callback on SseBroadcaster. After checking this, I think need to use byteman to simulate this in a always repeatable way. I removed the test for now, because there is no support in resteasy testsuite for byteman, I will send this in another PR.

PR #1436, opened 2018-03-02
[RESTEASY-1821] allow JAX-RS annotations in CDI @Stereotype

xstefank:
JIRA: https://issues.jboss.org/browse/RESTEASY-1821

This PR covers only Produces and Consumes. It was submitted only for the approach review, please don't merge before the inclusion of other JAX-RS annotations processing.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/LocatorRegistry.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/LocatorRegistry.java
@@ -26,17 +26,18 @@ public class LocatorRegistry
    public LocatorRegistry(Class<?> clazz, ResteasyProviderFactory providerFactory)
    {
       this.providerFactory = providerFactory;
+      ResourceBuilder resourceBuilder = providerFactory.getResourceBuilder();
       if (Proxy.isProxyClass(clazz))
       {
          for (Class<?> intf : clazz.getInterfaces())
          {
-            ResourceClass resourceClass = ResourceBuilder.locatorFromAnnotations(intf);
+            ResourceClass resourceClass = resourceBuilder.getLocatorFromAnnotations(intf);
             register(resourceClass);
          }
       }
       else
       {
-         ResourceClass resourceClass = ResourceBuilder.locatorFromAnnotations(clazz);
+         ResourceClass resourceClass = resourceBuilder.getLocatorFromAnnotations(clazz);
          register(resourceClass);
       }
    }
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
@@ -48,11 +48,13 @@ public class ResourceMethodRegistry implements Registry
    protected RootClassNode root = new RootClassNode();
    protected boolean widerMatching;
    protected RootNode rootNode = new RootNode();
+   protected ResourceBuilder resourceBuilder;
 
 
    public ResourceMethodRegistry(ResteasyProviderFactory providerFactory)
    {
       this.providerFactory = providerFactory;
+      this.resourceBuilder = providerFactory.getResourceBuilder();
    }
 
    public boolean isWiderMatching()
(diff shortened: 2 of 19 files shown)


xstefank:
Build passed, @asoldano @marekkopecky can you please review?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResourceBuilder.java
```diff
@@ -0,0 +1,36 @@
+package org.jboss.resteasy.spi;
+
+import org.jboss.resteasy.spi.metadata.ResourceClass;
+import org.jboss.resteasy.spi.metadata.ResourceConstructor;
+
+import java.lang.reflect.Method;
+
+public interface ResourceBuilder
```
marekkopecky: I guess that these classes shouldn't be used by RESTEasy users. But this PR change original non-abstract ResourceBuilder class without non-static method to interface. But original ResourceBuilder class was public and had public static methods and inner static classes. @asoldano WDYT? Is it ok to change it?

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
```diff
@@ -78,6 +85,47 @@ public void observeBeforeBeanDiscovery(@Observes BeforeBeanDiscovery event, Bean
       active = true;
    }
 
+   public <T> void observeStereotypes(@WithAnnotations({Stereotype.class}) @Observes ProcessAnnotatedType<T> event, BeanManager manager)
+   {
+      AnnotatedType<T> annotatedType = event.getAnnotatedType();
+      Set<Annotation> jaxRsAnnotations = collectJaxRsAnnotation(annotatedType.getAnnotations(), beanManager);
+
+      //TODO fix isJaxrsAnnotatedClass?
```
marekkopecky: I'm not sure about the details, but this probably needs to be clarified before merging, right?

marekkopecky:
I prefer to have more javadoc in the code for the future updating/improving of RESTEasy&&CDI integration

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
xstefank: sorry, I missed this in cleanup

marekkopecky:
Should stereotype works also with methods? CDI contains default Model stereotype (javax.enterprise.inject.Model) and this stereotype has "\@Target({ElementType.TYPE, ElementType.METHOD, ElementType.FIELD})", so I guess that stereotype should works with methods.
Anyway, stereotypes doesn't work with methods on this PR, I try this:

```
   @GET
   @Path("/produces/custom/object")
   @MediaTypeStereotype
   public CustomObject producesCustomObject()
   {
      return new CustomObject().setA(1).setB(2);
   }
```

marekkopecky:
I guess that this PR handles only Produces&&Consumes annotation. What is the plan about other annotations?

xstefank:
@marekkopecky I will update this PR with other annotations, if the approach will be approved. I don't think the method inclusion can be done in similar way as for classes but I will check

asoldano:
@xstefank @marekkopecky I haven't reviewed this yet, but I looked at https://github.com/resteasy/Resteasy/pull/1410 which would conflict with this and is possibly a more general solution this could be based on... or not?
In any case both this and that share the doubts on backward compatibility, we investigate a bit if what we'd be changing is something the users might be relying directly or not, as that affects our ability to merge to 3.5

xstefank:
@asoldano @marekkopecky Is #1410 going to be merged? I can rebase this PR, as I will need to tackle this to use stereotypes on methods nevertheless. To my knowledge, the approach does not require to change ResourceBuilder to an interface, but the processing of Produces and Consumes is done in the static method which makes it hard to extend/change. In that case I think 1410 would be sufficient for me. In other case, I think it may be possible to keep ResourceBuilder public API without change if I introduce some additional annotation processing in different class.

One more comment - the static methods of ResourceBuilder are used in the Wildfly - https://github.com/wildfly/wildfly/blob/master/jaxrs/src/main/java/org/jboss/as/jaxrs/DeploymentRestResourcesDefintion.java#L216 which would need to change

marekkopecky:
@xstefank This is a question for @asoldano, but I'm looking to #1410 and there is still some unanswered questions about backward compatibility in #1410.

asoldano:
@xstefank nice catch on those methods being used in WFLY. Let me see if #1410 can be improved a little to retain backward compatibility a little bit more.

asoldano:
@xstefank @marekkopecky , FYI #1410 is most likely going to be replaced by #1449

PR #1438, opened 2018-03-02
[RESTEASY-1820] Patch tests

kanovotn:


kanovotn:
@asoldano Thanks, fixed.

PR #1442, opened 2018-03-08
[RESTEASY-1826] added plugins to generated i18n files

rsearls:


asoldano:
@rsearls please note the build is failing

testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/category/Jaxrs21.java
```diff
@@ -0,0 +1,7 @@
+package org.jboss.resteasy.category;
+
+/**
+ * Marker interface for tests which are expected pass only with jax-rs 2.1
+ */
+public class Jaxrs21 {
```
asoldano: @rsearls this class is not needed here (master) and can be removed

rsearls:
File removed

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
```diff
@@ -53,7 +54,9 @@ public Object readFrom(Class<Object> type, Type genericType,
          return jsonb.fromJson(entityStream, genericType);
       } catch (Throwable e)
       {
-         throw new ProcessingException(Messages.MESSAGES.jsonBDeserializationError(e.toString()));
+         LogMessages.LOGGER.debugf(Messages.MESSAGES.jsonBDeserializationError(e.toString()));
```
asoldano: @rsearls  better doing something like:
if (LogMessages.LOGGER.isDebugEnabled()) {
  LogMessages.LOGGER.debugf(Messages.MESSAGES.jsonBDeserializationError(e.toString()));
}
to avoid generating the stacktrace string when not needed

rsearls:
change complete

PR #1444, opened 2018-03-09
[RESTEASY-1832] Eliminate System.out.println

rsvoboda:
Eliminate System.out.println and similar stuff from codebase

https://issues.jboss.org/browse/RESTEASY-1832


rsvoboda:
Update of AsyncBenchTest.java to use logger, it's ignored tests and make sense to see the output ;)
Thanks @asoldano for pointing that out

PR #1449, opened 2018-03-13
[RESTEASY-1805]

asoldano:
A proposal for a more backward compatible version of https://github.com/resteasy/Resteasy/pull/1410

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/resourcefactory/POJOResourceFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/resourcefactory/POJOResourceFactory.java
@@ -25,6 +25,7 @@ public class POJOResourceFactory implements ResourceFactory
    private ConstructorInjector constructorInjector;
    private PropertyInjector propertyInjector;
 
+   @Deprecated
    public POJOResourceFactory(Class<?> scannableClass)
    {
       this(new ResourceBuilder(), scannableClass);
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/resourcefactory/POJOResourceFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/resourcefactory/POJOResourceFactory.java
@@ -25,11 +25,21 @@ public class POJOResourceFactory implements ResourceFactory
    private ConstructorInjector constructorInjector;
    private PropertyInjector propertyInjector;
 
+   public POJOResourceFactory(Class<?> scannableClass)
+   {
+      this(new ResourceBuilder(), scannableClass);
+   }
+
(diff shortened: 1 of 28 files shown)


asoldano:
@chkal please have a look at the PR here. I'm willing to merge this instead of your #1410 , we have decided that these changes can be later merged to 3.5 branch without real backward compatibility risks for the WildFly users.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/resourcefactory/POJOResourceFactory.java
```diff
@@ -25,13 +25,23 @@
    private ConstructorInjector constructorInjector;
    private PropertyInjector propertyInjector;
 
+   public POJOResourceFactory(Class<?> scannableClass)
+   {
+      this(new ResourceBuilder(), scannableClass);
```
chkal: I also thought about simply instantiating a new `ResourceBuilder` in such places like you did here. The problem I saw back then is that this may lead to the situation that different classes may see different metadata for the same class, depending on whether they use the centrally managed builder instance (which may contain `ResourceClassProcessor` instances which could modify the metadata) or create their own builder (without any processors registered). I'm not completely sure if this could cause issues. But I guess it depends on which metadata the individual class requires from the class.

chkal: Answering to my own comment: If I understand it correctly, this method isn't actually called from the RESTEasy code any more bu MAY be used by existing application code? That's the reason for restoring it, correct?

I guess my comment is still valid in this case but the risk may be acceptable. But perhaps these methods should be deprecated in this case!?!

chkal:
Hey @asoldano. Thanks a lot. I posted a few thoughts in the comments.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/resourcefactory/POJOResourceFactory.java
asoldano: I see your point... we should explicitly state that this constructor is deprecated, with a comment explaining the reason. That way the user at least knows.

asoldano: OK, we came to the same consideration. And yes, it's only for applications outside there that could be relying on the removed stuff. The actual code in resteasy (except for tests) is and will use the new code.

chkal: Okay! Sounds great! I agree that keeping the old methods is fine if some deprecation notice is added.

PR #1456, opened 2018-03-18
[RESTEASY-1838]: SseEventSourceImpl.open() should not be blocking

NicoNes:
[https://issues.jboss.org/browse/RESTEASY-1838](https://issues.jboss.org/browse/RESTEASY-1838)

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcastTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcastTest.java
@@ -5,30 +5,23 @@ import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.logging.Logger;
 import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
-import org.jboss.resteasy.plugins.providers.sse.client.SseEventSourceImpl;
 import org.jboss.resteasy.test.providers.sse.resource.SseBroadcastResource;
-import org.jboss.resteasy.test.providers.sse.resource.SseReconnectResource;
 import org.jboss.resteasy.utils.PortProviderUtil;
 import org.jboss.resteasy.utils.TestUtil;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Assert;
 import org.junit.Test;
-import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 
 import javax.ws.rs.client.Client;
-import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
-import javax.ws.rs.core.Response;
-import javax.ws.rs.sse.InboundSseEvent;
 import javax.ws.rs.sse.SseEventSource;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Consumer;
 
 @RunWith(Arquillian.class)
 @RunAsClient
@@ -100,6 +93,8 @@ public class SseBroadcastTest {
             eventSource.open();
             eventSource2.open();
             eventSource3.open();
+            //Let's wait for all SseEventSource to be connected
+            TimeUnit.SECONDS.sleep(3);
 
             client.target(generateURL("/broadcast/start")).request()
                     .post(Entity.entity(textMessage, MediaType.SERVER_SENT_EVENTS));
@@ -139,6 +134,8 @@ public class SseBroadcastTest {
                 logger.error(ex.getMessage(), ex);
             });
             eventSource.open();
+            //Let's wait for SseEventSource to be connected
+            TimeUnit.SECONDS.sleep(3);
 
             client.target(generateURL("/broadcast/listeners")).request().get();
             client.target(generateURL("/broadcast/start")).request()
@@ -175,6 +172,8 @@ public class SseBroadcastTest {
                 logger.error(ex.getMessage(), ex);
             });
             eventSource.open();
+            //Let's wait for SseEventSource to be connected
+            TimeUnit.SECONDS.sleep(3);
 
             client.target(generateURL("/broadcast/listeners")).request().get();
             client.target(generateURL("/broadcast/startAndClose")).request()
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
@@ -86,6 +86,8 @@ public class SseTest
             throw new RuntimeException(ex);
          });
          eventSource.open();
+         //Let's wait for SseEventSource to be connected
+         TimeUnit.SECONDS.sleep(3);
 
          Client messageClient = new ResteasyClientBuilder().connectionPoolSize(10).build();
          WebTarget messageTarget = messageClient.target(generateURL("/service/server-sent-events"));
@@ -195,6 +197,10 @@ public class SseTest
       SseEventSource eventSource2 = SseEventSource.target(target2).build();
       eventSource2.register(checkConsumer);
       eventSource2.open();
+      
+      //Let's wait for all SseEventSource to be connected
+      TimeUnit.SECONDS.sleep(3);
+      
       //Test for eventSource subscriber
 
       client.target(generateURL("/service/server-sent-events/broadcast")).request()
@@ -270,6 +276,8 @@ public class SseTest
             throw new RuntimeException(ex);
          });
          eventSource.open();
+         //Let's wait for SseEventSource to be connected
+         TimeUnit.SECONDS.sleep(3);
 
          Client messageClient = new ResteasyClientBuilder().build();
          WebTarget messageTarget = messageClient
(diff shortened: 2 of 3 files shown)


asoldano:
@jimma I suggest also checking this and the recent SSE related PRs with the TCK

jimma:
@NicoNes  Looks we can remove this connectedLatch. Can the sleep time you added in the test be removed ?  I ran SseTest after remove this connectedLatch and get several failures, don't know if it the issue in test or something related this connectedLatch removal.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcastTest.java
```diff
@@ -175,6 +172,8 @@ public void testBroadcasterOnCloseCallbackCloseSinkOnServer() throws Exception {
                 logger.error(ex.getMessage(), ex);
             });
             eventSource.open();
+            //Let's wait for SseEventSource to be connected
+            TimeUnit.SECONDS.sleep(3);
```
NicoNes: Without this sleep time, invocations of  `client.target(generateURL("/broadcast/startAndClose")).request()
.post(Entity.entity(textMessage, MediaType.SERVER_SENT_EVENTS))` will throw a NPE on server side and will not try to broadcast message since `this.eventSink` will be null because of no SseEventSINK connected yet

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
```diff
@@ -86,6 +86,8 @@ public void testAddMessage() throws Exception
             throw new RuntimeException(ex);
          });
          eventSource.open();
+         //Let's wait for SseEventSource to be connected
+         TimeUnit.SECONDS.sleep(3);
```
NicoNes: Without this sleep time, invocations of `messageTarget.request().post(Entity.text(msg))` will throw a IllegalStateException on server side since no SseEventSINK is connected yet

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
```diff
@@ -195,6 +197,10 @@ public void testBroadcast() throws Exception
       SseEventSource eventSource2 = SseEventSource.target(target2).build();
       eventSource2.register(checkConsumer);
       eventSource2.open();
+      
+      //Let's wait for all SseEventSource to be connected
+      TimeUnit.SECONDS.sleep(3);
```
NicoNes: Without this sleep time, invocations of `client.target(generateURL("/service/server-sent-events/broadcast")).request()
.post(Entity.entity(textMessage, MediaType.SERVER_SENT_EVENTS));` will try to broadcast message too early since no SseEventSINK is connected yet.

NicoNes:
@jimma The sleep time I added to the tests cannot be removed.
Without those sleep times , few tests will fail. 
I 've just added explainaition for few of them.

jimma:
Thanks @NicoNes . So this is required after the connetedLatch removal, right ?

NicoNes:
@jimma right

jimma:
@NicoNes  I think more about this issue and looked at all test failure after remove connectedLatch. 
With awaitConnected() in open() , you don't need write more Thread.sleep to wait the EventSource really connected to server as the change you did in test case.  I think we can provide another non blocking mode open() method instead of remove it. WDYT ?

jimma:
@NicoNes  Just to let you know. I send an email to jaxrs spec list to ask/proposal add another non blocking  open method in EventSource.

NicoNes:
Hi @asoldano, @jimma .

As you said for now the spec requires the `open()` method to be sync/blocking. However the current implementation of this mehtod is not 100% sync/blocking.
Actually it will be sync for 30 seconds at max  waiting for a response body to be sent by the server. If the 30 seconds elapsed and nothing has been sent by the server the method will become async.
Is it the expected behavior for this method ?

jimma:
> Actually it will be sync for 30 seconds at max waiting for a response body to be sent by the server. If the 30 seconds elapsed and nothing has been sent by the server the method will become async.

This is sync method , and you have to wait the connection is created. As I commented to the jaxrs api spec issue, you have to wait the open return , then you can do the other things. Given there is already event consumer to trigger once event is received , this isn't necessary IMO.

```
eventSource.open();  //this is blocked and wait the sse connection
client.target("http://foo.bar/test").request().post(Entity.entity("msg", MediaType.TEXT_PLAIN_TYPE));
```

NicoNes:
Hi @jimma 

> Given there is already event consumer to trigger once event is received , this isn't necessary IMO.

Personaly I totally agree with you.

However since the current version of the spec requires this method to be sync, what I meant in my previous comment was about removing the 30 second timeout to be fully synch and blocking until the sse connection has been established and not just for 30 seconds at max.

```
      public void awaitConnected()
      {
         try
         {
            connectedLatch.await();
         }
         catch (InterruptedException ex)
         {
            Thread.currentThread().interrupt();
         }

}
```

NicoNes:
Hi @jimma @asoldano 

To move forward here too, I created a PR https://github.com/resteasy/Resteasy/pull/1535 just to remove the 30 seconds timeout since the method is suppose to be fully synch

ronsigal:
@jimma, @NicoNes, hey guys, I'm a little late here, but I have a thought. It seems to be that it's reasonable to impose a time limit on open(). It's good enough for sockets, after all. It's just that if awaitConnected() times out, it should throw an exception to indicate that the attempt to make a connection has failed. It's still completely synchronous. Also, I would make the timeout value configurable. Again, it's good enough for sockets.

NicoNes:
Hi @ronsigal ,

I agree with you about the time limit to impose on `open()`.

However instead of adding any new mechanism to handle it and throwing a kind of `TimeoutException` that is not declared by the spec in the signature of the `open()` method, why do we not rely on the timeout that are already configurable on `javax.ws.rs.client.ClientBuilder` such as `javax.ws.rs.client.ClientBuilder.connectTimeout(...)` or `javax.ws.rs.client.ClientBuilder.readTimeout(...)` ?
That is what I was saying to Marek [here ](url)https://github.com/resteasy/Resteasy/pull/1535.

Users will just have to set the timeout they want using  `javax.ws.rs.client.ClientBuilder.readTimeout(...)` on the `javax.ws.rs.Client` instance provided to the SseEventSource through the `WebTarget` instance.
Once it times out, the `open()` method will end and `onErrorConsumer` will be trigered with the exception. 
And by the way, in the same time the http connection will also be closed once the TimeoutException will be thrown without extra mech to do it.

I can be an easy and nice way to achieve what we want, don't you think ?

ronsigal:
Hey @NicoNes,

I semi-agree with you.

1. Using the ClientBuilder.connectTimeout() sounds nice.
2. It seems from https://github.com/eclipse-ee4j/jaxrs-api/issues/625 that there's some sentiment for having both synchronous and asynchronous versions of open(). It seems to me that onErrorConsumer fits well with the asynchronous version. On the other hand, I would think that throwing a timeout exception is more appropriate for the synchronous version. I see that the definition of Socket.connect() looks like
>public void connect(SocketAddress endpoint, int timeout) throws IOException

Maybe SseEventSource.open() can be given a throws clause in the future.

-Ron

NicoNes:
Hey @ronsigal,

> It seems to me that onErrorConsumer fits well with the asynchronous version. On the other hand, I would think that throwing a timeout exception is more appropriate for the synchronous version

I agree with you, throwing an exception is the best behavior for the synchronous version.
If a `throws` clause is added to `SseEventSource.open()` in the future, we will be able to use the `ClientBuilder.connectTimeout()` and throw the exception once request times out.
Let's see what spec guys decide about it.

> It seems from eclipse-ee4j/jaxrs-api#625 that there's some sentiment for having both synchronous and asynchronous versions of open()

It's my personal opinion but, as already discussed with Jimma, I don't undertsand why `open()` has to be synchronous. 
To my mind it should have been async and trigger the right handler on event:
- errors trigger the `onErrorConsumer`
- sse events trigger the `onEventConsumer`
- completion trigger the `onCompleteConsumer`

Anyway, to go forward on this PR and to deal with the way the spec is for now, can we:

1. remove the aritrary 30 second timeout

The problem with this 30 seconds timeout is that if the Sse connection is not established within 30 seconds, the `open()` method will not block any longer and then just becomes async. The user is not warn about any kind of timeout exception. More the http connection is not even killed.

Since we do not have other choice yet, let's suggest user to configure timeout with `ClientBuilder.connectTimeout()`. Timeout error will have to be handled with `onErrorConsumer` as described in `the register(...)` documentation since it is a SseEventSource unrecoverable error.

2. invoke  `connectedLatch.countDown()` when the connection is established. I mean on receiving the HTTP response code instead of doing it when the first SseEvent is sent by the server.

As far as I understand, `open` method should be sync/blocking until the Sse connection is established and not until the first SseEvent is received.
As I was saying in the [JIRA](https://issues.jboss.org/browse/RESTEASY-1838) the server initial response to a SSE connection, that let us know the connection is established, can only contains status code and headers but no content.
So if  `connectedLatch.countDown()` is invoked after invocation of `eventInput = request.method(verb, SseEventInputImpl.class);` we are not good since `request.method(verb, SseEventInputImpl.class);` will wait for a response body to be available before returning.

I made updates to [PR  1535](https://github.com/resteasy/Resteasy/pull/1535) to include this change.

Let me know what you think about thoses points.

Thanks

NicoNes:
Hey @ronsigal,

1. re: "remove the arbitrary 30 second timeout"

> Suppose the user sets a timeout with ClientBuilder.connectTimeout(). What do we do with it? I'm thinking that we do the same thing with connectedLatch.await(), except with the user's timeout value, right? And then call all onError() methods. But what about, as you mention, the http connection?

If user wants to set a timeout for the SseEventSourceClient, he just has to do this:

```
      Client client = ClientBuilder.newBuilder().readTimeout(1, TimeUnit.MILLISECONDS).build();
      WebTarget sseWebTarget = client.target("http://mySsEndpoint");
      SseEventSource sseEventSource = SseEventSource.target(sseWebTarget).build();
      sseEventSource.register(x->{}, y->{
         System.err.println("Error handler where in charge of dealing with error such as timeout");
      });
      sseEventSource.open();
```
Then the SseEventSourceImpl will use the given Client instance and its timeout settings to perform the Sse call at line [342](https://github.com/NicoNes/Resteasy/blob/76438634709248703087a31bf0351f60ebe7518a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L342):

`Response response = request.invoke();`

Then if server takes more then 1 millisecond to answer, a `javax.ws.rs.ProcessingException:RESTEASY004655: Unable to invoke request` will be thrown with `java.net.SocketTimeoutException: Read timed out` as a cause.
When this exception is thrown by the connection provider (Apache HTTP client in the default case), the underlying http connection has already been closed/aborted.

Then the exception will be handled at line [380](https://github.com/NicoNes/Resteasy/blob/76438634709248703087a31bf0351f60ebe7518a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L380) by SseEventSourceImpl. It will then enter in `CLOSED` state, `connectedLatch.countDown()` will be invoked and all registered error listeners will be notified (see Line [441.](https://github.com/NicoNes/Resteasy/blob/76438634709248703087a31bf0351f60ebe7518a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L441))

> In the absence of a user configured connect timeout, it doesn't seem unreasonable to me to have 30 seconds, or some other value, as a default connect timeout, as long as we take responsibility for doing the cleanup.

Personnaly I think that there is no need to set any default timeout for user if he does not explicitly set up one.
That is already what is done for standard request where no timeout is set by default (See [ResteasyClientBuilder](https://github.com/NicoNes/Resteasy/blob/76438634709248703087a31bf0351f60ebe7518a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java#L72) where all timeout are set to -1 by default) and I do not see why Sse request should be different.

2. re: "invoke connectedLatch.countDown() when the connection is established"

> It looks like you're replacing Builder.method() with Invocation.invoke(). Aren't they both getting the first Response?

With my changes here is the execution plan:

1 - Make the call
2 - get the Response
3 - countdown the latch
4 - read the  entity (SseEventInputImpl)

With the previous version:

1 - Make the call
2 - read the  entity (SseEventInputImpl)
3 - countdown the latch

With my version the latch is counted down as soon as the server reply with any 2XX response. And then after that the response body is read.

With the previous version, the body is first read and then the latch is counted down.
So it means that if the server initial response is a 200 but with no entity in the body and then 20 seconds later the first SseEvent is sent then the latch will be counted down 20 seconds later. It means that the `open()` method will block all this time.

ronsigal:
Hi @NicoNes,

re: countDown() before read(): Ah, yes, I see. Thanks.

re: no default timeout: Ok.

Basically, this is all making sense to me after your long explanation. Thanks!

asoldano:
OK, so considering the long discussion above, I'm closing this as it's replaced by #1535 . thanks

PR #1457, opened 2018-03-19
RESTEASY-1839: SseEventSourceImpl unexpected behavior

NicoNes:
https://issues.jboss.org/browse/RESTEASY-1839


jimma:
@NicoNes  I recently fixed couple of issues about reconnect and handle 204 response in SseEventSource . Can you please check if this works for the test case you add ?

NicoNes:
Hi @jimma,

I was your fix about reconnect and 204 and they does not pass the tests cases I added:

> We are expecting the SseEventSource connection to fail when a response other than 200 is received.
> In this case, it must be closed and must notify:
> - error listeners since it is an unrecoverable error.
> - completion listener since no further events will be received.

With your fix 201 for example is as a valid response. =>**KO**


> We are expecting the SseEventSource connection to fail when a 200 response with a Content-Type header unspecified
> or other than text/event-stream is received.
> In this case, it must be closed and must notify:
> - error listeners since it is an unrecoverable error.
> - completion listener since no further events will be received.

With your fix:
- a 200 response with Content-Type : `*/*` is a valid response => **KO**
- a 200 response with Content-Type: text/plain is not a valid repsonse this is OK but is does not notify any listeners nor close the SseEventSource=> **KO**

> We are expecting the SseEventSource to close itself and not try to reconnect on a 204 response.
> In this case, it must be closed and must notify:
> - completion listener since no further events will be received.
> Error listeners are not notified in this case since it is a normal behavior.

With your fix 204 response des not notify the completion listeners => **KO**

> We are expecting the SseEventSource to reconnect when the connection is closed (gracefully or not).
> In this case, no error listener will be notified since it is not an unrecoverable error since we are trying to reconnect.
> Completion listener will not be notified neither since other events will be received.

With your fix gracefull close does not trigger reconnect  => **KO**

jimma:
> With your changes 201 for example is as a valid response. =>KO

Can you explain why we need to handle like  this ?  The consumers for other media type all reads the response if the http response is 201 . 

> a 200 response with Content-Type : */* is a valid response => KO

It's MediaType compatible:
https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java#L329

> a 200 response with Content-Type: text/plain is not a valid repsonse this is OK but is does not notify any listeners nor close the SseEventSource=> KO

Why does it require to notify listener ?  is it a unrecoverable error ?  I didn't find jaxrs spec says something about this.  Will SseEventSource  reconnect for this target if it doesn't close ?

> With your changes 204 response des not notify the completion listeners => KO

There is no content and isn't started, do we need to call completion listener ?

> With your changes gracefull close does not trigger reconnect => KO

Spec group are discussing this: https://github.com/eclipse-ee4j/jaxrs-api/issues/599

NicoNes:
>     With your changes 201 for example is as a valid response. =>KO
> 
> Can you explain why we need to handle like this ? The consumers for other media type all reads the response if the http response is 201 .

https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model => "Any other HTTP response code not listed here, as well as the cancelation of the fetch algorithm by the user agent (e.g. in response to window.stop() or the user canceling the network connection manually) must cause the user agent to fail the connection." => 201, 401... for example are not valid.
Browsers work that way.

> 
>     a 200 response with Content-Type : / is a valid response => KO
> 
> It's MediaType compatible:
> https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/core/MediaType.java#L329

https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model => "HTTP 200 OK responses with a `Content-Type` header specifying the type `text/event-stream`, ignoring any MIME type parameters, must be processed line by line as described below....HTTP 200 OK responses that have a Content-Type specifying an unsupported type, or that have no Content-Type at all, must cause the user agent to fail the connection."
Browsers work that way.

> 
>     a 200 response with Content-Type: text/plain is not a valid repsonse this is OK but is does not notify any listeners nor close the SseEventSource=> KO
> 
> Why does it require to notify listener ? is it a unrecoverable error ? I didn't find jaxrs spec says something about this. Will SseEventSource reconnect for this target if it doesn't close ?

Yes it is an unrecoverable since the media type is not valid (according to the SSE spec https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model, the client must "fail the connection"). And once the connection is failed the eventsource will not reconnect (https://html.spec.whatwg.org/multipage/server-sent-events.html#fail-the-connection)=> unrecoverable
Browsers work that way and does  not perform reconnection

> 
>     With your changes 204 response des not notify the completion listeners => KO
> 
> There is no content and isn't started, do we need to call completion listener ? 

According to the Java doc the completion listener is to be called when no more events will be received.
And yes it is started since open() has been invoked.

jimma:
Please note  https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model is the spec for html5 javascript. And jaxrs 2.1 spec says it is inspired by that and there could be some change for different language, and SSE should work in jaxrs runtime with other jaxrs apis. 

>According to the Java doc the completion listener is to be called when no more events will be received. And yes it is started since open() has been invoked.

But http 204 implies no event is received from beginning, it doesn't make sense to call completion handler.

NicoNes:
> Please note https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model is the spec for html5 javascript. And jaxrs 2.1 spec says it is inspired by that and there could be some change for different language, and SSE should work in jaxrs runtime with other jaxrs apis.

I agree with you, but except for the handling of 503 that is not specificed in the HTML5 specs and that is only in the Java spec. Is there any reason not to follow this spec on the points I spotted?

- According to the JAVA spec (9.3): "A resource method that injects an SseEventSink and produces the media type text/event-stream is an SSE resource method." => so valid content-type seems to be strictly **text/event-stream** not all types compatible with it. =>Will work with other JAXRS api

- About handling response other than 200: The JAVA spec does not say nothing about this and does not provide any mech for user to set it or override it. So I guess it's up to implementor to decide. So since there is no way to overrride it  implementor should better use 200 since it the only response handled by browser. And if we are pretending to do SSE we should better ensure that at least browser will understand it as SSE.
So if server side should always send 200 why should we consider other response code (except 204, 503) as valid in SseEventSource.

>     According to the Java doc the completion listener is to be called when no more events will be received. And yes it is started since open() has been invoked.
> 
> But http 204 implies no event is received from beginning, it doesn't make sense to call completion handler.

SseeventSource works in an asynch fashion so there is to be a callback to notify user who invoked **eventSource.open()** that this event source is over and will not received any event more. To my mind this is one of the purpose of the completion hanlder. According to the java spec it must  be seen as an handlers for SseEventSource life-cycle
Suppose the following code with `d` a resource too be closed when it will no longer be used. Without completion notification how could we know whento close it:

```
 public static void main(String[] args)
      {
         SseEventSource eventSource = ...;
         MyCloseableDispatcher d= ...;
         eventSource.register(e -> {
            d.dispatch(e);
         }, t -> {
         }, () -> {
            d.close();
         });
      }

```

NicoNes:
Other questions: 

- Should 204 be treated as unrecoverable error ? => According to me it's not an error. **NO**
- Should 503 with retry-after set be treated as unrecoverable error ? => According to me it's an error with reconnect as a recovery mech so it's not unrecoverrable. **NO** 
-Should other non 2xx be treated as unrecoverable error ? => According to me  there are real uncoverrable errors with no reconnect. **YES**

Are you OK with the fact that just calling `state.set(State.CLOSED);` in `run()` method is not enough to close the SseEventSource ? We should also close the connection and shutdown the executor.

jimma:
> However the JAVA spec does not say nothing on the 204 neither and we choosed to handle it as the HTML5 specs says to. So maybe a spec clarification will be needed on this whole point too. WDYT?

Sure. We can always talk with jaxrs spec group to get some clarification and get it improved in the next release version. 

 About the completion listener/ onComplete callback, this is the new api when Flowable was introduced in jaxrs api: https://github.com/jax-rs/api/blob/nio/jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java#L238 .  It is intended to be called inside subscriber.onComplete()  and  jaxrs 2.1 spec has some javadoc for this callback : onComplete callback is invoked when there are no  further events to be received. onComplete callback is the thing to be invoked after the event consumer doesn't receive any event.  For http 204 response, it doesn't trigger any consumer, so we don't need to call this callback. I might be wrong on reading this.  But we can always ask spec group to get more clarification. 

> Are you OK with the fact that just calling state.set(State.CLOSED); in run() method is not enough to close the SseEventSource ? We should also close the connection and shutdown the executor.

I didn't think this in depth and I'll look at your change.

NicoNes:
Ok I'm waiting for your feedback on my changes and I'll ask the spec group for more clarification.

Thanks

NicoNes:
Clarification requested: https://github.com/eclipse-ee4j/jaxrs-api/issues/606

NicoNes:
@asoldano, @jimma I've asked expert group for more clarification on few points. So if you're OK I suppose that this PR can be put on hold for now.

Thanks

jimma:
@NicoNes  Sure. Let's wait the clarification from spec group.

NicoNes:
Hey guys,

The expert group stated about that (https://github.com/eclipse-ee4j/jaxrs-api/issues/606) and it is already packaged for a future release:

- https://github.com/eclipse-ee4j/jaxrs-api/pull/670
- https://github.com/eclipse-ee4j/jaxrs-api/pull/833

So if you agree I will review my PR based on what I've been stated and all the changes that was made in the code since I created.

Thanks

jamezp:

PR #1462, opened 2018-03-20
[RESTEASY-1840] Sync surefire version in whole TS, extract versions t…

kanovotn:
…o common version dependencies pom

* Downgrading to 2.18.1 version of surefire fixes the hpux crash issue
* Extract depedency versions from microprofile-tck testsuite module into shared dependency file and unify the surefire version used.
* Update arquillian bom


asoldano:
@kanovotn , I've merged this. Shouldn't we have this on master too, anyway?

PR #1467, opened 2018-03-23
[RESTEASY-1880] Parameter annotations with default names

asoldano:
Rebased / evolved version of #1374


marekkopecky:
More javadoc && community documentation would be useful, but it could be provided by another PR ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/InjectorFactoryImpl.java
```diff
@@ -130,15 +130,15 @@ public ValueInjector createParameterExtractor(Parameter parameter, ResteasyProvi
 
 
    @Override
-   public ValueInjector createParameterExtractor(Class injectTargetClass, AccessibleObject injectTarget, Class type,
+   public ValueInjector createParameterExtractor(Class injectTargetClass, AccessibleObject injectTarget, String defaultName, Class type,
```
marekkopecky: @asoldano This is backward incompatible change, right? I'm looking to TS changes and it seems that this should affect user's deploments ... maybe, it's not necessary to change signature of old public method, we can add new method with additional defaultName parameter, and old method can just call new method with additional defaultName parameter with some reasonable default ...

What are the plans for 3.6 branch?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/InjectorFactoryImpl.java
```diff
@@ -130,15 +130,15 @@ public ValueInjector createParameterExtractor(Parameter parameter, ResteasyProvi
...
                                                  Type genericType, Annotation[] annotations, ResteasyProviderFactory providerFactory)
    {
-      return createParameterExtractor(injectTargetClass, injectTarget, type, genericType, annotations, true, providerFactory);
+      return createParameterExtractor(injectTargetClass, injectTarget, defaultName, type, genericType, annotations, true, providerFactory);
    }
 
    @SuppressWarnings("deprecation")
    @Override
-   public ValueInjector createParameterExtractor(Class injectTargetClass, AccessibleObject injectTarget, Class type, Type genericType, Annotation[] annotations, boolean useDefault, ResteasyProviderFactory providerFactory)
+   public ValueInjector createParameterExtractor(Class injectTargetClass, AccessibleObject injectTarget, String defaultName, Class type, Type genericType, Annotation[] annotations, boolean useDefault, ResteasyProviderFactory providerFactory)
```
marekkopecky: Same here ^^^

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/PathSuppressionTest.java
```diff
@@ -82,7 +84,7 @@ public static void after() throws Exception {
      */
     @Test
     public void testInputSuppressPathDefault() throws Exception {
-        doTestInputViolations("default", "s", "t", "", "test.arg0");
+        doTestInputViolations("default", "s", "t", "", new String[]{"test.arg0", "test.u"});
```
marekkopecky: What is the reason for this change? I run this test and violation object still contains "test.arg0":
```
[PARAMETER]
[test.arg0]
[size must be between 4 and 6]
[c]
```

Btw, "u" is the name of affected parameter, but I don't know, what is the relationship between this name and violation report ...

testsuite/integration-tests/pom.xml
```diff
@@ -390,7 +394,13 @@
                         </execution>
                     </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <parameters>true</parameters>
```
marekkopecky: This line is not necessary, if we define "maven.compiler.parameters" property, right?

https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html

marekkopecky:
New annotations have different packages, but have the same name, eg.: javax.ws.rs.QueryParam x org.jboss.resteasy.annotations.jaxrs.QueryParam
This could be confusing for the users. But if some new jaxrs spec sync behaviour with this feature, it would be really easy to switch from custom resteasy annotation to its jaxrs variant. @asoldano WDYT?

@asoldano Additional tests: this PR doesn't contain test with end-point, that contains both old/jax-rs param annotations and new annotations together. Maybe some negative scenarios would be suitable too, eg. what happen if end point have both old&&new annotation with the same param name ... I can add these tests, if you agree. Or/And do you have also some other kind of tests in your mind?

marekkopecky:
@asoldano I did some EJB/CDI experiments on 3.6 variant of this PR:
* EJB/CDI needs public default constructor, resteasy is unable to inject request parameters to custom constructor, because custom constructor is never called. Is there some way to fix it, or is this by design?
* Resteasy is unable to replace class variable with param annotation, if end-point class is Stateless/Stateful/ApplicationScoped. This variable can be set only once by resteasy. Is there some reason for that? Or is it a bug? Or feature limitation? In my point of view, resteasy should at least print some warning ...
* The same for setters, resteasy call setters with new annotation only once, if Stateless/Stateful/ApplicationScoped is used. Resteasy doesn't call these setters again with another request.
* RequestScope&&(setters||class_variables) works correctly

I test it on [this commit](https://github.com/marekkopecky/Resteasy/commit/6ade48f0328a7084a1c4c233500edb15ac82b750)

resteasy-jaxrs/src/main/java/org/jboss/resteasy/annotations/jaxrs/QueryParam.java
```diff
@@ -0,0 +1,16 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Retention(RUNTIME)
+@Target({ FIELD, METHOD, PARAMETER })
+public @interface QueryParam
+{
+   String value() default "";
```
marekkopecky: @asoldano I assume that value of this annotation should behave like jax-rs QueryParam annotation (behaviour is not described in javadoc or jira). Anyway, I tried to use this parameter:

`@QueryParam("queryParam3") String queryParam3Updated`

And RESTEasy uses value of `queryParam3Updated` HTTP parameter. I guess that RESTEasy should use value of `queryParam3` HTTP parameter, right?

testsuite/integration-tests/pom.xml
FroMage: I didn't know about this property. We can try it.

FroMage:
> this PR doesn't contain test with end-point, that contains both old/jax-rs param annotations and new annotations together. Maybe some negative scenarios would be suitable too, eg. what happen if end point have both old&&new annotation with the same param name ... I can add these tests, if you agree. Or/And do you have also some other kind of tests in your mind?

Yes, those tests sound good. I assume for the last one, if old&new annotation have the same param name, it should behave the same as if you had twice an old annotation with the same name.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/annotations/jaxrs/QueryParam.java
FroMage: Yes, that's my understanding too.

FroMage:
> EJB/CDI needs public default constructor, resteasy is unable to inject request parameters to custom constructor, because custom constructor is never called. Is there some way to fix it, or is this by design?
Resteasy is unable to replace class variable with param annotation, if end-point class is Stateless/Stateful/ApplicationScoped. This variable can be set only once by resteasy. Is there some reason for that? Or is it a bug? Or feature limitation? In my point of view, resteasy should at least print some warning ...
The same for setters, resteasy call setters with new annotation only once, if Stateless/Stateful/ApplicationScoped is used. Resteasy doesn't call these setters again with another request.
RequestScope&&(setters||class_variables) works correctly

I'm not sure these questions are related to the new annotations or the old ones. Can you clarify?

marekkopecky:
@FroMage New annotations.

> EJB/CDI needs public default constructor, resteasy is unable to inject request parameters to custom constructor, because custom constructor is never called. Is there some way to fix it, or is this by design?

Example: 
```
@Path("/")
@Stateless
public class RESTEasyParamStatelessResource {
   private String queryParam0;


   public RESTEasyParamStatelessResource(@QueryParam String queryParam0) {
      this.queryParam0 = queryParam0;
   }

   public RESTEasyParamStatelessResource() {

   }
```
First constructor is never called, second constructor is necessary for EJB bean.

> Resteasy is unable to replace class variable with param annotation, if end-point class is Stateless/Stateful/ApplicationScoped. This variable can be set only once by resteasy.

Example:
```
@Path("/")
@Stateless
public class RESTEasyParamStatelessResource {
   @QueryParam
   private String queryParam1;
```

First request "localhost:8080/applicationName?queryParam1=a" sets `a` to `queryParam1` variable.
Second request "localhost:8080/applicationName?queryParam1=b" doesn't replace the value of `queryParam1`, so `queryParam1` variable still contains `a` value.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/PathSuppressionTest.java
asoldano: AFAIR, this was because of different versions of hibernate validator building messages differently

marekkopecky: Hmm ... so this PR changes hibernate validator version somehow? Or it was changed in some other PR?

testsuite/integration-tests/pom.xml
asoldano: @marekkopecky right, thanks, fixed

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/PathSuppressionTest.java
asoldano: No, afair it's basically that by adding the <maven.compiler.parameters>true</maven.compiler.parameters> configuration, we're exposing a difference in behavior of new and old hibernate validator versions.

asoldano:
`New annotations have different packages, but have the same name, eg.: javax.ws.rs.QueryParam x org.jboss.resteasy.annotations.jaxrs.QueryParam
This could be confusing for the users. But if some new jaxrs spec sync behaviour with this feature, it would be really easy to switch from custom resteasy annotation to its jaxrs variant. @asoldano WDYT?`
Actually I used the same name, with different packages, on purpose. The reason is to allow for easy switch between versions (basically, you usually just fix imports in the code) and because figuring out another set of decent names that would really make sense was quite hard.

`@asoldano Additional tests: this PR doesn't contain test with end-point, that contains both old/jax-rs param annotations and new annotations together. Maybe some negative scenarios would be suitable too, eg. what happen if end point have both old&&new annotation with the same param name ... I can add these tests, if you agree. Or/And do you have also some other kind of tests in your mind?`
Sure, feel free to provide tests if you have some time, I appreciate that. As for what happens if someone uses both old and new annotations... let's first say that it's a nonsense; when the old annotations only are available, I doubt anyone would try using them twice, regardless the fact that the compiler would complain as most (all?) of them are non repeatable. This said, we can try ensuring that standard (old) ones overrides the resteasy (new) ones and check that with a test, if that is a real concern.

asoldano:
@marekkopecky , regarding your questions on EJB/CDI, I also share @FroMage doubts (basically, is this really something related to the additions of the new annotations?).
In any case, look at the javadoc for QueryParam, for example, I found:
` * <p>Because injection occurs at object creation time, use of this annotation
 * on resource class fields and bean properties is only supported for the
 * default per-request resource class lifecycle. Resource classes using
 * other lifecycles should only use this annotation on resource method
 * parameters.</p>`
... which might explain the behavior you're seeing, right?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/annotations/jaxrs/QueryParam.java
asoldano: +1.
@marekkopecky , maybe you can provide a test on this too as part of the tests you're working on? Then we (I) have a look :)

marekkopecky:
> ... which might explain the behavior you're seeing, right?

@asoldano I agree, it sounds reasonable ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/annotations/jaxrs/QueryParam.java
marekkopecky: @asoldano Sure. I will prepare PR tomorrow or on Monday ...

marekkopecky: @asoldano I prepare tests in [my branch](https://github.com/marekkopecky/Resteasy/commits/RESTEASY-1880-master-tests) for master upstream branch. Custom values of new annotations are tested [here](https://github.com/marekkopecky/Resteasy/commit/ea872100cb47d1ab567a61b52bc8eb989ec2215e#diff-2923ffdb33f3ce8e5e45ef599d048b96R133). If you want, I can prepare PR or prepare branch with tests for 3.6.

marekkopecky: https://github.com/resteasy/Resteasy/pull/1510
RESTEasyParamBasicTest#customValuesTest is in  ExpectedFailing category in this PR ...

asoldano:
@marekkopecky  , @FroMage  I've fixed the failing test and re-enabled it. So given all concerns are basically solved, I'm merging this. Thanks!

asoldano:
@FroMage if you have time for that... it would be appreciated for sure :-) Otherwise I can do that.

FroMage:
If you tell me where you think it should go, I'll add it.

asoldano:
Well, looking at the index at http://docs.jboss.org/resteasy/docs/3.5.1.Final/userguide/html/index.html , I'd say that either a section on each custom annotation is added after the relevant vanilla annotation section, or a new section is created for all the new annotations.
The sources are at https://github.com/resteasy/Resteasy/tree/master/docbook/reference/en/en-US/modules

FroMage:
OK thanks.

FroMage:
Done: #1519

PR #1468, opened 2018-03-25
[RESTEASY-1846] dont append /* to application paths that already end in /*

radai-rosenblatt:
on the one hand, the spec says:
```
if the Application subclass is annotated with @ApplicationPath , implementations are REQUIRED to use the value of this annotation appended with ”/*” to define a mapping for the added server.
```
on the other the code already tries to handle the "/" suffix differently.
as-is right now though, the resulting path would end in `/*/*`, which is definitely wrong.


asoldano:
Thanks @radai-rosenblatt . This said, we require that all contributions be made under the terms of the ASL
Do you agree to these terms?

radai-rosenblatt:
@asoldano i agree to the terms.
anything else i need to file/submit ?

asoldano:
@radai-rosenblatt no, it's fine like this, thanks

PR #1480, opened 2018-04-08
Resteasy 1863: SseEventSourceImpl should not close Client instance

NicoNes:
Issue description: https://issues.jboss.org/browse/RESTEASY-1863

Hi @jimma 

I supposed that you will be interested in this Sse PR.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
@@ -13,15 +13,19 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
+import javax.ws.rs.ProcessingException;
 import javax.ws.rs.ServiceUnavailableException;
 import javax.ws.rs.client.Invocation;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status.Family;
 import javax.ws.rs.sse.InboundSseEvent;
 import javax.ws.rs.sse.SseEventSource;
 
 import org.apache.http.HttpHeaders;
 import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
+import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
 import org.jboss.resteasy.plugins.providers.sse.SseConstants;
 import org.jboss.resteasy.plugins.providers.sse.SseEventInputImpl;
 import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
@@ -41,7 +45,7 @@ public class SseEventSourceImpl implements SseEventSource
    private final ScheduledExecutorService executor;
 
    private enum State {
-      PENDING, OPEN, CLOSED
+      PENDING, CONNECTING, OPEN, CLOSED
    }
 
    private final AtomicReference<State> state = new AtomicReference<>(State.PENDING);
@@ -51,14 +55,16 @@ public class SseEventSourceImpl implements SseEventSource
    private final List<Consumer<Throwable>> onErrorConsumers = new CopyOnWriteArrayList<>();
 
    private final List<Runnable> onCompleteConsumers = new CopyOnWriteArrayList<>();
+   
+   // No need to be volatile since visibility is guarantees by the fact that it will always be set before setting 
+   // AtomicReference state to 'OPEN' and read only if AtomicReference state is set to 'OPEN'.
+   private Response response;
 
    protected static class SourceBuilder extends Builder
    {
       private WebTarget target = null;
 
       private long reconnect = RECONNECT_DEFAULT;
-
-      private String name = null;
 
       private boolean disableKeepAlive = false;
 
@@ -67,15 +73,9 @@ public class SseEventSourceImpl implements SseEventSource
          //NOOP
       }
 
-      public Builder named(String name)
-      {
-         this.name = name;
-         return this;
-      }
-
       public SseEventSource build()
       {
-         return new SseEventSourceImpl(target, name, reconnect, disableKeepAlive, false);
+         return new SseEventSourceImpl(target, reconnect, disableKeepAlive, false);
       }
 
       @Override
@@ -104,10 +104,10 @@ public class SseEventSourceImpl implements SseEventSource
 
    public SseEventSourceImpl(final WebTarget target, final boolean open)
    {
-      this(target, null, RECONNECT_DEFAULT, false, open);
+      this(target, RECONNECT_DEFAULT, false, open);
    }
 
-   private SseEventSourceImpl(final WebTarget target, String name, long reconnectDelay, final boolean disableKeepAlive,
+   private SseEventSourceImpl(final WebTarget target, long reconnectDelay, final boolean disableKeepAlive,
          final boolean open)
    {
       if (target == null)
@@ -118,10 +118,6 @@ public class SseEventSourceImpl implements SseEventSource
       this.reconnectDelay = reconnectDelay;
       this.disableKeepAlive = disableKeepAlive;
 
-      if (name == null)
-      {
-         name = String.format("sse-event-source(%s)", target.getUri());
-      }
       ScheduledExecutorService scheduledExecutor = null;
       if (target instanceof ResteasyWebTarget)
       {
@@ -168,7 +164,7 @@ public class SseEventSourceImpl implements SseEventSource
 
    public void open(String lastEventId)
    {
-      if (!state.compareAndSet(State.PENDING, State.OPEN))
+      if (!state.compareAndSet(State.PENDING, State.CONNECTING))
       {
          throw new IllegalStateException(Messages.MESSAGES.eventSourceIsNotReadyForOpen());
       }
@@ -180,7 +176,8 @@ public class SseEventSourceImpl implements SseEventSource
    @Override
    public boolean isOpen()
    {
-      return state.get() == State.OPEN;
+      State currentState = state.get();
+      return currentState == State.OPEN  || currentState == State.CONNECTING;
    }
 
    @Override
@@ -237,11 +234,21 @@ public class SseEventSourceImpl implements SseEventSource
    @Override
    public boolean close(final long timeout, final TimeUnit unit)
    {
-      if (state.getAndSet(State.CLOSED) != State.CLOSED)
+      State currentState = state.getAndSet(State.CLOSED);
+      if (currentState != State.CLOSED)
       {
-         ResteasyWebTarget resteasyWebTarget = (ResteasyWebTarget) target;
-         //close httpEngine to close connection
-         resteasyWebTarget.getResteasyClient().httpEngine().close();
+         if(currentState == State.OPEN)
+         {
+            try
+            {
+               //close response to close connection
+               //If already close it's fine since multiple invocation of close has no further effect.
+               response.close();
+            }
+            catch (ProcessingException e)
+            {
+            }
+         }
          executor.shutdownNow();
 
          onCompleteConsumers.forEach(Runnable::run);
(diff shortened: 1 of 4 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -406,30 +459,67 @@ private void onEvent(final InboundSseEvent event)
...
+         // for really long...
+         // One solution provided by Apache HTTPClient to do so is
+         // invoking CloseableHttpResponse.close() instead. But it will
+         // not try to do best effort like JAVA HttpUrlConnection, it
+         // will discard the connection even if the whole response
+         // content can be read.
+         // http://hc.apache.org/httpcomponents-client-4.4.x/tutorial/html/fundamentals.html#d5e145
+         // https://docs.oracle.com/javase/8/docs/technotes/guides/net/http-keepalive.html
+         request.header(HttpHeaders.CONNECTION, "close");
          return request;
```
jimma: @NicoNes  I still don't understand why it has to disable keep alive. How can we enable it ?

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -51,31 +53,25 @@
    private final List<Consumer<Throwable>> onErrorConsumers = new CopyOnWriteArrayList<>();
 
    private final List<Runnable> onCompleteConsumers = new CopyOnWriteArrayList<>();
+   
+   // No need to be volatile since visibility is guarantees by the fact that it will always be set before setting 
+   // AtomicReference state to 'OPEN' and read only if AtomicReference state is set to 'OPEN'.
```
jimma: Is this for AtomicReference<State> ?

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
NicoNes: Hi @jimma 

To be simple, without disabling keep alive, calling `response.close()` in `close()` method will block until the connection is closed on remote side.

That's what I'm testing in `SseEventSourceTest.testCloseSseEventSource()`.
If you take a look at the remote side, the connection is never closed in `SseSmokeResource.sentEventAndStayOpen()`. On the client side if keep alive was enabled `sseEventSource.close()` invoked by the try with resource will block indefinitely waiting for the remote side to close the connection.

The reason of such behavior can be found in `org.apache.http.impl.io.ChunkedInputStream` documentation:

>  Note that this class NEVER closes the underlying stream, even when close
 gets called.  Instead, it will read until the "end" of its chunking on
 close, which allows for the seamless execution of subsequent HTTP 1.1
 requests, while not requiring the client to remember to read the entire
 contents of the response.

And looking at the `org.apache.http.impl.io.ChunkedInputStream#close()` method we can see the blocking loop:

```
    @Override
    public void close() throws IOException {
        if (!closed) {
            try {
                if (!eof && state != CHUNK_INVALID) {
                    // read and discard the remainder of the message
                    final byte buff[] = new byte[BUFFER_SIZE];
        => blocking loop           while (read(buff) >= 0) {
                    }
                }
            } finally {
                eof = true;
                closed = true;
            }
        }
    }
```

So disabling keep alive allows to close response rigth away without blocking. Since the connection is not a persistent one and will not be reused, Apache HttpClient will not try to read the whole response content and will not block (that what I was trying to explain in the big comment sorry if it was not clear).

NicoNes:
This comment is just to say that I voluntarily  did not make `response` field volatile.
Its visibility in `close(...)` method will be ensured by the happen before relationship established by the `AtomicReference state` field.

In other words, if state == Open is true in close method, then thanks to happen before transitivity, reading of `response` field will read the most up to date version of that field.

This is ensured by the fact that we always write to `response` field before setting `state` field
to OPEN:

```
      @Override
      public void run()
      {
           ...
            // response must be set before switching to OPEN state to ensure that it will be visible
            // in close method.
            response = localResponse;
            // Let's switch state from CONNECTING to OPEN if possible and process event stream.
            if (state.compareAndSet(State.CONNECTING, State.OPEN))
           ...
        }
```

So in close method If I can see that `AtomicReference state` is OPEN I will also see any writes that happen prior to setting `AtomicReference state` to OPEN.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
NicoNes: This comment is just to say that I voluntarily  did not make `response` field volatile.
Its visibility in `close(...)` method will be ensured by the happen before relationship established by the `AtomicReference state` field.

In other words, if state == Open is true in close method, then thanks to happen before transitivity, reading of `response` field will read the most up to date version of that field.

This is ensured by the fact that we always write to `response` field before setting `state` field
to OPEN:

```
      @Override
      public void run()
      {
           ...
            // response must be set before switching to OPEN state to ensure that it will be visible
            // in close method.
            response = localResponse;
            // Let's switch state from CONNECTING to OPEN if possible and process event stream.
            if (state.compareAndSet(State.CONNECTING, State.OPEN))
           ...
        }
```

So in close method If I can see that `AtomicReference state` is OPEN I will also see any writes that happen prior to setting `AtomicReference state` to OPEN.

NicoNes:
Hi @kanovotn 

Your understanding is correct and I will also add:

- That this PR allows user to reuse the Client instance he provided to the SseEventSource (through WebTarget instance) **even after SseEventSource.close() has been invoked**.

Your concern about poor SSE connection performance VS regular rest request is also right.
Since the SSE connection is not persistent it will not be pooled nor reused so each invocation of `SseEventSource.open()` will create a new TCP connection.
So yes for sure there will be a performance gap. But contrary to standard rest request do we care that much if each SSE connection establishment is a bit slower ?
I explain myself:
For both SSE request and standard rest request we want the TCP connection creation time to be as short as possible so that it is negligible compared to the overall request time.
For standard rest request since keep-alive and pooling are enabled following ratio will tend toward 0: `TCP connection creation time/ overall request time`
For SSE request since SSE is a long lasting communication, TCP connection creation time will also be negligible compared to the overall request time and the same ratio will also tend toward 0.
WDYT ?

In addition to that, your comment spotted an issue with what I did.

Suppose that the user creates a single Client instance bounded to 2 pooled connection max for his whole application.
Then the user perform requests that fill the pool with 2 keep alived connections.
Then he creates a `SseEventSouce` and `open()` it. One of the two connections from the pool will be taken and used (no TCP connection creation in this case).
Then the user invoke `SseEventSouce.close()`  the connection will be closed, discarded and not added back to the pool.
Then if the user repeat the previous two last steps onemore time, the pool will be empty and the next standard rest request will pay the cost of creating the TCP connection since no connection is available in the pool.

So in order to avoid SSE operation to affect standard rest request performance I should create a dedicated RestEasy Client singleton instead of using the one provided by the user.
WDYT ?

jimma:
@NicoNes  I think this isn't a good idea that EventSource forces to disable keep-alive, especially when the connection is closed by server side.  Did you try close the http client's CloseableResponse ? Can it help ? 
Do you know is this some bug in apache http client and any reason that user can't force to release the connection ?

kanovotn:
>  For SSE request since SSE is a long lasting communication, TCP connection creation time will also be negligible compared to the overall request time and the same ratio will also tend toward 0.
WDYT ?

I think you are right, that SSE connections make more sense to be long lasting. But still disabling keep-alive seems to me something we rather shouldn't do.
Do you know how the other implementations - Jersey or CXF handle reusable client do they need to disable keep-alive?

> So in order to avoid SSE operation to affect standard rest request performance I should create a dedicated RestEasy Client singleton instead of using the one provided by the user.

For the special singleton client, well from all examples I was looking so far, user could use it's own instance of the client. What about client instances with some custom configuration, or custom HttpClient.
(http://docs.jboss.org/resteasy/docs/3.5.0.Final/userguide/html_single/index.html#transport_layer)

NicoNes:
Hi @jimma ,

Unfortunately, closing the http client's CloseableResponse will have the same effect: it will discard the connection and it will not be back into the pool. (http://hc.apache.org/httpcomponents-client-4.4.x/tutorial/html/fundamentals.html#d5e145)
As far as I know this is not a bug in Apache HTTP client and according to the doc it seems that this behavior is normal for them.
As I was saying in my code comment, JAVA HttpUrlConnection allows us to invoke inputStream.close() in a non blocking fashion. It will try to read the whole content in hurry mode without blocking. If it succeeds the connection will be reused else it will be discarded. (https://docs.oracle.com/javase/8/docs/technotes/guides/net/http-keepalive.html). So maybe one good option  will be to use `org.jboss.resteasy.client.jaxrs.engines.URLConnectionEngine` to take advantage of HttpUrlConnection behavior. WDYT ?

NicoNes:
Hi @kanovotn,

I don't know how other impementions do for SSE, if they use persistent connection or not. 
However by defaut Jersey does not use Apache Http client, it comes with an extra connector module and it seems that they are facing problems too.
They fixed (https://github.com/jersey/jersey/issues/3124) the blocking `response.close()` by closing the Apache CloseableResponse we were talking about with @jimma. But with this fix, connections are now discarded and not added back to the pool => https://github.com/jersey/jersey/issues/3358.
This issue seems to be open for a while now and for now I don't think there is any solution to fix it.

So as explained in the first comment of https://github.com/jersey/jersey/issues/3358:

> Currently it's better to just use Sun's sun.net.www.protocol.http.HttpURLConnection in the JDK which actually keeps the connections alive and reuses sockets.

Maybe one good solution will be to use `org.jboss.resteasy.client.jaxrs.engines.URLConnectionEngine` in our SSE impl instead of the apache engine in order to take advantage of JDK behavior to be able to perform non blocking close() and reuse connections.

In `org.jboss.resteasy.client.jaxrs.internal.ClientInvocation#invoke()`  following line: 

` if (response == null) response = client.httpEngine().invoke(this);`

Could be changed by something like:

```
if(this.headers.getAcceptableMediaTypes().contains(javax.ws.rs.core.MediaType.SERVER_SENT_EVENTS_TYPE))
{
 clientHttpEngine= new  URLConnectionEngine();
}
else
{
 clientHttpEngine=client.httpEngine();
}
if (response == null) response = clientHttpEngine.invoke(this);
```
WDYT ?

NicoNes:
Forget my last proposal about changing things in `org.jboss.resteasy.client.jaxrs.internal.ClientInvocation#invoke()`. This is definitively not the good place to make this change.
To my opinion the change should be made in `ApacheHttpClient4Engine` since it is a specific treatment for Apache Http Client. And the `URLConnectionEngine` will have to share the same connection settings as the `ApacheHttpClient4Engine` itself.
Let me know if it seems reasonnable to you and will try to make you a proposal if OK.

jimma:
@NicoNes @kanovotn  Just let you know, I am trying to talk with http client team to understand this issue more. So for I am told this isn't an issue:  if there is no terminating chunk and this connection is released and reused,  it's possible the dirty message from previous exchange will be get in the following http message exchange.

jimma:
@NicoNes  Nice finding.  The "hurried" state is what apache http client ChunkedInputStream doesn't have .It looks this finally goes into close the underlying connection too : http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/sun/net/www/http/HttpClient.java#HttpClient.closeServer%28%29 , right ? 
From the answer from apache http client team, the not expensive way to handle this is close the underlying http response.

NicoNes:
Hey @jimma,

Yes, if the response stream can't be fully consumed in hurry mode without blocking the underlying connection is closed and not reused.

> From the answer from apache http client team, the not expensive way to handle this is close the underlying http response.

They mean closing the  http client's CloseableResponse, right ?

jimma:
> They mean closing the http client's CloseableResponse, right ?

@NicoNes, yes.  This will discard the connection immediately. 
I  wrote some thing to initially implement/test this: https://github.com/resteasy/Resteasy/pull/1492
Can you please review ?

NicoNes:
@jimma  yes sure.

NicoNes:
So @jimma I suppose that I close this PR. right ?

Thanks

jimma:
@NicoNes  yes. please.

PR #1482, opened 2018-04-10
RESTEASY-1842

liweinan:
https://issues.jboss.org/browse/RESTEASY-1842

Create document to show how to inject JSON format data into @MultipartForm parameter correctly

### DIFF
--- a/docbook/reference/en/en-US/modules/Multipart.xml
+++ b/docbook/reference/en/en-US/modules/Multipart.xml
@@ -443,7 +443,7 @@ public interface CustomerPortal {
     @Path("issues/{id}")
     @Consumes("multipart/form-data")
     @PUT
-    public void putProblem(@MultipartForm CustomerProblemForm, 
+    public void putProblem(@MultipartForm CustomerProblemForm,
                            @PathParam("id") int id);
 }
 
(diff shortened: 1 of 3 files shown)


docbook/reference/en/en-US/modules/Multipart.xml
```diff
@@ -472,6 +472,120 @@ public class CustomerPortalServer {
...
 }
 </programlisting>
+       <para>In addition to the XML data format, you can also use JSON formatted data to represent your POJO classes.
+           To achieve this goal, you need to plug in a JSON provider into your project. For example,
+           you can add RESTEasy Jackson2 Provider into your project's dependency scope:
+       </para>
+<programlisting><![CDATA[
+<dependency>
+    <groupId>org.jboss.resteasy</groupId>
+    <artifactId>resteasy-jackson-provider</artifactId>
```
asoldano: This is the Jackson, not Jackson2 provider that is mentioned above. So I guess you should have resteasy-jackson2-provider instead.

liweinan: I have fixed this typo. Thanks for checking!

PR #1483, opened 2018-04-10
[RESTEASY-1844] Move to jettison 1.4.0 and jboss-marshalling-osgi 2.0…

asoldano:
….5.Final (due to JBMAR-216)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
         <surefire.system.args>${modular.jdk.args} ${modular.jdk.props}</surefire.system.args>
         <!-- Plugins versions -->
         <version.org.jacoco.plugin>0.7.9</version.org.jacoco.plugin>
-        <version.surefire.plugin>2.18.1</version.surefire.plugin>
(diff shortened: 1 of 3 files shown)


asoldano:
@kanovotn , @rsvoboda , maybe you want to have a look here. Notice I had to add an explicit dependency version for org.jboss.marshalling:jboss-marshalling-osgi in resteasy-cache/resteasy-cache-core/pom.xml. I was expecting the change in the resteasy-dependency-bom to be enough for controlling that transitive dependency of infinispan, but it seems that both infinispan and jboss-marshalling-osgi versions are not kept from the dependencyManagement block in the resteasy-dependencies-bom, at least here. Would you like to double check / have comments on this?
In any case, with the changes here, I can build / test resteasy-master with JDK 10.

kanovotn:
@asoldano When I use the jboss-marshalling-osgi version only in the resteasy-dependency-bom, the 
`mvn clean install dependency:tree`
shows:
`[INFO] +- org.jboss.marshalling:jboss-marshalling-osgi:jar:2.0.5.Final:provided`

PR #1487, opened 2018-04-12
[RESTEASY-1822]:Accept-patch header is missing in the response for th…

jimma:
…e OPTIONS request

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
@@ -19,10 +19,12 @@ import javax.ws.rs.NotSupportedException;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.ResponseBuilder;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
```diff
@@ -414,11 +422,26 @@ public Match match(List<Match> matches, String httpMethod, HttpRequest request)
...
+               {  
+                  String acceptPatch = "";
+                  first = true;
+                  for (MediaType mediaType : patchAccepts)
+                  {
+                     if (first)
+                        first = false;
+                     else
+                        acceptPatch += ", ";
+                     acceptPatch += mediaType.toString();
```
asoldano: Please turn 'acceptPatch' into a StringBuilder and append on it instead of doing bunch of string concatenations (for performance reasons).

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
```diff
@@ -367,7 +369,7 @@ public Match match(List<Match> matches, String httpMethod, HttpRequest request)
       for (MediaType accept : requestAccepts) weightedAccepts.add(WeightedMediaType.parse(accept));
 
       List<Match> list = new ArrayList<Match>();
-
+      Set<MediaType> patchAccepts = new HashSet<MediaType>();
```
asoldano: The HashSet here is backed by a HashMap of size 16. Maybe we can leave 'patchAccepts' as null here and assign it to an actual HashSet instance only when we really need it later (at that time we also know the initial size for the map btw) ?

jimma: OK. I changed it to create patchAccepts hash set with initialize size 8 only if the http method is "option" and "patch" is listed in the allowed method list.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
jimma: Fixed both for Accept-patch and Allow headers.

PR #1492, opened 2018-04-20
[Resteasy-1863]:SseEventSourceImpl should not close http client engine

jimma:


NicoNes:
@jimma actually closing the `CloseableHttpResponse` will have the same effect as setting the `connection:close` header right ?
- The connection will be closed without blocking but not added back to the pool.
- Same situation if the server-side close the connection.
- the user connection pool will lost connections affecting the performance of other standard rest request.

What about using `URLConnectionEngine` in `ApacheHttpClient4Engine` just to invoke SSE request ? We will take adavanatge of JDK mechanism and not affect the user connection pool if a connection is being discarded because the stream cannot be read in hurry mode.

jimma:
> @jimma actually closing the CloseableHttpResponse will have the same effect as setting the connection:close header right ?

If you mean force to close the eventsouce , it is the same effect.  But It's a bit different when the event sink is closed in server and terminating chunk is received before close the eventsouce , the connection can be reused, it's keep alive.

> - The connection will be closed without blocking but not added back to the pool.

Yes. That's the not expensive to do this. 

> - Same situation if the server-side close the connection.

Do you mean the even sink is closed before event source is closed ? This really depends the server side implementation. If http header is sent with "keep-alive" , server won't close the connection. 

>- the user connection pool will lost connections affecting the performance of other standard rest request.

For the chunked inputstream it has to discard the connection. And it seems the efficient way to do this. 

> What about using URLConnectionEngine in ApacheHttpClient4Engine just to invoke SSE request ?
 We will take adavanatge of JDK mechanism and not affect the user connection pool if a connection is being discarded because the stream cannot be read in hurry mode.

I only implement this for ApacheHttpClientEngine now and will have a look at URLConnectionEngine. What do you mean by not affect the user connection pool ? For jdk url connection,  connection is discarded(server socket close) and won't return back to pool.

NicoNes:
>  But It's a bit different when the event sink is closed in server and terminating chunk is received before close the eventsouce , the connection can be reused, it's keep alive.

Actually  I dont think that closing the CloseableHttpResponse will allow to reuse the TCP connection event if the terminating chunked has been received. According to the doc and the code it seems that the connection is always discarded and not added back to the pool to be reused.
That's the issue encountered by the jersey guys here https://github.com/jersey/jersey/issues/3358

So If the connection is always discarded not added back to the pool, later a new TCP connection will have to be established and this is the expensive operation.

NicoNes:
A simple code snipet as an example off what I was saying above:

```
	PoolingHttpClientConnectionManager poolingHttpClientConnectionManager = new PoolingHttpClientConnectionManager();
	poolingHttpClientConnectionManager.setMaxTotal(50);
	poolingHttpClientConnectionManager.setDefaultMaxPerRoute(50);
	RequestConfig.Builder requestConfigBuilder = RequestConfig.custom();
	requestConfigBuilder.setSocketTimeout(0);
	requestConfigBuilder.setConnectTimeout(0);
	requestConfigBuilder.setConnectionRequestTimeout(0);
	CloseableHttpClient closeableHttpClient = HttpClientBuilder.create()
			.setConnectionManager(poolingHttpClientConnectionManager)
			.setDefaultRequestConfig(requestConfigBuilder.build()).build();

	HttpGet httpGet = new HttpGet("http://localhost:9998/books/1");
	httpGet.addHeader("Connection", "keep-alive");
	httpGet.addHeader("Authorization", "Basic QURNSU46MTE=");
	httpGet.addHeader(HttpHeaders.ACCEPT, ContentType.APPLICATION_JSON.getMimeType());

	// No connection in the pool so a new TCP connection will be created and
	// it will be expensive (see 'execution time').
	// After closing the stream the connection will remain in the pool to
	// be reused (see 'pool stats').
	long start = System.currentTimeMillis();
	try {
		closeableHttpClient.execute(httpGet).getEntity().getContent().close();
		System.out.println("pool stats: " + poolingHttpClientConnectionManager.getTotalStats());
		System.out.println("execution time: " + (System.currentTimeMillis() - start));
	} catch (IOException e) {
		e.printStackTrace();
	}

	// One connection in the pool so it will be reused and since there is no
	// need to create any new TCP connection it will be fast (see 'execution
	// time').
	// After closing the stream the connection will remain in the pool to
	// be reused (see 'pool stats').
	start = System.currentTimeMillis();
	try {
		closeableHttpClient.execute(httpGet).getEntity().getContent().close();
		System.out.println("pool stats: " + poolingHttpClientConnectionManager.getTotalStats());
		System.out.println("execution time: " + (System.currentTimeMillis() - start));
	} catch (IOException e) {
		e.printStackTrace();
	}

	// One connection in the pool so it will be reused and since there is no
	// need to create any new TCP connection it will be fast (see 'execution
	// time').
	// After closing the stream the connection will not be added back to the
	// pool to
	// be reused (see 'pool stats').
	start = System.currentTimeMillis();
	try {
		closeableHttpClient.execute(httpGet).close();
		System.out.println("pool stats: " + poolingHttpClientConnectionManager.getTotalStats());
		System.out.println("execution time: " + (System.currentTimeMillis() - start));
	} catch (IOException e) {
		e.printStackTrace();
	}

	// No connection in the pool so a new TCP connection will be created and
	// it will be expensive (see 'execution time').
	// After closing the stream the connection will remain in the pool to
	// be reused (see 'pool stats').
	start = System.currentTimeMillis();
	try {
		closeableHttpClient.execute(httpGet).getEntity().getContent().close();
		System.out.println("pool stats: " + poolingHttpClientConnectionManager.getTotalStats());
		System.out.println("execution time: " + (System.currentTimeMillis() - start));
	} catch (IOException e) {
		e.printStackTrace();
	}
}
```

NicoNes:
Hi @jimma 

You're right it works  as you described it :+1: 
What about merging the content of new `closeHttpResponse()` method into existing `releaseConnection`
instead of having both `close()` and  `closeHttpResponse()`?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
```diff
@@ -494,6 +494,24 @@ public void testNoReconnectAfterEventSinkClose() throws Exception
...
+      try (SseEventSource source = sourceImpl2)
+      {
+         source.register(event -> results.add(event.readData()));
+         source.open();
+         Thread.sleep(1000);
+      }
+      catch (InterruptedException e)
+      {
+         // falls through
+         e.printStackTrace();
```
rsvoboda: Use logger pls

and maybe also do the switch of e.printStackTrace(); on line 495

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -54,6 +58,7 @@
    
    private boolean alwaysReconnect;
 
+   private ClientResponse response;
```
NicoNes: Should be volatile

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -304,10 +306,11 @@ public void run()
          long delay = reconnectDelay;
          try
          {
-            final Invocation.Builder request = buildRequest();
+            final  ClientInvocation invocation = (ClientInvocation) buildRequest().buildGet();
             if (state.get() == State.OPEN)
             {
-               eventInput = request.get(SseEventInputImpl.class);
+               response = (ClientResponse)invocation.invoke();
+               eventInput = ClientInvocation.extractResult(new GenericType<SseEventInputImpl>(SseEventInputImpl.class), response, null);
```
NicoNes: If current SseEventSource has been closed after getting the response but before reading the SseEventInputImpl, invoking ClientInvocation.extractResult(new GenericType<SseEventInputImpl>(SseEventInputImpl.class), response, null) will throw an ResponseProcessingException with IllegalStateException as root cause. This exception is not an error but a normal consequence of the SseEventSource being closed by user so let's just ignore it instead of sending to error consumers. WDYT ?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseTest.java
jimma: @rsvoboda  I'll fix this.  Thanks.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
jimma: I am not sure if this will happen . If there is exception, I think it doesn't matter to trigger the error consumers.

jimma:
> What about merging the content of new closeHttpResponse() method into existing releaseConnection
instead of having both close() and closeHttpResponse()?

We have https://issues.jboss.org/browse/RESTEASY-889 to take care.  That's why I added another closeHttpResponse method.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
NicoNes: OK

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
jimma: Can you please explain why we need volatile ?

jimma: I don't think this is necessary to be volatile and response is a private field of EventSource.  You have to get the EventSource object to close it.  Moreover, the response in close() won't be null. If it is not connected (not get a response) sucessfully, it will always throw an exception.

NicoNes: > Moreover, the response in close() won't be null

Imagine following cases:
**Case1:**
Thread1: user instantiate and `open()` a SseEventSourceImpl
Thread2: this thread starts execution and pause for thread scheduling purpose before executing `response = (ClientResponse)invocation.invoke()` 
( https://github.com/jimma/Resteasy/blob/4c839cdc670b09d461b76be33abd099b1632086d/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L312)
Thread1: user invoke `close()` and a NPE is thrown because `response` has no already be set by Thread2.

**Case2:**
Thread1: user instantiate and `open()` a SseEventSourceImpl
Thread2: this thread start its execution and invokes `response = (ClientResponse)invocation.invoke()` 
( https://github.com/jimma/Resteasy/blob/4c839cdc670b09d461b76be33abd099b1632086d/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L312).
For any reason the connection with the server takes time so that `(ClientResponse)invocation.invoke()` does not return immediately. This thread pause for thread scheduling purpose.
Thread1: user invoke `close()` and a NPE is thrown because `response` has no already be set by Thread2.

Actually many other cases can be found you can even simulate them with debugger.
Since `response` is set and read by two different threads, due to the way thread execution can be scheduled by the system `response` can be null in the reading thread.
And for the same reason even if writing thread set `response` before reading thread tries to read it, there is no guarantee that this thread will read the most up to date `response` value: it can read a null value or an already closed `reponse` in case of reconnect for example. And in this case it will lead to current response never closed.

jimma: For Case 1, can you show me some code or test case ?  As I said,   response is private and you can't create(write) in thread 1 and close in another thread. 
For Case 2,  volatile can't help fix the NPE if the close EventSource before it is connected successully, right ?  We need to add NPE guard to fix this.

NicoNes: >For Case 2, volatile can't help fix the NPE if the close EventSource before it is connected successully, right ? We need to add NPE guard to fix this

Right volatile won't help fixing that, null check will.

>For Case 1, can you show me some code or test case ? As I said, response is private and you can't create(write) in thread 1 and close in another thread.

To simulate the case you can add ` TimeUnit.MINUTES.sleep(2);` before `response = (ClientResponse)invocation.invoke()` in `run` method
( https://github.com/jimma/Resteasy/blob/4c839cdc670b09d461b76be33abd099b1632086d/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L312) to simulate the  thread pause.
And launch the following test:

```
   @Test
   @InSequence(14)
   public void test() throws Exception
   {
      Client client = ClientBuilder.newBuilder().build();
      final AtomicInteger errors = new AtomicInteger(0);
      WebTarget target = client.target(generateURL("/service/server-sent-events"));
      SseEventSourceImpl sourceImpl = (SseEventSourceImpl)SseEventSource.target(target).build();
      sourceImpl.open();
      sourceImpl.close();
   }
```
Thread 1 (test thread): open the  SseEventSource that starts thread 2 in the same time. Thread 1 then waits 30 seconds for the connected latch to be counted down.
Thread 2: pauses for 2 minutes before executing `response=(ClientResponse)invocation.invoke()`
Thread 1: the connected latch timeouts then 'sourceImpl.close();' is invokes then the NPE is thrown.

For both case 1 and case 2 for sure volatile won't fix the issue. The correct fix is to add a null check.
What I was saying about volatile is that: without it we can't say for sure that the closing thread will always read the most up to date value of `response` field.  It is needeed to guarantee that exact `response` set by thread 2 (`run()` method) will always be readable by thread 1 (`close()` method)

NicoNes: Hi @asoldano,

To my mind both NPE protection and volatile are required here.

NicoNes:
Hi @jimma, @asoldano ,

This PR is OK for me.

Thanks you

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -402,5 +402,10 @@ public void abortIfClosed()
    {
        if (bufferedEntity == null) super.abortIfClosed();
    }
+   
+   public void closeHttpResponse()
+   {
+
```
asoldano: This should probably be an abstract method (also considering the ClientResponse class here is abstract). A bit of javadoc would have to be added, so that implementors know what do in their classes extending this.

NicoNes: Personnaly I'm not really comfortable with having both `close()` , `closeHttpResponse()` and `releaseConnection()` methods. 
IMO from the implementators point of view `closeHttpResponse()` and  `releaseConnection()` are pretty similar. So why not just having a `releaseConnection(boolean consumeInputStream)` method ? The concept of `boolean consumeInputStream` is common to both Apache HTTP Client and JDK HttpUrlConnection so it will not be that disturbing for implementators.
We will have to keep existing `releaseConnection()` for compatibility reason but it will just invoke `releaseConnection(true)`. 
WDYT ?

NicoNes: Hi @asoldano , @jimma 

To move forward, I created a PR with the proposal I was talking about above: https://github.com/resteasy/Resteasy/pull/1534

asoldano: @NicoNes sorry for the late reply, do you mean you're proposing this PR to be closed and replaced with #1534 ?

NicoNes: Hi @asoldano,

Right, you can close this one. I've merged @jimma changes from this PR to the new one plus the changes I was talking about in the comment above.

PR #1495, opened 2018-04-23
[RESTEASY-1862] JettyClientEngine

asoldano:
Rebased version of https://github.com/resteasy/Resteasy/pull/1380/ (plus doc fixes)

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.client.jaxrs.engines;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.ByteBuffer;
@@ -24,7 +25,7 @@ import org.eclipse.jetty.client.HttpClient;
 import org.eclipse.jetty.client.api.Request;
 import org.eclipse.jetty.client.api.Response;
 import org.eclipse.jetty.client.api.Result;
-import org.eclipse.jetty.client.util.DeferredContentProvider;
+import org.eclipse.jetty.client.util.BytesContentProvider;
 import org.eclipse.jetty.http.HttpFields;
 import org.eclipse.jetty.io.ByteBufferPool;
 import org.jboss.resteasy.client.jaxrs.AsyncClientHttpEngine;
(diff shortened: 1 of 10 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,289 @@
...
+     * decoding infrastructure without buffering or spinning up auxiliary threads (arguably more expensive than buffering).
+     *
+     * @see AsyncClientHttpEngine#submit(ClientInvocation, boolean, InvocationCallback, org.jboss.resteasy.client.jaxrs.AsyncClientHttpEngine.ResultExtractor)
+     */
+    @Override
+    public <T> Future<T> submit(ClientInvocation invocation, boolean bufIn, InvocationCallback<T> callback, ResultExtractor<T> extractor) {
+        final boolean buffered;
+        if (!bufIn && extractor != null) {
+            if (!WARN_BUF.getAndSet(true)) {
+                System.err.println("TODO: ResultExtractor is synchronous and may not be used without buffering - forcing buffer mode.");
```
marekkopecky: @asoldano 
Logging is not used here ... but I know that some community users may want to use their custom logging instead of JBoss logging ...

Btw, what about plans of this todo?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
```diff
@@ -0,0 +1,289 @@
...
+                    } catch (Exception e) {
+                        onFailure(response, e);
+                    }
+                }
+            }
+
+            @SuppressWarnings("unchecked")
+            private void complete() {
+                completing.set(true);
+                // TODO: dangerous cast, see javadoc!
```
marekkopecky: What about plans for this todo?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
stevenschlansker: Yes, this could be turned into a proper log, or even removed.

The reason for this is that currently there is no asynchronous interface to `ResultExtractor`.  This means it is not possible to extract the content without either buffering, or spinning an auxiliary thread.

The solution here is to have an asynchronous interface to ResultExtractor.

Another solution is to use a coroutine technology like Quasar or Project Loom to allow suspending of the synchronous extractor.  (We have demonstrated this with Jackson, but I don't think the approach is Production Ready yet.)

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/JettyClientEngine.java
stevenschlansker: This might not actually be a TODO, you can only get an invalid cast here if the caller "uses" the class wrong.  But it might be nicer to throw an explicit error with a better message about what went wrong.

PR #1502, opened 2018-04-26
[RESTEASY-1862] Move JettyClient engine to separate module

asoldano:


rsvoboda:
Otherwise +1 for this change as it gives some flexibility for further changes without need to consider backward compatibility impact on WF/EAP

asoldano:
@rsvoboda thanks for the feedback, documentation updated.

PR #1504, opened 2018-04-27
[RESTEASY-1905] Async injection

FroMage:
Here's a preview of the work I did for pluggable async injection, FYI. Note that I still have to test this with a real-world integration (hopefully next week), so don't merge this as-is. I also need to iron out a few things and write docs.

To give a quick example of what this allows is that it's now possible to declare an injector for `Single<Foo>` and inject it via `@Context Foo` and the response will be automatically made asynchronous and the resource method will only be invoked once the `Single<Foo>` is resolved (non-blocking) to `Foo` and we can proceed. That's very huge for async users who want injection of async types.

- New `@Provider` type `ContextInjector` which allows users to declare pluggable injectors that can be injected with `@Context`.
- These injectors can be declared to provide injection of async types, such as `CompletionStage<X>`.
- If trying to inject `@Context X` and we have an async injector for `CompletionStage<X>` we properly wait for its resolution before injecting the resolved `X` (non blocking).
- Async injection supports pluggable async types using existing (unmodified) `AsyncResponseProvider` system to wrap pluggable async types into `CompletionStage`. We may have to rename it, though?
- Injection had to be rewritten to use `CompletionStage` everywhere. I initially wrote this using callbacks in order to avoid using `CompletionStage` but the API was not at all cleaner, and the implementation was just terribly fragile and error-prone. Using `CompletionStage` is invasive, but clean and trustworthy. I did verify that performance would not suffer, crudely, by measuring difference of running all Maven tests, and there's zero difference. Obviously that's not a real benchmark, but if you can point me to one, I'll measure it.
- Async injection is only attempted at points where async injection is permitted, such as for resource creation and resource method invocation. It is not enabled at points where the API does not allow for suspending the request (via `ResourceContext.getResource` for example).

### DIFF
--- a/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
+++ b/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
@@ -57,7 +57,23 @@ public class PatchMethodFilter implements ContainerRequestFilter
             throw new ProcessingException("Get method not found and patch method failed");
          }
          ResourceMethodInvoker methodInvoker = (ResourceMethodInvoker) resourceInovker;
+<<<<<<< HEAD
          Object object;
+=======
+         Object object = methodInvoker.invokeDryRun(request, response).toCompletableFuture().getNow(null);
+
+         ByteArrayOutputStream tmpOutputStream = new ByteArrayOutputStream();
+         MessageBodyWriter msgBodyWriter = ResteasyProviderFactory.getInstance().getMessageBodyWriter(
+               object.getClass(), object.getClass(), methodInvoker.getMethodAnnotations(),
+               MediaType.APPLICATION_JSON_TYPE);
+         msgBodyWriter.writeTo(object, object.getClass(), object.getClass(), methodInvoker.getMethodAnnotations(),
+               MediaType.APPLICATION_JSON_TYPE, new MultivaluedTreeMap<String, Object>(), tmpOutputStream);
+         ObjectMapper mapper = new ObjectMapper();
+         JsonNode targetJson = mapper.readValue(tmpOutputStream.toByteArray(), JsonNode.class);
+         JsonPatch patch = JsonPatch.fromJson(mapper.readValue(request.getInputStream(), JsonNode.class));
+
+         JsonNode result = null;
+>>>>>>> Fixed bug with PatchMethodFilter due to my changes
          try
          {
             object = methodInvoker.invokeDryRun(request, response);
--- a/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
+++ b/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
@@ -57,26 +57,10 @@ public class PatchMethodFilter implements ContainerRequestFilter
             throw new ProcessingException("Get method not found and patch method failed");
          }
          ResourceMethodInvoker methodInvoker = (ResourceMethodInvoker) resourceInovker;
-<<<<<<< HEAD
          Object object;
-=======
-         Object object = methodInvoker.invokeDryRun(request, response).toCompletableFuture().getNow(null);
-
-         ByteArrayOutputStream tmpOutputStream = new ByteArrayOutputStream();
-         MessageBodyWriter msgBodyWriter = ResteasyProviderFactory.getInstance().getMessageBodyWriter(
-               object.getClass(), object.getClass(), methodInvoker.getMethodAnnotations(),
-               MediaType.APPLICATION_JSON_TYPE);
-         msgBodyWriter.writeTo(object, object.getClass(), object.getClass(), methodInvoker.getMethodAnnotations(),
-               MediaType.APPLICATION_JSON_TYPE, new MultivaluedTreeMap<String, Object>(), tmpOutputStream);
-         ObjectMapper mapper = new ObjectMapper();
-         JsonNode targetJson = mapper.readValue(tmpOutputStream.toByteArray(), JsonNode.class);
-         JsonPatch patch = JsonPatch.fromJson(mapper.readValue(request.getInputStream(), JsonNode.class));
-
-         JsonNode result = null;
->>>>>>> Fixed bug with PatchMethodFilter due to my changes
          try
          {
-            object = methodInvoker.invokeDryRun(request, response);
+            object = methodInvoker.invokeDryRun(request, response).toCompletableFuture().getNow(null);
             ByteArrayOutputStream tmpOutputStream = new ByteArrayOutputStream();
             MessageBodyWriter msgBodyWriter = ResteasyProviderFactory.getInstance().getMessageBodyWriter(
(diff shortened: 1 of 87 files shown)


rsvoboda:
Hi, I will be on PTO next week and I can't promise I will look into it today.
I asked @marekkopecky and @kanovotn to take a look at this

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ContextInjector.java
```diff
@@ -0,0 +1,9 @@
+package org.jboss.resteasy.spi;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+public interface ContextInjector<WrappedType, UnwrappedType> {
+   // FIXME: remove rawType and genericType?
```
marekkopecky: What is the plan about this?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
```diff
@@ -335,14 +338,15 @@ public BuiltResponse invoke(HttpRequest request, HttpResponse response, Object t
       }
       uriInfo.pushCurrentResource(target);
       BuiltResponse rtn = invokeOnTarget(request, response, target);
-      return rtn;
+      // FIXME: async
```
marekkopecky: ?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -25,31 +30,62 @@
...
+         return CompletableFuture.completedFuture(new SseImpl());
       }
-      return createProxy();
+      return CompletableFuture.completedFuture(createProxy());
+   }
+
+   private CompletionStage<Object> unwrapIfRequired(HttpRequest request, Object contextData, boolean unwrapAsync)
+   {
+      if(unwrapAsync && rawType != CompletionStage.class && contextData instanceof CompletionStage) {
+         // FIXME: do not unwrap if we have no request?
```
marekkopecky: What about this? What is the plan for the future?

marekkopecky:
@asoldano This PR changes API of many inject/injectableArguments/.. public methods. This PR updates old tests in testsuite, because with this update testsuite fails. These changes are backward incompatible, real use-case (where users can have backward compatibility issues) is obvious from testsuite, see for example [this change](https://github.com/resteasy/Resteasy/pull/1504/files#diff-9c7348bdca89882dcf7b5f187b9a2044L27) + HttpRequestParameterInjectorTest + HttpRequestParameterInjectorResource.

Is it possible to create new new interfaces instead of changing old ones? For example for [this change](https://github.com/resteasy/Resteasy/pull/1504/files#diff-b188ce5e33e4bef689813f54f3a3c0efL18):
Would be possible to keep original interface as is and add new interface, for example ValueInjectorCompletionStage? Or do you have some different/better idea?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ContextInjector.java
FroMage: I guess I'd remove the types because I don't see their point, but I based this on `ContextResolver` from JAX-RS which has `Class` (and then added `Type`). @asoldano any idea why they have that `Class` param ? I couldn't guess any use for it, since an injector can only declare injection of one type, since it's duplicated in the class signature via the type parameter.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
FroMage: Doesn't look like something I should fix, I'll remove it.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
FroMage: No, that is outdated, I fixed that question with the `unwrapAsync` param. I'll remove it.

FroMage:
> Is it possible to create new new interfaces instead of changing old ones?

Perhaps I can add defaulted methods? Would that be enough for you?

marekkopecky:
@FroMage Which default methods do you want to add?

If you add new default `Object inject();` method to ValueInjector.java and add non-default `CompletionStage<Object> inject(boolean unwrapAsync);` method to ValueInjector, it means that old implementations of ValueInjector interface don't work, because these old implementations don't implement `CompletionStage<Object> inject(boolean unwrapAsync);` method, right? So this approach is still not backward compatible ...

FroMage:
I guess we could add the new (`CompletionStage`) methods as default methods that call the old ones, and make the old methods defaulted to call the new ones, this way old implementations keep working, and new implementations only have to implement the new ones. But that's pretty dangerous for new implementors because they won't get type-system help as to which ones to implement, and if they don't implement anything, there will be infinite recursion.

marekkopecky:
Well ... this approach would be backward compatible. But as you say, it could be confusing for RESTEasy users ... @asoldano What do you think about it?

FroMage:
Note that I'm totally fine with this only being merged to the 4.0 tree :)

asoldano:
@marekkopecky @FroMage , well, retaining backward compatibility would be a great plus, as it would allow backporting to 3.6 branch. However the proposed approach looks quite hacky to me... and I believe it would probably make future maintenance even more complex than having to deal with two diverged code branches because of the changes here. So I'm leaning more towards including this on master only.

asoldano:
@FroMage can you please rebase against master / fix the conflicts ? Thanks

asoldano:
OK, so I'm merging this to master only, it will be treated as a RESTEasy 4 feature. Should you have a bit of time for writing up some documentation, I would really appreciate, as usual ;-)

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
```diff
@@ -60,7 +60,7 @@ public void filter(ContainerRequestContext requestContext) throws IOException
          Object object;
          try
          {
-            object = methodInvoker.invokeDryRun(request, response);
+            object = methodInvoker.invokeDryRun(request, response).toCompletableFuture().getNow(null);
```
marekkopecky: @FroMage Wouldn't be better to wait some time and use `.get(some_reasonable_timeout)`? Same for all `getNow` usage in this PR. Is it possible, that default value (`null`) is returned, because data is not yet available?

FroMage: So, in fact this particular plugin just does not support async. It's almost never a good idea to wait on a timeout because that may block the very thread that would cause the async value to be emitted (on vertx for example).
I used `getNow` in all places where async wasn't supported.
Now, with the async filter support, it may be possible to make this particular filter async, but in general I'm wary of changing things unless there's user code to require this change.

PR #1507, opened 2018-05-07
[RESTEASY-1805] Tests for resource metadata SPI

marekkopecky:
[RESTEASY-1805] Tests for resource metadata SPI

jira: https://issues.jboss.org/browse/RESTEASY-1805

This PR contains "Category(ExpectedFailing.class)" two times, more details in jira ...

3.6 PR: https://github.com/resteasy/Resteasy/pull/1506


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/spi/ResourceClassProcessorNotAppliedTest.java
```diff
@@ -0,0 +1,87 @@
...
+ * @tpSince RESTEasy 3.6
+ */
+@RunWith(Arquillian.class)
+public class ResourceClassProcessorNotAppliedTest {
+
+    protected static final Logger logger = Logger.getLogger(ResourceClassProcessorNotAppliedTest.class.getName());
+
+    private static List<String> visitedProcessors = new ArrayList<>();
+
+    public static void addToVisitedProcessors(String item) {
```
asoldano: Shouldn't this be synchronized?

marekkopecky: AFAICS no, because processors are applied in sequence, see [java code](https://github.com/resteasy/Resteasy/pull/1449/files#diff-a7cd5f8090c07c60ab3f97314b8091bbR1064). But we may want to be super safe and synchronize this. So I add `synchronized` keyword ...

PR #1509, opened 2018-05-16
Checkstyle profile, checkstyle configuration

kanovotn:
The profile is enabled with '-Dcheckstyle'. The errors found by checkstyle needs to be fixed before enabling this by default.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,40 @@
                 </dependencies>
             </dependencyManagement>
         </profile>
+        <profile>
+            <id>checkstyle</id>
+            <activation>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -191,6 +191,68 @@
                 </dependencies>
             </dependencyManagement>
         </profile>
+        <profile>
+          <id>jdk9</id>
+          <activation>
+            <jdk>9</jdk>
+          </activation>
```
asoldano: This profile looks unrelated to the PR here, right?

PR #1510, opened 2018-05-18
[RESTEASY-1880] Parameter annotations with default names

marekkopecky:
Evolved version of #1467 and #1374 

jira: https://issues.jboss.org/browse/RESTEASY-1880

More javadoc && community documentation would be useful

RESTEasyParamBasicTest#customValuesTest always fails, so this test is in ExpectedFailing category. This test needs to be fixed.


asoldano:
Closing this as #1467 is now merged and includes the latest commit here too.

PR #1512, opened 2018-05-21
[RESTEASY-1725] Registering lambda context resolver results in ArrayIndexOutOfBoundsException

rnetuka:
JIRA: https://issues.jboss.org/browse/RESTEASY-1725

As stated by R Searls, Java doesn't support inspecting generic types inside lambdas. Reworking the whole ResteasyProviderFactory system for this one particular case would be very difficult. The IndexOutOfBoundsException is however highly confusing. It would be definitely better to throw some more descriptive excpetion instead.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
@@ -1411,6 +1411,10 @@ public class ResteasyProviderFactory extends RuntimeDelegate implements Provider
 
    protected void addContextResolver(ContextResolver provider, Class providerClass, boolean builtin)
    {
+      // RESTEASY-1725
+      if (providerClass.getName().contains("$$Lambda$")) {
+         throw new RuntimeException(Messages.MESSAGES.registeringContextResolverAsLambda());
+      }
       Type parameter = Types.getActualTypeArgumentsOfAnInterface(providerClass, ContextResolver.class)[0];
       addContextResolver(provider, parameter, providerClass, builtin);
    }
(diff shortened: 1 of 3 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -1272,6 +1272,10 @@ protected void addContextResolver(ContextResolver provider, boolean builtin)
 
    protected void addContextResolver(ContextResolver provider, Class providerClass, boolean builtin)
    {
+      // RESTEASY-1725
+      if (providerClass.getName().contains("$$Lambda$")) {
+         throw new ResteasyProviderFactoryException("Registering a context resolver doesn't support lambdas.");
```
asoldano: The message should actually be an i18n one.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -2810,4 +2818,13 @@ protected void addResourceClassProcessor(ResourceClassProcessor processor)
    public ResourceBuilder getResourceBuilder() {
       return resourceBuilder;
    }
+
+   public static class ResteasyProviderFactoryException extends RuntimeException {
+
+      public ResteasyProviderFactoryException(String message) {
```
asoldano: Do we really need this additional class? The ResteasyProviderFactory is already quite big and critical memory-wise, maybe we can still use a RuntimeException, but with the i18n message ?

rnetuka:
@asoldano Done

rnetuka:
Retest this please

PR #1494, opened 2018-04-23
RESTEASY-1418 RESTEasy should be able to provide trace information about request

liweinan:
It's just a preview of the feature. Here are the todo list:

1. We need to put log all the events in RESTEasyServerTracingEvent
2. We need to have ON_DEMAND tracing supported
3. Client side tracing events need to be designed and implemented
4. resteasy-tracing module needs to be removed, because tracing features need to be inside resteasy-jaxrs module.
5. tests...

I'll keep updating this branch.

PR #1516, opened 2018-05-22
RESTEasy Tracing Feature | Part 1

liweinan:
This commits is the first commit for:

- https://issues.jboss.org/browse/RESTEASY-1418 RESTEasy should be able to provide trace information about request #1494

And it contains the implementation of requirements of the following issues:

- https://issues.jboss.org/browse/RESTEASY-1891 Implement the ON_DEMAND mode of tracing feature
- https://issues.jboss.org/browse/RESTEASY-1893 Client side needs to get tracing info
- https://issues.jboss.org/browse/RESTEASY-1899 Move tracing logger from global scope to per-request scope
- https://issues.jboss.org/browse/RESTEASY-1909 Add tests to the tracing feature


liweinan:
@asoldano okay let me rethink about the signatures of these methods, to see if I could avoid to add logger in the method.

liweinan:
@asoldano Here are my analysis to the usages of three modified public methods:

## `public Response executeExactExceptionMapper(Throwable exception, RESTEasyTracingLogger logger)`

The above method is used by this method:

`public Response handleException(HttpRequest request, Throwable e)`

---

## `public Response executeExceptionMapperForClass(Throwable exception, Class clazz, RESTEasyTracingLogger logger)`

The above method is used by following methods:

- `protected Response handleApplicationException(HttpRequest request, ApplicationException e, RESTEasyTracingLogger logger)`
-  `protected Response handleWriterException(HttpRequest request, WriterException e, RESTEasyTracingLogger logger)`
-  `protected Response handleReaderException(HttpRequest request, ReaderException e, RESTEasyTracingLogger logger)`

---

## `public Response executeExceptionMapper(Throwable exception)`

The above method is used by this method:

`public Response handleException(HttpRequest request, Throwable e)`

---

In conclusion:

1. The above three methods are only used by `org.jboss.resteasy.core.ExceptionHandler` internally.
2. `handleException()` method is actually the `entry point` of `org.jboss.resteasy.core.ExceptionHandler`, and the three `public` methods in above are actually called by `protected` methods inside the `org.jboss.resteasy.core.ExceptionHandler` class.

---

The reason we must inject loggers into the above three methods is that actually these three methods are quite `internal` so they deal with some internal data that we need for logging.

Taking `executeExceptionMapperForClass()` for example:

```
public Response executeExceptionMapperForClass(Throwable exception, Class clazz, RESTEasyTracingLogger logger)
   {
      ExceptionMapper mapper = providerFactory.getExceptionMappers().get(clazz);
      if (mapper == null) return null;
      mapperExecuted = true;
      long timestamp = logger.timestamp(RESTEasyServerTracingEvent.EXCEPTION_MAPPING);
      Response resp = mapper.toResponse(exception);
      logger.logDuration(RESTEasyServerTracingEvent.EXCEPTION_MAPPING, timestamp, mapper, exception, exception.getLocalizedMessage(), resp);
      return resp;
   }
```

As we can see above, the method will get the `mapper` properly and generate the `response` objects. And these object are needed by the tracing logger to trace them.

Based on above analyze, I changed the above three methods from ‘public` to `protected`. I’ll leave this open for discussion.

liweinan:
A better solution would be leave three public methods as-is and mark them is deprecated, and the protected methods are actually used internally.

liweinan:
I have added protected methods to inject loggers properly, and mark the public methods as deprecated.

asoldano:
@liweinan also please rebase on top of master, so that the PR can be merged when it's ready.

liweinan:
Because the PR affects many files, so I'd like to save this branch in case I need the commit histories in future, and I created a new branch to rebase and here is the PR: https://github.com/resteasy/Resteasy/pull/1533

This PR should not be merged.

liweinan:
Here is the new commit: https://github.com/resteasy/Resteasy/pull/1533

asoldano:
OK, I'm closing this PR as it's superseded by #1533

PR #1517, opened 2018-05-22
[RESTEASY-1805] Add support for priorities and fix issue with local R…

asoldano:
…esteasyProviderFactory instances


asoldano:
@chkal FYI

PR #1520, opened 2018-05-23
[RESTEASY-1865] SynchronousDispatcher #getResource not returning Resource Instance

rnetuka:
JIRA: https://issues.jboss.org/browse/RESTEASY-1865

dispatcher.getRegistry().addSingletonResource(...) works for ResourceMethodInvoker (returned object is the proper singleton), but not for resourceContext.getResource(...). Here, the object being returned is simply constructed one by ResteasyProviderFactory using a constructor.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
@@ -366,6 +366,10 @@ public class ResourceMethodRegistry implements Registry
             rootNode.addInvoker(fullpath, locator);
          else root.addInvoker(classExpression, fullpath, locator);
       }
+
+      if (rf instanceof SingletonResource) {
+         providerFactory.registerSingletonResource((SingletonResource) rf);
+      }
    }
 
    /**
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
@@ -20,6 +20,7 @@ import org.jboss.resteasy.plugins.delegates.MediaTypeHeaderDelegate;
 import org.jboss.resteasy.plugins.delegates.NewCookieHeaderDelegate;
 import org.jboss.resteasy.plugins.delegates.UriHeaderDelegate;
 import org.jboss.resteasy.plugins.providers.RegisterBuiltin;
+import org.jboss.resteasy.plugins.server.resourcefactory.SingletonResource;
 import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
 import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 import org.jboss.resteasy.specimpl.LinkBuilderImpl;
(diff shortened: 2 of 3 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -2436,6 +2450,14 @@ public T getContext(Class type)
...
       Constructor<?> constructor = PickConstructor.pickSingletonConstructor(clazz);
       Object obj = null;
+
+      if (isSingletonResource(clazz))
+      {
+         SingletonResource factory = singletonResourceFactories.get(clazz);
+         obj = factory.createResource(request, response, this);
+      }
+      else
+      {
```
rnetuka: I've not indented the else branch as it would have caused very confusing diff

rnetuka:
@dansiviter This solution works and was tested for SingletonResource only. To have more generic solution, I advice to create it as a complex enhancement since we don't have any more relevant test cases.provided by the bug reporter.

rnetuka:
@asoldano Done

rnetuka:
Fixed. Please, review again. Particulary this line:
https://github.com/resteasy/Resteasy/pull/1520/files#diff-be425f1e75ab8a4f6b0a674dcc8d5136R2647

(within changed files)
ResteasyProviderFactory
obj = factory.createResource(request, response, this)**.toCompletableFuture().getNow(null);**

asoldano:
@rnetuka thanks, sorry for the late reply, I've been on PTO and then catching up. I don't see problems in the line that you mentioned. Now that I think more about this PR, though, I'm thinking that the hashmap you're adding should be made a thread safe... WDYT?

ronsigal:
Hi @rnetuka,

I don't see anything in the JAX-RS spec (2.0 or 2.1) that suggests that the resource returned by ResourceContext.getResource() should be any particular iinstance. The javadoc says
>Get ****a**** resolved instance of a resource or sub-resource class.

[My bold emphasis.] Could you elaborate?

-Ron

ronsigal:
Hi @rnetuka, there are a couple of outstanding questions here. Any thoughts?

rsearls:
Hi @rnetuka,
Please rebase again.
@ronsigal, Once rebased, this looks good to merge.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ResourceMethodRegistry.java
```diff
@@ -366,6 +366,10 @@ protected void processMethod(ResourceFactory rf, String base, ResourceLocator me
             rootNode.addInvoker(fullpath, locator);
          else root.addInvoker(classExpression, fullpath, locator);
       }
+
+      if (rf instanceof SingletonResource) {
```
ronsigal: Hi @rnetuka,

I'm questioning whether this should be in
```
   protected void register(ResourceFactory rf, String base, ResourceClass resourceClass)
   {
      for (ResourceMethod method : resourceClass.getResourceMethods())
      {
         processMethod(rf, base, method);
      }
      for (ResourceLocator method : resourceClass.getResourceLocators())
      {
         processMethod(rf, base, method);
      }
   }
```
Otherwise, you'll be calling registerSingletonResource() for each method, no?

-Ron

jamezp:

PR #1533, opened 2018-06-01
[RESTEASY-1346 / RESTEASY-1418] RESTEasy Tracing Feature | Part 1

liweinan:
This is the rebased version of https://github.com/resteasy/Resteasy/pull/1516


liweinan:
This is the rebased commit to sum up all the work in https://github.com/resteasy/Resteasy/pull/1516 and https://github.com/resteasy/Resteasy/pull/1494

liweinan:
@asoldano Here is the final version(hopefully) of the first commit for resteasy tracing feature. Could you please help to check it?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/tracing/RESTEasyTracingLoggerImpl.java
```diff
@@ -0,0 +1,145 @@
...
+    public long timestamp(final RESTEasyTracingEvent event) {
+        if (isEnabled(event.level())) {
+            return System.nanoTime();
+        }
+        return -1;
+    }
+
+    @Override
+    public void flush(final MultivaluedMap<String, Object> headers) {
+        final String[] messages = tracingInfo.getMessages();
```
asoldano: Do we need any check here on in the callers of the flush method here to ensure the processing is performed only when needed? Or maybe we need this in any case? I'm just thinking about performances.

liweinan: @asoldano `flush()` method flushes the data in `tracingInfo` and put it into response headers, and `tracingInfo` is injected by `logImpl` method of `RESTEasyTracingLoggerImpl`:

```
tracingInfo.addMessage(message);
```

This is the single place that `tracingInfo` is added, and `tracingInfo` itself is private.

So the checking point is in `logImpl(...)` method:

```java
if (isEnabled(event.level())) {
...
    final RESTEasyTracingMessage message = new RESTEasyTracingMessage(event, duration, messageArgsStr);
    tracingInfo.addMessage(message);
...
```

So to use the tracingLogger, taking the usages in `SynchronousDispatcher` for example:

```java
...
tracingLogger.log(RESTEasyServerTracingEvent.FINISHED, response.getStatus());
...
tracingLogger.flush(response.getOutputHeaders());
...
```

tracingLogger will always call `log(...)` for server side tracing *and* add the info into `tracingInfo` according to the level set and mode set. This is controlled by `log(...)` method internally.

So if the level is to disable the tracing, then `tracingInfo` will be empty.

In conclusion:

- `tracingInfo` is manipulated insided `logImpl(...)` function and it's private data of `RESTEasyTracingLoggerImpl`.
- `logImpl(...)` takes care of the tracing threshold and tracing mode.
- `tracingInfo` is empty, if tracing level is set to OFF or when controlling by ON_DEMAND mode.
- `flush(...)` method will do nothing when `tracingInfo` is empty.

asoldano: OK, makes sense, thanks.

PR #1535, opened 2018-06-02
[RESTEASY-1838]: Remove the SseEventSource.open() 30 seconds timeout to be fully synch

NicoNes:
Please read the discussion on #1456 on the reasoning that led to this PR

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
@@ -21,7 +21,6 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.sse.InboundSseEvent;
 import javax.ws.rs.sse.SseEventSource;
 
-import org.apache.http.HttpHeaders;
 import org.jboss.resteasy.client.jaxrs.ResteasyWebTarget;
 import org.jboss.resteasy.plugins.providers.sse.SseConstants;
 import org.jboss.resteasy.plugins.providers.sse.SseEventInputImpl;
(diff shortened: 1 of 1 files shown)


marekkopecky:
@asoldano AFAICS, there is no way to set disableKeepAlive variable to true, so it is save to remove it, right?

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -208,12 +199,6 @@ public boolean isOpen()
       return state.get() == State.OPEN;
    }
 
-   @Override
```
marekkopecky: If this method is removed, default jax-rs close method is used, right? Default method from spec uses 5s timeout:
```java
    @Override
    default void close() {
        close(5, TimeUnit.SECONDS);
    }
```

marekkopecky:
If resteasy doesn't use any timeout, we need to rely on HTTP timeout, or server timeout, right? Wouldn't be better (for example) to allow to parametrize this timeout? I can imagine that some users would like to have some timeout on client. And (for example) if client doesn't return any event, users can try to send new request again. @asoldano, @jimma  WDYT?

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
NicoNes: Yes, it's right

NicoNes:
Hi @marekkopecky,

> If resteasy doesn't use any timeout, we need to rely on HTTP timeout, or server timeout, right?

Right

>  Wouldn't be better (for example) to allow to parametrize this timeout?

IMO, users can still set either `javax.ws.rs.client.ClientBuilder.connectTimeout(...)` or `javax.ws.rs.client.ClientBuilder.readTimeout(...)` to configure the `javax.ws.rs.client.Client` instance used by the SSE client to set up the timeout the way they want.

PR #1538, opened 2018-06-06
Async filter exception bug

FroMage:
If we have an async response (returning `CompletionStage` for example, and a filter throws an exception while writing the response, the exception used to be silently ignored by `CompletionStage`, which meant that the client would hang waiting for the response, and nothing would be logged.

I fixed this, but it also means I don't throw `UnhandledException` to the container anymore in such cases, because it's very hard (if not impossible) to know at that point if the container is still in our call stack, or if we're in another thread entirely, in which case the container would never get the exception, and the exception would be caught by whatever happened to complete the async response, which is just wrong.

So I had to adapt a test which used to assume we'd get a stack trace from the container in the response, while now we get a 500 response with no stack trace. I would have changed the RESTEasy error response to include the stack trace, but I think that's a security risk unless we have a setting to turn this on/off. WDYT @asoldano @ronsigal ?


FroMage:
OK I will.

FroMage:
Done: https://issues.jboss.org/browse/RESTEASY-1921

PR #1540, opened 2018-06-07
[RESTEASY-1920] Add 13.0.0.Final server version

marekkopecky:
[RESTEASY-1920] Add 13.0.0.Final server version

Jira: https://issues.jboss.org/browse/RESTEASY-1920

3.6 PR: https://github.com/resteasy/Resteasy/pull/1541


marekkopecky:
> I would remove the 10.1.0.Final target

+1

> either we exclude all failing tests against 13.0.0.Final, or we wait for the testsuite to be fixed before enabling the new target container

I prefer to wait. I just create this PR (together with #1541) as reminder for WF13 testing ...

PR #1543, opened 2018-06-07
[RESTEASY-1927] ClientResponseContextImpl: getEntityStream returns possibly exhausted inputStream

stevenschlansker:
Currently, a response filter that chooses to read the response context's `getEntityStream` is returning the `inputStream`.
This is fine if the stream has not been read already.  But if it ends up being buffered, either internally or via `bufferEntity()`, the input stream is exhausted.

Returning the entity stream from the response fixes this issue.

### DIFF
--- a/resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientResponse.java
+++ b/resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientResponse.java
@@ -24,6 +24,7 @@ class JettyClientResponse extends ClientResponse {
     @Override
     protected void setInputStream(InputStream is) {
         stream = is;
+        resetEntity();
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


stevenschlansker:
cc @mikebell90 @pennello @strykerd @rushalias

stevenschlansker:
Thanks for sharing your hunch, I think you might be right.  I've attempted to fix this by resetting the buffered entity, e.g.

```diff
     @Override
     protected void setInputStream(InputStream is) {
         stream = is;
+        streamFullyRead = false;
+        entity = null;
     }
```

but this additional change appears to break some tests which check buffered exception responses.  I'll continue digging in.

stevenschlansker:
Hm, I cannot reproduce the test failures I saw yesterday.  @asoldano do you think the fix of resetting the entity, so it re-processes the stream, is correct for the issues you highlighted above?  It seems reasonable to me but I understand this is probably a touchy area of the code.

stevenschlansker:
Ok, I understand the test failure now.  `AbortedResponse` should not reset the entity, since it might be provided directly rather than being set by stream.

asoldano:
Thanks!

PR #1551, opened 2018-06-11
[RESTEASY-1897] Moving YamlProvider to different package for JDK9 mod…

asoldano:
…ular support


asoldano:
@rsvoboda , just to be sure we're on the same page, note that actually we're talking about a class that was deprecated at least in a previous minor.
From a container (WFLY/EAP) point of view, the class here is the only class in the org.jboss.resteasy.resteasy-yaml-provider which is marked as private, so we're allowed to apply any change in it.
The reason we have the #1552 is just to keep 3.6 and master aligned for some more time. If the PR here is not approved/ merged, I'll close #1552 and drop the resteasy-yaml completely on master (given it's deprecated and for a good a reason).

rsvoboda:
The safest approach is to keep 3.6 untouched, we can drop resteasy-yaml completely on master.

asoldano:
OK, closing then.

PR #1566, opened 2018-06-15
[Resteasy 1911]:JSONB provider is used even if Jackson2 provider should be used

jimma:


rsvoboda:
Asking @marekkopecky to take a look as he filed the original issue.

Have some comments in general
 * PR description is missing details, at least link to https://issues.jboss.org/browse/RESTEASY-1911
 * ID of PR should be fixes for it gets linked to https://issues.jboss.org/browse/RESTEASY-1911
 * https://issues.jboss.org/browse/RESTEASY-1911 state should be updated

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
```diff
@@ -37,14 +43,52 @@
 @Priority(Priorities.USER-100)
 public class JsonBindingProvider extends AbstractJsonBindingProvider
         implements MessageBodyReader<Object>, MessageBodyWriter<Object> {
-
+   private static final boolean JSONB_DISABLED = Boolean.getBoolean("resteasy.jsonb.disable");
```
rsvoboda: Existing deployments do not have `resteasy.jsonb.disable` defined so the `JSONB_DISABLED` will be false.

Will those deployments work correctly without need for any changes in them ?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/JacksonDatatypeTest.java
```diff
@@ -8,10 +11,18 @@
 import org.jboss.resteasy.client.jaxrs.ResteasyClient;
 import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
 import org.jboss.resteasy.test.core.basic.resource.ApplicationTestScannedApplication;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonJAXBResource;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonProduct;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonResource;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonXmlProduct;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonXmlResource;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonXmlResourceWithJAXB;
+import org.jboss.resteasy.test.providers.jackson.resource.JacksonXmlResourceWithJacksonAnnotation;
```
rsvoboda: Fix above imports (not used)

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
asoldano: There are some deployments that are going to be affected and need changes. That's also why Jim had to fix some tests in the PR here.
Note, we *can't* default to JSONB_DISABLED = true as that would break EE8 compliance.
So, IOW, there're still some scenarios in which JAX-RS 2.1/EE8 is forcing us to break backward compatibility (and this is also why I'm adding you as reviewer here)

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
```diff
@@ -37,14 +43,52 @@
...
+      if (JSONB_DISABLED)
+      {
+         return false;
+      }
+      ResteasyConfiguration context = ResteasyProviderFactory.getContextData(ResteasyConfiguration.class);
+      if (context != null && Boolean.parseBoolean(context.getParameter("resteasy.jsonb.disable")))
+      {
+         return false;
+      }
+      if (context != null && Boolean.parseBoolean(context.getParameter("resteasy.jsonp.enable")))
```
rsvoboda: `Boolean.parseBoolean(context.getParameter("resteasy.jsonp.enable")` is used twice, it could be put into property, similar to JSONB_DISABLED.

rsvoboda: `resteasy.jsonp.enable` is mentioned in docs just in relation to Jackson2JsonpInterceptor

Should be docs updated also for JsonBindingProvider section to mention it ?

rsvoboda: `resteasy.jsonb.disable`  vs.  `resteasy.jsonp.enable` can be confusing

similar property names, jsonb disable, but jsonp enable. This can be misleading for end users, I switched them when I was typing this comment

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
rsvoboda: `resteasy.jsonb.disable` is not documented

rsvoboda: Shouldn't we introduce something like `resteasy.EE7.like.behavior` property?

This is not the first change which may require customers actions with deployments.
Customers do not want to touch deployments. And when we think about OpenShift, it's another reason why it should be good to have one system property for backward compatible behavior.

jimma: This line reads system property "resteasy.jsonb.disable" if user needs a global settings for all resouces/deloyments.

jimma: This is in my todo list. I'll update the documentation soon.

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
jimma: resteasy.jsonp.enable is read from context if context is settings and It's different from JSONB_DISABLED reading from system property.
I failed to find some documentation about resteasy.jsonp.enable and we need to fix this.
About the confusing property name , is the full name like resteay.jsonbindng.disable better ?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/JacksonDatatypeTest.java
jimma: Fixed. Thanks.

PR #1568, opened 2018-06-16
[RESTEASY-1879] Move MP REST Client 1.0 impl to separate module (take II)

asoldano:


asoldano:
@sberyozkin , please check here

resteasy-client-microprofile/pom.xml
```diff
@@ -23,5 +23,26 @@
...
             <version>${project.version}</version>
         </dependency>
+        
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-cdi</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        
+        <dependency>
```
rsvoboda: resteasy-cdi has dependency on this - https://github.com/resteasy/Resteasy/blob/master/resteasy-cdi/pom.xml#L29

Is it needed to be explicitly declared here ?

asoldano: The MP* classes are needed to have a custom version of the ClientInvocation and ResteasyProviderFactory , which included changes for the MP Rest Client.
I'll check the cdi dependency thing later/tomorrow, thanks

asoldano: OK, checked, that dependency is needed because the javax.enterprise:cdi-api dependency defaults to provided scope due to the way it's declared in resteasy-dependencies-bom/pom.xml

asoldano:
@rsvoboda right, thanks, I've just fixed both.

kanovotn:
@asoldano @rsvoboda Tests didn't reveal any regressions/backward compatibility issues. LGTM

PR #1573, opened 2018-06-21
[RESTEASY-1861] RESTEasy ignores @ConstrainedTo in some situations

jurakp:
Jira: https://issues.jboss.org/browse/RESTEASY-1861

The approach is to not call configure method if the runtime environment does not match the one mentioned in javax.ws.rs.ConstrainedTo annotation (similar to dynamic features).

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
@@ -2010,15 +2010,17 @@ public class ResteasyProviderFactory extends RuntimeDelegate implements Provider
       }
       if (isA(provider, Feature.class, contracts))
       {
+         ConstrainedTo constrainedTo = (ConstrainedTo) provider.getAnnotation(ConstrainedTo.class);
          int priority = getPriority(priorityOverride, contracts, Feature.class, provider);
          Feature feature = injectedInstance((Class<? extends Feature>) provider);
-         if (feature.configure(new FeatureContextDelegate(this)))
-         {
-            enabledFeatures.add(feature);
+         if (constrainedTo == null || constrainedTo.value() == getRuntimeType()) {
+            if (feature.configure(new FeatureContextDelegate(this)))
+            {
+               enabledFeatures.add(feature);
+            }
          }
          featureClasses.add(provider);
          newContracts.put(Feature.class, priority);
-
       }
       if (isA(provider, RxInvokerProvider.class, contracts))
       {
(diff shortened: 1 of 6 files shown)


resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -2010,13 +2010,30 @@ protected void processProviderContracts(Class provider, Integer priorityOverride
       }
       if (isA(provider, Feature.class, contracts))
       {
+         ConstrainedTo constrainedTo = (ConstrainedTo) provider.getAnnotation(ConstrainedTo.class);
          int priority = getPriority(priorityOverride, contracts, Feature.class, provider);
          Feature feature = injectedInstance((Class<? extends Feature>) provider);
-         if (feature.configure(new FeatureContextDelegate(this)))
-         {
-            enabledFeatures.add(feature);
+         if (constrainedTo != null && constrainedTo.value() == RuntimeType.SERVER && getRuntimeType() == RuntimeType.SERVER) {
```
asoldano: How about replacing this and the following if clauses with:
if (contrainedTo == null || getRuntimeType() == constrainedTo.value())

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/custom/resource/CustomClientConstrainedFeature.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A feature constrained to the client runtime.
+ * 
+ * @author pjurak
+ */
+@Provider
+@ConstrainedTo(RuntimeType.CLIENT)
+public class CustomClientConstrainedFeature implements Feature {
+   private static boolean invoked = false;
```
asoldano: This should be volatile

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/custom/resource/CustomServerConstrainedFeature.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A feature constrained to the server runtime.
+ * 
+ * @author pjurak
+ */
+@Provider
+@ConstrainedTo(RuntimeType.SERVER)
+public class CustomServerConstrainedFeature implements Feature {
+   private static boolean invoked = false;
```
asoldano: This should be volatile

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
jurakp: Yep, I originally thought about it, but I wanted to keep the same style as for dynamic features. Will change it then. Thanks!

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -2010,15 +2010,17 @@ protected void processProviderContracts(Class provider, Integer priorityOverride
...
          int priority = getPriority(priorityOverride, contracts, Feature.class, provider);
          Feature feature = injectedInstance((Class<? extends Feature>) provider);
-         if (feature.configure(new FeatureContextDelegate(this)))
-         {
-            enabledFeatures.add(feature);
+         if (constrainedTo == null || constrainedTo.value() == getRuntimeType()) {
+            if (feature.configure(new FeatureContextDelegate(this)))
+            {
+               enabledFeatures.add(feature);
+               featureClasses.add(provider);
```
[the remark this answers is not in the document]
jurakp: It is the same for featureInstances. Since there is a lack of docs, I assume that purpose of this method is to get all defined enabled feature instances/classes (was not able to find any class which calls this method though). But let me put it outside the if block, so that the change is small.

asoldano:
Thanks @jurakp

PR #1589, opened 2018-06-28
[RESTEASY-1890] Implement all the server side tracing events.

liweinan:
Server tracing events are all implemeted, and tests are added. Some minor bugs are fixed.


resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/registry/SegmentNode.java
```diff
@@ -74,6 +76,11 @@ public Match(MethodExpression expression, Matcher matcher)
    public ResourceInvoker match(HttpRequest request, int start)
    {
       String path = request.getUri().getMatchingPath();
+
+      LogMessages.LOGGER.pathInfo(request.getUri().getPath());
```
asoldano: is this moved here from SynchronousDispatcher on purpose?

liweinan: Oops. I moved it for testing purpose, and I should put it back to `SynchronousDispatcher`, or the `pathInfo` message will be logged to many times :-) Let me revert this change.

PR #1592, opened 2018-06-29
[RESTEASY-1932]

ronsigal:
Fixed ClientResponseFilterExceptionTest, which was passing for the wrong reason.


ronsigal:
@asoldano Race condition in the tests. Aarrgghh. Fixed.

PR #1594, opened 2018-07-02
[JBEAP-10146] added permissions to deployed archive

rsearls:


asoldano:
@rsearls please see comment from @kanovotn

rsearls:
All permissions added were based upon an exception being thrown for it.

kanovotn:
I meant for example:
`// Needs to read applicationContext.xml file`
`new FilePermission("<<ALL FILES>>", "read"),` `// Curious why it needs to read permission for all files and not only for the one needed.`

I'm not going to obstruct here, so if there is consensus on the approach in this commit please go ahead for merge.

asoldano:
I've added the requires rebase label as we now have a new version of the spec api jar pulled on master head and that contains security manager related fixes.

rsearls:
rebase complete

rsearls:
Once you merge the changes for RESTEASY-1946, I can rebase the changes into this branch and everything will be upto date.

asoldano:
@rsearls can this be fixed now?

rsearls:
These tests remain broken but not due to permissions.  See https://issues.jboss.org/browse/RESTEASY-1994

asoldano:
@rsearls , please see my comments on RESTEASY-1994. Moreover, would then the testsuite here expected to fail?

asoldano:
Btw, I suggest rebasing against master, just in case, to be sure the latest sources are used.

rsearls:
@asoldano, @kanovotn  I've rebased to master.   All tests run on unix when using jdk-10.  Test  org.jboss.resteasy.test.asynch.AsyncInjectionTest  hangs when run with jdk-8, but this issue is unrelated to
any permissions issue.

PR #1596, opened 2018-07-05
[RESTEASY-1943] Async filter resume with exception issues

FroMage:
This fixes the following bugs:

- `SuspendableRequestContainerContext.resume(Throwable)` and `SuspendableResponseContainerContext.resume(Throwable)` were ill-specified
- `SuspendableRequestContainerContext.resume(Throwable)` and `SuspendableResponseContainerContext.resume(Throwable)` did not call async callbacks
- `SuspendableRequestContainerContext.resume(Throwable)` did not always call exception mapper and response filters
- `SuspendableResponseContainerContext.resume(Throwable)` should not call exception mapper and response filters

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AbstractAsynchronousResponse.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AbstractAsynchronousResponse.java
@@ -38,6 +38,7 @@ public abstract class AbstractAsynchronousResponse implements ResteasyAsynchrono
    protected TimeoutHandler timeoutHandler;
    protected List<CompletionCallback> completionCallbacks = new ArrayList<CompletionCallback>();
    protected Map<Class<?>, Object> contextDataMap;
+   private boolean callbacksCalled;
 
    protected AbstractAsynchronousResponse(SynchronousDispatcher dispatcher, HttpRequest request, HttpResponse response)
    {
@@ -151,6 +152,10 @@ public abstract class AbstractAsynchronousResponse implements ResteasyAsynchrono
    @Override
    public void completionCallbacks(Throwable throwable)
    {
+      // make sure we only call them once
+      if(callbacksCalled)
+         return;
+      callbacksCalled = true;
       for (CompletionCallback callback : completionCallbacks)
       {
          callback.onComplete(throwable);
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/interception/jaxrs/ContainerResponseContextImpl.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/interception/jaxrs/ContainerResponseContextImpl.java
@@ -29,6 +29,7 @@ import org.jboss.resteasy.specimpl.BuiltResponse;
 import org.jboss.resteasy.spi.ApplicationException;
 import org.jboss.resteasy.spi.HttpRequest;
 import org.jboss.resteasy.spi.HttpResponse;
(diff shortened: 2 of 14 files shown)


FroMage:
Corresponding JIRA: https://issues.jboss.org/browse/RESTEASY-1943

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/AbstractAsynchronousResponse.java
```diff
@@ -38,6 +38,7 @@
    protected TimeoutHandler timeoutHandler;
    protected List<CompletionCallback> completionCallbacks = new ArrayList<CompletionCallback>();
    protected Map<Class<?>, Object> contextDataMap;
+   private boolean callbacksCalled;
```
asoldano: Should this be volatile?

FroMage: I don't think so, but honestly I'm not sure. At least I don't hope concurrent threads will resume the same request. Is that even specified?

FroMage:
> there're some failures on Travis CI that look related to the changes here

Yeah, and it only fails on CI for some config… I can't reproduce it locally. I'll try to switch JDK to see if that changes anything.

FroMage:
Just what I thought: it runs locally on both Java 8 and 9 (using undertow and WF-12).

Is there any setting in the build I need to tweak to try another version of WF?

FroMage:
I tried with Java 8 and 10 (both oracle), Java 8 and 9 (openjdk) with undertow and it passes. Must be related to the WF version/jdk.

asoldano:
@FroMage check the .travis.yml config for info on the jobs; in any case you can use -Dserver.version=... to specify the WF version to test against.
Regarding the failures, I haven't had time to look at the properly, but in general we happen to have failure that are only reproducible on Travis from time to time. Usually, that's because it's way slower that our laptop and/or doing thread scheduling differently, etc. We also happened to have issues with entropy generation and random numbers, but that's definitely not the case here.

FroMage:
Thanks. I've tried with WF 11, 12, 13. Tried with shorter sleeps in the test, longer sleeps, no sleeps and still can't reproduce this bug.

asoldano:
@FroMage , first of all, I've just re-started the job, let's see how it goes.
When we happened to have similar situations in the past, I've tried to isolate the problem on Travis, by creating a new personal PR, excluding as much unrelated tests as possible (in some cases just running the failing test won't reproduce the problem), printing server and client logs at the end of the run (need to do that in the .travis.yml), adding some debug logs, etc.

FroMage:
OK I just created #1598 to try running the single test on jdk8 (seems like the best failing candidate) and WF 12.

FroMage:
OK, thanks for your help. I found the issue, and it's related to `AsyncResponse.resume` and `CompletionCallback` because this is the ordering that happens:

- Client makes a request A
- Server suspends request and spawns a thread to handle request A
- Thread resumes request A with exception (should trigger completion callbacks)
- Server sends response to client
- Client makes a request B to see if the callback was called
- Server returns that the callback was not called (yet)
- Client marks test as failing
- Thread (from Request A) continues, and calls the callbacks (too late)

Not sure I can do better than make the client sleep some more between requests A and B, WDYT @asoldano ? The spec confirms callbacks should be called after the response is sent, so that's proper.

asoldano:
Thanks @FroMage

FroMage:
Thanks!

Any chance I can get a snapshot published? :)

asoldano:
@FroMage I've just pushed a new snapshot :-)

PR #1598, opened 2018-07-06
Async filter resume with exception issues debug PR 

FroMage:
DO NOT MERGE. This is for debugging #1596 on CI server


FroMage:
OK found it.

PR #1478, opened 2018-04-05
[RESTEASY-1798]

ronsigal:
Implement proxies for Single and Observable.

[RESTEASY-1798]

Remvoving RxCompletionStageProxyTest; adding @Ignore to
resteasy-rxjava2/RxTest.

[RESTEASY-1798]

Deleted RxCompletionStageProxyTest.

PR #1435, opened 2018-03-01
[RESTEASY-1816] OnComplete callback registered via SseEventSource.register() is not called

rnetuka:
JIRA: https://issues.jboss.org/browse/RESTEASY-1816

Test case added separately in https://github.com/kanovotn/Resteasy/commit/4cc3f12f862b45212bfe6a2fc2a3780cb6be90c0#diff-46bd7bf2136ed3b55bf449e8b0a9ee77R121

PR #1599, opened 2018-07-07
OnComplete callback registered via SseEventSource.register() must be called only once

NicoNes:
This pull request is to remove multiple invocation of SseEventSource completion handlers and enable test about it.

@ronsigal You might be interested by this PR since it seem's that the regression has been introduced by https://github.com/resteasy/Resteasy/pull/1478 

@rnetuka and @kanovotn You also might be interested by this PR as an enrichment to what have been done in https://github.com/resteasy/Resteasy/pull/1435/files


asoldano:
Thanks @NicoNes !

PR #1600, opened 2018-07-08
[RESTEASY-1946] jboss-modules requires more permissions on tests

rsearls:
These changes are only in effect when running with the security manager flag, (-Dsecurity.manager).  This PR only addresses the test failures due to missing security permissions. There are other tests failing due to issues other than permissions and only show when running with the flag above.  Those test failures will be addressed by separate jira(s).


testsuite/integration-tests/pom.xml
```diff
@@ -274,6 +274,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
asoldano: What's the link between tests failing with security manager enabled and the need for explicit hibernate-validator dependency here? Is it to control the exact version being used? And if that's the case, is it really a problem to have code that runs regardless the version of the hibernate-validator that comes from the selected target container?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/asynch/AsyncPostProcessingTest.java
```diff
@@ -46,15 +46,15 @@
...
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+                new ReflectPermission("suppressAccessChecks"),
                 new PropertyPermission("arquillian.*", "read"),
                 new PropertyPermission("ipv6", "read"),
                 new PropertyPermission("node", "read"),
                 new PropertyPermission("org.jboss.resteasy.port", "read"),
                 new RuntimePermission("accessDeclaredMembers"),
                 new RuntimePermission("getenv.RESTEASY_PORT"),
-                new SecurityPermission("insertProvider"),
+                new RuntimePermission("getClassLoader"),
```
asoldano: Just as a confirmation, is the getClassloader permission needed here and everywhere else in this PR because of jboss-modules? or maybe RESTEasy is actually needing it?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientResponseRedirectTest.java
```diff
@@ -46,10 +46,12 @@
         WebArchive war = TestUtil.prepareArchive(ClientResponseRedirectTest.class.getSimpleName());
         war.addClass(ClientTestBase.class);
         // Use of PortProviderutil in the deployment
-        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(new PropertyPermission("node", "read"),
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+                new PropertyPermission("node", "read"),
                 new PropertyPermission("ipv6", "read"),
                 new RuntimePermission("getenv.RESTEASY_PORT"),
-                new PropertyPermission("org.jboss.resteasy.port", "read")), "permissions.xml");
+                new PropertyPermission("org.jboss.resteasy.port", "read")
```
asoldano: Actually, now that I notice this, I wondering if running the System.getProperty("org.jboss.resteasy.port") in a privileged block wouldn't be better than asking users to explicitly add the permissions...

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/RedirectTest.java
```diff
@@ -36,6 +39,12 @@
    {
       WebArchive war = TestUtil.prepareArchive(RedirectTest.class.getSimpleName());
       war.addClasses(PortProviderUtil.class);
+      war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+              new PropertyPermission("ipv6", "read"),
+              new RuntimePermission("getenv.RESTEASY_PORT"),
```
asoldano: Same concerns as with "org.jboss.resteasy.port" property permission

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/spi/ResourceClassProcessorNotAppliedTest.java
```diff
@@ -47,7 +51,17 @@ public static synchronized void addToVisitedProcessors(String item) {
         war.addClass(ResourceClassProcessorNotAppliedTest.class);
         war.addClass(PortProviderUtil.class);
         war.addClass(ResourceClassProcessorNotAppliedImplementation.class);
-
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+                new SocketPermission(PortProviderUtil.getHost(), "connect,resolve"),
+                new PropertyPermission("org.jboss.resteasy.port", "read"),
+                new RuntimePermission("getenv.RESTEASY_PORT"),
+                new PropertyPermission("ipv6", "read"),
```
asoldano: Same concerns as with "org.jboss.resteasy.port" property permission

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jaxb/JaxbMarshallingSoakTest.java
```diff
@@ -80,7 +80,9 @@ public void after() throws Exception {
         Map<String, String> contextParam = new HashMap<>();
         contextParam.put("resteasy.async.job.service.enabled", "true");
         // Arquillian in the deployment use if TimeoutUtil in the deployment
-        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(new ReflectPermission("suppressAccessChecks"),
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+                new RuntimePermission("modifyThread"),
```
asoldano: Out of curiosity, is jboss-modules needing this permission?

rsearls: The call to threadPool.shutdown(); at line 162 in test is causing this permission requirement.

	at org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:295)
	at org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:192)
	at java.util.concurrent.ThreadPoolExecutor.checkShutdownAccess(ThreadPoolExecutor.java:735)
	at java.util.concurrent.ThreadPoolExecutor.shutdown(ThreadPoolExecutor.java:1387)
	at org.jboss.resteasy.test.providers.jaxb.JaxbMarshallingSoakTest.compare(JaxbMarshallingSoakTest.java:162)

asoldano: OK, so in this case it really makes sense to add the permission, it's testcase eventually needing that, fine.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientResponseRedirectTest.java
rsearls: The test calls PortProviderUtil in arquillian-utils.  This class is calling 
System.getenv and/or System.getProperty however putting these calls in a privileged 
block is of no help because the test itself is not running in a security manager.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/RedirectTest.java
rsearls: The test calls PortProviderUtil in arquillian-utils.  This class is calling 
System.getenv and/or System.getProperty however putting these calls in a privileged 
block is of no help because the test itself is not running in a security manager.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/spi/ResourceClassProcessorNotAppliedTest.java
rsearls: same a previous 2 comments

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientResponseRedirectTest.java
rsearls: Hmmm. If the client is not running in the SM why is the permission being requested.    This is curious.

rsearls: It looks like wildfly security implements a rule that System.getProperty and
System.getenv calls require explicit permissions to exec the cmd.
15:27:43,036 ERROR [stderr] (default task-1) 	at org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:295)
15:27:43,036 ERROR [stderr] (default task-1) 	at org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:192)

This means the existing permission 
          new PropertyPermission("node", "read"),
          new PropertyPermission("ipv6", "read"),
          new RuntimePermission("getenv.RESTEASY_PORT"),
          new PropertyPermission("org.jboss.resteasy.port", "read")

must stay unless there is some other way to direct wildfly to allow them.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/asynch/AsyncPostProcessingTest.java
asoldano: Just tried, this is the stacktrace: https://paste.fedoraproject.org/paste/bg8Xppg3l1-nrP6TdSRP8Q/raw
So javax.ws.rs.ext.FactoryFinder.getModuleClassLoader method is calling getClassloader through reflection and that's failing because WildFlySecurityManager checks for the permissions before allowing that.
If we want to avoid asking users to provide the permission here (and I think we should, it's really a detail out of user control), we should likely fix our jboss-jaxrs-api project: https://github.com/jboss/jboss-jaxrs-api_spec/blob/master/src/main/java/javax/ws/rs/ext/FactoryFinder.java#L255 .

This is basically the same that has been done in the past for jaxws at https://github.com/jboss/jboss-jaxws-api_spec/commit/c1eb89a088255f071f01bbb7561c77cb3052239d#diff-37f7e6b2d322553ab40f0cc7f1290683L178 (see https://issues.jboss.org/browse/JBEE-176 ).

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientResponseRedirectTest.java
asoldano: Interesting, can you please investigate this a little bit more? Does it have anything to do with arquillian being used to run the tests?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/asynch/AsyncPostProcessingTest.java
rsearls: Making the code change in https://github.com/jboss/jboss-jaxrs-api_spec FactoryFinder
resolved the "getClassLoader" issue there but a new "getClassLoader" issue is now
present.  See  https://paste.fedoraproject.org/paste/86J2ZO4cXvWaHKjjU75YkQ 

The call in org.jboss.resteasy.core.ContextParameterInjector.createProxy(ContextParameterInjector.java:168)
is to java.lang.reflect.Proxy which does perform a SecurityManager check. 

Suggestions for addressing this one?

asoldano: Judging from the stacktrace, the call resulting in the getClassloader permission check being performed is the "rawType.getClassLoader()" one that's used to compute the first parameter for the proxy creation. So I think you could use a privileged action to get that classloader, right?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientResponseRedirectTest.java
asoldano: OK, right. What's not clear to me is why the permission check fails if you wrap the invocation (in PortProviderUtil) into a PrivilegedAction; there has to be an explanation (I assume that's on purpose, maybe someone from core team can help?)

PR #1602, opened 2018-07-11
[RESTEASY-1922] Exclude interfaces

asoldano:


asoldano:
@mkouba this is what you were suggesting to do, right?

mkouba:
Yep, exactly. Thanks!

PR #1604, opened 2018-07-12
Master cleanup for version 4 - Phase I

asoldano:
…nding test modules


rsvoboda:
@marekkopecky it's worth to check this PR, we can discuss it f2f once you have chance to see this.

PR #1605, opened 2018-07-14
Resteasy 1958: connect delay override using last received 'retry' field value should not be a one time override

NicoNes:


jimma:
@NicoNes  I think respected retry field as one-time retry value is correct way to handle this.  If the retry field is set for server scheduled disconnect event stream for some period time and later there is connection lost by local network, it doesn't makes sense to still use this retry value to reconnect.

NicoNes:
Hey @jimma,

> If the retry field is set for server scheduled disconnect event stream for some period time and later there is connection lost by local network, it doesn't makes sense to still use this retry value to reconnect.

I'm not sure to understand your point well. Can you please detail ? Thanks

jimma:
@NicoNes  For example, retry fieldy set by server is intended to let EventSource wait 30 minutes to reconnect; after 30 minutes the connection is established again and EventSource starts to consume the event events, during consuming events if network connection drops,  it doesn't make sense it still wait 30 minutes to reconnect.

NicoNes:
@jimma Thank you for the details.

Actually from my undertsanding of the doc the last "Retry" field received must override the SseEventSource default retry delay in  an "irrevocable" way not a one-time. 

So if the last Retry field received was set to 30 minutes it will become the default retry delay until the server set retry field to something else.

For information and for what it worth this is the behavior of HTML eventSource() object.

Anyway,  again I asked the spec guys for clarification about the expected behavior.

Thanks

NicoNes:
Here is the issue I opened to get clarification: https://github.com/eclipse-ee4j/jaxrs-api/issues/644

NicoNes:
PR updated here https://github.com/resteasy/Resteasy/pull/1995

PR #1607, opened 2018-07-17
[RESTEASY-1934] Use existing methods in ResteasyProviderFactory to pr…

rsearls:
…event leaks

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/basic/DecoratorsTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/basic/DecoratorsTest.java
@@ -4,6 +4,8 @@ import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.logging.Logger;
+import org.jboss.resteasy.plugins.providers.RegisterBuiltin;
+import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.test.cdi.basic.resource.DecoratorsBookReader;
 import org.jboss.resteasy.test.cdi.basic.resource.DecoratorsBookReaderDecorator;
 import org.jboss.resteasy.test.cdi.basic.resource.DecoratorsBookReaderInterceptor;
@@ -32,6 +34,8 @@ import org.jboss.resteasy.utils.PortProviderUtil;
 import org.jboss.resteasy.utils.TestUtil;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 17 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/basic/DecoratorsTest.java
```diff
@@ -71,6 +74,19 @@
         return war;
     }
 
+    private ResteasyProviderFactory factory;
+    @Before
+    public void setup() {
+        // Create an instance and set it as the singleton to use
+        factory = ResteasyProviderFactory.newInstance();
```
asoldano: I believe _RegisterBuiltin.register(factory);_ should also be executed, in order to have a provider factory that behaves the expected way. WDYT? (this applies to all similar code blocks in this PR)

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/ClientDynamicFeaturesTest.java
```diff
@@ -60,7 +60,7 @@
      */
     @Test
     public void testDynamicFeatures() throws Exception {
-        ResteasyProviderFactory factory =  ResteasyProviderFactory.getInstance();
+        ResteasyProviderFactory factory =  ResteasyProviderFactory.newInstance();
```
asoldano: Same comment as before, regarding the need to call _RegisterBuiltin.register(factory);_

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/priority/resource/ProviderPriorityResource.java
```diff
@@ -32,7 +32,7 @@ public String paramConverter(@PathParam("foo") ProviderPriorityFoo foo) throws E
    @GET
    @Path("register")
    public Response register() throws Exception {
-      ResteasyProviderFactory factory = ResteasyProviderFactory.getInstance();
+      ResteasyProviderFactory factory = ResteasyProviderFactory.newInstance();
```
asoldano: I have some doubts about the change here. In this case, the attempt is exactly to alter the current ResteasyProviderFactory instance (yes, I agree, that's not a best practice at all). By creating a new instance and not setting it in the singleton reference within ResteasyProviderFactory, you're altering the test. I suspect this is not spotted by Travis CI because the test here is likely not running due to the exclusions for RESTEASY-1935.

asoldano: Do you think you can let the test do what it currently does and cleanup/restore the initial status afterwards?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/HeaderDelegateTest.java
```diff
@@ -102,7 +102,7 @@ public void lastModifiedTest() throws Exception {
      */
     @Test
     public void localTest() throws Exception {
-        ResteasyProviderFactory factory = ResteasyProviderFactory.getInstance();
+        ResteasyProviderFactory factory = ResteasyProviderFactory.newInstance();
```
asoldano: Similar comment to the previous one. The test wants to check the current instance, not a new one.

rsearls: All changes complete.

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/providers/ProviderFactoryTest.java
```diff
@@ -127,7 +127,7 @@ public void testDeploymentStart()
             @Override
             public void configure(ResourceInfo resourceInfo, FeatureContext context)
             {
-               if (ResteasyProviderFactory.getInstance().isRegistered(MyInterceptor.class))
+               if (ResteasyProviderFactory.newInstance().isRegistered(MyInterceptor.class))
```
asoldano: This should probably not be changed, judging from the test code here around

rsearls:
Fixed and all tests run now.

PR #1617, opened 2018-07-26
[RESTEASY-1930] NullEntityProxyTest, VariantAcceptTest

kanovotn:


providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
```diff
@@ -71,9 +72,11 @@ public Object readFrom(Class<Object> type, Type genericType,
                                  MultivaluedMap<String, String> httpHeaders,
                                  InputStream entityStream) throws java.io.IOException, javax.ws.rs.WebApplicationException {
       Jsonb jsonb = getJsonb(type);
-      try
-      {
+      try {
          return jsonb.fromJson(entityStream, genericType);
+         // If null is returned, considered to be empty stream
+      } catch (NullPointerException ex) {
+         return null;
```
kanovotn: This is based on behavior of ResteasyJackson2Provider, it also allows to return null from readFrom in case the entity stream is empty.

PR #1620, opened 2018-07-26
[RESTEASY-1965] permission added to allow read,write anything in app

rsearls:


rsearls:
See comment and stackTrace in https://issues.jboss.org/browse/RESTEASY-1965

asoldano:
@rsearls , OK, commented on jira. It's because of rx-java.

PR #1625, opened 2018-07-27
[RESTEASY-1975] test rewritten with contents of original RESTEASY-108…

rsearls:
…2.war


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/basic/CDIResourceTest.java
```diff
@@ -24,29 +30,58 @@
...
+        war.addClasses(FooResource.class,
+                TestApplication.class,
+                TestServlet.class);
+        war.addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+        war.addAsWebInfResource(CDIResourceTest.class.getPackage(),
+                "web-resteasy1082.xml", "web.xml");
+
+        war.addAsManifestResource(PermissionUtil.createPermissionsXmlAsset(
+                new SecurityPermission("insertProvider"),
+                new SecurityPermission("insertProvider.BC"),
```
asoldano: These 2 permissions are weird. How is it that this test needs security permissions for the security provider, and BouncyCastle in particular?

asoldano: I mean what makes this different from others in the testsuite, considering it does not even look like a security related test?

rsearls:
Good catch.   I can't regenerated the need for these permissions.    I have removed them.

PR #1637, opened 2018-08-01
[RESTEASY-1933] Provide an alternative way to set and get ObjectWrite…

asoldano:
…rModifier (based on TCCL, to be used with servlet filters)


asoldano:
The failure on one run is unrelated, merging

PR #1649, opened 2018-08-06
[RESTEASY-1695] Add GRAMMARS into RESTEasy WADL

liweinan:


liweinan:
work in progress

liweinan:
Design:

image

liweinan:
External grammar can be included now:

image

The left tasks is allow server to serve the external schema.

liweinan:
`ResteasyWadlGrammar` is the class for grammar storing(both included grammars or generated grammars):

class diagram3

And it's designed to be thread-safe.

liweinan:
Sample included schema output: 

image

The usage should be simple:

```java
contextBuilder.getDeployment().getActualResourceClasses().add(ResteasyWadlExtendedResource.class);
```

```java
ResteasyWadlGrammar.includeGrammars("application-grammars.xml");
```

As shown above, the included grammars should be put into classpath and set via `includeGrammars` method.

liweinan:
Now the schema generation is working:

image

It will scan the resource methods and its parameters(also return types), and collect all `@XmlRootElement` annotated classes, and generate schemas for them.

liweinan:
tests added. code cleanup done.

rsvoboda:
Just scanned the PR a bit, no deep dive.

 * please remove `System.out.println`
 * hard-coded http://127.0.0.1 can bring issues in the future
    * HttpServer is used instead of ARQ in resteasy-wadl module

liweinan:
@rsvoboda @asoldano To move tests to `integration-tests` and using ARQ instead of the current standalone tests, we have to add `resteasy-wadl` module into `jboss-modules`:

image

wdyt?

resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/ResteasyWadlGrammar.java
```diff
@@ -0,0 +1,200 @@
...
+
+    // the JAXB annotated classes need to be included to generate schemas
+    public static Set<Class> schemaClasses = Collections.synchronizedSet(new HashSet<>());
+
+    static ClassLoader loader = Thread.currentThread().getContextClassLoader();
+
+    private static AtomicBoolean generateSchema = new AtomicBoolean(false);
+
+
+    public static boolean hasGrammars() {
```
asoldano: Shouldn't this be syncronized too?

asoldano:
@liweinan , as for the resteasy-wadl new module, yes, that will have to be added

liweinan:
thanks for reviewing everyone! Let me do the refactor work.

liweinan:
class diagram1


Refactor:

- `ResteasyWadlGrammar` is no longer a singleton class and it's now inner class of `ResteasyWadlWriter`
- `ResteasyWadlResourceHelper` is added to hold the singletion instance of `ResteasyWadlWriter`(the instance is moved out of `ResteasyWadlDefaultResource`).
- `ResteasyWadlDefaultResource` and `ResteasyWadlExtendedResource` are provided for default WADL usages, and they use the writer inside `ResteasyWadlResourceHelper`.
- `ResteasyWadlWriter` supports caching now(And it's turned off by default).

## Usage Becomes

Add `ResteasyWadlDefaultResource` and `ResteasyWadlExtendedResource` into deployment:

```java      contextBuilder.getDeployment().getActualResourceClasses().add(ResteasyWadlDefaultResource.class);     contextBuilder.getDeployment().getActualResourceClasses().add(ResteasyWadlExtendedResource.class);
```

Actually the usages of`ResteasyWadlDefaultResource`, `ResteasyWadlExtendedResource` and `ResteasyWadlResourceHelper` are all optional. But it's convenient to use them because it helps users to setup almost everything. And it will provide two paths: `/application.xml` and  `/wadl-extended/{path}` to the users.

To use grammar support, it's necessary to create an instance of `ResteasyWadlGrammar`. The recommended way is  to put it into `ResteasyWadlResourceHelper.apiWriter`:

```java
      ResteasyWadlWriter.ResteasyWadlGrammar wadlGrammar = new ResteasyWadlWriter.ResteasyWadlGrammar();
      wadlGrammar.includeGrammars("application-grammars.xml");
      wadlGrammar.enableSchemaGeneration();
      ResteasyWadlResourceHelper.getApiWriter().setWadlGrammar(wadlGrammar);
```

Then `ResteasyWadlDefaultResource` and `ResteasyWadlExtendedResource` can use it properly.

liweinan:
Relative PRs:

- RESTEASY-1360 - Create RESTEasy WADL Feature pack for WildFly
- RESTEASY-2043 - Update jboss-modules to include restasy-wadl module

asoldano:
@liweinan nevermind, the PR now shows no conflicts

liweinan:
@asoldano Thanks for checking! Let me:

- Review the design and make necessary refactor
- Provide the integration tests with Wildfly
- Fix the imports

liweinan:
@rsvoboda Yeah that's a problem. Let me redesign this part.

liweinan:
# Changes

- Fix the original global instance design of wadl writers, and now its instance could be put into per-deployment scope.
- Add wildfly integration test.
- Deprecate `ResteasyWadlServlet` and `ResteasyWadlServletWriter`, because it doesn't support grammars.
- Improve `ResteasyWadlDefaultResource`, for it can be used in both embedded container and servlet container.
- Using logging error instead of runtime exceptions inside wadl writer.
- Other minor fixes.

liweinan:
@asoldano The refactor is done. Could you please help to review it? Thanks!

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/WadlTestApplication.java
```diff
@@ -0,0 +1,100 @@
...
+      return s.hasNext() ? s.next() : "";
+   }
+
+   /**
+    * @see javax.ws.rs.core.Application#getClasses()
+    */
+   @Override
+   public Set<Class<?>> getClasses() {
+      if (classes.isEmpty()) {
+         for (String clazz : getClassesFromDeployment("classes")) {
```
asoldano: Can you please clarify me a bit what you're doing in this Application class and why? In particular, I'm concerned with you reading from the classes.txt file, as AFAIR that's something our TestUtil class adds to deployment just for our testsuite. Is this a shortcut or something? What would be the other approach for achieving the same if that's the case? Maybe I'm simply missing something obvious here, though.

resteasy-wadl-undertow-connector/src/main/java/org/jboss/resteasy/wadl/WadlUndertowConnector.java
```diff
@@ -30,11 +30,11 @@ public UndertowJaxrsServer deployToServer(UndertowJaxrsServer server, Class<? ex
...
-      ServletInfo resteasyWadlServlet = servlet("ResteasyWadlServlet", ResteasyWadlServlet.class)
-            .setAsyncSupported(false)
-            .setLoadOnStartup(1)
-            .addMapping("/application.xml");
-      di.addServlet(resteasyWadlServlet);
+//      ServletInfo resteasyWadlServlet = servlet("ResteasyWadlServlet", ResteasyWadlServlet.class)
+//            .setAsyncSupported(false)
+//            .setLoadOnStartup(1)
+//            .addMapping("/application.xml");
+//      di.addServlet(resteasyWadlServlet);
```
asoldano: Can we please get rid of this and other commented out stuff in this PR please?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/WadlTestApplication.java
liweinan: It seems our current testsuite uses `classes.txt` and `singleton.txt` inside the bundled war file to load them properly.

I copied it from `org.jboss.resteasy.utils.TestApplication`, and it's written by `tterem`:

```bash
$ git blame ./arquillian-utils/src/main/java/org/jboss/resteasy/utils/TestApplication.java | grep classes
48ac2ebcd3 (tterem          2018-09-25 12:32:30 +0200 18)    public static Set<Class<?>> classes = new HashSet<Class<?>>();
48ac2ebcd3 (tterem          2018-09-25 12:32:30 +0200 22)     * Load resources from classes.txt file from deployment
48ac2ebcd3 (tterem          2018-09-25 12:32:30 +0200 64)       if (classes.isEmpty()) {
48ac2ebcd3 (tterem          2018-09-25 12:32:30 +0200 65)          for (String clazz : getClassesFromDeployment("classes")) {
48ac2ebcd3 (tterem          2018-09-25 12:32:30 +0200 68)                   classes.add(Class.forName(clazz));
48ac2ebcd3 (tterem          2018-09-25 12:32:30 +0200 75)       return classes;
```

Our current tests are mainly based on above scheme, but I need to create my own singleton instance of `WadlDefaultResource` for testing, so I have to create `WadlTestApplication` to override the behavior in `TestApplication`.

resteasy-wadl-undertow-connector/src/main/java/org/jboss/resteasy/wadl/WadlUndertowConnector.java
liweinan: yes sir!

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/WadlTestApplication.java
asoldano: OK, how about extending TestApplication and calling super.getSingletons() in the overridden getSingletons() method?

resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/ResteasyWadlServlet.java
```diff
@@ -18,12 +18,19 @@
 
 /**
  * @author <a href="mailto:l.weinan@gmail.com">Weinan Li</a>
+ * This servlet does not support grammars.
+ * Use @org.jboss.resteasy.wadl.ResteasyWadlDefaultServlet instead.
  */
+@Deprecated
```
asoldano: On master we'll likely remove this instead of deprecating (reminder)

resteasy-wadl/src/test/java/org/jboss/resteasy/test/nextgen/wadl/WadlTests.java
```diff
@@ -44,24 +43,25 @@ public void setClient(Client client) {
...
       contextBuilder.getDeployment().getActualResourceClasses().add(BasicResource.class);
-      contextBuilder.getDeployment().getActualResourceClasses().add(ResteasyWadlDefaultResource.class);
+//      contextBuilder.getDeployment().getActualResourceClasses().add(ResteasyWadlDefaultResource.class);
+//      contextBuilder.getDeployment().getActualResourceClasses().add(ResteasyWadlExtendedResource.class);
       contextBuilder.getDeployment().getActualResourceClasses().add(RESTEASY1246.class);
+      contextBuilder.getDeployment().getActualResourceClasses().add(ExtendedResource.class);
+
       contextBuilder.bind(httpServer);
-      ResteasyWadlDefaultResource.getServices().put("/", ResteasyWadlGenerator.generateServiceRegistry(contextBuilder.getDeployment()));
+//      defaultResource.getServices().put("/", ResteasyWadlGenerator.generateServiceRegistry(contextBuilder.getDeployment()));
```
asoldano: Can we also remove the commented out lines here above or do they have some special meaning that justify their presence?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/DeploymentTest.java
```diff
@@ -0,0 +1,81 @@
...
+//      WebArchive war = ShrinkWrap.create(WebArchive.class, DeploymentTest.class.getSimpleName() + ".war");
+//      war.addClass(WadlTestApplication.class);
+
+      WebArchive war = TestUtil.prepareArchiveWithApplication(DeploymentTest.class.getSimpleName(), WadlTestApplication.class);
+      war.addPackages(true, "org.jboss.resteasy.wadl");
+
+      TestUtil.finishContainerPrepare(war, null, ExtendedResource.class, ListType.class);
+
+//      war.as(ZipExporter.class).exportTo(new File("/tmp/" + war.getName()), true);
+      return war;
```
asoldano: Please remove commented out lines here above

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/WadlTestApplication.java
liweinan: We can't extend `TestApplication`, because it's marked as `@Provider` and`@ApplicationPath("/")`. The extended class won't work properly:

```
Failed tests:
  DeploymentTest.testBasic:65 expected:<200> but was:<404>
```

resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/ResteasyWadlServlet.java
liweinan: added TODO comments to remind this

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/DeploymentTest.java
liweinan: done

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/wadl/WadlTestApplication.java
liweinan: I also tried to create an abstract class `BasicApplication extends Application`, and then make `TestApplication` and `WadlTestApplication` to extend it, but it also not work properly. It should be relative to the way `Application` is processed.

resteasy-wadl/src/main/java/org/jboss/resteasy/wadl/ResteasyWadlWriter.java
```diff
@@ -232,4 +283,188 @@ private void setType(Param param, ResteasyWadlMethodParamMetaData paramMetaData)
...
+                       results.add(result);
+                       return result;
+                   }
+               });
+
+               if (grammars != null) {
+                   Iterator<Include> iter = grammars.getInclude().iterator();
+                   while (iter.hasNext()) {
+                       for (Doc doc : iter.next().getDoc()) {
+                           if ("Generated".equals(doc.getTitle())) ;
```
tterem: @liweinan This if statement is useless because of the semicolon at the end of line. Following line is executed every time. Is this a bug? 
cc @asoldano @marekkopecky

liweinan: yes it is.  But this part of code is redundant by current implementation, so it doesn't cause tests to fail. Let me create a JIRA issue to track this bug: https://issues.jboss.org/browse/RESTEASY-2058. Thanks for checking!

PR #1654, opened 2018-08-08
[RESTEASY-1981] Prevent multiple instance registration of same provider

ronsigal:
type


NicoNes:
@ronsigal, Seems ok for me. :+1:

PR #1655, opened 2018-08-13
[RESTEASY-1993] Removing jsonb api dependency from resteasy-jaxrs

asoldano:


asoldano:
@rsvoboda here #1656 is the 3.6 PR...

PR #1664, opened 2018-08-15
[RESTEASY-1987] keep independent list of Future objects to cancel on …

rsearls:
…shutdownNow request

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
@@ -5,8 +5,11 @@ import java.util.Date;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -42,6 +45,8 @@ public class SseEventSourceImpl implements SseEventSource
    private enum State {
       PENDING, OPEN, CLOSED
    }
+
+   private Future<?> fEventHandler;
 
    private final AtomicReference<State> state = new AtomicReference<>(State.PENDING);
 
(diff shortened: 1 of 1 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -291,7 +294,15 @@ private void internalClose()
             });
          }
       }
-      executor.shutdownNow();
+
+      // showdownNow only the objects registered by this class
+      for(Future<?> f : futureEventHandlerList) {
+         if (!f.isDone()) {
+            f.cancel(true);
```
jimma: This looks good to me. Just two minor things: 
From Future javadoc, looks f.cancel() checks if the task is already completed. Can we remove f.isDone() check?
After cancel the future,  do we still need executor.awaitTermination(timeout, unit)?

rsearls: I've removed f.isDone.
We don't want to call executor.awaitTermination(timeout, unit).  If the user is using executor for other tasks we do not want canceling the EventHandler task to terminate all other task.  The user's code should call executor.awaitTermination(timeout, unit) before exiting his app.

rsearls:
@jimma I think I've addressed concerns with these changes.   What do you think?

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -255,10 +258,9 @@ public boolean close(final long timeout, final TimeUnit unit)
       internalClose();
       try
       {
-         if (!executor.awaitTermination(timeout, unit))
-         {
-            return false;
-         }
+         // mimic executor.awaitTermination giving time for tasks to complete
+         long duration = TimeUnit.MILLISECONDS.convert(timeout, unit);
+         Thread.sleep(duration);
```
jimma: Is it possible to run internalClose() with schduleExecutor.scheduleWithFixedDelay() without thread.sleep?   Is it better ?

rsearls:
@jimma I made the change requested.   I have a concern that here is no type of exception that will be thrown that allows us to catch and execute the "onErrorConsumers" block of code.   That code has been removed from this particular code change.  What are your thoughts on this?

rsearls:
@asoldano, @jimma I've made the change from array to variable, however with both solutions I am getting random failures on org.jboss.resteasy.test.providers.sse.SseSmokeTest when running TravisCI.  In some runs 1 configuration will fail.  In other runs multiple configurations will fail.  There is no consistent failure on a JDK version or wfly version.  It appears the registered event does not get a chance to run before it is cancel.  I have not found a solution to resolve this timing issue.  What would you guys suggest?

NicoNes:
@rsearls, @jimma, @asoldano I allowed myself to propose another solution to fix problem describe in the JIRA.
Let me know what you think about it: https://github.com/resteasy/Resteasy/pull/1677

asoldano:
@rsearls considering the comments above, are you ok with us going with #1677 changes and closing this PR?

rsearls:
Yes.  Jim's solution works.  Closing this PR.

PR #1665, opened 2018-08-15
RESTEASY-1983: Code review

NicoNes:


NicoNes:
Hey @ronsigal,

This PR is just a little code review for https://github.com/resteasy/Resteasy/pull/1648, in order to be consistent with what is done in initialize() method  and keep on using getters to access parent properties instead of accessing them directly. 
Since there is no reason to do things in a different way and not using getters let's fix it to prevent any misunderstanding.

ronsigal:
@NicoNes, Ok. Strange how everything is a "Registry" except "clientResponseFilters". I wouldn't change it now, though, since getClientResponseFilters() is a public method.

PR #1666, opened 2018-08-16
RESTEASY-1997: Make ResteasyProviderFactory spec compliant

NicoNes:
Jira: https://issues.jboss.org/browse/RESTEASY-1997

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
@@ -2561,18 +2561,7 @@ public class ResteasyProviderFactory extends RuntimeDelegate implements Provider
    @Override
    public boolean isEnabled(Feature feature)
    {
-      Collection<Feature> enabled = getEnabledFeatures();
-      //logger.info("********* isEnabled(Feature): " + feature.getClass().getName() + " # enabled: " + enabled.size());
-      for (Feature f : enabled)
-      {
-         //logger.info("  looking at: " + f.getClass());
-         if (f == feature)
-         {
-            //logger.info("   found: " + f.getClass().getName());
-            return true;
-         }
-      }
-      return false;
+      return getEnabledFeatures().contains(feature);
    }
 
    @Override
@@ -2603,12 +2592,7 @@ public class ResteasyProviderFactory extends RuntimeDelegate implements Provider
    @Override
    public boolean isRegistered(Class<?> componentClass)
    {
-      if (getProviderClasses().contains(componentClass)) return true;
-      for (Object obj : getProviderInstances())
-      {
-         if (obj.getClass().equals(componentClass)) return true;
-      }
-      return false;
+      return getClassContracts().containsKey(componentClass);
    }
 
    @Override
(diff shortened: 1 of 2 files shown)


asoldano:
@ronsigal please see above

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -2561,18 +2561,7 @@ public RuntimeType getRuntimeType()
...
-      {
-         //logger.info("  looking at: " + f.getClass());
-         if (f == feature)
-         {
-            //logger.info("   found: " + f.getClass().getName());
-            return true;
-         }
-      }
-      return false;
+      return getEnabledFeatures().contains(feature);
```
ronsigal: @NicoNes, hmmm. You know, I think there used to be a bug, and I think there still is. Consider
```
@Test
public void testEnabled() throws Exception {
   Feature f1 = new TestFeature();
   Feature f2 = new TestFeature();
   ResteasyProviderFactory factory = ResteasyProviderFactory.getInstance();
   factory.register(f1);
   boolean b = factory.isEnabled(f2);
   System.out.println("b: " + b);
}
```
The answer should be "true", right? But it will be false because the code, both old and new, is testing == instead of equals().

Note that the javadoc says
```
Method returns true only in case an instance equal to the feature instance is already present
```
I guess the use of the word "equal" is somewhat ambiguous, but I think it means equals(). What do you think?

-Ron

ronsigal:
@asoldano, @NicoNes,

As for isRegistered():

1. @NicoNes, am I right in thinking that your change is for efficiency, rather than correctness? I mean, the original looks correct to me. Am I missing something?

2. As for why it was done the way it was done, I can just conjecture that the classContracts map wasn't originally intended for this purpose. It was for storing priorities. But it's easier and faster to use classContracts for this new purpose. I'm thinking that it was @NicoNes who originally noticed that, but I'm not sure.

-Ron

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
NicoNes: Hi @ronsigal ,

I agree with you, the word "equal" means "equals()" for me too and that's why I changed the code to use `getEnabledFeatures().contains(feature)`.
The contains() method will first check the equality using == and then using equals() if needed.
That's what I'm testing at line https://github.com/resteasy/Resteasy/pull/1666/files#diff-60eb6107f10fea897504eba5e2a8ff10R216.
Am I missing something ?

ronsigal: @NicoNes, forget what I just said. Collection.contains() used equals(). Sorry.

The update looks good.

ronsigal: @NicoNes, ah, I missed your response.

NicoNes:
@ronsigal, @asoldano 

About isRegistered(), you're right the change is just for efficiency and nothing else.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
NicoNes: :+1:

PR #1676, opened 2018-08-28
[RESTEASY-1997] return immutable Sets as required by Configuration intf

rsearls:


rsearls:
(#1666) super seeds this PR.  Closing PR as not needed.

PR #1677, opened 2018-08-28
[RESTEASY-1987] SseEventSourceImpl assumes it owns the executor

NicoNes:


NicoNes:
This PR provides a way for SseEventSourceImpl to not close executor it does not created.
In other word it should not invoke `shutdownNow()` on an executor instance it does not created. 
All the changes done in this PR rely on the usage of Phaser in order to be able to detect when task are no longer running so that `SseEventSourceImpl .close(timeout,unit)` can keep on working as specified.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java
```diff
@@ -0,0 +1,167 @@
...
+
+   boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException
+   {
+      if (!this.closed.get())
+      {
+         return false;
+      }
+      try
+      {
+         this.phaser.awaitAdvanceInterruptibly(0, timeout, unit);
```
jimma: If shutdownNow() and phaser.arriveAndDeregister() are executed, this.phaser.awaitAdvanceInterruptibly(0, timeout, unit) always returns immediately. I don't get what's the point of adding this check.

NicoNes:
Hi @jimma,

The method  `SseEventSourceScheduler.awaitTermination(long timeout, TimeUnit unit)` follows the exact same rule as `ExecutorService.awaitTermination(long timeout, TimeUnit unit)`.
It means that it will return true only if a `SseEventSourceScheduler.shutdownNow()` has been invoked and if all tasks have completed execution before the timeout occurs.

So in our case suppose that T2 is a running thread executing line  https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R97. Since `SseEventSourceScheduler.shutdownNow()` has not been invoked yet `registrationPhase` is equal to 0.
 So T2 will arrive at line https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R108 but without executing it and will be pause for thread scheduling purpose.
Now T1 invoke `SseEventSourceScheduler.shutdownNow()` successfully. So phaser now only has 1 party, I mean the one registered by T2. And even if scheduledExecutorService.shutdownNow() has been invoked, it still has a running task: the one running in T2. 

So at this step a call to  `scheduledExecutorService.awaitTermination(long timeout, TimeUnit unit)` will never return true since T2 is not resuming.
So `SseEventSourceScheduler.awaitTermination(long timeout, TimeUnit unit)` should behave the same.
Thanks to phaser that what is done: `this.phaser.awaitAdvanceInterruptibly(0, timeout, unit)` will never return true since we are still at phase 0 because of the remaining party, the phaser is not terminated yet. It will return true once T2 will resume and finish as required.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java
NicoNes: Hi @jimma,

The method SseEventSourceScheduler.awaitTermination(long timeout, TimeUnit unit) follows the exact same rule as ExecutorService.awaitTermination(long timeout, TimeUnit unit).
It means that it will return true only if a SseEventSourceScheduler.shutdownNow() has been invoked and if all tasks have completed execution before the timeout occurs.

So in our case suppose that T2 is a running thread executing line https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R97. Since SseEventSourceScheduler.shutdownNow() has not been invoked yet registrationPhase is equal to 0.
So T2 will arrive at line https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R108 but without executing it and will be pause for thread scheduling purpose.
Now T1 invoke SseEventSourceScheduler.shutdownNow() successfully. So phaser now only has 1 party, I mean the one registered by T2. And even if scheduledExecutorService.shutdownNow() has been invoked, it still has a running task: the one running in T2.

So at this step a call to scheduledExecutorService.awaitTermination(long timeout, TimeUnit unit) will never return true since T2 is not resuming.
So SseEventSourceScheduler.awaitTermination(long timeout, TimeUnit unit) should behave the same.
Thanks to phaser that what is done: this.phaser.awaitAdvanceInterruptibly(0, timeout, unit) will never return true since we are still at phase 0 because of the remaining party, the phaser is not terminated yet. It will return true once T2 will resume and finish as required.

jimma:
```
So T2 will arrive at line https://github.com/resteasy/Resteasy/pull/1677/files
/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R108 but without 
executing it and will be pause for thread scheduling purpose.
```
Sorry. I don't understand this.  is  without executing blocked by phraser's unarrived parties or something else ? If I understand correctly, this will be executed without pause.

NicoNes: Sorry I was not clear enough. It has nothing to deal with the phaser.
What I mean is:

So T2 will arrive at line https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R108 and lets suppose that the underlying system where the program is running decides to pause this thread at this line to let other threads execute.

NicoNes: Hi @rsearls,

Yes this class supports the scenario you are talking about.

In case where scheduledExecutorService is provided, invoking `sseEventSourceScheduler.shutdownNow()` will not shutdown the scheduledExecutorService. So user's tasks in scheduledExecutorService will not be affected at all.

Actually in this case only the sse tasks will be affected:
- The current sse task running (https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R108) will be stopped since `SseEventSourceImpl.close()` close the response stream.
- The sse tasks that are scheduled but not already running will exit as soon as they start running (since the scheduledExecutorService has not been sutted down)  thanks to lines https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R93
and https://github.com/resteasy/Resteasy/pull/1677/files/f8a2bdfc85831ad269e514f1de2e9c9bbdf9678c#diff-ef11a0ff038fff5fc91418bec89c4962R104

Hope it's clearer.

rsearls: @NicoNes, @jimma Great.   Jim I'm onboard for having this merged.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java
```diff
@@ -0,0 +1,167 @@
...
+                  return;
+               }
+               int registrationPhase = SseEventSourceScheduler.this.phaser.register();
+               try
+               {
+                  // If this phaser is terminated (registrationPhase < 0)
+                  // (no more registered parties) or the current
+                  // registration phase is other than 0 it means that
+                  // shutdownNow has been called already.
+                  if (registrationPhase != 0)
```
jimma: @NicoNes  I  tried this locally and it turns out registrationPhase here can only be 0 before phaser is terminated because each job will be arriveAndDeregistered. 
In shutdownNow(),   phaser.arriveAndDeregister() can't guarantee it successfully terminates the phaser the scheduled job is running without call the phaser.arriveAndDeregiste(). 
```
 void shutdownNow()
   {
      if (this.closed.compareAndSet(false, true))
      {
         this.phaser.arriveAndDeregister();
         if (this.shutdownExecutorService)
         {
            this.scheduledExecutorService.shutdownNow();
         }
      }
   }
```
So IMO change this to phaser.forceTermination() and check phaser.isTerminated() here could be better and more guarantee. WDYT ?

jimma: @NicoNes  sorry. I still didn't get what is the actual reason for the phase can only be 0. If I understand correctly, phaser.arriveAndDeregister() can't increase the arrived party number(or phaser.getPhase()) . So in your change, there is no code to increase the arrived party( or make the phaser.getPhase() larger than 0) , right ?
Regarding the without forceTermination,  I agree with you and it's better. 
For the onAdcance() method, if it is not called(phaser is not advanced), is it a problem without running the onTerminated runnable method ?
```
this.phaser = new Phaser(1)
      {
         @Override
         protected boolean onAdvance(int phase, int registeredParties)
         {
            boolean terminated = super.onAdvance(phase, registeredParties);
            if (terminated)
            {
               if (onTerminated != null)
               {
                  onTerminated.run();
               }
            }
            return terminated;
         }
      };
   ```

Comment (NicoNes) [reply resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java]: Hey @jimma,

> I still didn't get what is the actual reason for the phase can only be 0. If I understand correctly, phaser.arriveAndDeregister() can't increase the arrived party number(or phaser.getPhase()) . So in your change, there is no code to increase the arrived party( or make the phaser.getPhase() larger than 0) , right ?

`phaser.arriveAndDeregister()` will increase the number of arrived parties for the phaser current phase and decrease the number of registered parties needed for future phases.
When all expected parties are arrived for phaser current phase, `onAdvance()` will be invoked. 
This method will say if the phaser is terminated or if it can advance to next phase.
The default `super.onAdvance()` will return true, indicating that the phaser is terminated if there is no more registered parties, else it will return false;

> For the onAdcance() method, if it is not called(phaser is not advanced), is it a problem without running the onTerminated runnable method ?

If `onAdvance()` is not called, it means that the phaser is not terminated and there are still running tasks. So no need to invoke onTerminated runnable method.

In our case as said above the `onAdvance()` will be called only once, when all registered parties are arrived at phaser current state. And since we have deregistered all arrived parties the onAdvance will return true and the phaser will be terminated

Here is a code sample to better understand:

```
		Phaser phaser = new Phaser(1) {
			@Override
			protected boolean onAdvance(int phase, int registeredParties) {
				boolean terminated = super.onAdvance(phase, registeredParties);
				if (terminated) {
					System.out.println("OnAdvance() invoked and phaser is terminated");
				} else {
					// WILL never happen
					throw new RuntimeException("OnAdvance() invoked and phaser is not terminated");
				}
				return terminated;
			}
		};
		String phaserStateTemplate = "\tRegistered parties: %s\n\tArrived parties still registered: %s\n\tUnArrived parties missing to advance to next phase: %s\n\tPhase: %s\n";
		System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
				phaser.getArrivedParties(), phaser.getUnarrivedParties(), phaser.getPhase()));

		// Following lines are to simulate the sse taks running before
		// shutdownNow method is invoked
		for (int i = 0; i < 2; i++) {
			System.out.println("Beginning of task " + i + ":");
			int currentPhase = phaser.register();
			System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
					phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));
			System.out.println("End of task " + i + ":");
			currentPhase = phaser.arriveAndDeregister();
			System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
					phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));
		}

		System.out.println("Shutdown invocation");
		// Following lines are to simulate the invocation of shutdownNow method
		int currentPhase = phaser.arriveAndDeregister();
		System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
				phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));

		System.out.println("Rejected task invocation");
		// Following lines are to simulate the sse taks running after
		// shutdownNow method is invoked
		currentPhase = phaser.arriveAndDeregister();
		System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
				phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));
	}
```

Comment (jimma) [reply resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java]: @Thanks for the code example and it better explains. I added the new line in code example to try to explain what I meant that phaser.register() after phaser.arriveAndDeregister() doesn't increase the arrived party number from my local result:
registrationPhase = SseEventSourceScheduler.this.phaser.register();
(https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97)
```
      Phaser phaser = new Phaser(1) {
          @Override
          protected boolean onAdvance(int phase, int registeredParties) {
              boolean terminated = super.onAdvance(phase, registeredParties);
              if (terminated) {
                  System.out.println("OnAdvance() invoked and phaser is terminated");
              } else {
                  // WILL never happen
                  throw new RuntimeException("OnAdvance() invoked and phaser is not terminated");
              }
              return terminated;
          }
      };
      String phaserStateTemplate = "\tRegistered parties: %s\n\tArrived parties still registered: %s\n\tUnArrived parties missing to advance to next phase: %s\n\tPhase: %s\n";
      System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
              phaser.getArrivedParties(), phaser.getUnarrivedParties(), phaser.getPhase()));

      // Following lines are to simulate the sse taks running before
      // shutdownNow method is invoked
      for (int i = 0; i < 2; i++) {
          System.out.println("Beginning of task " + i + ":");
          int currentPhase = phaser.register();
          System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
                  phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));
          System.out.println("End of task " + i + ":");
          currentPhase = phaser.arriveAndDeregister();
          
          System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
                  phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));
          //NEW ADDED & Below is the code to explain registrationPhase is always zero
          // int registrationPhase = SseEventSourceScheduler.this.phaser.register();
          System.out.println("*** phaser.register() after arriveAndDeregister *** " + phaser.register());
          System.out.println("*** phaser.getPhase() after arriveAndDeregister *** " + phaser.getPhase());
          phaser.arriveAndDeregister();
          
      }

      System.out.println("Shutdown invocation");
      // Following lines are to simulate the invocation of shutdownNow method
      int currentPhase = phaser.arriveAndDeregister();
      System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
              phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));

      System.out.println("Rejected task invocation");
      // Following lines are to simulate the sse taks running after
      // shutdownNow method is invoked
      currentPhase = phaser.arriveAndDeregister();
      System.out.println(String.format(phaserStateTemplate, phaser.getRegisteredParties(),
              phaser.getArrivedParties(), phaser.getUnarrivedParties(), currentPhase));
  }
```

> If onAdvance() is not called, it means that the phaser is not terminated and there are still running tasks. So no need to invoke onTerminated runnable method.

If I understand correctly, the onTerminated runnable method calls the all metehods registered onCompleteConsumers of SseEventSource in onAdance(). These runnables are all called after the EventSource is set to closed. Why we need to change this behavior ?

Comment (NicoNes) [reply resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java]: @jimma, happy to ear that the code sample helps. I also get your point about the registration phase that will always be 0 before invocation of `shutdownNow()`. 
So I guess that we are done on this part.

Regarding the changes about calling onCompleteConsumers in `onAdvanceMethod()`:

I could have let the code as it was and do something like:

  ```
 private void internalClose()
    {
       if (state.getAndSet(State.CLOSED) == State.CLOSED)
       {
          return;
       }
      ...
      sseEventSourceScheduler.shutdownNow();
      onCompleteConsumers.forEach(Runnable::run);    
   }
```

But, IMO, there is a problem  with this approach.
To my mind, the `onCompleteConsumers.forEach(Runnable::run);`  is supposed to be invoked once no SSE task is running anymore. In other words `onCompleteConsumers.forEach(Runnable::run);` should be invoked once `SseEventSourceScheduler.awaitTermination(long timeout, TimeUnit unit)` returns `true`.
In code above, invocation of `sseEventSourceScheduler.shutdownNow()` does not guarantee for sure that there is no sse task running anymore. So we should not invoke `onCompleteConsumers.forEach(Runnable::run); ` here.
Instead if done in `onAdvance()` once `super.onAdvance()` return `true`,  we can say for sure that no sse task is/will be running anymore.
WDYT?

jimma: >@jimma, happy to ear that the code sample helps. I also get your point about the registration phase that will always be 0 before invocation of shutdownNow().
So I guess that we are done on this part.

I am happy too that I finally make me understood. :-)   Let me give my more thoughts about this topic : 
 If phase that will always be 0 before invocation of shutdownNow(), does that mean this line can't be reached in any case :
``` 
 if (registrationPhase != 0) {
   return;
}
```
considering there is already closed check before this  and only shutdownNow can change registrationPhase to a negative value
```
     if (SseEventSourceScheduler.this.closed.get())
               {
                  return;
               }
```
From spec, onCompleteConsumers.forEach(Runnable::run)  should be invoked when there is no events received .  
I think we call it after setting closed state should be enough.   
https://github.com/eclipse-ee4j/jaxrs-api/blob/master/jaxrs-api/src/main/java/javax/ws/rs/sse/SseEventSource.java#L191

NicoNes: Let's consider 2 cases with two threads T1 and T2:

- 1st case: 
T1  invokes `SseEventSourceScheduler.shutdownDown()` in one go without being interrupted by the underlying system and exits.
T2 (thread that will run the sse task) performs the closed check as you said and then exits since closed==true.
So in this case, you're right the check on `registrationPhase` is useless.

- 2nd case:
T1 (thread that will run the sse task) performs the closed check. Since `SseEventSourceScheduler.shutdownDown()`  has not been invoked yet, closed=false and T1 is allowed to continue. At the discretion of the underlying system where the java code is running, T1 is paused juste before running line (https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97) so that other threads can run.
T2  invokes `SseEventSourceScheduler.shutdownDown()` in one go without being interrupted by the underlying system and exits.
T1 is 'waken up' by the underlying system and line https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97 is executed.  At this moment, without any check on registration phase T1 would have been allowed to go ahead even if the phaser is terminated => this is not what we want.
So thank to the check on `registrationPhase` T1 will exit since `registrationPhase<0` because phaser is terminated.

So, yes, the second check is useful as seen above in case 2.

>I think we call it after setting closed state should be enough.

I will make the change. Thanks

jimma: >2nd case:
T1 (thread that will run the sse task) performs the closed check. Since SseEventSourceScheduler.shutdownDown() has not been invoked yet, closed=false and T1 is allowed to continue. At the discretion of the underlying system where the java code is running, T1 is paused juste before running line (https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97) so that other threads can run.
T2 invokes SseEventSourceScheduler.shutdownDown() in one go without being interrupted by the underlying system and exits.
T1 is 'waken up' by the underlying system and line https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97 is executed. At this moment, without any check on registration phase T1 would have been allowed to go ahead even if the phaser is terminated => this is not what we want.
So thank to the check on registrationPhase T1 will exit since registrationPhase<0 because phaser is terminated.

Can you explain why T1 is paused and resumed ?

NicoNes: Because, for any reason depending on the undelrying system (a higher priority thread becomes runnable, the time allotment reached for a given thread...).  a thread execution can be paused (and then resume) at any time during its execution for thread scheduling purpose in order to let other threads run.

NicoNes: If you're not confortable with pause/resume, let's say that T1 and T2 are racing together.
T1 pass the check close and just before executing line https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97, T2 at the same time invokes `phaser.arriveAndDeregister()` in `SseEventSourceScheduler.shutdownDown()`.
T1 executes line https://github.com/resteasy/Resteasy/pull/1677/files#diff-ef11a0ff038fff5fc91418bec89c4962R97. At this moment, without any check on registration phase T1 would have been allowed to go ahead even if the phaser is terminated => this is not what we want.
So thank to the check on registrationPhase T1 will exit since registrationPhase<0 because phaser is terminated.

Globaly when multiple threads are racing, you cannot know in advance the way thread execution will be scheduled by the system (unless using specific mechanism such as mutual exclusion for instance) so we have to take all cases into account.

jimma: @NicoNes  Thanks for clarification, and what you meant is cpu time/OS thread scheduling, not the java thread thing. From my understanding, Phaser or Phaser.register() is thread safe. Am I wrong ?

NicoNes: >and what you meant is cpu time/OS thread scheduling, not the java thread thing

Yes exactly, that's it.

>From my understanding, Phaser or Phaser.register() is thread safe. Am I wrong ?

Yes you're rigth, T1 and T2 can share and use the same phaser instance safely

NicoNes: @jimma I've just made the change about completion listeners notification.

PR #1683, opened 2018-09-03
[RESTEASY-2002] Move towards WildFly 14

rsvoboda:
Move towards WildFly 14
https://issues.jboss.org/browse/RESTEASY-2002


rsvoboda:
@jbliznak is already working on fix to address the lock issue, so will wait for his fix.

asoldano:
@rsvoboda the issues should now be solved by the RESTEASY-2003 fix. Merging

PR #1684, opened 2018-09-03
[RESTEASY-2003] Isolate container standalone dir in tests

jbliznak:
https://issues.jboss.org/browse/RESTEASY-2003

To isolate the instance specific stuff I had to introduce few more steps for copying the standalone directory and retarget few plugin runs to different phases so that all is executed in proper order.

### DIFF
--- a/testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/utils/LogCounter.java
+++ b/testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/utils/LogCounter.java
@@ -20,16 +20,27 @@ public class LogCounter {
     private boolean onServer;
 
 
-    public LogCounter(String message, boolean onServer) {
+    /**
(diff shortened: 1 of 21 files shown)


jbliznak:
We are not copying whole server (WF_HOME) here, just the standalone dir (WF_HOME/standalone). For vanilla distribution that was never started it is under 1MB so I guess this is ok.

Sure, we can introduce alternative solution, for WF14 the minimum seems to be own standalone/data directory (-Djboss.server.data.dir) per instance because that is the place where the lock files is  ++ own standalone-full.xml ++ specify instance specific log file or log dir (-Djboss.server.log.dir). This could save some space in exchange for more complex configuration in pom.xml.

Given we didn't even need the isolated data dir for WF13-, who knows what can change in future releases. I believe I chose the second safest way (the first being copying the whole server distribution).

PR #1690, opened 2018-09-06
JDK11 java.se module and WF14 upstream testing

marekkopecky:
**https://issues.jboss.org/browse/RESTEASY-2004**

Add java.se module to JDK11 testsuite runs

I add the same jdk11 args as common.sh script: https://github.com/wildfly/wildfly-core/blob/master/core-feature-pack/src/main/resources/content/bin/common.sh#L9-L12

This PR remove modular.jdk.props, this property is currently not used in WF: https://github.com/wildfly/wildfly/blob/master/pom.xml#L6985-L6997

master PR: https://github.com/resteasy/Resteasy/pull/1689

----

**https://issues.jboss.org/browse/RESTEASY-2002**

Move towards WildFly 14

master PR: https://github.com/resteasy/Resteasy/pull/1683

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -115,7 +115,7 @@
               </property>
             </activation>
             <properties>
-                <server.version>12.0.0.Final</server.version>
+                <server.version>14.0.0.Final</server.version>
             </properties>
         </profile>
         <profile>
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -142,6 +142,21 @@
                 <version.resteasy.testsuite>${version.resteasy.testsuite}</version.resteasy.testsuite>
             </properties>
         </profile>
+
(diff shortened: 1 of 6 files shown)


testsuite/pom.xml
```diff
@@ -142,6 +142,21 @@
                 <version.resteasy.testsuite>${version.resteasy.testsuite}</version.resteasy.testsuite>
             </properties>
         </profile>
+
+        <!-- https://issues.jboss.org/browse/MODULES-372 - JDK 11 - module not found java.se -->
+        <profile>
+            <id>modular-jdk</id>
+            <activation>
+                <jdk>[9,)</jdk>
```
rsvoboda: Is it really Java 9+ or 11+ ?

jbliznak: We in sync with WF, https://github.com/wildfly/wildfly/blob/master/pom.xml#L6985-L6997

PR #1691, opened 2018-09-06
[RESTEASY-2012] Getting rid of old Apache HttpClient 4.0 support, def…

asoldano:
…aulting to closing httpclient when the engine is closed

Documentation changes still todo


rsvoboda:
Do we have migration guide from 3.x to 4.x series ?
Time to write down some noteworthy items ;)

asoldano:
There's something on the doc regarding migration already, I need to update that too (will do before merging this pr)

asoldano:
Apache HTTP client engine documentation updated; as for the migration guide, I've created https://issues.jboss.org/browse/RESTEASY-2014 .

PR #1692, opened 2018-09-06
[RESTEASY-2013] Move stats to a separate module

asoldano:


pom.xml
```diff
@@ -109,6 +109,7 @@
                 <module>resteasy-jsapi</module>
                 <module>resteasy-cdi</module>
                 <module>resteasy-servlet-initializer</module>
+                <module>resteasy-stats</module>
```
rsvoboda: How will be this change mapped to wildfly modules ?
If this change will get to WF one day ...

asoldano: Users will pull the dependency in their project. No need to ship those classes in WildFly.

PR #1694, opened 2018-09-11
[RESTEASY-1916]:Fix InboundSseEventImpl doesn't getReader from Client…

jimma:
…Configuration


jimma:
@asoldano I tried and it doesn't work. This is due to before we read response to an InboundSseEventInputImpl (line 4), the context providers is already poped (line 1) in SseEventSourceimpl:
```           
 1  response = (ClientResponse) request.invoke();
 2  if (Family.SUCCESSFUL.equals(response.getStatusInfo().getFamily()))
 3      {   onConnection();
 4           eventInput = response.readEntity(SseEventInputImpl.class);
```  
The direct and simple way is manually setting this providers.

PR #1710, opened 2018-09-24
PatchMethodFilter doesn't handle request of MediaType application/jso…

benjamin-chapoton:
…n-patch+json if MediaType have argument


asoldano:
@benjamin-chapoton , we require that all contributions be made under the terms of the ASL
Do you agree to these terms?

benjamin-chapoton:
> 
> 
> @benjamin-chapoton , we require that all contributions be made under the terms of the ASL
> 2.0 License: http://www.apache.org/licenses/LICENSE-2.0
> Do you agree to these terms?

Yes I agree, thanks for your reactivity.

PR #1715, opened 2018-09-26
[RESTEASY-2016] updated message format

rsearls:


asoldano:
@marekkopecky would this be ok / enough to you? Also considering the requirements on the size of the logging in case of errors.

marekkopecky:
@asoldano Basically, this PR is just copy-paste diff from description of [RESTEASY-2016](https://issues.jboss.org/browse/RESTEASY-2016). As I mentioned in description of [RESTEASY-2016](https://issues.jboss.org/browse/RESTEASY-2016), we need to consider more use-cases.

Currently, [jsonBDeserializationError](https://github.com/resteasy/Resteasy/pull/1715/files#diff-8c175b8ad2c04da2724fd3bc3d9bc9afR17) contains `@Cause Throwable element`, [jsonBSerializationError](https://github.com/resteasy/Resteasy/pull/1715/files#diff-8c175b8ad2c04da2724fd3bc3d9bc9afR20) not, we should probably unified that? We may consider to print stacktrace in DEBUG (in both server/client use-cases), WDYT?

I guess that if we pass "caused by" exception to ProcessingException, the stacktrace will be printed by [this](https://github.com/resteasy/Resteasy/blob/master/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java#L196) on both server and client side, right? (I'm not sure now, I can check it tomorrow) Or would be better to just print the stacktrace in debug right in JsonBindingProvider?

Main motivation is to prevent situations like [this](https://developer.jboss.org/message/984944#984923)

asoldano:
@rsearls please see / follow up on latest comments from @marekkopecky

rsearls:
"unify logging of exceptions for server/client side and writeTo/readFrom use-cases"
This is no trivial task.  Writing the tests to force different exceptions in order
to evaluate the existing exception messages on server/client and to test generation 
of revised messages is a major non-trivial task.  Our tests currently don't test for 
this negative scenario.  We would also need to create a mechanism to extract and 
evaluate the message written to the server.log.  

If this work is to be performed I would recommend a design document to detail 
the scope of the work. Guidelines for the message format.  Risk assessment of
breaking backward compatibility. Guidelines for the testsuite tests.  Recommendations 
for the utility to compare the control message to the generated message in the 
server.log.  Once the scope of the work is understood, an implementation schedule 
could be proposed.

asoldano:
@marekkopecky @rsearls , how do we move forward from here?

asoldano:
Having heard nothing, I've merged this. If additional improvements are expected, please follow-up on jira / through a new PR. Thanks.

PR #1716, opened 2018-09-26
[RESTEASY-1142] ExceptionHandler: catch IllegalStateException from ClientResponse

lyind:
Fix for the following issue: https://issues.jboss.org/browse/RESTEASY-1142

The resteasy-client's ClientResponse throws an IllegalStateException when the ExceptionMapper calls ClientResponse.getEntity().

This issue is a blocker for my use-case.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
@@ -165,13 +165,14 @@ public class ExceptionHandler
 
       if (unwrappedException instanceof WebApplicationException) {
          WebApplicationException wae = (WebApplicationException) unwrappedException;
-         if (wae.getResponse() != null) {
+         Response response = wae.getResponse();
+         if (response != null) {
             try {
-               if (wae.getResponse().getEntity() != null) return wae.getResponse();
+               if (response.getEntity() != null) return response;
             }
             catch(IllegalStateException ise) {
-               // ignore
                // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
+               return response;
             }
          }
       }
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
@@ -165,7 +165,15 @@ public class ExceptionHandler
 
       if (unwrappedException instanceof WebApplicationException) {
          WebApplicationException wae = (WebApplicationException) unwrappedException;
-         if (wae.getResponse() != null && wae.getResponse().getEntity() != null) return wae.getResponse();
+         if (wae.getResponse() != null) {
+            try {
+               if (wae.getResponse().getEntity() != null) return wae.getResponse();
+            }
+            catch(IllegalStateException ise) {
+               // ignore
+               // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
+            }
+         }
       }
 
       jaxrsResponse = executeExceptionMapper(unwrappedException, logger);
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
@@ -172,7 +172,6 @@ public class ExceptionHandler
             }
             catch(IllegalStateException ise) {
                // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
-               return response;
             }
          }
       }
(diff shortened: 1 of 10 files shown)


asoldano:
@ronsigal , you self assigned RESTEASY-1142 last year, can you please evaluate this PR for it? Thanks!

asoldano:
@lyind thanks for your PR. While we review it, we require that all contributions be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0
Do you agree to these terms?

lyind:
@asoldano Yes, please accept my PR under the terms of the ASL 2.0 license.

@ronsigal Regarding the PR's background:

ClientResponse throwing IllegalStateException in the "isClosed == true" case is expected behavior.
See: https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.html#hasEntity--

We are already using a RESTEasy build based on Beta5 with the patch applied.

asoldano:
@ronsigal ping :-)

lyind:
@ronsigal Thank you for your suggestions, you are right.

There was more work required to complete the fix and catch all instances of this issue. Please review my addition to the PR.

I also converted the provided test to a standard integration test `ClosedResponseHandling` which tests all the changes.

resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
```diff
@@ -165,7 +165,16 @@ protected Response unwrapException(HttpRequest request, Throwable e, RESTEasyTra
...
          WebApplicationException wae = (WebApplicationException) unwrappedException;
-         if (wae.getResponse() != null && wae.getResponse().getEntity() != null) return wae.getResponse();
+         Response response = wae.getResponse();
+         if (response != null) {
+            try {
+               if (response.getEntity() != null) return response;
+            }
+            catch(IllegalStateException ise) {
+               // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
+               return response;
```
ronsigal: @lyind, it seems to me that the flow of control should drop through here, rather than returning the response. The spec (3.3.4 Exceptions, item 1) says
```
If the response property of the exception does not contain an entity and an exception mapping
provider (see Section 4.4) is available for WebApplicationException or the corresponding sub-
class, an implementation MUST use the provider to create a new Response instance, otherwise the
response property is used directly.
```

If the Response is closed, then, in effect, it doesn't have an entity.  Falling through gives us a chance to check if "an exception mapping provider (see Section 4.4) is available for WebApplicationException or the corresponding sub-class".

-Ron

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/exception/ClosedResponseHandlingTest.java
```diff
@@ -0,0 +1,160 @@
...
+    * @tpPassCrit A NotSupportedException is returned
+    * @tpSince RESTEasy 4.0.0.CR1
+    */
+   @Test(expected = NotSupportedException.class)
+   public void testNotSupportedTraced() {
+      new ResteasyClientBuilderImpl().build().target(generateURL("/testNotSupportedTraced")).request().get(String.class);
+   }
+
+   @Path("")
+   public static class TestResource {
```
ronsigal: Hi @lyind.

The convention in resteasy-integration-tests is to put supporting classes like TestResource, PleaseMapException, and PleaseMapExceptionMapper in the related .../resource directory, org.jboss.resteasy.test.exception.resource in this case. The names would be related to the test class name; e.g., ClosedResponseHandlingResource, ClosedResponseHandlingPleaseMapException, etc.

ronsigal:
@lyind, my general reaction is that this is definitely better now. I have a couple of observations.

1. There are 14 references to Response.getEntity() in resteasy-core. To really do a complete treatment, all of them should be examined.
2. In principle, there should be a test case to verify each of these corrections. Some of them might be pretty hard to create, but at least the common ones should be tested.

Again, thank you very much for your contributions.
-Ron

resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
lyind: Right, corrected in bd16649e45b4177a61a6f6c4d882b8002b66b584.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/exception/ClosedResponseHandlingTest.java
lyind: Layout re-factored in dd2ebd4f16c8c6994a0fcfe6c877d1a840f1891c.

ronsigal:
Hi @lyind,

I'm convinced. I tried to find some loopholes, but I couldn't. ;-)

Well done!!

-Ron

PR #1717, opened 2018-10-01
[RESTEASY-891] added check for user provided UnhandledException mapper

rsearls:
If we choose to support this requested feature here is the patch.  There is not good way to provide a test case for this.  It is a low risk fix because it is a very specific situation.  Normal handling continues if the user does not provide a mapper for this.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
@@ -264,47 +264,69 @@ public class ExceptionHandler
       // See if there is an ExceptionMapper for the exact class of the exception instance being thrown
       if ((jaxrsResponse = executeExactExceptionMapper(e, logger)) != null) return jaxrsResponse;
 
-      // These are wrapper exceptions so they need to be processed first as they map e.getCause()
-      if (e instanceof ApplicationException) {
-         return handleApplicationException(request, (ApplicationException) e, logger);
-      } else if (e instanceof WriterException) {
-         return handleWriterException(request, (WriterException) e, logger);
-      } else if (e instanceof ReaderException) {
-         return handleReaderException(request, (ReaderException) e, logger);
-      }
-
-      /*
-       * 					If the response property of the exception does not
-       * 					contain an entity and an exception mapping provider
-       * 					(see section 4.4) is available for
-       * 					WebApplicationException an implementation MUST use the
-       * 					provider to create a new Response instance, otherwise
-       * 					the response property is used directly.
-       */
-      if (e instanceof WebApplicationException) {
-         WebApplicationException wae = (WebApplicationException) e;
-         if (wae.getResponse() != null && wae.getResponse().getEntity() != null) {
-            Response response =  wae.getResponse();
-            return response;
-         }
-      }
-
-      // First try and handle it with a mapper
+      try
       {
-         jaxrsResponse = executeExceptionMapper(e, logger);
-         if (jaxrsResponse != null) {
+         // These are wrapper exceptions so they need to be processed first as they map e.getCause()
+         if (e instanceof ApplicationException)
+         {
+            return handleApplicationException(request, (ApplicationException) e, logger);
+         } else if (e instanceof WriterException)
+         {
+            return handleWriterException(request, (WriterException) e, logger);
+         } else if (e instanceof ReaderException)
+         {
+            return handleReaderException(request, (ReaderException) e, logger);
+         }
+
+         /*
+          * 					If the response property of the exception does not
+          * 					contain an entity and an exception mapping provider
+          * 					(see section 4.4) is available for
+          * 					WebApplicationException an implementation MUST use the
+          * 					provider to create a new Response instance, otherwise
+          * 					the response property is used directly.
+          */
+         if (e instanceof WebApplicationException)
+         {
+            WebApplicationException wae = (WebApplicationException) e;
+            if (wae.getResponse() != null && wae.getResponse().getEntity() != null)
+            {
+               Response response = wae.getResponse();
+               return response;
+            }
+         }
+
+         // First try and handle it with a user provided mapper for the exception class
+         {
+            jaxrsResponse = executeExceptionMapper(e, logger);
+            if (jaxrsResponse != null)
+            {
+               return jaxrsResponse;
+            }
+         }
+
+         // Otherwise do specific things
+         if (e instanceof WebApplicationException)
+         {
+            return handleWebApplicationException((WebApplicationException) e);
+         } else if (e instanceof Failure)
+         {
+            return handleFailure(request, (Failure) e);
+         }
+
+         LogMessages.LOGGER.unknownException(request.getHttpMethod(), request.getUri().getPath(), e);
+         throw new UnhandledException(e);
+
+      } catch (UnhandledException une) {
+
+         // Allow user a chance to handle an unidentified error with their custom
+         // ExceptionMapper<UnhandledException>
+         jaxrsResponse = executeExceptionMapperForClass(une, UnhandledException.class, logger);
+         if (jaxrsResponse != null)
+         {
             return jaxrsResponse;
          }
+         throw une;
       }
-
-      // Otherwise do specific things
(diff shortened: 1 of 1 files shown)


domtoupin:
@asoldano I took a look at the patch and this does not fit with what I had in mind. From what I understand, the patch would allow to handle the exception as if it was an UnhandledException; however, at that point the exception was not wrapped yet in an UnhandledException so I believe this may not work at all...

What we want to achieve is to handle all Exceptions with an ExceptionMapper except those inheriting from WebApplicationException and Failure because these are already properly handled by RESTEasy... Our biggest issue is that because a 204 is returned in executeExceptionMapper when the Exception is not handled (i.e. null is returned), then we are not able to defers the handling of WebApplicationException and Failure to RESTEasy...

rsearls:
I've restructured this such that Resteasy can throw an UnhandledException
and the user by providing a ExceptionMapper<UnhandledException> can evaluate
it and return a response object or null.  When no response object is provided the
original UnhandledException is rethrown.

domtoupin:
@rsearls @asoldano The updated version of the PR is an improvement but does not meet our requirements.

We are using RESTEasy in an agnostic way, i.e. with access only to the JSR311 api classes. We cannot therefore throw an `UnhandledException` in our `ExceptionMapper` because that class is not part of JSR311 but rather a class specific to RESTEasy.

At this point, I see two solutions:
1. Remove this code block from `protected Response executeExceptionMapper(Throwable exception, RESTEasyTracingLogger logger)` (lines 136-138)
```java
          if (jaxrsResponse == null) {
             jaxrsResponse = Response.status(204).build();
          }
```
This would allow one to let exceptions such as `WebApplicationException` or RESTEasy `Failure` go through a generic ExceptionMapper without having to deal with them.
2. Instead of catching `UnhandledException`, attempt to catch `Exception`or some other exception class available in the JRE or JSR311.

rsearls:
As I understand your jira request it is to enable your code to handle any
exception for which Resteasy provides no builtin ExceptionMapper and for which 
the user code has provided no custom ExceptionMapper.  This PR allows the
user code first chance to create a response for an unknown exception before 
Resteasy throws an UnhandledException and the way that is done is by the
user providing a custom ExceptionMapper<UnhandledException>.

Resteasy does not provide any builtin ExceptionMappers.  This means in the scenario
you describe there must be a custom ExceptionMapper that has been register and
that mapper is not returning a Response object.  It is returning null.

Lets look more closely at the code.
lines 122-127

     Class causeClass = exception.getClass();
      while (mapper == null) {
         if (causeClass == null) break;
         mapper = providerFactory.getExceptionMappers().get(causeClass);
         if (mapper == null) causeClass = causeClass.getSuperclass();
      }

Lets consider this contrived example of a exception class' "cause" tree.

    cause: java.io.IOException
    cause: java.security.GeneralSecurityException
    cause: java.lang.Exception
    cause: java.lang.Throwable
    cause: java.lang.Object

IOException is the initial exception.  Resteasy already performed a lookup
for a mapper for this class and did not found one, so now it proceeds to look
for a mapper for superClasses GeneralSecurityException, Exception, Throwable
and Object in that order.  When Resteasy does not find a mapper for any of the 
causes in the tree it returns null, allowing exception evaluation to continue.  
But when a mapper is found, it is processed, lines 129-140.

In your scenario a custom mapper was found for one of the "cause" class types.
Line 133 the mapper is called and some value is returned.
Line 136 the returned value is checked.  The value is determined to be null.
By the mapper returning null it is indicating that "No Content" is being returned to the caller, a 204 status.
This is in accordance with the spec section 3.3.3.

      if (mapper != null) {
         mapperExecuted = true;

         final long timestamp = logger.timestamp("EXCEPTION_MAPPING");
         Response jaxrsResponse = mapper.toResponse(exception);
         logger.logDuration("EXCEPTION_MAPPING", timestamp, mapper, exception, exception.getLocalizedMessage(), jaxrsResponse);

         if (jaxrsResponse == null) {
            jaxrsResponse = Response.status(204).build();
         }
         return jaxrsResponse;
      }

So the question becomes, in your application whose exception mapper is returning 
a null Response object and why.  As I mentioned earlier Resteasy does not register 
any builtin mappers.  I'm not aware of any 3rd party components your code might 
be using that register their own ExceptionMappers but lets assume that is happening.  
You can provide a ExceptionMapper that takes priority over another ExceptionMapper 
of the same type by adding a @Priority annotation to your ExceptionMapper provider.  
Unfortunately you are going to have to identify what ExceptionMapper<Clazz> is 
returning a null response in order know what you need to create with a higher priority.

rsearls:
@domtoupin
Why would you need an ExceptionMapper for every single class extending Exception?
UnhandledException is a wrapper class around any exception Resteasy does not 
identify or handle.  You just need one ExceptionMapper<UnhandledException> provider.
There are 2 places in class ExceptionHandler that create a new instance of
UnhandledException at line 103 in method handleApplicationException and
line 318 in method handleException.  Note that in both cases the Thrown
"cause" is registered in UnhandledException.  Your toResponse method input param 
sees a single class UnhandledException that provides access to the full exception
"cause" tree.  Your code can process the stack just as is done by lines 122-127
in our previous discussion.

Code lines 320-330 catch either of the thrown UnhandledException (above) and allows
your impl first crack at processing it.  If your UnhandledException mapper returns 
null, then Resteasy rethrows the original UnhandledException.

    } catch (UnhandledException une) {

             // Allow user a chance to handle an unidentified error with their custom
             // ExceptionMapper<UnhandledException>
             jaxrsResponse = executeExceptionMapperForClass(une, UnhandledException.class, logger);
             if (jaxrsResponse != null)
             {
                return jaxrsResponse;
             }
             throw une;
    }

As to postponing the generation of a 204 response, I don't think that would be
a proper solution.  You need to determine what mapper in your app is returning
null and why.

domtoupin:
@rsearls We know exactly which `ExceptionMapper` returns a null and why; we created our own `ExceptionMapper` to handle `Exception` and let the response be null if the original exception is a `WebApplicationException` (tested using `instanceof` since that class is known at compile time) or a Resteasy `Failure` (tested using `ExceptionMapper.class.getClassLoader().loadClass("org.jboss.resteasy.spi.Failure").isInstance(exc)` because `org.jboss.resteasy.spi.Failure` is only known at runtime).

The reason we are not on board with the current solution is that it requires us to add a compile time requirement on a Resteasy class (`UnhandledException`) which would defeat our goal of sticking to the JSR311 API.

I thought about this overnight and I am wondering the following:

You previously mentioned that Resteasy doesn't provide any built-in `ExceptionMapper` to which I said this was something we were aware of and we were also aware that `WebApplicationException` and `Failure` had special treatments in Resteasy if they were not previously handed to an `ExceptionMapper`; what if `handleFailure` and `handleWebApplicationException` were turned into actual built-in `ExceptionMapper`? That would work with us because then `WebApplicationException` and Failure` would be dealt with before reaching our generic `ExceptionMapper` since `WebApplicationException` and `Failure` are "specialized" `Exception`...

rsearls:
@domtoupin
Providing builtin mappers for Failure and WebApplicationException that implement
the code for handleFailure and handleWebApplicationException respectively would be
an elegant solution but I have not been able to make it work.  The biggest snag
is handleWebApplicationException's instanceof check for 
"(!(wae instanceof NoLogWebApplicationException))".  There is no way to check
for that in a mapper class because there is no access to subclasses of
input type WebApplicationException in the toResponse method.

What I can do is refactor method handleException so that WebApplicationException
and Failure exceptions are evaluated and processed before other exception types.

That still leaves 2 hurdles, the 204 response object and the 
ExceptionMapper<UnhandledException> requirement.  I might be allowed to 
check for a 204 response and postpone returning it until your mapper 
has a chance to evaluate the exception.  I would need to discuss further with
the team the spec compliance issues this might introduce.  I don't see how to
get around the requirement for a custom ExceptionMapper<UnhandledException>.
I know of no spec compliant class that allows Resteasy to hand your mapper the
root exception.  Barring that, you would need to provide a mapper for every
conceivable class type.

If you absolutely can't have a ExceptionMapper<UnhandledException> in your
code I don't see that the other proposed changes will be of any help.  Should
I proceed with this or abandon the effort?

rsearls:
Very old and obsolete

PR #1721, opened 2018-10-05
[RESTEASY-1996] Added default ParamConverter for multiple element

ronsigal:
params.


ronsigal:
Don't merge yet: I want to add a couple more tests.

ronsigal:
@marekkopecky, @NicoNes: you guys might be interested in this. I've created a default ParamConverter for multiple element parameters, as per @marekkopecky's request.

docbook/reference/en/en-US/modules/StringConverter.xml
```diff
@@ -665,5 +665,152 @@ public static class TestResource {
...
+    <sect1>
+   <title>Default multiple valued <classname>ParamConverter</classname></title>
    
+   <para>
+   RESTEasy includes two built-in <classname>ParamConverter</classname>s in the resteasy-jaxrs module,
+   one for <classname>Collection</classname>s:
+   </para>
+   
+<programlisting>
+   <classname>org.jboss.resteasy.plugins.providers.MultiValuedArrayParamConverter</classname>,
```
NicoNes: You mean `org.jboss.resteasy.plugins.providers.MultiValuedCollectionParamConverter` right ?

docbook/reference/en/en-US/modules/StringConverter.xml
```diff
@@ -665,5 +665,152 @@ public static class TestResource {
...
+</programlisting>
+
+   <para>
+   and one for arrays:
+   </para>
+   
+<programlisting>
+   <classname>org.jboss.resteasy.plugins.providers.MultiValuedCollectionParamConverter</classname>,
```
NicoNes: You mean `org.jboss.resteasy.plugins.providers.MultiValuedArrayParamConverter` right ?

NicoNes:
Hi @ronsigal @marekkopecky ,

I've just started reading the `StringConverter.xml` for now and it looks really interesting to me. I'll let you know me feedback for sure. :+1:

docbook/reference/en/en-US/modules/StringConverter.xml
```diff
@@ -665,5 +665,152 @@ public static class TestResource {
...
+!"#$%&amp;'()*+,-./:;&lt;=&gt;?@[\]^_`{|}~
+</programlisting>
+
+      or a class of punctuation symbols like "[-,;]".
+      </listitem>
+      
+      <listitem>
+      For either of these <classname>ParamConverter</classname>s to be available for a given parameter, that
+      parameter must be annotated with <classname>@Separator</classname>.
+      </listitem>
```
NicoNes: According to what you've said above, the annotation `@Separator` can also target a method `@Target({ElementType.METHOD, ElementType.PARAMETER})`. 
Can you please clarify, what does this annoation mean when just annotating a method ?

docbook/reference/en/en-US/modules/StringConverter.xml
ronsigal: Yep. Thanks. :-P

docbook/reference/en/en-US/modules/StringConverter.xml
ronsigal: And yep.

docbook/reference/en/en-US/modules/StringConverter.xml
ronsigal: @NicoNes, it was a careless cut-and-paste. Thanks, again.

docbook/reference/en/en-US/modules/StringConverter.xml
```diff
@@ -665,5 +665,152 @@ public static class TestResource {
...
+   to transform <classname>Collection</classname>s or arrays into a string representation.
+   </para>
+   
+   <para>
+   String representations are determined by <classname>org.jboss.resteasy.annotations.Separator</classname>,
+   a parameter annotation in the resteasy-jaxrs module:
+   </para>
+   
+<programlisting>
+@Target({ElementType.METHOD, ElementType.PARAMETER})
```
NicoNes: You forgot to remove the `ElementType.METHOD ` here

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,189 @@
...
+ */
+@Provider
+@Priority(Integer.MAX_VALUE)
+public class MultiValuedArrayParamConverter implements ParamConverter<Object>
+{
+   private Class<?> rawType;
+   private StringParameterInjector stringParameterInjector;
+   private String separator;
+
+   public MultiValuedArrayParamConverter(Class<?> rawType, Annotation[] annotations, StringParameterInjector stringParameterInjector, String separator)
```
NicoNes: Quick clean up for `annotations `parameter that is not used here

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+@Provider
+@Priority(Integer.MAX_VALUE)
+public class MultiValuedCollectionParamConverter implements ParamConverter<Collection<?>> 
+{
+   private Constructor<?> constructor;
+   private StringParameterInjector stringParameterInjector;
+   private String separator;
+
+   public MultiValuedCollectionParamConverter(Class<?> rawType, Annotation[] annotations, Constructor<?> constructor, StringParameterInjector stringParameterInjector, String separator)
```
NicoNes: Quick clean up for both `rawType` and `annotations` parameters that are not used here

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,189 @@
...
+         throw new RuntimeException(Messages.MESSAGES.expectedArray(value.getClass().getName()));
+      }
+      Object[] values = new Object[Array.getLength(value)];
+      if (values.length == 0)
+      {
+         return "";
+      }
+      if (rawType.getComponentType().isPrimitive())
+      {
+         copyPrimitiveArray(values, value);
```
NicoNes: If I understand well, `copyPrimitiveArray(...)` method is to filter array elements whose type is not equal to the type that this param converter can handle (`rawType.getComponentType()`) right ?

But if I'm right in which situation such case can happen ?

For example if I have a proxy interface like that:
```
public interface Client{
	Response getItems(int[]itemsId);
}
```
When user will use proxy instance to invoke `getItems(int[]itemsId)` he will only be able to provide a int[] parameter not an Object nor Object[]. So filtering done by the param converter will be useless. 

And by the way if we are doing this kind of filtering for primitive type we have to do it for Object type as well.

As I said above I don't know in which situation this case can happen but if we want to add a check for it, the same way we are checking for parameter validity by verifying that it is an array as expected, will it not be good to also test for the component type ?
```

else if (rawType.getComponentType().isAssignableFrom(value.getClass().getComponentType()))
{
 throw new RuntimeException()
}
```

What do you think? Am I missing something?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,189 @@
...
+      try
+      {
+         for (int i = 0; i < elements.length; i++)
+         {
+            Array.set(array, i, stringParameterInjector.extractValue(elements[i]));
+         }
+      }
+      catch (Exception e)
+      {
+         throw new RuntimeException("Ooops");
```
NicoNes: Ooops you forgot to change this one :D

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,189 @@
...
+         {
+            saveAsString = values[0].getClass().getMethod("saveAsString");
+         }
+         catch (Exception e)
+         {
+            // ok
+         }  
+      }
+      char separatorChar = getSeparatorChar();
+      StringBuffer sb = new StringBuffer();
```
NicoNes: Better use a StringBuilder instance here

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,189 @@
...
+   @SuppressWarnings({ "rawtypes", "unchecked" })
+   private String stringify(Object[] values)
+   {
+      ParamConverter converter = stringParameterInjector.getParamConverter();
+      Method saveAsString = null;
+      if (converter == null)
+      {
+         try
+         {
+            saveAsString = values[0].getClass().getMethod("saveAsString");
```
NicoNes: Better look for a `stringParameterInjector.getHeaderDelegate()` before looking for `saveAsString()` method. This way, in case the parameter is a `@HeaderParam` and a `HeaderrDelegate<T>` has been registered for it let's use it.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,189 @@
...
+import org.jboss.resteasy.core.StringParameterInjector;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
+
+/**
+ * @author Marek Kopecky mkopecky@redhat.com
+ * @author Ron Sigal rsigal@redhat.com
+ */
+@Provider
+@Priority(Integer.MAX_VALUE)
+public class MultiValuedArrayParamConverter implements ParamConverter<Object>
```
NicoNes: What about ParamConverter<Object[]> here instead of ParamConverter<Object> ?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
```diff
@@ -0,0 +1,139 @@
...
+            Object o = value.iterator().next();
+            saveAsString = o.getClass().getMethod("saveAsString");
+         }
+         catch (Exception e)
+         {
+            // ok
+         }  
+      }
+      char separatorChar = getSeparatorChar();
+      StringBuffer sb = new StringBuffer();
```
NicoNes: Better use a StringBuilder instance here

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
```diff
@@ -0,0 +1,139 @@
...
+import org.jboss.resteasy.core.StringParameterInjector;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
+
+/**
+ * @author Marek Kopecky mkopecky@redhat.com
+ * @author Ron Sigal rsigal@redhat.com
+ */
+@Provider
+@Priority(Integer.MAX_VALUE)
+public class MultiValuedCollectionParamConverter implements ParamConverter<Collection<?>>
```
NicoNes: Why not create an Abstract class`MultiValuedParamConverter` to provide base impl for construtor with `StringParameterInjector` and `separator` and impl for `getSeparatorChar()` and `stringify()`.
It will be shared betwen `MultiValuedCollectionParamConverter` and `MultiValuedArrayParamConverter`.
WDYT?

docbook/reference/en/en-US/modules/StringConverter.xml
ronsigal: Yep. Thanks.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Ok, got it.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
ronsigal: Done.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Well, the point of copyPrimitiveArray() was to work around Java's unfortunate distinction between primitive values and Objects. So toString() ends up with an array of Objects in both cases. I guess the alternative would be to have a different stringify() for each primitive type.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Ooops. ;-)

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Good. Thanks.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Ok, I'm working on it.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
NicoNes: When reading your answer first time I did not understand your point very well.
Now it's clearer and it also makes me understand why you make this class extends from `ParamConverter<Object>` instead of `ParamConverter<Object[]>`.
I forgot this java distiction about primitive  array and object array.
Thanks :+1:

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
NicoNes: Please forget about this comment. 
As said below, I get the point of the java disctinction between primitive array and object array :+1:

resteasy-core/src/main/java/org/jboss/resteasy/core/ResteasyProviderFactoryImpl.java
```diff
@@ -1249,6 +1256,20 @@ public String toString(Object object, Class clazz, Type genericType, Annotation[
       }
       return object.toString();
    }
+   
+   public String toString() {
```
NicoNes: Hi ron,

I don't get the point of this `toString()` impl. 
If I understand well, in addition to the java standard `toString()` representation it will list all `ParamConverterProvider` registered right ? What for ?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/processors/AbstractCollectionProcessor.java
```diff
@@ -1,35 +1,71 @@
...
       else if (object.getClass().isArray())
       {
-         if (object.getClass().getComponentType().isPrimitive())
+         ParamConverter<Object> paramConverter = config.getProviderFactory().getParamConverter(object.getClass(), type, annotations);
+         if (paramConverter != null)
+         {
+            object = paramConverter.toString(object);
+            target = apply(target, object);
+         }
+         else if (object.getClass().getComponentType().isPrimitive())
```
NicoNes: Can this `else if` section and the next `else` section be merged this way:

```
 int length= Array.getLength(object);
 for (int i = 0; i < length; i++)
 {
     target = apply(target, Array.get(object, i));
  }
```

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedParamConverterProvider.java
```diff
@@ -0,0 +1,173 @@
...
+   @SuppressWarnings("unchecked")
+   @Override
+   public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations)
+   {
+      if (!isApplicable(annotations))
+      {
+         return null;
+      }
+      String separator = getSeparator(annotations);
+      Class<?> paramType = getHeaderParam(annotations);
```
NicoNes: The `paramType` here should never be null to avoid NPE in `StringParameterInjector.getParamSignature()` wich is used many times in `StringParameterInjector` class.
So I suggest to change and rename `getHeaderParam()` this way:

```
 private Class<?> getParamType(Annotation[] annotations)
   {
      if (annotations == null)
      {
         return null;
      }
      for (Annotation a : annotations)
      {
         if (a instanceof HeaderParam)
         {
            return HeaderParam.class;
         }
         else if (a instanceof CookieParam)
         {
            return CookieParam.class;
         }
         else if (a instanceof PathParam)
         {
            return PathParam.class;
         }
         else if (a instanceof MatrixParam)
         {
            return MatrixParam.class;
         }
         else if (a instanceof QueryParam)
         {
            return QueryParam.class;
         }
         else if (a instanceof FormParam)
         {
            return FormParam.class;
         }
      }
      return null;
   }
```

And then if we go further, this `MultiValuedParamConverterProvider` is applicable if the `@Separator` is present and the `paramType` is not null so the code can be changed to:

```
@Provider
public class MultiValuedParamConverterProvider implements ParamConverterProvider
{

@SuppressWarnings("unchecked")
@Override
public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations)
{
  Separator separatorAnnotation = getSeparatorAnnotation(annotations);
  Class<?> paramType = getParamType(annotations);
  if(separatorAnnotation == null || paramType == null){
	 return null;
  }
  String separator = getSeparator(separatorAnnotation,paramType);
  ...
}

...

private Separator getSeparatorAnnotation(Annotation[] annotations)
{
  for (Annotation a : annotations)
  {
	 if (a instanceof Separator)
	 {
		return (Separator) a;
	 }
  }
  return null;
}

private String getSeparator(Separator separatorAnnoation, Class<?> paramType)
{
  String separator = separatorAnnoation.value();
  if ("".equals(separator))
  {
	 separator = CookieParam.class.equals(paramType) ? "-" : ",";
  }
  return separator;
}

...
}
```

WDYT?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
```diff
@@ -0,0 +1,214 @@
...
+            objectArray[i] = Array.getFloat(primitiveArray, i);
+         }
+         else if (double.class.equals(primitiveClass))
+         {
+            objectArray[i] = Array.getDouble(primitiveArray, i);
+         }
+      }
+   }
+   
+   private boolean isHeader()
```
NicoNes: Instead of iterating again on the annotations array can we better add a getter in `StringParameterInjector` to use the `paramType` already stored here ?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
```diff
@@ -0,0 +1,138 @@
...
+   {
+      for (String param : params)
+      {
+         c.add(stringParameterInjector.extractValue(param));
+      }
+      return c;
+   }
+
+   @SuppressWarnings({ "rawtypes", "unchecked" })
+   private String stringify(Collection<?> value)
```
NicoNes: You forgot to find and use the `headerDelegate` if any here

NicoNes:
Hey @ronsigal,

Apart from my comments everything looks good to me in the code.  I'll have a look to the tests.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Great!!

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: Done.

resteasy-core/src/main/java/org/jboss/resteasy/core/ResteasyProviderFactoryImpl.java
ronsigal: Hmmm, it must be something I put in for testing and forgot to remove. Good catch.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/processors/AbstractCollectionProcessor.java
ronsigal: @NicoNes, not sure I follow. The section
```
      else if (object.getClass().isArray())
      {
         ParamConverter<Object> paramConverter = config.getProviderFactory().getParamConverter(object.getClass(), type, annotations);
         if (paramConverter != null)
         {
            object = paramConverter.toString(object);
            target = apply(target, object);
         }
```
is meant to transform the array as a whole by way of a single ParamConverter ... .

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
ronsigal: @NicoNes, yeah, I hadn't gotten that far. It's in there now (but not committed yet).

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
ronsigal: @NicoNes, yes, now I'm just pulling HeaderDelegate out of the StringParameterInjector. isHeader() is gone.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedParamConverterProvider.java
ronsigal: @NicoNes, well, I care only about @HeaderParam. But I've changed getParamSignature() to
```
   public String getParamSignature()
   {
      return (paramType != null ? paramType.getName() : "") + "(\"" + paramName + "\")";
   }
```
Also, now that I think of it, I changed
```
       if (HeaderParam.class.equals(paramType))
       {
          delegate = factory.getHeaderDelegate(baseType);
          if (delegate != null)
          {
        	  return true;
          }
       }
```
to
```
       if (HeaderParam.class.equals(paramType) || org.jboss.resteasy.annotations.jaxrs.HeaderParam.class.equals(paramType))
       {
          delegate = factory.getHeaderDelegate(baseType);
          if (delegate != null)
          {
        	  return true;
          }
       }
```

ronsigal:
I need to take another look at the User Guide.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedParamConverterProvider.java
NicoNes: Ok :+1:

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
NicoNes: :+1:

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
NicoNes: Looks good to me now.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
ronsigal: @NicoNes, yeah, ok. I used Arrays.asList().

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/processors/AbstractCollectionProcessor.java
ronsigal: @NicoNes, ah, I see. Yes, I think that works. I think it's clear enough as it is, though.

docbook/reference/en/en-US/modules/StringConverter.xml
```diff
@@ -665,5 +665,152 @@ public static class TestResource {
...
+"\\p{Punct}|\\[\\p{Punct}+\\]"
+</programlisting>
+
+      That is, it must be either a single instance of a punctuation symbol, i.e., a symbol in the set
+<programlisting>
+!"#$%&amp;'()*+,-./:;&lt;=&gt;?@[\]^_`{|}~
+</programlisting>
+
+      or a class of punctuation symbols like "[-,;]".
+      </listitem>
```
marekkopecky: We can also mention here that separator value is limited also by other specifications. Eg, "," or ";" can't be used with CookieParam because of [cookie syntax defined by RFC](https://tools.ietf.org/html/rfc6265#section-4.1.1). Another limitation is that separator can't be used in items in array/collection, (right?), eg if I have list `["a","b;c","d"]` and use separator ";", the result would be probably `["a", "b", "c", "d"]`. It is obvious, but it would be fine to have this in doc ...

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedAbstractParamConverter.java
```diff
@@ -0,0 +1,107 @@
...
+   @SuppressWarnings({ "unchecked" })
+   protected String stringify(Collection<?> value)
+   {
+      Method saveAsString = null;
+      if (paramConverter == null && headerDelegate == null)
+      {
+         try
+         {
+            Object o = value.iterator().next();
+            saveAsString = o.getClass().getMethod("saveAsString");
```
marekkopecky: What about to add description of `saveAsString` method to docs? (StringConverter.xml)

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/MultiValuedParamDefaultParamConverterProxyTest.java
```diff
@@ -0,0 +1,536 @@
...
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+/**
+ * @tpSubChapter Parameters
+ * @tpChapter Integration tests
+ * @tpTestCaseDetails Test for RESTEasy extended support for multivalue @*Param (RESTEASY-1566 + RESTEASY-1746)
```
marekkopecky: s/RESTEASY-1566 + RESTEASY-1746/RESTEASY-1996/g

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/MultiValuedParamDefaultParamConverterProxyTest.java
```diff
@@ -0,0 +1,536 @@
...
+ *                    org.jboss.resteasy.plugins.providers.MultiValuedArrayParamConverter and
+ *                    org.jboss.resteasy.plugins.providers.MultiValuedArrayParamConverter are used
```
marekkopecky: s/MultiValuedArrayParamConverter/MultiValuedCollectionParamConverter/g

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.jboss.resteasy.core.StringParameterInjector;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
+
+/**
+ * @author Marek Kopecky mkopecky@redhat.com
+ * @author Ron Sigal rsigal@redhat.com
+ */
+@Provider
+@Priority(Integer.MAX_VALUE)
```
marekkopecky: https://github.com/resteasy/Resteasy/pull/1721/files#diff-505f9bfa494b22e39f4249e79a979f41R811
```
These built-in ParamConverters have the lowest priority ...
```
MIN_VALUE would be probably better, right? (if I don't missed something ...)

Anyway, I believe that `@Provider` and `@Priority` annotation shouldn't be in corverter classes (same for `MultiValuedArrayParamConverter` class), but in MultiValuedParamConverterProvider class, right? I see this approach for example in [PriorityEqualityTest](https://github.com/resteasy/Resteasy/blob/master/testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/providers/PriorityEqualityTest.java#L79)

NicoNes: > Anyway, I believe that @Provider and @Priority annotation shouldn't be in corverter classes (same for MultiValuedArrayParamConverter class), but in MultiValuedParamConverterProvider class, right?

I agree with you on this point.

> MIN_VALUE would be probably better, right? (if I don't missed something ...)

I worked on this point in another PR and if I 'm not wrong, I can say that to achieve following behavior:

> These built-in ParamConverters have the lowest priority, so any user supplied  ParamConverters will be tried first

the `@Priority(Integer.MAX_VALUE)` on built-in provider will not be taken into account.
 Instead this behavior will be achieved by the `ResteasyProviderFactoryImpl.ExtSortedKey.compareTo()` method that says that user supplied  `ParamConverters `are less than built-in `ParamConverters`.

`@Priority(Integer.MAX_VALUE)` on built-in provider will only be used to sort built-in `ParamConverters` between them. And the provider with the lowest priority is used first.

docbook/reference/en/en-US/modules/StringConverter.xml
ronsigal: @marekkopecky, done. Thanks.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/MultiValuedParamDefaultParamConverterProxyTest.java
ronsigal: @marekkopecky, done. Good eye.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/MultiValuedParamDefaultParamConverterProxyTest.java
ronsigal: @marekkopecky done.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedArrayParamConverter.java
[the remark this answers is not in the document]
ronsigal: @NicoNes, I eliminated saveAsString(). Now I just look for toString().

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedAbstractParamConverter.java
ronsigal: @marekkopecky, I eliminated saveAsString(). I just look for toString() now.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/MultiValuedCollectionParamConverter.java
ronsigal: @marekkopecky, @NicoNes, you're both right. I took out all of the @Priority annotations. Thank you! This is why it's good to be surrounded by smart people. ;-)

ronsigal:
Besides all of the corrections mentioned above, I've added one paragraph to the User Guide:

```
The MultiValued*ParamConverters depend on existing facilities for handling the individual 
elements. On the server side, once it has parsed the incoming string into substrings, 
MultiValued*ParamConverter turns each substring into an Java object according to Section 3.2 
"Fields and Bean Properties" of the JAX-RS specification. On the client side, 
MultiValued*ParamConverter turns a Java object into a string as follows:

    1. look for a ParamConverter;
    2. if there is no suitable ParamConverter and the parameter is labeled @HeaderParam, look for a HeaderDelegate; or
    3. call toString(). 
```

ronsigal:
At this point, I think I've addressed all of your suggestions and corrections. I really appreciate all of the help.

If there are no further problems, I think it's ready to commit. Let me know. Thanks, guys.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfiguration.java
```diff
@@ -57,7 +57,7 @@ public void setProperties(Map<String, Object> newProps)
       providerFactory.setProperties(newProps);
    }
 
-   protected ResteasyProviderFactory getProviderFactory()
+   public ResteasyProviderFactory getProviderFactory()
```
marekkopecky: subjective comment:

I guess that the only reason for this is adding HeaderDelegate, right?

```java
ResteasyProviderFactory factory = ((ClientConfiguration)client.getConfiguration()).getProviderFactory();
factory.addHeaderDelegate(HeaderDelegateClass.class, new HeaderDelegate());
```

^^^ this looks quite ... complicated, WDYT? What about to use just:
```java
((ClientConfiguration)client.getConfiguration()).addHeaderDelegate(HeaderDelegateClass.class, new HeaderDelegate());
```
?

marekkopecky:
> At this point, I think I've addressed all of your suggestions and corrections. I really appreciate all of the help.
> 
> If there are no further problems, I think it's ready to commit. Let me know. Thanks, guys.

Please give me more time to review this, I'll approve this PR once I will finish the review. I need to do more tests, maybe add some new tests, etc.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfiguration.java
ronsigal: @marekkopecky, ClientConfiguration doesn't have an addHeaderDelegate() method. Are you saying that we should add one? 

In fact, I found it surprisingly hard to add HeaderDelegates, on both the client and server sides. Why isn't it classified as a provider, so that it can be registered easily? I don't know.

ronsigal:
@marekkopecky, re: "Please give me more time"

Of course. Thank you for looking.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfiguration.java
ronsigal: @marekkopecky,

re: "I guess that the only reason for this is adding HeaderDelegate, right?"

That's probably true.

asoldano:
@ronsigal ... and please rebase the PR, so that when @marekkopecky is fine with it, it can be merged.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfiguration.java
marekkopecky: > ClientConfiguration doesn't have an addHeaderDelegate() method. Are you saying that we should add one?

I don't say that we should add one. But I see these advantages of this approach:
* easier adding of HeaderDelegate to client
* we don't provide access to all ResteasyProviderFactory methods, but just to the one method, that is really needed (`addHeaderDelegate`)

I currently doesn't see any disadvantage of this approach. Do you see some? WDYT?

ronsigal: Hi @marekkopecky,

re: "I currently doesn't see any disadvantage of this approach. Do you see some?"

I've created RESTEASY-2059 "Simplify registration of HeaderDelegates".

ronsigal:
@asoldano, rebase done.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfiguration.java
marekkopecky: @ronsigal Yes, we can deal with this later, but if we make getProviderFactory() method public now, it could be complicated to make it private later ...

Anyway, this is PR for master branch, so it is not important for current  WF ...

ronsigal: @marekkopecky, I've reverted getProviderFactory() to protected and added getParamConverter() and addHeaderDelegate().

ronsigal:
Everything has been resolved, I believe, so I'll merge this one.

PR #1739, opened 2018-10-22
[RESTEASY-2045] refactored method

rsearls:
The code has been reorganized to show a more logical grouping of the actions to be take on "instanceof" classes.  
Before the flow was
    check for specific class
    check instanceof ApplicationException
    check instanceof WriterException
    check instanceof ReaderException
    check instanceof WebApplicationException  handle case one
    check exception cause and super classes of cause 
    check instanceof WebApplicationException  handle case two
    check instanceof Failure

The refactored flow is
  check for specific class
  check instanceof WebApplicationException
         handle case one
         check exception cause and super classes of cause
        handle case two
  check instanceof Failure   
        check instanceof WriterException   - class is a subclass of Failure
        check instanceof ReaderException  - class is a subclass of Failure
        check for mapper for some other subclass of Failure
  check instanceof ApplicationException   
  check exception cause and super classes of cause

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
@@ -255,56 +255,74 @@ public class ExceptionHandler
       Response response = wae.getResponse();
       return response;
    }
-
-
+   
    public Response handleException(HttpRequest request, Throwable e) {
+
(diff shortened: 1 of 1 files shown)


asoldano:
@rsearls , for future reference, can you write in the description above what is actually changing here? the formatting fixes actually makes it difficult to understand that. Thanks

PR #1743, opened 2018-10-24
[RESTEASY-2038] CacheInterceptor uses content type of response as mediatype instead of accept header of request

jurakp:
Use media type from the request, if not present then fallback to response content type.

Jira: https://issues.jboss.org/browse/RESTEASY-2038


jimma:
Thanks for the work. Just a minor thing to improve: can you please improve the test to demonstrate the cache really works for the scenario :  accept is application/json and response content type is application/json;charset=UTF-8 ?

jurakp:
@jimma Added message created time millis, so now we should be good.

mihxil:
How about 'Accept' headers with multiple media-types? I think these now cause an error in CacheInterceptor, because it assumes 'Accept' to contain just one simple value.

PR #1747, opened 2018-10-24
[RESTEASY-1749] Fix validation behavior for EJBs

ronsigal:


providers/resteasy-validator-provider-11/src/main/java/org/jboss/resteasy/plugins/validation/SimpleViolationsContainer.java
```diff
@@ -25,4 +37,314 @@ public SimpleViolationsContainer(Set<ConstraintViolation<Object>> cvs)
...
+            }
+         }
+         if (!skip)
+         {
+             getViolations().add(cv1);
+         }
+      }
+   }
+
+   public static boolean compareConstraintViolation(ConstraintViolation<?> cv1, ConstraintViolation<?> cv2)
```
asoldano: @ronsigal does this really need to be 'public'?

ronsigal: Nope. Thanks.

PR #1758, opened 2018-10-29
[RESTEASY-2050] updated springframework version to 5.1.1.RELEASE (on …

rsearls:
…master branch)


rsearls:
I took the 2nd option.  Because spring tests are disabled.  A migration effort might be needed to move to 5.1.1.RELASE and work will be needed to reenable the tests

PR #1770, opened 2018-11-12
Reduce memory allocation on client side because of ResteasyProviderFactory#initialize

asoldano:
Revisiting ResteasyProviderFactory implementations to reduce memory usage on client side.
The changes reduce the number of invocations to ResteasyProviderFactory#initialize() of 70% when running the current RESTEasy testsuite.


resteasy-core/src/main/java/org/jboss/resteasy/core/ResteasyProviderFactoryImpl.java
```diff
@@ -2361,6 +2367,7 @@ public void injectProperties(Object obj, HttpRequest request, HttpResponse respo
    }
 
    // Configurable
+   //TODO!! problem with listeners...
```
marekkopecky: Which problem?

asoldano: uh, stale comment. Basically, as the method returns a live collection, invocations to such method are to be treated the same as the methods that potentially change the ResteasyProviderFactory. That's already the case, though, in DelegateLazyResteasyProviderFactory (I didn't have that class yet when I wrote the comment)

marekkopecky:
@asoldano Do you have a RESTEasy jira for this PR?

asoldano:
@marekkopecky not yet. I need to backport this to 3.6, have the performance team try it and then if it's worth keeping the changes I'll create a jira and merge the PRs

marekkopecky:
@asoldano +1

asoldano:
Closing, a different approach (#1879) is now being evaluated.

PR #1771, opened 2018-11-12
[RESTEASY-2061] exclude jaxb-api archive from resteasy-multipart-prov…

rsearls:
…ider and resteasy-jaxb-provider

### DIFF
--- a/providers/jaxb/pom.xml
+++ b/providers/jaxb/pom.xml
@@ -38,6 +38,12 @@
         <dependency>
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jboss.spec.javax.servlet</groupId>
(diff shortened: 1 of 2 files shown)


resteasy-bom/pom.xml
```diff
@@ -40,6 +40,12 @@
                 <groupId>org.jboss.resteasy</groupId>
                 <artifactId>resteasy-jaxb-provider</artifactId>
                 <version>${project.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>javax.xml.bind</groupId>
+                        <artifactId>jaxb-api</artifactId>
+                    </exclusion>
+                </exclusions>
```
asoldano: Why not excluding this for the jaxb-runtime dependency in resteasy-dependencies/pom.xml  ?

rsearls: This is no jaxb-runtime in the BOM.  I checked that first.

asoldano: https://github.com/resteasy/Resteasy/blob/master/resteasy-dependencies-bom/pom.xml#L225 ?

jbliznak: @asoldano what is the purpose to have two distinct BOMs anyway? 
https://github.com/resteasy/Resteasy/blob/master/resteasy-bom/pom.xml
https://github.com/resteasy/Resteasy/blob/master/resteasy-dependencies-bom/pom.xml

If used for differrent purposes then they both should handle excludes in the same way, right?

PR #1776, opened 2018-11-15
[RESTEASY-2068] Improve Client Proxy @Encoded handling

lyind:
Implement enhancement explained at: [RESTEASY-2068](https://issues.jboss.org/browse/RESTEASY-2068)

This PR finishes the work done for two issues that have already been closed:
[RESTEASY-724](https://issues.jboss.org/browse/RESTEASY-724)
[RESTEASY-1076](https://issues.jboss.org/browse/RESTEASY-1076)

The test `org.jboss.resteasy.test.client.proxy.EncodedPathProxyTest` has been slightly modified to also test for the minor enhancements. (The unmodified test also still passed with the changes in `PathParamProcessor`.)

This enhancement allows me to fix a critical issue in one of my projects: passing `';'` character via path parameter to a RESTEasy endpoint is impossible. With this PR applied one can now work around this by annotating the parameter in question with `@Encoded` and replacing all `';'` characters with `"%3B"` prior to passing to the proxy client.


asoldano:
@ronsigal , please check this :-)

ronsigal:
Hi @lyind,

I believe that your changes are correct. Thank you for your contribution.

I just have a couple of suggestions.

  1. I think it would be better if this facility were available to all of the @*Param types.
  2. Consequently, the tests should be extended to cover the other @*Param types.
  3. This new facility should be documented in the RESTEasy User Guide.

There's one other thing worth mentioning, I think. The "normal" way of using this facility would be something like
```
    @Path("")
    public static interface TestResource {

        @POST
        @Path("encode/{p}")
        public String testEncode(@Encoded @PathParam("p") String p);
    }

    @Path("")
    public static class TestResourceImpl {

        @POST
        @Path("encode/{p}")
        public String testEncode(@Encoded @PathParam("p") String p) {
            return p;
        }
    }
```
where @Encoded appears on both sides. Note that the output of
```
        ResteasyClient client = new ResteasyClientBuilderImpl().build();
        ResteasyWebTarget target = client.target(generateURL(""));
        TestResource proxy = target.proxy(TestResource.class);
        String response = proxy.testEncode("a%3bb");
        System.out.println("response: " + response);
```
would be
```
response: a%3bb
```
But it's also interesting to consider
```
    @Path("")
    public static interface TestResource {

        @POST
        @Path("encode/{p}")
        public String testEncode(@Encoded @PathParam("p") String p);
        
        @POST
        @Path("decode/{p}")
        public String testDecode(@Encoded @PathParam("p") String p);
    }

    @Path("")
    public static class TestResourceImpl {

        @POST
        @Path("encode/{p}")
        public String testEncode(@Encoded @PathParam("p") String p) {
            return p;
        }
        
        @POST
        @Path("decode/{p}")
        public String testDecode(@PathParam("p") String p) {
            return p;
        }
    }
```
where, for the testDecode() method, @Encoded appears only on the client side. The output  of
```
        ResteasyWebTarget target = client.target(generateURL(""));
        TestResource proxy = target.proxy(TestResource.class);
        response = proxy.testDecode("a%3bb");
        System.out.println("response: " + response);
```
is
```
response: a;b
```
Here, the entity "a%3bb" remains unencoded on the client side, but it is decoded on the server side, which might be convenient.

@lyind, what do you think? I could help with the documentation, by the way.

-Ron

asoldano:
Hi @lyind , did you perhaps have time to extend this as mentioned in the previous comment? Happy new year btw :-)

lyind:
Thank you @asoldano, happy new year, too.

I did implement client proxy `@Encoded` support for `@QueryParam` and noticed `@MatrixParam` already got its own encoding handling. I am not sure about `@FormParam` in connection with `@Encoded`, the complexity seems high and I'm not sure anybody needs it.

I'm now waiting for the tests to complete and will fix remaining issues. I had some problems running the tests locally, not sure if due to my local setup or issues with the code (JettyClientEngineTest failures, RxObservableTest test hanging forever).

asoldano:
@lyind sorry for the late reply. Thanks for your work here, let's have @ronsigal check this too (he's off for few days) than I believe we can merge this.

ronsigal:
@lyind, @asoldano, OK, I will look soon.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/proxy/EncodedParamsProxyTest.java
```diff
@@ -0,0 +1,114 @@
...
+    */
+   @Test
+   public void testNoencodeProxy() throws Exception
+   {
+      ResteasyWebTarget target = client.target(generateBaseUrl());
+      EncodedParamsProxyInterface proxy = target.proxy(EncodedParamsProxyInterface.class);
+      Response response = proxy.noencode("t;hawkular/f%3Bjk-feed", "100%-%25ig", null);
+      Assert.assertEquals(200, response.getStatus());
+      String uri = response.readEntity(String.class);
+      Assert.assertEquals(generateBaseUrl() + "/test/noencode/t;hawkular/f%253Bjk-feed?query=100%25-%2525ig", uri);
```
ronsigal: Hi @lyind,

I don't understand this. '/' is not encoded, but '%', when not part of an encoded character, is.

-Ron

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/proxy/EncodedParamsProxyTest.java
```diff
@@ -0,0 +1,114 @@
...
+    */
+   @Test
+   public void testMatrixNoencodeProxy() throws Exception
+   {
+      ResteasyWebTarget target = client.target(generateBaseUrl());
+      EncodedParamsProxyInterface proxy = target.proxy(EncodedParamsProxyInterface.class);
+      Response response = proxy.noencodeMatrix("t;hawkular/f%3Bjk-feed", "100%-%25ig", null);
+      Assert.assertEquals(200, response.getStatus());
+      String uri = response.readEntity(String.class);
+      Assert.assertEquals(generateBaseUrl() + "/test/noencode-matrix;matrix1=t%3Bhawkular%2Ff%3Bjk-feed;matrix2=100%25-%25ig", uri);
```
ronsigal: @lyind , this result looks the same as the result in testMatrixEncodeProxy().

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/proxy/EncodedParamsProxyTest.java
lyind: @ronsigal The method `noencode()` tests the parameter value handling without `@Encoded` annotation. All the tests in `EncodedParamsProxyTest` focus on the proxy client which is why they return the whole URI instead of the parameter values processed by the server.

## Regarding '/'
A `/` in the path parameter value isn't encoded in the original design, which I see as a deliberate decision complementing allowing pattern matching for parameters instead of forcing a fixed 1:1 mapping between path segments and parameters. This was probably done for flexibility reasons.

Examples:
1. ```
   @Path("/{param1:.?}")
   collapsesSlashes(@PathParam("param1") String param1)
   // call via proxy: collapsesSlashes("a/b/c") -> param1 == "a/b/c"
   ```
2. ```
   @Path("/{param1}/{param2}")
   honoursSlashes(@PathParam("param1") String param1, @PathParam("param2") String param2)
   // call via proxy: honoursSlashes("a", "b") -> param1 == "a" && param2 == "b"
   // BUT:
   // call via proxy: honoursSlashes("a/b/c", "d") -> param1 == "a" && param2 == "b"
   ```
3. ```
   @Path("/{param1}/{param2}")
   honoursSlashesEncoded(@Encoded @PathParam("param1") String param1, @PathParam("param2") String param2)
   // call: honoursSlashesEncoded("a/b/c", "d") -> param1 == "a%2Fb%2Fc" && param2 == "b"
   ```

To still be able to use a service with methods looking like in example #3 with *an unknown number of slashes* in parameter values, I made sure the proxy client encodes the slashes to get them through in one piece if it encounters an `@Encoded` `@PathParam`.

To clarify, `noencode()` correctly tests behavior as in example #1.

## Regarding '%'
The `%` must always be encoded according to https://www.ietf.org/rfc/rfc1738.txt :
 1. "The character "%" is unsafe because it is used for encodings of other characters."
 2. "Octets must be encoded if they have no corresponding graphic character within the US-ASCII coded character set, if the use of the corresponding character is unsafe, or if the corresponding character is reserved for some other interpretation within the particular URL scheme."

Since `@Encoded` is not present on `noencode()`'s' `@PathParam` or `@QueryParam` the proxy client may safely assume the values passed are not pre-encoded and just encode the whole string without preserving encoded chars.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/proxy/EncodedParamsProxyTest.java
lyind: @ronsigal I didn't change the `@Encoding` handling for `@MatrixParam` as it was working just fine where `@PathParam` and `@QueryParam` weren't working for me. To document that and to guard against future regressions, I still added the tests `testMatrixEncodeProxy()` and `testMatrixNoencodeProxy()`.

I also noticed `@MatrixParam` behaves just the same in the encoded and non-encoded case, from the proxy client's point of view. Apart from *appearing* odd or redundant in the tests I can't find anything wrong about it.

Do you have a suggestion what should be changed in the proxy client's `@MatrixParam` handling?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/proxy/EncodedParamsProxyTest.java
lyind: @ronsigal I just saw your extended explanation at [RESTEASY-2068](https://issues.jboss.org/browse/RESTEASY-2068?focusedCommentId=13699650&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13699650).

I think confusion is caused by the special handling of `/` in the encoding methods.

Values passed as arguments for `@Encoded` annotated parameters should be encoded already and thus not contain the unencoded slash at all. The example I provided above also missed that fact (I edited it).

I will modify the code to be in line with the handling for non `@Encoded` annotated methods (regarding the slash).

lyind: @ronsigal Detailed discussion updated in Issue Tracker: [RESTEASY-2068](https://issues.jboss.org/browse/RESTEASY-2068?focusedCommentId=13703720&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13703720)

lyind: @ronsigal Can you please have a look at the summary/suggestions I put together in the Jira ticket?

dmsleptsov:
Hi @ronsigal  @lyind,

This PR will be merge? Or this issue resolved in other PR?

lyind:
I still have to use an EncodedPath class to wrap strings including paths, when using the proxy client, today.

We still need this PR.

dmsleptsov:
@lyind Thanks for your reply and maybe you know workaround how pass already encoded path-param?

jamezp:

lyind:
Someone should really pick this up, since `@Encoded` is still broken in Resteasy.

In my opinion it's an example of a major shortcoming that drives switches from Resteasy to more simple and incomplete REST frameworks or even languages (golang for example).

Granted, most APIs won't have to pass characters like `;` or `/` in a single parameter. But if one does and then has to jump through unnecessary hoops, because of `@Encoded` not working correctly...

PR #1778, opened 2018-11-16
[RESTEASY-2065] provided boolean to flag use of TrustSelfSignedStrate…

rsearls:
…gy or not.


asoldano:
@rsearls all runs are failing, due to a checkstyle rule. I suggest rebasing your branch on top of latest master, so that you can get the same failures reported here.

rsearls:
Issue resolved.

PR #1780, opened 2018-11-18
[RESTEASY-1142] ExceptionHandler: catch IllegalStateException from ClientResponse

lyind:
Backport of #1716 for 3.6 branch.

@ronsigal This does not include the tracing logger fix, as that feature doesn't exist on 3.6.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
@@ -125,7 +125,15 @@ public class ExceptionHandler
 
       if (unwrappedException instanceof WebApplicationException) {
          WebApplicationException wae = (WebApplicationException) unwrappedException;
-         if (wae.getResponse() != null && wae.getResponse().getEntity() != null) return wae.getResponse();
+         Response response = wae.getResponse();
+         if (response != null) {
+            try {
+               if (response.getEntity() != null) return response;
+            } catch(IllegalStateException ise) {
+               // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
+               return response;
+            }
+         }
       }
 
       if ((jaxrsResponse = executeExceptionMapper(unwrappedException)) != null) {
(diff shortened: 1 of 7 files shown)


asoldano:
@lyind thanks for the PR. I see checkstyle failures preventing the integration-testsuite to run on Travis CI. Would you mind fixing them please?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
```diff
@@ -125,7 +125,15 @@ protected Response unwrapException(HttpRequest request, Throwable e)
...
       if (unwrappedException instanceof WebApplicationException) {
          WebApplicationException wae = (WebApplicationException) unwrappedException;
-         if (wae.getResponse() != null && wae.getResponse().getEntity() != null) return wae.getResponse();
+         Response response = wae.getResponse();
+         if (response != null) {
+            try {
+               if (response.getEntity() != null) return response;
+            }
+            catch(IllegalStateException ise) {
+               // IllegalStateException from ClientResponse.getEntity() means the response is closed and got no entity
```
rsvoboda: exception is swallowed, i'd rather see log message on trace or debug level.

lyind: While I share your opinion that it is good style not to catch and swallow exceptions in general application code, the point here is to ignore an expected exception **in the ExceptionHandler**. This affair could certainly have been avoided if the specification would have mentioned a Response.isClosed() method or handled the `isClosed == true` state differently.

The actual exception the user/admin/developer cares about is going to be extracted and forwarded, together with the ClientResponse's entity (if such is present).

In almost all other parts of RESTEasy or the actual application code I would certainly handle/log any exception for good reason. In this specific instance I would dare make an exception, though. ;)

In case one would insist on logging something here:
What useful information could one log here? That the response is closed? That would mean two exceptions logged with stacktraces whenever ONE client exception happens.

lyind:
@asoldano I fixed the checkstyle issue and noticed a commit was missing and only present on master. I added it, tested locally and squashed the fix commits to make it more readable on the 3.6 branch history. Are you fine with that approach?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
ronsigal: Yes, this is a special case, and I think @lyind's solution makes sense here.

ronsigal:
Looks good to me.

PR #1782, opened 2018-11-20
[RESTEASY-1619] Do not set compression quality for compression algori…

asoldano:
…thms not supporting it


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/iioimage/IIOImageProviderTest.java
```diff
@@ -40,7 +39,8 @@
 @RunAsClient
 public class IIOImageProviderTest {
    static ResteasyClient client;
-   static final String testPngResource = "test.png";
+   static final String testPngResource1 = "test1.png";
```
jbliznak: Just a nit, can the filenames or names of the properties suggest why different image files are checked? Something like "test_image_jdk8.png" and "test_image_jdk9plus.png" and/or comment, WDYT?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/iioimage/IIOImageProviderTest.java
```diff
@@ -70,8 +70,8 @@ private static String generateURL(String path) {
     * @tpSince RESTEasy 3.0.16
     */
    @Test
-   @Ignore("[RESTEASY-1724] java.lang.IllegalStateException: Compression mode not MODE_EXPLICIT!")
    public void testPostJPEGIMage() throws Exception {
```
jbliznak: Can we change the method name too? There is a little mismatch JPEG vs PNG ;)

asoldano: method name changed

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/iioimage/IIOImageProviderTest.java
asoldano: added a comment, ok

PR #1785, opened 2018-11-21
[RESTEASY-1821] allow JAX-RS annotations in CDI @Stereotype

xstefank:
https://issues.jboss.org/browse/RESTEASY-1821

This is a first PR in a series for all JAX-RS annotations. It contains only Produces and Consumes annotations to verify the approach taken and also make the review easier.


xstefank:
@asoldano thanks for the review! I think you're right about potential memory leak, I will try to catch this somehow in the undeployment if possible. Thank you

marekkopecky:
@xstefank Please add java-docs ...

xstefank:
@asoldano @marekkopecky @ronsigal PR updated and ready for review again

xstefank:
The PR now also updates org.jboss.arquillian.container:arquillian-weld-embedded to version 2.0.1.Final. Please let me know if I should create a separate issue for this upgrade.

jbliznak:
@xstefank it looks like rebase/amend got a little scrumbled, I would bet you don't want to use RESTEASY-1884 commit with your contriution amended.

xstefank:
@jbliznak probably :) thanks for the notice

asoldano:
@xstefank , thanks I'll review in details in the following days. I'm moving this to 4.1 as it's something I'd rather not rush into 4.0 due to the additional annotation scanning it implies (4.0 is what we have in Quarkus).
Btw, I'm thinking that this is probably to be considered more as a new feature than as a bug fix. WDYT @marekkopecky ? In that case, in order for having this included in 3.7, we should change the jira type and go through the process for WildFly RFE (basically, we need a dev analysis document)

marekkopecky:
bq. I'm thinking that this is probably to be considered more as a new feature than as a bug fix

Agree

xstefank:
Closing this PR. I will open a new one for the feature request

PR #1792, opened 2018-11-27
[RESTEASY-2081] Prepare test for application packaging when Application subclass is not present

tterem:
Prepare test for application packaging when Application subclass is not present

Jira: https://issues.jboss.org/browse/RESTEASY-2081
3.6: https://github.com/resteasy/Resteasy/pull/1793


testsuite/integration-tests/src/test/resources/org/jboss/resteasy/test/core/basic/BasicWeb.xml
```diff
@@ -0,0 +1,12 @@
+<web-app version="3.0" xmlns="http://java.sun.com/xml/ns/javaee"
```
marekkopecky: Can you please rename this file to NoApplicationSubclassWeb.xml? To be clear that this web.xml is related with NoApplicationSubclassTest class ...

tterem: Done.

PR #1796, opened 2018-11-29
[RESTEASY-2083] Fix issue when SseBroadcasterImpl.notifyOnCloseListeners removes wrong sink from the outputQueue

Ingvord:
Due to GenericType.equals implementation currently (as per 3.6.2.Final) SseBroadcasterImpl always removes first item from the outputQueue in notifyOnCloseListeners method when client closes connection. Resulting in other clients stop receive events.

Test project that reproduces the issue and suggest workaround: [link](https://github.com/Ingvord/resteasy-sse-bug-report)

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
@@ -315,4 +315,13 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
       return false;
    }
 
+   @Override
+   public boolean equals(Object o) {
+      return this == o;
+   }
+
+   @Override
+   public int hashCode() {
+      return ((Object)this).hashCode();
+   }
 }
--- a/testsuite/integration-tests/pom.xml
+++ b/testsuite/integration-tests/pom.xml
@@ -105,6 +105,19 @@
     </profiles>
 
     <dependencies>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.9.5</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.hamcrest</groupId>
+                    <artifactId>hamcrest-core</artifactId>
+                </exclusion>
+            </exclusions>
+            <scope>test</scope>
+        </dependency>
+
         <!-- TODO Workaround dependency for arquillian to work with container using Remoting 5. Remove when updated version of
         wildfly-arquillian-container-managed is available -->
         <dependency>
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java
@@ -1,18 +1,26 @@
 package org.jboss.resteasy.test.providers.sse;
 
-import java.io.IOException;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
+import org.jboss.resteasy.core.ResteasyContext;
+import org.jboss.resteasy.plugins.providers.sse.OutboundSseEventImpl;
+import org.jboss.resteasy.plugins.providers.sse.SseBroadcasterImpl;
+import org.jboss.resteasy.plugins.providers.sse.SseEventOutputImpl;
+import org.jboss.resteasy.spi.HttpRequest;
+import org.jboss.resteasy.spi.ResteasyAsynchronousContext;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
 
 import javax.ws.rs.sse.OutboundSseEvent;
 import javax.ws.rs.sse.SseEventSink;
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.concurrent.*;
 
-import org.jboss.resteasy.plugins.providers.sse.OutboundSseEventImpl;
-import org.jboss.resteasy.plugins.providers.sse.SseBroadcasterImpl;
-import org.junit.Assert;
-import org.junit.Test;
+import static junit.framework.TestCase.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
 
 /***
  *
@@ -94,8 +102,8 @@ public class SseBroadcasterTest
       {
          Assert.fail("All close listeners should have been notified");
       }
-      Assert.assertTrue(sseEventSink1.isClosed());
-      Assert.assertTrue(sseEventSink2.isClosed());
+      assertTrue(sseEventSink1.isClosed());
+      assertTrue(sseEventSink2.isClosed());
    }
 
    // We are expecting this test to invoke both close and error listeners when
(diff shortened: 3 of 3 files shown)


asoldano:
@Ingvord thanks for the contribution. @jimma will review the PR; in the mean time, we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0
Do you agree to these terms?

Ingvord:
@asoldano 

> we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0

sure, no objections from my side!

jimma:
@Ingvord  Appreciate your detailed analysis and patch for this issue.  This looks OK to me. It would better that we have a test for this important fix. Can you please add your reproducer/test to this PR ?

Ingvord:
@jimma test case has been added in 35ffb25 

While looking through the test class I have noticed the following code snippet:

```java
sseBroadcasterImpl.onClose(ses -> {
   Assert.fail("Close listeners should not have been notified");
});
```

AFAIK _fail_ will never happen because _onClose_ is invoked from [async thread](https://github.com/resteasy/Resteasy/blob/e20f09f02fd4054251d83b28d872dafd4dffb07f/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java#L145) and exceptions thrown there are ignored. Could you please correct me if I am wrong?

asoldano:
@Ingvord thanks!
@jimma please see last comment above. Any thoughts? is the test ok and we can proceed? Thanks

NicoNes:
Hi guys,

Abouth that:

> AFAIK fail will never happen because onClose is invoked from async thread and exceptions thrown there are ignored. Could you please correct me if I am wrong?

To me @Ingvord is right, the exception thrown by `Assert.fail("Close listeners should not have been notified");` will be ignored.

This exception will be reflected in `CompletionStage` that will be completed exceptionally  at line https://github.com/resteasy/Resteasy/blob/master/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java#L150: 
``` 
outputImpl.send(event).whenComplete((object, err) -> {
if (err != null)
{
 notifyOnErrorListeners(eventSink, err);
}
})
```
But since we are not doing nothing with this `CompletionStage`, this exception will never appear in our test results.

This exception will prevent `onErrorConsumers` from being invoked and thus the `countDownLatch` will not be counted down. 
So it's OK the test will fail, but with the wrong message/cause.
Instead of having "Close listeners should not have been notified" we will get "All error listeners should have been notified"

So to get it works properly I suggest to change the test as follow:

   ```
// We are expecting this test to only invoke error listeners on broadcasting
   // error other than IOException
   @Test
   public void testErrorListeners() throws Exception
   {
      SseBroadcasterImpl sseBroadcasterImpl = new SseBroadcasterImpl();

      SseEventSink sseEventSink = newSseEventSink(new RuntimeException());
      sseBroadcasterImpl.register(sseEventSink);
      Assert.assertFalse(sseEventSink.isClosed());

      AtomicBoolean onCloseListenerInvoked = new AtomicBoolean(false);
      sseBroadcasterImpl.onClose(ses -> {
         onCloseListenerInvoked.set(true);
      });
      
      CountDownLatch countDownLatch = new CountDownLatch(2);
      sseBroadcasterImpl.onError((ses, error) -> {
         countDownLatch.countDown();
      });
      sseBroadcasterImpl.onError((ses, error) -> {
         countDownLatch.countDown();
      });

      CompletionStage<?> broadcast = sseBroadcasterImpl
            .broadcast(new OutboundSseEventImpl.BuilderImpl().data("Test").build()).toCompletableFuture().com;
      if (!countDownLatch.await(5, TimeUnit.SECONDS))
      {
         Assert.fail("All error listeners should have been notified");
      }
      if (onCloseListenerInvoked.get())
      {
         Assert.fail("Close listeners should not have been notified");
      }
   }
```

**Comment by jimma** [comment]
> But since we are not doing nothing with this CompletionStage, this exception will never appear in our test results.
This exception will prevent onErrorConsumers from being invoked and thus the countDownLatch will not be counted down.

I am not sure I understand this. Do you mean the exception thrown from SseEventOutputImpl.send(event) prevents invoking onErrorConsumers ?

**Comment by NicoNes** [comment]
> I am not sure I understand this. Do you mean the exception thrown from SseEventOutputImpl.send(event) prevents invoking onErrorConsumers ?

No the exception I'm talking about is the one thrown by `onCloseListener` in the test at line https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java#L173:

```
sseBroadcasterImpl.onClose(ses -> {
   Assert.fail("Close listeners should not have been notified");
});
```

More detailled explanation:

The goal of `SseBroadcasterTest.testErrorListeners` is to be sure that only **onErrorConsumers** are invoked when an error occured while broadcasting and this error is other than an IOException.
If for any reason **onCloseConsumers** are also invoked when such error occured, I would like the test to fail with an explicit error like "Close listeners should not have been notified".

So now suppose that in the future someone changes actual `SseBroadcasterImpl.notifyOnErrorListeners() ` to this:

```
private void notifyOnErrorListeners(SseEventSink eventSink, Throwable throwable)
{
  notifyOnCloseListeners(eventSink);
  onErrorConsumers.forEach(consumer -> {
	 consumer.accept(eventSink, throwable);
  });
}
```

With this change in case of error, both **onCloseConsumers** and **onErrorConsumers** are invoked no matter the type of the exception.  So when running test `SseBroadcasterTest.testErrorListeners`, I'm expecting the test to fail with message "Close listeners should not have been notified" => unfortunately with actual version of the test it will fail with message "All error listeners should have been notified".

Now let's see what happen in the code in this scenario:

1- Test `testErrorListeners ` is running. Line https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java#L182 is invoked:
`sseBroadcasterImpl.broadcast(new OutboundSseEventImpl.BuilderImpl().data("Test").build());`

2- Then we will arrive at `SseBroadcasterImpl .broadcast()` line https://github.com/resteasy/Resteasy/blob/master/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java#L150: 

```
outputImpl.send(event).whenComplete((object, err) -> {
if (err != null)
{
 notifyOnErrorListeners(eventSink, err);
}
})
```
`outputImpl.send(event)` => the CompletionStage result of this invocaton is completed exceptionally  (see https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java#L215).

`...whenComplete((object, err) -> {...})` => since the previous CompletionStage is completed exceptionally => `err!=null`  =>  `notifyOnErrorListeners(eventSink, err);` is invoked

3 - With the changes described above in `SseBroadcasterImpl.notifyOnErrorListeners()` , `notifyOnCloseListeners(eventSink);` will be invoked. So **onCloseConsumers** defined in test at line https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java#L172 will be invoked and will throw a JUNIT exception with message "Close listeners should not have been notified".

And this is this exception that will prevent **onErrorConsumers**  to be invoked (see `SseBroadcasterImpl.notifyOnErrorListeners()` with change described above).

4- Don't forget that this JUNIT exception is thrown form within `...whenComplete((object, err) -> {...})` so it will not be thrown, it will just be catched and ignored by the CompletionStage mechanism since the `CompletionStage`  from `outputImpl.send(event)`  was already completed exceptionally.

5- So `SseBroadcasterImpl .broadcast()` will end without throwing no exception. We will arrive at line https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/sse/SseBroadcasterTest.java#L183 in test:
```
if (!countDownLatch.await(5, TimeUnit.SECONDS))
{
 Assert.fail("All error listeners should have been notified");
}
```
And since no **onErrorConsumers**  has been invoked the `countDownLacth` was never counted down => The test only fail with message "All error listeners should have been notified" and message "Close listeners should not have been notified" is never displayed.

Longer but I hope it's clearer. Let me know

PR #1800, opened 2018-12-10
[RESTEASY-2097]: ClientResponse.getEntity() and ClientResponse.bufferEntity() do not completely follow the spec

NicoNes:
Jira: https://issues.jboss.org/browse/RESTEASY-2097

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
@@ -43,8 +43,8 @@ public abstract class ClientResponse extends BuiltResponse
    protected Map<String, Object> properties;
    protected ClientConfiguration configuration;
    protected byte[] bufferedEntity;
-   protected boolean streamRead;
-   protected boolean streamFullyRead;
+   protected volatile boolean streamRead;
+   protected volatile boolean streamFullyRead;
    protected RESTEasyTracingLogger tracingLogger;
 
    @Deprecated
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
@@ -43,6 +43,7 @@ public abstract class ClientResponse extends BuiltResponse
    protected Map<String, Object> properties;
    protected ClientConfiguration configuration;
    protected byte[] bufferedEntity;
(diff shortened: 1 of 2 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -43,6 +43,7 @@
    protected Map<String, Object> properties;
    protected ClientConfiguration configuration;
    protected byte[] bufferedEntity;
+   protected boolean streamRead;
    protected boolean streamFullyRead;
```
asoldano: While reviewing this PR, I noticed both the new streamRead and already existing streamFullyRead attributes are not volatile and perhaps should; comments? AFAICS, they're accessed from the InputStreamWrapper that's returned to users, so potentially we could have memory visibility issues, couldn't we?

NicoNes: I think you're right.
User can invoke `Response.read(InputStream.class)` or just `Response.getEntity()` that return the `InputStreamWrapper` in the main thread  and decide to read it in another thread. Then after that when calling  `Response.getEntity()` or  `Response.bufferEntity()` in the main thread we can face a visibility issue. 
If every one OK i'll make the change.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientResponse.java
```diff
@@ -401,7 +408,7 @@ public boolean bufferEntity()
    {
       abortIfClosed();
       if (bufferedEntity != null) return true;
-      if (entity != null) return false;
+      if (streamRead) return false;
       if (metadata.getFirst(HttpHeaderNames.CONTENT_TYPE) == null) return false;
       InputStream is = getInputStream();
```
NicoNes: Based on @asoldano 's comment here https://github.com/resteasy/Resteasy/pull/1244#discussion_r133709884, don't you think guys that we also have to  make `bufferEntity()` method `synchronized` since it also access and read the stream and concurrency can occurs with `finalize()` method that close this stream?

asoldano: mmh... I would leave the ProcessingException there, the javadoc says that it's meant to be thrown if there was an error while buffering the stream (it's a RuntimeException, so it's not declared)

asoldano:
Thanks @NicoNes for the fix!

PR #1806, opened 2018-12-22
[RESTEASY-1709] Inject javassist proxies for Application objects.

ronsigal:


ronsigal:
@asoldano, it turns out that the status of multiple Applications in a single WAR, and the status of @Context injection itself, are being debated in the JAX-RS Expert Group.

A likely outcome will be that the use of javassist is no longer relevant. I will close this PR for now, and possibly forever.

-Ron

PR #1808, opened 2018-12-26
RESTEASY-2106 Don't swallow JSON binding exceptions

gsmet:
* https://issues.jboss.org/browse/RESTEASY-2106

### DIFF
--- a/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
+++ b/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
@@ -85,7 +85,7 @@ public class JsonBindingProvider extends AbstractJsonBindingProvider
             return null;
          }
          // detail text provided in logger message
-         throw new ProcessingException(Messages.MESSAGES.jsonBDeserializationError(e, e.getMessage()));
+         throw new ProcessingException(Messages.MESSAGES.jsonBDeserializationError(e, e.getMessage()), e);
       }
    }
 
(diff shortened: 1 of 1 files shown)


gsmet:
BTW, maybe we should backport this to the current stable as it's a real problem when trying to use JSON-B instead of Jackson.

marekkopecky:
Current master branch doesn't print yasson part of stacktrace in "server receive", "server send", "client receive", "client send" cases.

This PR fix this for "server send", "client receive", "client send" cases. If server receives corrupted data ("server receive" case), server prints `11:03:43,788 INFO  [stdout] (default task-1) null`, or client receive for example `javax.ws.rs.ProcessingException: RESTEASY008200: JSON Binding deserialization error: Error deserialize JSON value into type: class java.lang.Integer.` message without any stacktrace information.

@asoldano What about to print stacktrace in debug log level only?

Btw, server prints exception in "server send" case twice (this is not regression on this PR, this behaviour is in current master).

asoldano:
@marekkopecky , I'm going to merge this PR. Regarding your comments, I agree we can make some effort for unification and I would be fine with debug level logs. Can you provide a PR with tests covering all the scenarios you have in mind? This would basically mean allowing @rsearls to finish up what was started in RESTEASY-2016. Thanks

marekkopecky:
@asoldano (@rsearls): I prepared test for all 4 scenarios https://github.com/marekkopecky/Resteasy/commit/0d5c3f37936b2c86054a10061680ffd1fb47498b , but there are still some open questions (like DEBUG or ERROR level, what to log on client, etc). This questions are represents by TBD sections in my tests. I have 4.x tests only, but I can prepare also 3.6 tests (cc @rsvoboda, @jbliznak, @tterem)


	
Btw, if you decide to use debug level, we should describe enableing of debug logs on client side in docs (basically fix [RESTEASY-1731 bug](https://issues.jboss.org/browse/RESTEASY-1731)).

asoldano:
@rsearls can you please take over from here using the tests that @marekkopecky has provided?

@marekkopecky thanks a lot

PR #1810, opened 2019-01-06
Update dependency version

automatictester:


automatictester:
Hi @asoldano I have noticed you are using one of the tools I've created. As the groupId [has changed](https://github.com/automatictester/wiremock-maven-plugin/commit/53845878aa9b3c62cce60365cf2390e7553644ba#diff-600376dffeb79835ede4a0b285078036) a good while ago, I thought it would be good if I let you know there is a new version with a new groupId. Also, in case you want to upgrade version in future, this is the groupId under which new versions are released.

asoldano:
OK, thanks!

PR #1811, opened 2019-01-08
[RESTEASY-2109] [3.6] Add nullary constructor to fix native image generation

triceo:
For details, see linked JIRA.

PR #1812, opened 2019-01-08
[RESTEASY-2109] [master] Add nullary constructor to fix native image generation

triceo:
For details, see the referenced JIRA.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/DocumentProvider.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/DocumentProvider.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.plugins.providers;
 
+import org.jboss.resteasy.core.ResteasyContext;
 import org.jboss.resteasy.plugins.server.servlet.ResteasyContextParameters;
 import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
 import org.jboss.resteasy.spi.ReaderException;
@@ -43,6 +44,12 @@ public class DocumentProvider extends AbstractEntityProvider<Document>
    private boolean expandEntityReferences = false;
    private boolean enableSecureProcessingFeature = true;
    private boolean disableDTDs = true;
+
+   public DocumentProvider()
+   {
+      // nullary constructor so that GraalVM's native-image is able to process this file
+      this(ResteasyContext.getContextData(ResteasyConfiguration.class));
+   }
 
    public DocumentProvider(final @Context ResteasyConfiguration config)
    {


triceo:
@asoldano Sure, didn't notice the newline changes. 

What's the proper line termination in the project? IDEA reports differences across the files I randomly checked.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ServerFormUrlEncodedProvider.java
```diff
@@ -1,52 +1,57 @@
...
+ * @version $Revision: 1 $
+ */
+@Produces("application/x-www-form-urlencoded")
+@Consumes("application/x-www-form-urlencoded")
+@ConstrainedTo(RuntimeType.SERVER)
+public class ServerFormUrlEncodedProvider extends FormUrlEncodedProvider
+{
+   protected boolean useContainerParams;
+
+   public ServerFormUrlEncodedProvider()
```
marekkopecky: @triceo Can you please share the details about this change? This doesn't seems to be related with Java's ServiceLoader SPI described in [RESTEASY-2109](https://issues.jboss.org/browse/RESTEASY-2109) ...

marekkopecky:
Jira: https://issues.jboss.org/browse/RESTEASY-2109
3.6 PR: https://github.com/resteasy/Resteasy/pull/1811

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ServerFormUrlEncodedProvider.java
triceo: @marekkopecky I went through all the SPI uses, looking for a situation where:

- Some constructor is provided,
- while no nullary constructor is provided.

Since this would cause the default nullary constructor not to be inserted by the JVM, `native-image` would have failed with the same issue after the earlier issue was fixed.

However, now that I'm double-checking my work, I see that the `ServerFormUrlEncodedProvider` is not actually made accessible through the SPI and this change is therefore unnecessary. Thanks for pointing this out.

I will remove this change when I'm updating the PR to fix the whitespace changes.

marekkopecky:
@triceo Line termination should be unified, but it is not a functional bug to have different line termination ([RESTEASY-2115](https://issues.jboss.org/browse/RESTEASY-2115)). For the purpose of this PR, it should be OK to just do necessary changes and do not change the line termination of DocumentProvider.java file. Line termination should be unified in another PR.

asoldano:
Thanks @marekkopecky

triceo:
@asoldano @marekkopecky I have addressed your comments and force-pushed the new changes. In my opinion, this is ready to merge.

PR #1815, opened 2019-01-09
[RESTEASY-1404] Client SSL tests sample test

tterem:
Jira: https://issues.jboss.org/browse/RESTEASY-1404

cc: @marekkopecky

### DIFF
--- a/testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/utils/PortProviderUtil.java
+++ b/testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/utils/PortProviderUtil.java
@@ -13,6 +13,8 @@ import java.net.URL;
 public class PortProviderUtil {
    private static final int DEFAULT_PORT = 8080;
 
+   private static final int HTTPS_PORT = 8443;
+
    private static final String PORT_ENV_VAR_NAME = "RESTEASY_PORT";
 
    private static final String PORT_PROPERTY_NAME = "org.jboss.resteasy.port";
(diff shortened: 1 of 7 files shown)


tterem:
Do not merge this yet.

asoldano:
@tterem @marekkopecky as a general feedback on this PR, wouldn't it be safer to have predefined server configurations for the tests needing SSL? I don't like the idea of copying/restoring server configs in each test. How about something similar to what we do in https://github.com/jbossws/jbossws-cxf/blob/master/modules/testsuite/cxf-tests/src/test/java/org/jboss/test/ws/jaxws/cxf/clientConfig/CXFDefaultClientConfigurationTestCase.java ?

marekkopecky:
>  wouldn't it be safer to have predefined server configurations for the tests needing SSL?

Looking to JBWS TS, AFAICS some maven-plugin creates new xml configuration based on standalone.xml and some groovy-scripts.

AFAIK, CLI commands should be backward compatible. And CLI command is recommended way how to configure the server.

> I don't like the idea of copying/restoring server configs in each test

Makes sence. So what about this:
* new container in arquillian.xml
* ssl will be setup in first ssl test (let's say that all ssl-tests checks new container and sets ssl if ssl is not set). 
    * (+) CLI commands (recommended on WF)
    * (+) non-127.0.0.1 IP address support (certificates will be still created by bash scripts)
    * (-) one reload is still needed for sll setup

WDYT?

If you don't want to use CLI at all, I guess we can also try to use offline commands from [Creaper](https://github.com/wildfly-extras/creaper), AFAIK these offline commands are based on some groovy xml transformations (similar like JBWS).

Btw WDYT about bash script for generating of certificates? This allow to use non-loopback IP adresses.

asoldano:
OK, I'm fine with a dedicated arquillian conf for this. I thought it could be one for all tests, but if that's not possible I'm fine with using creaper/cli . Using bash scripts for generating certs is ok. Thanks

PR #1818, opened 2019-01-10
[RESTEASY-2055] Fixed Javadoc

jurakp:
Jira: https://issues.jboss.org/browse/RESTEASY-2055


jurakp:
@asoldano I think each committer should check also javadoc to avoid such situation in the future. Maybe some CI job for Javadoc verification is worth doing :)

jurakp:
> @jurakp right, I agree. Do you want to do that too (additional CI job for javadoc) ? Otherwise feel free to create a jira, thanks.

I can try it if I have permissions to do it.

PR #1819, opened 2019-01-10
Resteasy 2062 default value first request

rsearls:
StringParameterInjector enhanced to validate setting of params with @DefaultValue.
Unit-test ParamInjectionExceptionTest added to test proper exception throwing.
FormParamTest altered. Code change caused it to fail on deployment. Test scenario
covered by unit-test.


rsearls:
The solution for this jira can not reside in Resteasy.  It MUST reside in Wildfly because errors are to be flagged at deployment time.  Resteasy is never called until a client calls the service.

marekkopecky:
@rsearls + @asoldano (cc @rsvoboda @jbliznak @tterem ) What about to handle this issue in RESTEasy, but server/container/WF activates RESTEasy during the deploying of the application?

PR #1821, opened 2019-01-11
[RESTEASY-2055] Added CI run for Javadoc

jurakp:
Added CI run for Javadoc.
Jira: https://issues.jboss.org/browse/RESTEASY-2055

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 install:
  - mvn clean; java -version
 script:
- - travis_wait 60 mvn -B -Ptravis  -fae -Dserver.version=$SERVER_VERSION install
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,7 @@
 install:
  - mvn clean; java -version
 script:
+ - mvn -B -Dmaven.javadoc.skip=false javadoc:javadoc
```
jbliznak: I guess this should go after the build is done, right? So we should switch the lines. Also do we need to execute it with every WF in matrix combinations? Perhaps something like https://docs.travis-ci.com/user/build-stages/matrix-expansion/ ?

jurakp: Made it as another stage. The first run is test in the test stage, the second run is javadoc in javadoc stage. Unfortunately I have to fix javadoc for Java 11.

PR #1830, opened 2019-01-16
[RESTEASY-2108] Making JsonBindingProvider more restrictive in terms …

asoldano:
…of declared @Consume/@Provide media-types; also optimizing the checks on isReadble/isWriteable a bit


asoldano:
The javadoc failure is unrelated to the changes here. I'm merging this, so that the work on RESTEASY-2107 can be performed on top of this.

PR #1832, opened 2019-01-17
[RESTEASY-2129] Java 11 as default release target

rsvoboda:
Java 11 as default release target

https://issues.jboss.org/browse/RESTEASY-2129

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,25 +5,13 @@ script:
 
 language: java
 jdk:
-  - openjdk8
   - oraclejdk11
 env:
-  - SERVER_VERSION=13.0.0.Final
   - SERVER_VERSION=14.0.0.Final
   - SERVER_VERSION=15.0.0.Final
 jobs:
-  exclude:
-    - jdk: oraclejdk11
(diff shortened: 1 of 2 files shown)


rsvoboda:
reverted parent bump as it caused regression in test compilation, will look into it

rsvoboda:
JavaDoc failed 
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.0:javadoc (default-cli) on project resteasy-core-spi: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.
```

rsvoboda:
I built and tried 3.1.0-SNAPSHOT of javadoc plugin but still the javadoc execution fails.

I found https://bugs.openjdk.java.net/browse/JDK-8212233 and several more via google, but didn't find the fix ...

https://github.com/rsvoboda/Resteasy/tree/java11-default-plus-parent-update branch to have also latest jboss-parent, just fyi

PR #1835, opened 2019-01-17
Resteasy 1991 response builder built response

rsearls:


asoldano:
@rsearls , I'll let you finish the conversation with Sergey on JIRA before looking at this in details. In the mean time, please make sure to rebase the PR against later master. Thanks

rsearls:
This rebased against the latest master.  I did it just before submitting it.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
```diff
@@ -247,9 +270,244 @@ public void setGenericType(Type genericType)
...
+      {
+         if (is != null)
+         {
+            if (consumeInputStream)
+            {
+               while (is.read() > 0)
+               {
+               }
+            }
+            is.close();
```
sberyozkin: can you please add
```
is = null'
```

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
```diff
@@ -262,8 +520,30 @@ public boolean hasEntity()
    public boolean bufferEntity()
    {
       abortIfClosed();
-      // no-op
-      return false;
+
+      if (bufferedEntity != null) return true;
+      if (streamRead) return false;
+      if (metadata.getFirst(HttpHeaderNames.CONTENT_TYPE) == null) return false;
```
sberyozkin: Please remove the media type check; there is no requirement that Content-Type header is available for bufferEntity to work

rsearls:
Done.

sberyozkin:
@rsearls thank you :-)

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
```diff
@@ -247,9 +265,245 @@ public void setGenericType(Type genericType)
 
    public <T> T readEntity(Class<T> type, Type genericType, Annotation[] anns)
    {
```
ronsigal: @rsearls, can't this code be removed from ClientResponse now?

rsearls: No it must remain.  It is used by BuiltResponseEntityNotBacked.  This class was provided to
support other select cases where entity should not be read.

ronsigal: Hi @rsearls,

I'm confused. ClientResponse is derived from BuiltResponse, and it looks like this code (readEntity(), etc.) was moved from ClientResponse up to BuiltResponse. And BuiltResponseEntityNotBacked is derived from BuiltResponse ...

-Ron

rsearls: Look at prior versions of the code.  You will see that BuiltResponse line 248
has always provided the root definition for this method.

   public <T> T readEntity(Class<T> type, Type genericType, Annotation[] anns)
   {
      throw new IllegalStateException(Messages.MESSAGES.entityNotBackedByInputStream());
   }

ClientResponse overrides this method with actual code (lines 237-301)

    public synchronized <T> T readEntity(Class<T> type, Type genericType, Annotation[] anns)
       {
          abortIfClosed();
          if (entity != null)
          {
             if (type.isInstance((this.entity)))
          ................

These classes create new instances of BuiltResponse, 

    line 182  AsyncResponseConsumer
    line 471  ResourceMethodInvoker
    line 184 ResteasyView 
    line 45  MessageSanitizerMediaTypeTest

so the current code should be maintained.

ronsigal: Hi @rsearls,

I'm still confused. I understand that BuiltResponseEntityNotBacked is replacing BuiltResponse because it restores the original version of readEntity(). But I see almost identical versions of readEntity() in BuiltResponse and ClientResponse, so why can't ClientResponse just inherit readEntity() from BuiltResponse? 

-Ron

rsearls: Here is more information on this.

BuiltResponse implements the 4 methods declared in Response interface.
   
   public <T> T readEntity(Class<T> type)
   public <T> T readEntity(GenericType<T> entityType)
   public <T> T readEntity(GenericType<T> entityType)
   public <T> T readEntity(GenericType<T> entityType, Annotation[] annotations)

Each of the methods above calls method 

    public <T> T readEntity(GenericType<T> entityType, Annotation[] annotations)

which is declared in BuiltResponse.

org.jboss.resteasy.client.jaxrs.internal.ClientResponse extends BuiltResponse.
This class does not override the first 4 methods.  It does override method

  public <T> T readEntity(GenericType<T> entityType, Annotation[] annotations).

org.jboss.resteasy.client.jaxrs.internal.AbortedResponse extends ClientResponse
and overrides method 

  public <T> T readEntity(GenericType<T> entityType, Annotation[] annotations)

unit-test org.jboss.resteasy.test.client.resource.NoContentStreamingCloseTestResponse 
extends BuiltResponse and overrides method 

  public <T> T readEntity(GenericType<T> entityType, Annotation[] annotations)

org.jboss.resteasy.core.ServerResponse extends BuiltResponse.  It does not
override method readEntity.  In the future it may want to override this method.

It's possible a customer may want to override this method in the future as well.

If this method is removed the first 4 methods will need to be implemented in
some way and the classes that extend from BuiltResponse might need to provide
their own implementations of these 4 methods.

ronsigal: Hi @rsearls,

re: "If this method is removed the first 4 methods will need to be implemented in
some way and the classes that extend from BuiltResponse might need to provide
their own implementations of these 4 methods."

Maybe this is the problem. I meant that the implementation of that last readEntity() could be removed from ClientResponse. Maybe you think I meant it could be removed from BuiltResponse? If so,then I apologize for being unclear.

-Ron

rsearls: Currnetly this method can not simply be removed.  The key issue is that there are some
class variables that are being shared between ClientResponse and BuiltResponse and
some are not, in particular entity and bufferedEntity are causing an issue.

Do you want me to attempt to refactor this code so method readEntity can be removed from ClientResponse?

ronsigal:
Hey @rsearls, I see. It seems like a shame to have so much duplicate code in BuiltResponse and ClientResponse. Is it a big deal to fix it?

rsearls:
I'll give it a go.

ronsigal:
Thanks, @rsearls. I can imagine a year from now, somebody coming along and not noticing that the code is duplicated, and making a change to just one version.

rsearls:
Refactoring complete

rsearls:
Done

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
```diff
@@ -38,488 +31,254 @@
...
 
-   @Override
-   public Set<Link> getLinks()
+   protected void setInputStream(InputStream is)
    {
-      LinkHeaders linkHeaders = getLinkHeaders();
-      Set<Link> links = new HashSet<Link>();
-      links.addAll(linkHeaders.getLinks());
-      return links;
+      this.is = is;
```
ronsigal: Hi @rsearls,

I see that setInputStream() in ApacheHttpClient43Engine is
```
         protected void setInputStream(InputStream is)
         {
            stream = is;
            resetEntity();
         }
```
Does it make sense to have resetEntity() here?

-Ron

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
```diff
@@ -38,488 +31,254 @@
...
+      }
+      return is;
    }
 
    @Override
-   public Link getLink(String relation)
+   public void releaseConnection() throws IOException
    {
-      return getLinkHeaders().getLinkByRelationship(relation);
+      releaseConnection(false);
```
ronsigal: @rsearls, I think releaseConnection() could be removed from ApacheHttpClient43Engine$ClientResponse() [not sure of the how to name that inner class.]

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
ronsigal: Like @asoldano said, there seem to be a lot of purely stylistic changes. For example,
```
@Override
public synchronized Object getEntity()
{
   abortIfClosed();
   Object entity = super.getEntity();
   if (entity != null)
   {
      return checkEntityReadAsInputStreamFullyConsumed(entity);
   }
   return checkEntityReadAsInputStreamFullyConsumed(getEntityStream());
}
```
starts in column 0. Not a big deal, but it increases the degree of difficulty ...

rsearls: The ClientResponse object in ApacheHttpClient43Engine is sometimes
called on a "get()" request which results in the input stream being
read and variable "streamFullyRead" set to TRUE.  This value needs to
be FALSE when a "getEntity()" call is make.  See testsuite case 
ClientFilterResponseBuilderTest and class ResponseBuilderCustomResponseFilter.
When I remove the call to resetEntity(); this test breaks.

This call could be replaced with setStreamFullyRead(false), but there
might be other use cases in which any of the other 4 variables in
resetEntity() need to be reset.

I would recommend leaving this as is.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
rsearls: No, releaseConnection() can't be removed from ApacheHttpClient43Engine.
This code performs tasks very different from BuiltResponse.releaseConnection()
A bunch of test break when this code is removed.

rsearls:
format adjusted

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
ronsigal: Sorry, I seem to be having trouble expressing myself clearly these days. I meant to say, "Does it make sense to add resetEntity() to BuiltResponse.setInputStream()?"

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
ronsigal: Ah, sorry, I see now that they're not the same.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/BuiltResponse.java
rsearls: This could be done.  None of our tests are negatively affected by adding it.   But do we really need it?  Customers have never reported any failures with the existing code and this code has been around a long time.  

Do you want me to add it any way?

ronsigal: re: "Customers have never reported any failures with the existing code and this code has been around a long time."

Hmmm. Prior to this pull request, setInputStream() was abstract in ClientResponse, so it was implemented only in classes derived from ClientResponse. So its implementation in BuiltResponse is something new. Given that resetting the inputStream, in effect, changes the content of the response, doesn't it make sense to reset the entity? WDYT?

asoldano:
@ronsigal the eclipse conf files for formatting are at https://github.com/resteasy/Resteasy/tree/master/config

ronsigal:
Thanks, @asoldano . I was thinking there should be a link to the style files in the resteasy.github.io pages, somewhere under "Contributing to RESTEasy". I thought there was a section like that, but I don't see it now. Maybe I'm imagining it.

asoldano:
@ronsigal I can't find that in the website... feel free to fix the pages and/or create a jira for doing that ;-)

asoldano:
Are we fin with merging this? @rsearls , can you please rebase again (sorry) ?

rsearls:
rebase done

ronsigal:
I think this is ready.

PR #1837, opened 2019-01-19
 [RESTEASY-2128] JDK-8212233 workaround

jurakp:
Jira: https://issues.jboss.org/browse/RESTEASY-2128

Re-enabling javadoc for JDK 11 by reverting commit f97ba8cce.
Thanks @rsvoboda for the found workaround.


asoldano:
Thanks @rsvoboda and @jurakp

PR #1844, opened 2019-01-23
[RESTEASY-2100] - The byte[] type of input parameter in proxy/MP-rest-client will cause large JVM stack used.

MMarus:
RESTEASY issue: https://issues.jboss.org/browse/RESTEASY-2100

Enhanced speed of the QueryParam processing in the proxy client. It is achieved by using StringBuilder and passing the data as array to the ResteasyUriBuilderImpl.java.


asoldano:
Reverted with https://github.com/resteasy/Resteasy/commit/e2736313fbf7beb8f24ab6e2f3dbe65da36c1253 , check jira for details.

PR #1847, opened 2019-01-25
[RESTEASY-2107]:Fix Bean Validation with JSON-B

jimma:


asoldano:
Or maybe we could simply avoid supporting the property for the precedence here in the client builder and expect the user to set a proprerly programmatically configured ResteasyProviderFactory if he really didn't want to ensure just one if the provider modules is in the classpath/classloader.

jimma:
>I've been thinking a bit about this and frankly speaking I'm not fully convinced about the proposed approach. I'm concerned with the need to add yet another thing into a ThreadLocal (the ResteasyContext), in particular the ResteasyProviderFactory, and never remove it.

Unless there is need to say we set one LocalResteasyProviderFactory in one client thread and another thread uses another LocalResetasyProvdierFacotry, this looks not a problem.  For user's api, LocalResteasyProviderFactory is built when client is created.  For all following service invocation with threads, it all happens under this LocalResteasyProviderFactory whether we set this LocalResteasyProviderFactory to ThreadLocal or not.

> Generally speaking, I'm thinking that we should probably try to move the responsibility of setting the precedence between Jackson2 and JsonB from the JsonBindingProvider to the ResteasyProviderFactory#register methods . In there, the current ResteasyProviderFactory properties should be checked and used to set a priority override for the provider being registered.
For the client side scenario, the ResteasyClientBuilderImpl#getProviderFactory() should simply be doing
providerFactory = new LocalResteasyProviderFactory(ResteasyProviderFactory.newInstance().setProperties(properties))
to pass its properties into the factory being created. No need to use threadlocal references to the RPF.

I am not sure I understand this. ClientBuilder.property() eventually does the same thing to set these property to ResteasyProviderFactory. 
`ClientBuilder builder = ClientBuilder.newBuilder();	      builder.property(ResteasyContextParameters.RESTEASY_PREFER_JACKSON_OVER_JSONB, true);`

The reason I add to ResteasyContext is we can use this api to retrieve it in JsonBindingProvider: 

`ResteasyProviderFactory providerFactory = ResteasyContext.getContextData(ResteasyProviderFactory.class);`

asoldano:
Superceded by https://github.com/resteasy/Resteasy/pull/1849

PR #1848, opened 2019-01-25
Fix invokeAsync with @PathParam annotations

fabienrousseau:


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -144,13 +144,16 @@ protected Object invokeAsync(final Object[] args)
...
+      ClientInvocation request = (ClientInvocation)requestBuilder.build(httpMethod);
+      request.setClientInvoker(this);
+
+      ClientResponse response = request.invoke();
       ClientContext context = new ClientContext(request, response, entityExtractorFactory);
       return extractor.extractEntity(context);
    }
 
-   protected ClientInvocation createRequest(Object[] args)
+   protected ClientInvocationBuilder createRequest(Object[] args)
```
marekkopecky: We need to be careful about backward compatibility in 3.6 PR. This changes protected method signature from public class from public WF module.

@asoldano + @ronsigal WDYT about this?

marekkopecky:
Master PR would be useful as well ...

I check 3.6 branch, proxy + MP REST client, Single + CompletionStage, RESTEasy works as expected in my checks, IllegalArgumentException is no more thrown

asoldano:
Thanks @marekkopecky ; yes, if possible it would be really better not changing that method signature. The class is in a package named "internal", still we should try avoiding changing it this way.
Besides that, the PR here needs a test.
@fabienrousseau if you still expect us to merge this PR, can you please work on the two mentioned things? Otherwise we'll likely include the changes in another revisited PR. Thanks

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
ronsigal: Good point, @marekkopecky. It's probably not a problem, but it's good to be safe. @fabienrousseau, can you just move the line
```
builder.setInvocation(request);
```
to the end of createRequest(), where it would become
```
builder.setInvocation(clientInvocation);
```
?

asoldano:
Replaced by #1900 , thanks anyway for the contribution!

PR #1849, opened 2019-01-27
RESTEASY-2107 2nd proposal

asoldano:
Another proposal for fixing RESTEASY-2107, with minimum disruption. The idea is to avoid introducing yet another mechnism for controlling precedences in json providers on client side. The user can either build the ResteasyProviderFactory with the providers he wants or he can control the classpath deciding which provider module is available.

### DIFF
--- a/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
+++ b/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/JsonBindingProvider.java
@@ -8,7 +8,6 @@ import java.lang.reflect.Type;
 
 import javax.annotation.Priority;
 import javax.json.bind.Jsonb;
-import javax.json.bind.annotation.JsonbAnnotation;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.Priorities;
 import javax.ws.rs.ProcessingException;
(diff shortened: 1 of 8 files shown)


asoldano:
@jimma when the disable-jsonb-client profile is executed, the ValidationJaxbTestCase should be using jackson2 on client side, shouldn't it? See https://github.com/resteasy/Resteasy/pull/1849/commits/f47f0a880020a9b2cfc1cb5ad87cf72d6ae4881c#diff-8c7ea03eb619e46657b96e8fe8d4f09eR505

jimma:
Adding this test to disable-jsonb-client profile makes  validation test running with Jackson2Provider client and JsonBindingProvider server. We need to remove the JsonBindingProvider setting server setting.

asoldano:
>  We need to remove the JsonBindingProvider setting server setting

Do we really need? I see the test try running with 2 different server configuration, with the deployment either using jsonb or jackson. As the client follows what comes from the classpath, having added it to both default and disable-jsonb profile should ensure we test all 4 combinations. Am I missing something? Perhaps feel free to show what you're proposing to do in addition to the PR here? Thanks

jimma:
@asoldano  I meant we don't need to test validation with jackson client and jsonbindingProvider server.

jimma:
@asoldano  Please go ahead to merge

asoldano:
@jimma thanks

PR #1851, opened 2019-02-05
[RESTEASY-1952] Unable to set proxy via JAX-RS API

jurakp:
Jira: https://issues.jboss.org/browse/RESTEASY-1952

Includes small piece of doc so that this is also documented.


ronsigal:
Looks good to me.

PR #1856, opened 2019-02-12
RESTEASY-2149 Comparing incompatible types for equality in resteasy-l…

liweinan:


rsvoboda:
@asoldano @liweinan this needs fix asap, the PR is wrong :(

https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#getInteger-java.lang.String-  expects its argument to name a system property !!!

To parse a string, call Integer.valueOf - https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#valueOf-java.lang.String-

PR #1860, opened 2019-02-13
[RESTEASY-2148] Enable utf8 encoding of filename only when explicitly…

asoldano:
… requested, disabled by default


marekkopecky:
I confirm that functionality of this fix is correct. I miss some tests here, tests would be useful if this commit will be backport to older streams. We can assume future backport, because original jira is blocker ...

asoldano:
Thanks @marekkopecky . The configuration option is actually tested in EncodingMimeMultipartFormProviderTest , but yes, I agree we could have a more thorough coverage. Given the tight schedule, I'll move on with the merging and backporting to 3.6, as yes, this is needed and blocking for the WFLY16 release. We can add further tests later on.

PR #1870, opened 2019-02-15
[RESTEASY-2056] 2 logging calls added to write exception and stacktra…

rsearls:
…ce to server.log. Test code added.


rsearls:
Because wfly-13 will be removed from resteasy travisCI when wfly-16 goes final in a week,and there are test failures due to wfly-13's use of jackson rather than jsonB, this PR will be withdrawn and reissued at a later time.

PR #1871, opened 2019-02-18
[RESTEASY-2084] @Provider annotation doesn't work with ClientRequestF…

xstefank:
…ilter

https://issues.jboss.org/browse/RESTEASY-2084

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
@@ -317,6 +317,11 @@ public class ResteasyClientBuilderImpl extends ResteasyClientBuilder
          // create a new one
          providerFactory = new LocalResteasyProviderFactory(ResteasyProviderFactory.newInstance());
          RegisterBuiltin.register(providerFactory);
+
+         if (ResteasyProviderFactory.peekInstance() != null)
+         {
+            providerFactory.initializeClientProviders(ResteasyProviderFactory.getInstance());
+         }
       }
       return providerFactory;
    }
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/interceptor/ClientRequestFilterRegistrationTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/interceptor/ClientRequestFilterRegistrationTest.java
@@ -1,0 +1,61 @@
+package org.jboss.resteasy.test.interceptor;
+
+
+import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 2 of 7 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
```diff
@@ -317,6 +317,11 @@ public ResteasyProviderFactory getProviderFactory()
          // create a new one
          providerFactory = new LocalResteasyProviderFactory(ResteasyProviderFactory.newInstance());
          RegisterBuiltin.register(providerFactory);
+
+         if (ResteasyProviderFactory.peekInstance() != null)
+         {
+            providerFactory.initializeClientProviders(ResteasyProviderFactory.getInstance());
+         }
       }
```
ronsigal: Hi @xstefank,

This looks good, but, now that I see it, I'm wondering why we didn't just build in the first place providerFactory as
```
providerFactory = new LocalResteasyProviderFactory(ResteasyProviderFactory.getInstance());
```
That would also handle, for example, clientMessageBodyReaders and clientMessageBodyWriters.

-Ron

xstefank: Hi @ronsigal, this would take into account all filters including server ones which I think is unnecessary for client provider factory. I would also like to add all other client providers - [RESTEASY-2163](https://issues.jboss.org/browse/RESTEASY-2163)

ronsigal: Yeah, @xstefank, good point.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/interceptor/ClientRequestFilterRegistrationTest.java
```diff
@@ -0,0 +1,61 @@
...
+   @After
+   public void close() {
+      client.close();
+   }
+
+   @Test
+   public void filterRegisteredTest() throws Exception {
+      WebTarget base = client.target(generateURL("/") + "testIt");
+      Response response = base.request().get();
+      Assert.assertEquals(404, response.getStatus());
```
ronsigal: Hi @xstefank,

This is a minor point, but I would suggest using a status like 456, something that wouldn't arise otherwise. You could get a 404 by getting the URL wrong.

-Ron

xstefank: surely, I will update the PR

xstefank:
@ronsigal PR updated with status change

ronsigal:
Thanks, @xstefank.

PR #1873, opened 2019-02-19
[RESTEASY-1364] Changed exception handling code for constructor metho…

rsearls:
…d.  Added NotFoundException.  Added new test case.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ConstructorInjectorImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ConstructorInjectorImpl.java
@@ -11,6 +11,8 @@ import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.spi.ValueInjector;
 import org.jboss.resteasy.spi.metadata.ConstructorParameter;
 import org.jboss.resteasy.spi.metadata.ResourceConstructor;
+import org.jboss.resteasy.spi.BadRequestException;
+import org.jboss.resteasy.spi.NotFoundException;
 
 import javax.ws.rs.WebApplicationException;
 
@@ -60,7 +62,9 @@ public class ConstructorInjectorImpl implements ConstructorInjector
    }
 
    @Override
-   public CompletionStage<Object[]> injectableArguments(HttpRequest input, HttpResponse response, boolean unwrapAsync)
+   public CompletionStage<Object[]> injectableArguments(HttpRequest input,
+                                                        HttpResponse response,
+                                                        boolean unwrapAsync)
    {
       if (params != null && params.length > 0)
       {
@@ -70,7 +74,9 @@ public class ConstructorInjectorImpl implements ConstructorInjector
          for (ValueInjector extractor : params)
          {
             int ifinal = i++;
-            stage = stage.thenCompose(v -> extractor.inject(input, response, unwrapAsync).thenAccept(value -> args[ifinal] = value));
+            stage = stage.thenCompose(v ->
+                    extractor.inject(input, response, unwrapAsync)
+                            .thenAccept(value -> args[ifinal] = value));
          }
          return stage.thenApply(v -> args);
       }
@@ -101,7 +107,19 @@ public class ConstructorInjectorImpl implements ConstructorInjector
    {
       return injectableArguments(request, httpResponse, unwrapAsync)
       .exceptionally(e -> {
-         throw new InternalServerErrorException(Messages.MESSAGES.failedProcessingArguments(constructor.toString()), e);
+         //CompletionStage does not support rethrow of exception.
+         //Must create new exception object and throw it.
+         Throwable t = e.getCause();
+         if (t != null) {
+            if (t instanceof NotFoundException) {
+               throw new NotFoundException(t.getMessage(), t.getCause());
+            } else if (t instanceof BadRequestException) {
+               throw new BadRequestException(t.getMessage(), t.getCause());
+            }
+         }
+         throw new InternalServerErrorException(
+                    Messages.MESSAGES.failedProcessingArguments(constructor.toString()), e);
+
       }).thenApply(args -> {
          try
          {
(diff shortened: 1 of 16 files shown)


rsearls:
I think this is addressed now.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/constructor/ResourceConstructorParamsTest.java
```diff
@@ -0,0 +1,72 @@
...
+        return PortProviderUtil.generateURL(path,
+                ResourceConstructorParamsTest.class.getSimpleName());
+    }
+
+    @Test
+    public void mixedTest() {
+        Response response = client.target(generateURL("/mixed/get"))
+                .request()
+                .get();
+        Assert.assertEquals("Incorrect status code", 500,
```
marekkopecky: @rsearls Thank you, my previous comments were fixed, server returns correct response code new. But I'm looking to the test and it seems that this testcase is unrelated with original issue (RESTEASY-1364). This testcases tests "500 response code behaviour", that I mentioned in my [previous comment](https://github.com/resteasy/Resteasy/pull/1873/#pullrequestreview-206199383), but this issue needs "400/404 response code behaviour" testcase. Can you provide correct testcase?

rsearls: Done

resteasy-core/src/main/java/org/jboss/resteasy/core/MatrixParamInjector.java
```diff
@@ -7,7 +7,7 @@
 import org.jboss.resteasy.spi.ValueInjector;
 
 import javax.ws.rs.MatrixParam;
-import javax.ws.rs.NotFoundException;
+import org.jboss.resteasy.spi.NotFoundException;
```
asoldano: Is this by mistake or on purpose? AFAIR org.jboss.resteasy.spi.NotFoundException is deprecated, btw.

resteasy-core/src/main/java/org/jboss/resteasy/core/PathParamInjector.java
```diff
@@ -9,7 +9,7 @@
 import org.jboss.resteasy.spi.ValueInjector;
 import org.jboss.resteasy.spi.util.Types;
 
-import javax.ws.rs.NotFoundException;
+import org.jboss.resteasy.spi.NotFoundException;
```
asoldano: Same as above

resteasy-core/src/main/java/org/jboss/resteasy/core/QueryParamInjector.java
```diff
@@ -7,7 +7,7 @@
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.spi.ValueInjector;
 
-import javax.ws.rs.NotFoundException;
+import org.jboss.resteasy.spi.NotFoundException;
```
asoldano: Same as above

resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -25,7 +25,7 @@
 import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
 
-import javax.ws.rs.BadRequestException;
+import org.jboss.resteasy.spi.BadRequestException;
```
asoldano: Similarly to the comments above, why is this changed? Is it by mistake maybe?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/asynch/JaxrsAsyncServletTest.java
```diff
@@ -27,7 +27,7 @@
 import org.junit.runner.RunWith;
 
 import javax.ws.rs.ForbiddenException;
-import javax.ws.rs.NotFoundException;
+import org.jboss.resteasy.spi.NotFoundException;
```
asoldano: Same as above

resteasy-core/src/main/java/org/jboss/resteasy/core/ConstructorInjectorImpl.java
```diff
@@ -11,6 +11,8 @@
 import org.jboss.resteasy.spi.ValueInjector;
 import org.jboss.resteasy.spi.metadata.ConstructorParameter;
 import org.jboss.resteasy.spi.metadata.ResourceConstructor;
+import org.jboss.resteasy.spi.BadRequestException;
+import org.jboss.resteasy.spi.NotFoundException;
```
asoldano: Similar comments to the other below on the exception imports (sorry, noticed them before this)

rsearls: Yes org.jboss.resteasy.spi.NotFoundException use is intentional.  I created it.  Just like 
org.jboss.resteasy.spi.BadRequestException which is pre-existing and used in several places and
by design carries the http error code in it.  I created org.jboss.resteasy.spi.NotFoundException to do the same.

I don't see that  org.jboss.resteasy.spi.BadRequestException or org.jboss.resteasy.spi.NotFoundException is deprecated?

Was it your intention to replace all uses of org.jboss.resteasy.spi.BadRequestException with
javax.ws.rs.BadRequestException ..etc.   That would require some adjustments in resteasy's
response and error messaging code.

asoldano: @rsearls well, the class was deprecated by Bill Burke in May 2013 (https://github.com/resteasy/Resteasy/commit/6e549642d1aa8b9d9f4f25dc8574c7a7a8b1644a#diff-192636fff108ec83395dee4ce70b425fR10), when he basically made changes to use the corresponding exception from jaxrs spec (I believe that was because previously there was no such exception in old JAX-RS api version). The deprecated exception was removed from master years after (I think I did that when started the cleanup on master).
So your addition would basically bring back something that was previously deprecated and removed; the question is really whether that's actually needed or we can go with the JAX-RS standard annotation...

rsearls: Alessio and I have discussed this.  We agreed to accept this as is.  I will create a new jira
to replace this exception and BadRequestException with the jax-rs classes and make needed
adjustments in the resteasy

PR #1875, opened 2019-02-20
[RESTEASY-2166] Review elytron testing

jbliznak:
https://issues.jboss.org/browse/RESTEASY-2166

These commits allow for testing with Elytron security domains against recent WildFly builds and running this configuration with JDK11.

The change of configuration of travis testing is just for a verification of the changes and serves as an example, I don't think we need to double the testing execution time this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 install:
- - mvn clean
+ - mvn -B clean
 script:
- - travis_wait 60 mvn -B -Ptravis -fae -Dserver.version=$SERVER_VERSION install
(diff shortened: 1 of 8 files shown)


jbliznak:
If accepted, I will prepare PR against master.

marekkopecky:
If we need to pick a single job, WF15&&JDK8 would be probably the best. But what about to use WF15&&(JDK8||JDK11)?

jbliznak:
PR for master: https://github.com/resteasy/Resteasy/pull/1877

jbliznak:
Rebased. 
@asoldano if you are OK with only WF15 && JDK8 for elytron testing then this is ready for merging.

PR #1869, opened 2019-02-15
RESTEASY-2161 remove @NoJackson annotation

liweinan:
- remove @NoJackson annotation (because jettison is removed)
- clean Jettison related tests
- update documents

PR #1878, opened 2019-02-25
RESTEASY-2161 remove @NoJackson annotation #1869 (rebased)

liweinan:
rebased version of: https://github.com/resteasy/Resteasy/pull/1869


docbook/reference/en/en-US/modules/Jaxb.xml
```diff
@@ -390,209 +376,6 @@ public class Book
...
-      <para>
-
-<programlisting>
-{ "book" :
-    {
-        "@title":"EJB 3.0",
-        "author":"Bill Burke",
-        "ISBN":596529260
-    }
-}
```
ronsigal: Hey @liweinan,

Having removed the Jettison example, maybe you could include something like "The JSON output produced by Jackson would look like this:" with a Book in JSON format.

liweinan: yes sir!

ronsigal: Lol.

ronsigal:
Ok, looks good. Thanks, Wei Nan!!

PR #1879, opened 2019-02-25
ResteasyProviderFactory refactoring / optimization

asoldano:
This PR proposes a refactoring of the huge ResteasyProviderFactory(impl) class. With the distinction of public spi and core, we can afford to split the class into smaller components without too much concerns of users having access to them. Moreover, the ResteasyProviderFactory really covers client and server aspects together, while it's quite common to need an instance for client side usage only that results in memory/cpu waste for building and populating multiple server-side-related collections. Splitting the ResteasyProviderFactory isolating client and server only aspects allows avoiding that inefficiency in some scenarios.
Besides that, whenever needing a new ResteasyProviderFactory on client side (to be wrapped into a LocalResteasyProviderFactory), instead of creating a new instance from scratch, we can cache and reuse the default ResteasyProviderFactory for each ThreadContextClassLoader (as the initial configuration is effectively function of the classloader, which controls the providers discovery); this should reduce memory allocation. 
Speaking of memory allocation reduction, I've created few singletons for the HeaderDelegate that are added to almost all ResteasyProviderFactory instance.
Finally, I've added a simple cache to the MediaTypeHeaderDelegate to reduce the resources spent on converting media type strings into the MediaType immutable objects.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
@@ -14,6 +14,7 @@ import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
+import javax.ws.rs.RuntimeType;
 import javax.ws.rs.core.Configuration;
 
 import java.security.KeyStore;
@@ -315,8 +316,7 @@ public class ResteasyClientBuilderImpl extends ResteasyClientBuilder
       if (providerFactory == null)
       {
          // create a new one
-         providerFactory = new LocalResteasyProviderFactory(ResteasyProviderFactory.newInstance());
-         RegisterBuiltin.register(providerFactory);
+         providerFactory = new LocalResteasyProviderFactory(RegisterBuiltin.getClientInitializedResteasyProviderFactory(Thread.currentThread().getContextClassLoader()));
 
          if (ResteasyProviderFactory.peekInstance() != null)
          {
@@ -424,7 +424,13 @@ public class ResteasyClientBuilderImpl extends ResteasyClientBuilder
    @Override
    public ResteasyClientBuilderImpl withConfig(Configuration config)
    {
-      providerFactory = new LocalResteasyProviderFactory(new ResteasyProviderFactoryImpl());
+      providerFactory = new ResteasyProviderFactoryImpl() {
+         @Override
+         public RuntimeType getRuntimeType()
+         {
+            return RuntimeType.CLIENT;
+         }
+      };
       providerFactory.setProperties(config.getProperties());
       for (Class clazz : config.getClasses())
       {
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
@@ -8,7 +8,7 @@ import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
 import org.jboss.resteasy.client.jaxrs.engines.ApacheHttpAsyncClient4Engine;
 import org.jboss.resteasy.client.jaxrs.engines.ClientHttpEngineBuilder43;
 import org.jboss.resteasy.client.jaxrs.i18n.Messages;
-import org.jboss.resteasy.core.ResteasyProviderFactoryImpl;
+import org.jboss.resteasy.core.providerfactory.ResteasyProviderFactoryImpl;
 import org.jboss.resteasy.plugins.providers.RegisterBuiltin;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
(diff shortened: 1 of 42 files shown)


asoldano:
Oh, there's also a commit to move the checkstyle into a maven profile (still enabled by default), so that it can be disabled when running tests locally with partially modified code ;-)

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
```diff
@@ -8,12 +8,13 @@
 import org.jboss.resteasy.client.jaxrs.engines.ApacheHttpAsyncClient4Engine;
 import org.jboss.resteasy.client.jaxrs.engines.ClientHttpEngineBuilder43;
 import org.jboss.resteasy.client.jaxrs.i18n.Messages;
-import org.jboss.resteasy.core.ResteasyProviderFactoryImpl;
+import org.jboss.resteasy.core.providerFactory.ResteasyProviderFactoryImpl;
```
jbliznak: Isn't that against java naming conventions? AFAIK package name should be all lower case

asoldano: Right, fixed

marekkopecky:
@asoldano Do you plan to backport this to 3.7?

pom.xml
```diff
@@ -220,7 +220,41 @@
                     </plugin>
                 </plugins>
             </build>
-      </profile>
+        </profile>
+        <profile>
+            <id>checkstyle</id>
+            <activation>
+                <property>
+                    <name>!no-checkstyle</name>
```
marekkopecky: @asoldano No functional objections to this, but ... if the only reason for this is an option to disable checkstyle (`-Dno-checkstyle`), the same behaviour can be reached with `-Dcheckstyle.skip=true` (see [checkstyle docs](https://maven.apache.org/plugins/maven-checkstyle-plugin/checkstyle-mojo.html#skip)) == no regression, no improvement, more lines of pom file ...

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/RegisterBuiltin.java
```diff
@@ -141,18 +179,22 @@ public Class run() throws ClassNotFoundException
             LogMessages.LOGGER.classNotFoundException(line, entry.getValue(), ex);
          }
       }
-      if (AccessController.doPrivileged(new PrivilegedAction<Boolean>() {
+      if (isGZipEnabled()) {
```
marekkopecky: s/isGZipEnabled()/gzipForCachedFactories/g ?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/RegisterBuiltin.java
```diff
@@ -30,6 +35,39 @@
...
 public class RegisterBuiltin
 {
+   private static final Map<ClassLoader, ResteasyProviderFactory> configuredClientFactories = new WeakHashMap<>();
+   private static final boolean gzipForCachedFactories = isGZipEnabled();
+
+   public static synchronized ResteasyProviderFactory getClientInitializedResteasyProviderFactory(ClassLoader cl)
+   {
+      ResteasyProviderFactory rpf = null;
+      final boolean gzip = isGZipEnabled();
+      if (gzipForCachedFactories == gzip) {
```
marekkopecky: When this can be false? (I'm probably missing some information ...)

pom.xml
asoldano: eh, I did not notice that, thanks. I'll remove the commit with this profile

asoldano:
> @asoldano Do you plan to backport this to 3.7?

No

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/RegisterBuiltin.java
asoldano: it can be false in case the resteasy.allowGzip property is changes after this class has been loaded

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/ResteasyProviderFactory.java
```diff
@@ -133,7 +133,7 @@ public static ResteasyProviderFactory newInstance()
       try
       {
          return (ResteasyProviderFactory) Thread.currentThread().getContextClassLoader()
-               .loadClass("org.jboss.resteasy.core.ResteasyProviderFactoryImpl").newInstance();
+               .loadClass("org.jboss.resteasy.core.providerfactory.ResteasyProviderFactoryImpl").newInstance();
```
jbliznak: can we change the deprecated (since java 9)  call `.newInstance()` to `.getDeclaredConstructor().newInstance()` (and extend catch clause)?

asoldano: done

ronsigal:
Hi @asoldano,

I've been convincing myself that clients on the server side are properly configured, and now I'm convinced. This is nice.

-Rono

asoldano:
Thanks @ronsigal .

@ronsigal , @marekkopecky , I've added a very simple perf test; once enabled, here it runs in ~24% less time then on current master, so the changes look promising (and the AsyncBenchTest which it was inspired from and which has no in-container client creation is also running in ~10% less time).

ronsigal:
@asoldano, that's awesome.

PR #1880, opened 2019-02-27
Misc minor perf improvements

asoldano:
Some tentative changes to try improving performances after having looked at few recent jfr recordings; this is possibly related to https://issues.jboss.org/browse/RESTEASY-2000


asoldano:
If the changes are verified to be actual improvements, this could be backported to 3.7.

PR #1890, opened 2019-03-02
[RESTEASY-754] Ported code from 2014 PR into current master.  Add mod…

rsearls:
…ule to jboss-modules for mime4j

### DIFF
--- a/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaround.java
+++ b/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaround.java
@@ -1,0 +1,228 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+package org.jboss.resteasy.plugins.providers.multipart;
+
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+
+import org.apache.james.mime4j.MimeException;
+import org.apache.james.mime4j.MimeIOException;
+import org.apache.james.mime4j.codec.DecodeMonitor;
+import org.apache.james.mime4j.dom.Message;
+import org.apache.james.mime4j.field.DefaultFieldParser;
+import org.apache.james.mime4j.field.LenientFieldParser;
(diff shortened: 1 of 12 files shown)


providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaround.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param is
+     * @return
+     * @throws IOException
+     * @throws MimeIOException
+     * @see org.apache.james.mime4j.message.DefaultMessageBuilder#parseMessage(java.io.InputStream)
+     */
+    public static Message parseMessage(InputStream is) throws IOException, MimeIOException {
+        try {
+            MessageImpl message = new MessageImpl();
+            MimeConfig cfg = new MimeConfig();
```
asoldano: By looking at the org.apache.james.mime4j.message.DefaultMessageBuilder class, I wonder if there's a reason for doing new MimeConfig() instead of using MimeConfig.DEFAULT

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaround.java
```diff
@@ -0,0 +1,77 @@
...
+            boolean strict = cfg.isStrictParsing();
+            DecodeMonitor mon = strict ? DecodeMonitor.STRICT : DecodeMonitor.SILENT;
+            BodyDescriptorBuilder bdb = new DefaultBodyDescriptorBuilder(null, strict ? DefaultFieldParser.getParser() : LenientFieldParser.getParser(), mon);
+            BodyFactory bf = new BasicBodyFactory();
```
asoldano: I have a feeling this could be too much of a simplification; we used to have a feature for writing big messages to the disk instead of using the memory, which I believe this PR would remove if merged as is.

asoldano: See org.apache.james.mime4j.storage.StorageBodyFactory for example

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/MultipartInputImpl.java
```diff
@@ -76,117 +51,6 @@
...
-            MimeStreamParser parser = new MimeStreamParser(null);
-
-            StorageProvider storageProvider;
-            if (System.getProperty(DefaultStorageProvider.DEFAULT_STORAGE_PROVIDER_PROPERTY) != null) {
-               storageProvider = DefaultStorageProvider.getInstance();
-            } else {
-               StorageProvider backend = new CustomTempFileStorageProvider();
-               storageProvider = new ThresholdStorageProvider(backend, 1024);
-            }
-            parser.setContentHandler(new BinaryOnlyMessageBuilder(this, storageProvider));
```
asoldano: This is the stuff that seems to be lost (storage based message builder)

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/MultipartInputImpl.java
```diff
@@ -507,131 +383,4 @@ public void setProviders(Providers providers)
    {
       savedProviders = providers;
    }
-
-   /**
-    * A custom TempFileStorageProvider that do no set deleteOnExit on temp files,
-    * to avoid memory leaks (see https://issues.apache.org/jira/browse/MIME4J-251)
```
asoldano: Note this jira is still open, so we probably still need to be careful and provide a workaround for it

rsearls:
I find no MimeConfig.DEFAULT in the mime4j 0.6 or 0.7.2 code base.

rsearls:
Storage support added

asoldano:
Thanks!

rsearls:
Please file a request for this in https://issues.jboss.org/browse/RESTEASY

glazzari:
Thanks @rsearls https://issues.jboss.org/browse/RESTEASY-2239

yogeshdabhi:
Is it possible to apply this patch on Resteasy-3.0.12.final? 
If yes, then which resteasy modules I need to compile? And after compilation which jars I need to add in my application? 
We already used latest mime4j 0.8.1 in our application. We cant move to latest Resteasy-4.0.0.final or Resteasy-3.7.0.final due to huge dependencies code.

PR #1891, opened 2019-03-04
[RESTEASY-2102] Renaming resteasy-validator-provider-11 to resteasy-v…

asoldano:
…alidator-provider (but keeping existing jboss modules name to be able to still deploy on existing containers), upgrading to Hibernate 6 / BV 2.0


asoldano:
> In jboss-modules/build.xml
> line 187
> 
> I don't see that you have changed the directory name.
> This would also require a rename of directory.
> jboss-modules/src/main/resources/modules/org/jboss/resteasy/resteasy-validator-provider-11
> 
> And I would think you would want to change file
> jboss-modules/src/main/resources/modules/org/jboss/resteasy/resteasy-validator-provider-11/main/module.xml
> contents accordingly.
> 
> I don't see that here. Have I missed something?

The problem is that the way resteasy distribution is installed on top of an existing wfly container does not allow _removing_ a module, we can only _add_ a new one or _modify_ existing ones. So we have no way to remove the old resteasy-validator-provider-11. Moreover, that module is referenced by the org.jboss.as.jaxrs module, which we can't modify from resteasy distribution as it includes the jaxrs subsytem jar (well, we could if we had a more advanced installation mechanism as we have in JBossWS project...)
So, what I decided to do is to let wildfly keep on using the original resteasy-validator-provider-11 module name, while the actual artifact from RESTEasy is changed to resteasy-validator-provider in the jar filename and Maven artifact name.
Does this make more sense now?

rsearls:
ok

asoldano:
@rsearls @ronsigal , after our discussion and consultation with Brian, I've added an additional commit here to also change the name of the wildfly module; the old module is turned into an alias one, referring to the new one.

ronsigal:
Ok.

PR #1894, opened 2019-03-05
RESTEASY-2009

liweinan:
Restore / migrate to Undertow excluded tests in resteasy-spring


liweinan:
work in progress, don't merge yet.

server-adapters/pom.xml
```diff
@@ -18,6 +18,7 @@
         <module>resteasy-netty4</module>
         <module>resteasy-netty4-cdi</module>
         <module>resteasy-undertow</module>
+        <module>resteasy-undertow-spring</module>
```
asoldano: @liweinan  What's this? why is it needed?

liweinan: Our current `UndertowJaxrsServer` just wrap a `HttpServlet30Dispatcher` inside:

```java
   public DeploymentInfo undertowDeployment(ResteasyDeployment deployment, String mapping)
   {
...
      ServletInfo resteasyServlet = servlet("ResteasyServlet", HttpServlet30Dispatcher.class)
              .setAsyncSupported(true)
              .setLoadOnStartup(1)
              .addMapping(mapping);
```

My thought is to provide a `UndertowJaxrsSpringServer` that extends the above `UndertowJaxrsServer`, and overrides the `undertowDeployment(...)` method above to prepare the ``org.springframework.web.servlet.DispatcherServlet` properly, and it also need to include our `springmvc-resteasy.xml` provided by `resteasy-spring` which setup the container properly), and then the spring related tests can use it to do the testing. And users get `UndertowJaxrsSpringServer` for their own usage.

If we don't provide `UndertowJaxrsSpringServer`, then I can just wrap the undertow setting up tasks into tests directly. But IMHO providing a `UndertowJaxrsSpringServer` is more useful to users. wdyt?

liweinan: The integration code inside `UndertowJaxrsSpringServer` should be something like the following:

```java
DeploymentInfo servletBuilder = Servlets.deployment()
        .setClassLoader(UndertowMain.class.getClassLoader())
        .setContextPath(CONTEXT_PATH).setDeploymentName(PKG_NAME)
        .addServlet(createDispatcherServlet(context))
        .addListener(createContextLoaderListener(context));
...
private static ListenerInfo createContextLoaderListener(WebApplicationContext context) {
    InstanceFactory<ContextLoaderListener> factory = new ImmediateInstanceFactory<>(new ContextLoaderListener(context));
    return new ListenerInfo(ContextLoaderListener.class, factory);
}

private static ServletInfo createDispatcherServlet(WebApplicationContext context) {
    InstanceFactory<DispatcherServlet> factory = new ImmediateInstanceFactory<>(new DispatcherServlet(context));
    return Servlets.servlet("DispatcherServlet", DispatcherServlet.class, factory)
            .addMapping(SPRING_DISPATCHER_MAPPING_URL)
            .setLoadOnStartup(1);
}
```

I'll work on it.

asoldano: OK, I see. Well, I thought it might probably be too much creating yet another module for a couple of classes, but if you think that makes sense and can help users, fair enough :-)

liweinan: @asoldano okay!

liweinan:
this branch needs rebase, I'll create a new PR.

PR #1898, opened 2019-03-05
RESTEASY-1419: add support for injecting Optional<T> parameter types

stevenschlansker:
https://issues.jboss.org/browse/RESTEASY-1419

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/InjectorFactoryImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/InjectorFactoryImpl.java
@@ -3,6 +3,8 @@ package org.jboss.resteasy.core;
 import org.jboss.resteasy.annotations.Form;
 import org.jboss.resteasy.annotations.Query;
 import org.jboss.resteasy.spi.ConstructorInjector;
+import org.jboss.resteasy.spi.HttpRequest;
+import org.jboss.resteasy.spi.HttpResponse;
 import org.jboss.resteasy.spi.InjectorFactory;
 import org.jboss.resteasy.spi.MethodInjector;
 import org.jboss.resteasy.spi.PropertyInjector;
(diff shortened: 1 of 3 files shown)


stevenschlansker:
Done.  Thanks in advance for the review to come :)

resteasy-core/src/main/java/org/jboss/resteasy/core/InjectorFactoryImpl.java
```diff
@@ -248,4 +265,73 @@ else if (useDefault)
...
+            this.empty = empty;
+            this.present = present;
+        }
+
+
+        static ValueInjector wrap(Type paramType, Function<Type, ValueInjector> injectorFactory) {
+            return Optional.ofNullable(optionalTypes.get(Types.getRawType(paramType)))
+                    .<ValueInjector>map(oi -> new DelegatingInjector(oi, injectorFactory.apply(oi.valueType.apply(paramType))))
+                    .orElseGet(() -> injectorFactory.apply(paramType));
+        }
```
ronsigal: Hi @stevenschlansker,

Mostly this looks pretty cool. I'm still ruminating about wrap(). You're creating an Optional<T> from T and then extracting T, right? Is that the best way to do it? Maybe it will be clear in the morning. ;-)

-Ron

stevenschlansker: Originally, a ValueInjector is responsible for creating the object that will be injected as the parameter.
This takes a type and a ValueInjector factory (loosely, the "original behavior").  If the injected type is an Optional type (as encoded in the enum "data table") this will "wrap" the original type in a DelegatingInjector that is responsible for adding the Optional wrapper.

If you have any suggestions on how to improve the clarity I'm happy to make improvements!

asoldano:
@stevenschlansker , I've merged this on master, thanks! @ronsigal might follow-up and/or propose changes in future, but I wanted to start including the changes, so that they start being tested / used.

stevenschlansker:
Awesome!  I'm of course excited to see people using it and welcome future improvements.  Congratulations on the 4.0.0.Final, we are already running it in production 😸

PR #1899, opened 2019-03-05
[RESTEASY-1715] Allow avoiding engine finalizer

asoldano:
I've modified a lot of tests to avoid them relying on finalizers to close the client, so that it's easier to figure out where the finalizer is really needed.
Then I've added a warning in the apache http client 4.3 engine finalizer, so that the user knows if that really has to close the client for the user, or not.
Finally, I've provided a copy of that engine that has no finalizer, for users brave enough to use that instead of the vanilla one and avoid the overhead of the finalizer.


docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
```diff
@@ -795,10 +795,18 @@ response.close();
...
-        Finally, if your javax.ws.rs.client.Client class has created the engine automatically for you, you should
+        If your javax.ws.rs.client.Client class has created the engine automatically for you, you should
         call Client.close() and this will clean up any socket connections.
     </para>
 
+    <para>
+        Finally, given having explicit <methodname>finalize()</methodname> methods can badly affect performances, the 
+        <classname>org.jboss.resteasy.client.jaxrs.engines.ManualClosingApacheHttpClient43Engine</classname> flavour
+        of <classname>org.jboss.resteasy.client.jaxrs.engines.ApacheHttpClient43Engine</classname> can be used. With
+        that the user is aways responsible for calling <methodname>close()</methodname> as not <methodname>finalize()</methodname>
```
jbliznak: typo "aways"

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/crypto/PKCS7SignatureSmokeTest.java
```diff
@@ -64,7 +64,6 @@ private String generateURL() {
     */
    @Test
    public void encodedData() throws Exception {
-      Client client = ClientBuilder.newClient();
```
jbliznak: You need to remove the ```client.close()``` below too since this test is using ```@AfterClass``` for closing client

docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
asoldano: will fix, thanks... I'll also fix the other typo in the same line ;-)

asoldano: Done

PR #1902, opened 2019-03-07
[RESTEASY-1955] - MicroprofileClientBuilder can't handle nested @BeanParam's

MMarus:
Resteasy issue: https://issues.jboss.org/browse/RESTEASY-1955

The verifyInterface() method in MicroprofileClientBuilder didn't take into account that @PathParam's can be also defined in the @BeanParam annotated class. Therefore this PR adds behaviour to populate @PathParam's from the @BeanParam as well in the verifyInterface() method.

Cherrypicked the https://github.com/resteasy/Resteasy/pull/1895 to the master branch as well.


asoldano:
Thanks @MMarus , I actually already cherrypicked your changes into master, see https://github.com/resteasy/Resteasy/commit/02afa0a1139cf37f8c76bf3a065d496c7fa4a629

PR #1909, opened 2019-03-08
[RESTEASY-2103] Remove distinction in validation between field and

ronsigal:
property.


ronsigal:
I accidentally named the source branch 2102b instead of 2103.

PR #1910, opened 2019-03-08
[RESTEASY-2171] JsonBindingProvider use incorrect charset encoding

iweiss:
Issue: https://issues.jboss.org/browse/RESTEASY-2171

### DIFF
--- a/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
+++ b/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
@@ -9,6 +9,8 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.ext.ContextResolver;
 import java.nio.charset.Charset;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 
 
 /**
@@ -18,6 +20,8 @@ public class AbstractJsonBindingProvider extends JsonBindingProvider {
 
    private static final String JSON = "json";
    private static final String PLUS_JSON = "+json";
+   private static final Boolean DISABLE_CHARSET = AccessController.doPrivileged((PrivilegedAction<Boolean>) () ->
+           Boolean.valueOf(System.getProperty("org.jboss.resteasy.jsonb.disable.charset", "true")));
 
    @Context
    javax.ws.rs.ext.Providers providers;
@@ -41,6 +45,11 @@ public class AbstractJsonBindingProvider extends JsonBindingProvider {
    }
 
    public static Charset getCharset(final MediaType mediaType) {
+      if (DISABLE_CHARSET)
+      {
+         return Charset.forName("utf-8");
+      }
+
       if (mediaType != null)
       {
          String charset = mediaType.getParameters().get("charset");
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java
@@ -1,7 +1,5 @@
 package org.jboss.resteasy.test;
 
-import java.io.File;
-
 /**
  * Various contants for containers defined in arquillian.xml. Keep this file in sync with the values there.
  */
@@ -11,6 +9,10 @@ public class ContainerConstants {
 
    public static final String GZIP_CONTAINER_QUALIFIER = "jbossas-manual-gzip";
 
+   public static final String ENCODING_CONTAINER_QUALIFIER = "jbossas-managed-encoding";
+
    public static final int GZIP_CONTAINER_PORT_OFFSET = 1000;
+
+   public static final int ENCODING_CONTAINER_PORT_OFFSET = 7000;
 
 }
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jsonb/basic/JsonBindingTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jsonb/basic/JsonBindingTest.java
@@ -1,9 +1,12 @@
 package org.jboss.resteasy.test.providers.jsonb.basic;
 
 import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.OperateOnDeployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.container.test.api.TargetsContainer;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.logging.Logger;
+import org.jboss.resteasy.test.ContainerConstants;
 import org.jboss.resteasy.test.providers.jsonb.basic.resource.Cat;
 import org.jboss.resteasy.test.providers.jsonb.basic.resource.JsonBindingCustomRepeaterProvider;
 import org.jboss.resteasy.test.providers.jsonb.basic.resource.JsonBindingResource;
@@ -27,6 +30,9 @@ import javax.ws.rs.core.Response;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
 
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
@@ -52,21 +58,34 @@ public class JsonBindingTest {
 
    private static final String WAR_WITH_JSONB = "war_with_jsonb";
    private static final String CUSTOM_JSON_PROVIDER = "custom_json_provider";
+   private static final String WAR_WITH_JSONB_ASCII_ENCODING = "war_with_jsonb_ascii_encoding";
 
    @Deployment(name = WAR_WITH_JSONB)
    public static Archive<?> deployWithJsonB() {
-      return deploy(WAR_WITH_JSONB, true);
+      return deploy(WAR_WITH_JSONB, true, false);
    }
 
    @Deployment(name = CUSTOM_JSON_PROVIDER)
    public static Archive<?> deployWithoutJsonB() {
-      return deploy(CUSTOM_JSON_PROVIDER, false);
+      return deploy(CUSTOM_JSON_PROVIDER, false, false);
    }
 
-   public static Archive<?> deploy(String archiveName, boolean useJsonB) {
+   @Deployment(name = WAR_WITH_JSONB_ASCII_ENCODING)
+   @TargetsContainer(ContainerConstants.ENCODING_CONTAINER_QUALIFIER)
+   public static Archive<?> deploymentWithJsonBAsciiEncoding() {
+      return deploy(WAR_WITH_JSONB_ASCII_ENCODING, true, true);
+   }
+
+   public static Archive<?> deploy(String archiveName, boolean useJsonB, boolean testEncoding) {
       WebArchive war = TestUtil.prepareArchive(archiveName);
       war.addClass(JsonBindingTest.class);
       war.addClass(Cat.class);
+      if (testEncoding) {
+         war.addAsManifestResource("jboss-deployment-structure-json-b.xml", "jboss-deployment-structure.xml");
+         Map<String, String> contextParams = new HashMap<>();
+         contextParams.put("org.jboss.resteasy.jsonb.disable.charset", "true");
+         return TestUtil.finishContainerPrepare(war, contextParams, JsonBindingResource.class);
+      }
       if (useJsonB) {
          war.addAsManifestResource("jboss-deployment-structure-json-b.xml", "jboss-deployment-structure.xml");
          return TestUtil.finishContainerPrepare(war, null, JsonBindingResource.class);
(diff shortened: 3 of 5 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jsonb/basic/JsonBindingTest.java
```diff
@@ -137,6 +137,31 @@ public void jsonbOnClientTest() throws Exception {
...
+    *                server returns json data with a value in a variable with @JsonbTransient annotation
+    *                client should not ignore @JsonbTransient annotation and should not receive a value in this variable
+    *
+    *                This test covers RESTEASY-2171.
+    *
+    * @tpPassCrit The resource returns object with correct values
+    * @tpSince RESTEasy 3.5
+    */
+   @Test
+   public void jsonbOnClientTestWithoutEncoding() throws Exception {
```
marekkopecky: This test doesn't check non-UTF8 encoding. Test needs to be fixed

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
```diff
@@ -46,7 +46,7 @@ public static Charset getCharset(final MediaType mediaType) {
          String charset = mediaType.getParameters().get("charset");
          if (charset != null) return Charset.forName(charset);
       }
-      return Charset.defaultCharset();
+      return Charset.forName("utf-8");
```
marekkopecky: Thinking about this fix again, it seems that this doesn't fully fix [RESTEASY-2171](https://issues.jboss.org/browse/RESTEASY-2171) (cc @asoldano @ronsigal ). Consider this end-point:
```java
    @GET
    @Produces("application/json;charset=us-ascii")
    @Path("a")
    public CustomObject a() throws Exception {
        return new CustomObject().setA("öü");
    }
```
And consider this RFC7159 part:

>  Note: No "charset" parameter is defined for this registration. Adding one really has no effect on compliant recipients.

Jackson2 returns UTF8 encoding, but JSON-B returns us-ascii encoding in this specific case.

testsuite/integration-tests/src/test/resources/arquillian.xml
```diff
@@ -27,6 +27,18 @@
                 <property name="managementPort">${container.management.port.manual.gzip}</property>
             </configuration>
         </container>
+        <container qualifier="${container.qualifier.managed.encoding}" mode="class">
```
marekkopecky: `mode="class"` is wrong here. We can't start this container repeatedly for each test class in TS (~491 times in integration-tests)

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java
```diff
@@ -11,6 +9,10 @@
 
    public static final String GZIP_CONTAINER_QUALIFIER = "jbossas-manual-gzip";
 
+   public static final String ENCODING_CONTAINER_QUALIFIER = "jbossas-managed-encoding";
+
    public static final int GZIP_CONTAINER_PORT_OFFSET = 1000;
 
+   public static final int ENCODING_CONTAINER_PORT_OFFSET = 2000;
```
marekkopecky: FYI: Offset 2000 is ok for 3.6.1.SPx branch, but this could cause some cherry-pick conflicts in master, because master currently uses 6 additional containers: https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java#L32

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
iweiss: I understand then that Jackson2 is wrong. If you clearly state the charset then it should be used, as JSON-B is doing.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jsonb/basic/JsonBindingTest.java
iweiss: It shouldn't be tested as the fix is for the specific case of the encoding not being provided.

testsuite/integration-tests/src/test/resources/arquillian.xml
iweiss: What should be the proper mode? Manual?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java
iweiss: OK. When I bring this to master I'll offset by 7000, thanks for the alert.

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
marekkopecky: See RFC7159: Adding one (charset) really has no effect on compliant recipients. @asoldano, @ronsigal WDYT about this?

testsuite/integration-tests/src/test/resources/arquillian.xml
marekkopecky: Manual LGTM

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jsonb/basic/JsonBindingTest.java
marekkopecky: Well, I add this my comment 3 days ago and I believe that this is fixed now, new container uses us-ascii, see https://github.com/resteasy/Resteasy/pull/1910/files#diff-49db9a69f91d736f22ab33486a4c65bbR36

Btw, If the fix is for some specific case, tests should checks just this specific case ...

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
asoldano: Honestly, I'm not super sure here, but I'd stick with the behavior that Jackson2 provider had so far (returning UTF8), which didn't lead to complaints afaiu (we're seeing the problem with JSON-B now). And the RFC7159 seems to confirm this.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java
asoldano: How about using 7000 here too?

testsuite/integration-tests/src/test/resources/arquillian.xml
asoldano: +1

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/ContainerConstants.java
asoldano: At the end of the day, this is not making much difference. I would go with the same on all branches, but I'm fine with the other solution too.

iweiss:
@marekkopecky I need your help. I'm trying to adapt Arquillian to pass `org.jboss.resteasy.jsonb.disable.charset` to the container but it's not working. I used both `javaVmArguments` and `jbossArguments` but RestEasy doesn't pick it up. If I run the test with `-Dorg.jboss.resteasy.jsonb.disable.charset=true` than the test passes.

marekkopecky:
@ronsigal Can you please check previous @iweiss question?

ronsigal:
@marekkopecky, @iweiss,

We've always passed parameters by way of context parameters, and we know it works. org.jboss.resteasy.test.core.basic.AppConfigTest is one example I've picked at random.

Note, by the way, that we've integrated MicroProfile Config on the master branch, so there are many options there. And here's a fragment from a test resource on master branch:
```
@Path("/")
public class MicroProfileConfigResource {

   static {
     System.setProperty("system", "system-system");
   }
```
So using System.setProperty() also seems to be an option. But I think a context parameter would be more consistent with current usage. The context parameters are documented here: http://docs.jboss.org/resteasy/docs/3.6.3.Final/userguide/html/Installation_Configuration.html#configuration_switches .

-Ron

iweiss:
Thanks @ronsigal. This is not working for me. I'll update the PR and would like to ask you to take a peek if you can.

providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
```diff
@@ -41,6 +45,12 @@ protected Jsonb getJsonb(Class<?> type) {
    }
 
    public static Charset getCharset(final MediaType mediaType) {
+      System.out.println("== org.jboss.resteasy.jsonb.disable.charset == " + DISABLE_CHARSET);
```
marekkopecky: @ronsigal This line looks like a typo (should be removed/moved to debug log level)

iweiss: It is a typo indeed. PR updated.

ronsigal: @iweiss, what's the status? Are you still having trouble with configuration, and should I be looking at something? Sorry - got distracted.

iweiss: @ronsigal This has been fixed on the latest push. The status for now is that the patch, including the system property, works as expected. The automated Arquillian test, on the other hand, doesn't unless I run the test using Maven with `-Dorg.jboss.resteasy.jsonb.disable.charset=true` set.

ronsigal: Ok, @iweiss, I'll play with it.

ronsigal: @iweiss, I'm still playing with this test. When I export the WAR, copy it manually to Wildfly, and run the failing tests, it works fine. For some reason, though, the context variable doesn't get set when I run the original Arquillian test. I think there's some Arquillian magic going on that I don't understand. I'll keep trying. 

Btw, I modified AbstractJsonBindingProvider to get the property either from a context variable or a system property. 

-Ron

iweiss: Thanks for the feedback @ronsigal, I'm glad I'm not the only one having issues with Arquillian.

ronsigal:
Here's an update. On the client side, TestUtil creates
```
<web-app version="3.0" xmlns="http://java.sun.com/xml/ns/javaee" 
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
 xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd"> 
<context-param>
<param-name>org.jboss.resteasy.jsonb.disable.charset</param-name>
<param-value>true</param-value>
</context-param>
</web-app>
```
On the server side, I see jbossas-managed-encoding/tmp/vfs/temp/temp7d42f54c213c7717/content-95fdab9b745f85e0/WEB-INF/web.xml:
```
<web-app version="3.0" xmlns="http://java.sun.com/xml/ns/javaee" 
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
 xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd"> 
<context-param>
<param-name>org.jboss.resteasy.jsonb.disable.charset</param-name>
<param-value>true</param-value>
</context-param>
</web-app>
```
So far, so good. But when AbstractJsonBindingProvider.getCharset() tries to access the context-params, it gets
```
&amp#27;[0m&amp#27;[31m17:06:19,496 ERROR [stderr] (default task-1) ServletConfig: io.undertow.servlet.spec.ServletConfigImpl@4f2237cd
&amp#27;[0m&amp#27;[31m17:06:19,496 ERROR [stderr] (default task-1)     resteasy.servlet.mapping.prefix->/
&amp#27;[0m&amp#27;[31m17:06:19,496 ERROR [stderr] (default task-1)     javax.ws.rs.Application->org.jboss.resteasy.utils.TestApplication
```
where "resteasy.servlet.mapping.prefix" and "javax.ws.rs.Application" are the only context-params.

@iweiss , @marekkopecky, does this suggest anything to you? Or, before I spend " 'ye whole evening of my sinking life" [1] studying Arquillian, do you know anyone who is an Arquillian guru?

[1] 1601, Mark Twain

ronsigal:
On, the other hand, there's this. I added
```
   @Path("disable")
   @GET
   public void disable() {
	   System.setProperty("org.jboss.resteasy.jsonb.disable.charset", "true");
   }
   
   @Path("enable")
   @GET
   public void enable() {
	   System.setProperty("org.jboss.resteasy.jsonb.disable.charset", "false");
   }
```
to JsonBindingResource and modified the two relevant tests:
```
   @Test
   @OperateOnDeployment(WAR_WITH_JSONB_ASCII_ENCODING)
   public void jsonbOnClientTestWithoutEncoding() throws Exception {
      URI url = new URI("http://" + PortProviderUtil.getHost() + ":"
              + (PortProviderUtil.getPort() + ContainerConstants.ENCODING_CONTAINER_PORT_OFFSET)
               + "/" + WAR_WITH_JSONB_ASCII_ENCODING + "/test/jsonBinding/");
      WebTarget target = client.target(url);
      try {
         target.path("disable").request().get();
         MediaType mediaType = MediaType.APPLICATION_JSON_TYPE;
         Entity<Cat> entity = Entity.entity(
               new Cat("Graça", "brazilian", "gray", true, JsonBindingResource.CLIENT_TRANSIENT_VALUE), mediaType);
         Cat response = target.path("repeater").request().post(entity, Cat.class);

         Assert.assertThat("Failed to return the correct name", response.getName(), is("Graça"));
         Assert.assertThat("Variable with JsonbTransient annotation should be transient, if JSON-B is used",
               response.getTransientVar(), is(Cat.DEFAULT_TRANSIENT_VAR_VALUE));
      } finally {
         target.path("disable").request().get();
      }
   }

   @Test
   @OperateOnDeployment(WAR_WITH_JSONB_ASCII_ENCODING)
   public void jsonbOnClientTestWithAsciiEncoding() throws Exception {
      URI url = new URI("http://" + PortProviderUtil.getHost() + ":"
              + (PortProviderUtil.getPort() + ContainerConstants.ENCODING_CONTAINER_PORT_OFFSET)
              + "/" + WAR_WITH_JSONB_ASCII_ENCODING + "/test/jsonBinding/");
      WebTarget target = client.target(url);
      try {
         target.path("disable").request().get();
         MediaType mediaType = MediaType.APPLICATION_JSON_TYPE.withCharset("us-ascii");
         Entity<Cat> entity = Entity.entity(
               new Cat("Graça", "brazilian", "gray", true, JsonBindingResource.CLIENT_TRANSIENT_VALUE), mediaType);
         Cat response = target.path("repeater").request().post(entity, Cat.class);

         Assert.assertThat("Failed to return the correct name", response.getName(), is("Graça"));
         Assert.assertThat("Variable with JsonbTransient annotation should be transient, if JSON-B is used",
               response.getTransientVar(), is(Cat.DEFAULT_TRANSIENT_VAR_VALUE));
      } finally {
         target.path("enable").request().get();
      }
   }
```
and now the tests all pass. I guess you don't need the special case Wildfly instance any more.

iweiss:
@ronsigal @marekkopecky That solution looks very cumbersome for something that should just work, don't you think? I wonder if there's some proper, better way to make Arquillian just comply with passing the system property and not filter it out.

ronsigal:
@iweiss, re "looks very cumbersome"

Well, actually, I think it's easier to understand that the use of a special server, but I guess that's because I hadn't looked at that infrastructure before. At this point, I'm happy with a test that works. If you can get an Arquillian expert to help, that's fine. I would like to understand what's failing. I just don't want to spend a lot more time on it right now.

ronsigal:
Hallelujah!!

PR #1912, opened 2019-03-08
Resteasy 2177 bad request exception not found exception

rsearls:


resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/BadRequestException.java
```diff
@@ -4,7 +4,10 @@
 
 /**
  * Thrown by RESTEasy when HTTP Bad Request (400) is encountered
+ *
+ * This class is deprecated use javax.ws.rs.BadRequestException
```
jbliznak: this should also use javadoc ```@deprecated``` tag, https://docs.oracle.com/javase/8/docs/technotes/guides/javadoc/deprecation/deprecation.html#javadoc_tag

asoldano:
@rsearls sorry, can you please rebase?

rsearls:
rebase complete

PR #1917, opened 2019-03-11
[RESTEASY-1427] Use JDK8 Base64

asoldano:


jbliznak:
Even after these changes there are still some usages of another implementation ```org.apache.commons.codec.binary.Base64``` in these files:

- https://github.com/resteasy/Resteasy/blob/95e39f301d9625dcd13353a56c9bbcbcb01caef3/server-adapters/resteasy-netty/src/main/java/org/jboss/resteasy/plugins/server/netty/RequestDispatcher.java
- https://github.com/resteasy/Resteasy/blob/95e39f301d9625dcd13353a56c9bbcbcb01caef3/server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/RequestDispatcher.java
- https://github.com/resteasy/Resteasy/blob/95e39f301d9625dcd13353a56c9bbcbcb01caef3/server-adapters/resteasy-vertx/src/main/java/org/jboss/resteasy/plugins/server/vertx/RequestDispatcher.java
- https://github.com/resteasy/Resteasy/blob/95e39f301d9625dcd13353a56c9bbcbcb01caef3/server-adapters/resteasy-jdk-http/src/main/java/org/jboss/resteasy/plugins/server/sun/http/BasicAuthFilter.java
- https://github.com/resteasy/Resteasy/blob/95e39f301d9625dcd13353a56c9bbcbcb01caef3/resteasy-core/src/main/java/org/jboss/resteasy/util/BasicAuthHelper.java

asoldano:
Thanks @jbliznak , I've added another comment to deal with what you noticed, please check :-)

resteasy-core/src/main/java/org/jboss/resteasy/util/BasicAuthHelper.java
```diff
@@ -13,7 +14,7 @@ public static String createHeader(String username, String password)
    {
       StringBuffer buf = new StringBuffer(username);
       buf.append(':').append(password);
-      return "Basic " + Base64.encodeBytes(buf.toString().getBytes(StandardCharsets.UTF_8));
+      return "Basic " + java.util.Base64.getEncoder().encodeToString(buf.toString().getBytes(StandardCharsets.UTF_8));
```
jbliznak: there is no need for using FQDN because only one Base64 is imported now

asoldano: Fixed,thaks

PR #1922, opened 2019-03-12
resteasy-core/MockHttpResponse: Fix charset detection

sventorben:
Given a Content-Type header with additional parameters after the charset
parameter, then the detected charset also contains the other parameters.
For example 'application/json;charset=UTF-8;version=1.0' would lead to a
charset of 'UTF-8;version=1.0'.
This commit fixes this issues by using the semicolon as a separator.


asoldano:
Thanks @sventorben for the contribution; the changes look fine.
Now, before merging, we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0. Do you agree to these terms?

PR #1936, opened 2019-03-16
[RESTEASY-1887] Additional JPMS fixes: lefovers of RESTEASY-2185, tra…

asoldano:
…nsaction api

As part of this PR's proposal, I'm reverting the changes that Rebecca applied for RESTEASY-2185: the javax.activation-api clearly is an API only subset of activation, but unfortunately the implementation classes are needed too in resteasy-multipart; so when I tried excluding the activation transitive dependency from javax.mail (which was the last remaining JPMS failure for RESTEASY-1887), I started getting NCDFE in resteasy-multipart.
With the changes in the PR I'm able to properly list the modules of a sample client project as per RESTEASY-1887 description.

### DIFF
--- a/providers/multipart/pom.xml
+++ b/providers/multipart/pom.xml
@@ -46,6 +46,10 @@
             <artifactId>apache-mime4j-storage</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.activation</groupId>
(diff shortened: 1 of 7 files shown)


asoldano:
Replaced by https://github.com/resteasy/Resteasy/pull/1948

PR #1940, opened 2019-03-18
RESTEASY-2194 fixing CDI property injector to always work on contextu…

istudens:
…al instance instead of client proxy

https://issues.jboss.org/browse/RESTEASY-2194


asoldano:
I believe we don't need fixes on 3.6 branch for now

PR #1941, opened 2019-03-18
RESTEASY-2194 fixing CDI property injector to always work on contextu…

istudens:
…al instance instead of client proxy

https://issues.jboss.org/browse/RESTEASY-2194

### DIFF
--- a/resteasy-cdi/pom.xml
+++ b/resteasy-cdi/pom.xml
@@ -61,6 +61,11 @@
             <artifactId>jboss-logging-processor</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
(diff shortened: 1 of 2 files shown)


resteasy-cdi/pom.xml
```diff
@@ -61,6 +61,12 @@
             <artifactId>jboss-logging-processor</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-api</artifactId>
+            <version>3.1.Final</version>
```
asoldano: You should avoid setting the version here and let it be driven by the dependencies bom (which I believe already includes the org.jboss.weld:weld-api dependency).

istudens:
Done. Thanks for review.

PR #1949, opened 2019-03-18
[RESTEASY-2195] replace sun security code with bouncycastle code to g…

rsearls:
…enerate V1 X509Certificate

### DIFF
--- a/security/resteasy-crypto/src/main/java/org/jboss/resteasy/security/KeyTools.java
+++ b/security/resteasy-crypto/src/main/java/org/jboss/resteasy/security/KeyTools.java
@@ -1,49 +1,116 @@
 package org.jboss.resteasy.security;
 
+import org.bouncycastle.asn1.x500.X500Name;
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier;
+import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo;
+import org.bouncycastle.cert.X509CertificateHolder;
+import org.bouncycastle.cert.X509v1CertificateBuilder;
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.crypto.util.PrivateKeyFactory;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.DefaultDigestAlgorithmIdentifierFinder;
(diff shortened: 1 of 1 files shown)


security/resteasy-crypto/src/main/java/org/jboss/resteasy/security/KeyTools.java
```diff
@@ -1,49 +1,91 @@
...
-      X509V1CertificateGenerator certGen = new X509V1CertificateGenerator();
+   /**
+    * Generate version 1 self signed {@link java.security.cert.X509Certificate}
+    *
+    * @param caKeyPair the CA key pair
+    * @param subject the subject name
+    * @return the x509 certificate
+    * @throws Exception the exception
+    */
+   public static X509Certificate generateV1SelfSignedCertificate(KeyPair caKeyPair,
```
asoldano: Can you please make this method private?

security/resteasy-crypto/src/main/java/org/jboss/resteasy/security/KeyTools.java
```diff
@@ -1,49 +1,91 @@
...
    }
 
+   /**
+    * Creates the content signer for generation of Version 1
+    * {@link java.security.cert.X509Certificate}.
+    *
+    * @param privateKey the private key
+    * @return the content signer
+    */
+   public static ContentSigner createSigner(PrivateKey privateKey) {
```
asoldano: Same as above, can you make this private?

security/resteasy-crypto/src/main/java/org/jboss/resteasy/security/KeyTools.java
```diff
@@ -1,49 +1,91 @@
...
  */
 public class KeyTools
 {
-   static
-   {
+   static {
       BouncyIntegration.init();
    }
 
-   public static X509Certificate generateTestCertificate(String subject, String issuer, KeyPair pair) throws InvalidKeyException,
```
asoldano: This method (with the signature having 3 parameters) is gone after the changes. It's probably unused, but if we want to be 100% on the safe side with backward compatibility and backport this to 3.7, it would be worth retaining this method signature too.

rsearls: Done

PR #1956, opened 2019-03-19
[RESTEASY-2195] 3.7 branch upgrade from sun security to Bouncycastle …

rsearls:
…certificate generation


asoldano:
@rsearls , thanks for the PR; it looks like the javadoc generation is failing on JDK11, can you please check and fix? Btw, something strange is why the basically equivalent changes do not cause the same failure on master? Can you please verify? Thanks!

PR #1958, opened 2019-03-20
[RESTEASY-2131] Implements use of MicroProfile Config.

ronsigal:


resteasy-core/src/main/java/org/jboss/resteasy/microprofile/config/FilterConfigSource.java
```diff
@@ -0,0 +1,55 @@
...
+      }
+      return config.getInitParameter(propertyName);
+   }
+
+   @Override
+   public String getName() {
+      ServletContext servletContext = ResteasyContext.getContextData(ServletContext.class);
+      FilterConfig filterConfig = ResteasyContext.getContextData(FilterConfig.class);
+      ServletConfig servletConfig = ResteasyContext.getContextData(ServletConfig.class);
+      return (servletContext != null ? servletContext.getServletContextName() : null) +
```
asoldano: I'm wondering whether this is something that should be computed once, stored and then reused. The reason being basically 1) stability of the result (is it possible that the name changes between multiple invocations?) 2) performances (there're multiple string concat involved in computing the name). WDYT?

resteasy-core/src/main/java/org/jboss/resteasy/microprofile/config/ServletConfigSource.java
```diff
@@ -0,0 +1,55 @@
...
+      }
+      return config.getInitParameter(propertyName);
+   }
+
+   @Override
+   public String getName() {
+      ServletContext servletContext = ResteasyContext.getContextData(ServletContext.class);
+      FilterConfig filterConfig = ResteasyContext.getContextData(FilterConfig.class);
+      ServletConfig servletConfig = ResteasyContext.getContextData(ServletConfig.class);
+      return (servletContext != null ? servletContext.getServletContextName() : null) +
```
asoldano: Same comment as above about name computation

resteasy-core/src/main/java/org/jboss/resteasy/microprofile/config/ServletContextConfigSource.java
```diff
@@ -0,0 +1,55 @@
...
+      }
+      return context.getInitParameter(propertyName);
+   }
+
+   @Override
+   public String getName() {
+      ServletContext servletContext = ResteasyContext.getContextData(ServletContext.class);
+      FilterConfig filterConfig = ResteasyContext.getContextData(FilterConfig.class);
+      ServletConfig servletConfig = ResteasyContext.getContextData(ServletConfig.class);
+      return (servletContext != null ? servletContext.getServletContextName() : null) +
```
asoldano: Same comment as above about name computation

resteasy-core/src/main/java/org/jboss/resteasy/microprofile/config/FilterConfigSource.java
ronsigal: Fair enough. I was thinking it was something that wouldn't get called a lot, but you never know. Also, I simplified the names. For example, in RESTEasy, starting with a Servlet and starting with a Filter are mutually exclusive.

resteasy-core/src/main/java/org/jboss/resteasy/microprofile/config/FilterConfigSource.java
```diff
@@ -0,0 +1,62 @@
...
+         return null;
+      }
+      return config.getInitParameter(propertyName);
+   }
+
+   @Override
+   public String getName() {
+      if (name == null) {
+         synchronized(this) {
+            if (name == null) {
```
asoldano: @ronsigal , shouldn't the "name" field be volatile for this to work properly?

ronsigal: D'oh. I knew there was something bothering me about that. Fixed. Thanks.

PR #1961, opened 2019-03-21
Others# Cannot parse MultivaluedMap in restful service

terryyrliang:
1. From org.jboss.resteasy.core.MessageBodyParameterInjector.inject(HttpRequest, HttpResponse, boolean)
it call HttpServletInputMessage.formParametersRead before calling HttpServletInputMessage.getDecodedFormParameters, but the first value assigment of variable decodedFormParameters is in method HttpServletInputMessage.getDecodedFormParameters.
2. The resteasy would not parse the MultivaluedMap parameter because the parameter check without parameter initialization

Change-Id: I5dc2ccaf9bb070b185645ea780079bb1fb0574f3


jamezp:

PR #1965, opened 2019-03-22
[RESTEASY-2131] Allow ConfigProvider to resolve RESTEasy's configured…

asoldano:
… ConfigSource classes even when the TCCL is not set


asoldano:
Replaced by #1966

PR #1968, opened 2019-03-22
[RESTEASY-1432] - Interface with generic type does not work with client proxy

MMarus:
RESTEASY issue: https://issues.jboss.org/browse/RESTEASY-1432

There was a problem with getting the GenericType from methods inside interfaces which use generics.


ronsigal:
Nice. Thanks, @MMarus.

PR #1969, opened 2019-03-26
[RESTEASY-2201] update jackson in jose-jwt

asoldano:


marekkopecky:
WF module.xml needs to be updated as well

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/jose/JWSTest.java
```diff
@@ -47,8 +48,8 @@
    @Test
    public void testHeaderSerialization() throws Exception {
       ObjectMapper mapper = new ObjectMapper();
-      mapper.setSerializationInclusion(JsonSerialize.Inclusion.NON_NULL);
```
marekkopecky: This is backward incompatible change,but jose-jwt module is private (see [module.xml](https://github.com/wildfly/wildfly/blob/master/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/resteasy/jose-jwt/main/module.xml#L28)), so WF users shouldn't use this directly.

asoldano: Yep, exactly what I was thinking, but I'd like to have your (QE) go/no-go on this before merging.

asoldano:
> WF module.xml needs to be updated as well
> 
> https://github.com/wildfly/wildfly/blob/master/feature-pack/src/main/resources/modules/system/layers/base/org/jboss/resteasy/jose-jwt/main/module.xml

That would be done when upgrading the resteasy version in wfly, yep

asoldano:
Having heard no real objections, I'm merging this (cc @marekkopecky )

PR #1985, opened 2019-04-06
ResteasyJackson2Provider: fix ConcurrentHashMap contention

asoldano:
Master version of #1984 

Replace Jackson's AnnotationBundleKey with an implementation that uses `.equals()` instead of `==` to compare annotations. AnnotationBundleKey assumed annotation references were unique which breaks because the JRE doesn't cache parameter annotations, returning new references each time parameter annotations are queried. This was causing a severe performance degradation over time since matching ClassAnnotationKey instances had the same `hashCode` while failing equality checks, leading to a large and growing single bucket with identical objects and lock contention.


jvmccarthy:
@asoldano thanks for the PR to master! To repeat here, I agree to contributing under the terms of the Apache 2.0 license.

asoldano:
Thanks @jvmccarthy

PR #1988, opened 2019-04-08
[RESTEASY-2196] CompletionStageRxInvokerImpl should not block on a th…

anilgursel:
…read.


asoldano:
Thanks @anilgursel and sorry for the late reply, I've been busy lately. Today I'm reviewing your PR which definitely looks interesting. I might possibly propose you an improvement over your proposal later today/tomorrow, please let me work a bit on this.
In the mean time, we require that all contributions are made under the terms of the ASL2.0 License: http://www.apache.org/licenses/LICENSE-2.0 Do you agree to these terms?
Thanks!

asoldano:
@anilgursel , I've created #1990 , I'm wondering if you can review it. I still need to have it benchmark tested in any case, to see if it really gives benefits.

anilgursel:
@asoldano Yes, I agree on the Apache License 2.0 terms.

asoldano:
Thanks @anilgursel ; btw I'm closing this as we're going on with the discussion on #1990

PR #1990, opened 2019-04-09
[RESTEASY-2196] CompletionStageRxInvokerImpl should not block on a thread

asoldano:
A revisited version of #1988 ; I've removed a class that is not needed and refactored the ClientInvocation methods a bit to avoid allocation of functions and lambdas when not needed and reuse methods a bit.


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/CompletionStageInvoke.java
```diff
@@ -1,176 +0,0 @@
...
-import javax.ws.rs.core.Response;
-import java.util.concurrent.CompletionStage;
-
-/**
- * An invoker for asynchronous invocation of HTTP methods that returns {@link CompletionStage}
- *
- * @author <a href="mailto:anilgursel@gmail.com">Anil Gursel</a>
- * @version $Revision: 1 $
- */
-public class CompletionStageInvoke
```
anilgursel: The reason I created this class was just to align with current design of RestEasy where it has `AsynchronousInvoke`. But, I see no issue getting rid of it.  It is cleaner not to have it.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
```diff
@@ -589,9 +545,92 @@ public T extractResult(ClientResponse response)
...
+      }
+   }
+
+   private <T> Function<AsyncClientHttpEngine, Future<T>> createAsyncSubmitFunction(final boolean buffered,
+         final InvocationCallback<T> callback, final AsyncClientHttpEngine.ResultExtractor<T> extractor)
+   {
+      return asyncClientHttpEngine -> asyncSubmit(
+            ext -> (callback != null
+                  ? asyncClientHttpEngine.submit(this, buffered, callback, ext)
+                  : asyncClientHttpEngine.submit(this, buffered, ext, client.asyncInvocationExecutor())),
```
anilgursel: For the scenarios where a `Future` would be returned but the `callback` is not passed, it would cause the `CompletableFuture` path to be taken.  It may not be the desirable thing to do as the default implementation in `AsyncHttpClientEngine` for `CompletableFuture` blocks on `Future`.  So, for the engines that do not override `CompletableFuture` version of `submit`, the execution will now be blocking.

asoldano: OK, please bear with me and help me understand better.
Are you basically saying that we should really keep separate `submit` / `submitCF` methods in `ClientInvocation`? (in the refactoring commit I've removed the `submitCF` ones except for the one that accept a callback as parameter).
We haven't overridden the `CompletableFuture` versions of submit in `AsyncHttpClientEngine` anywhere. I was actually a bit puzzled about that earlier, with the changes for this jira we're removing the blocking get in the `CompletionStageRxInvokerImpl`, but adding it in the default methods in `AsyncHttpClientEngine`: can you clarify how is that resulting in an improvement? Thanks

anilgursel: > ... how is that resulting in an improvement?

It allows `AsyncClientHttpEngine` implementation to override the following and not block:

```java
default <T> CompletableFuture<T> submit(ClientInvocation request, boolean buffered, ResultExtractor<T> extractor)
```

For instance, I am working on a Reactor Netty based `ReactorNettyClientHttpEngine` which will implement the `default` method in `AsyncClientHttpEngine` and it won't block in `default <T> CompletableFuture<T> submit` method (I will send a PR in a few weeks).

You are right, with the current `ApacheHttpAsyncClient4Engine`, it does not make a difference.  But, in later versions, `ApacheHttpAsyncClient4Engine` can implement the `default <T> CompletableFuture<T> submit` method, and not block in it.  It should not be much work and can be achieved as part of this effort.  However, there is the `JettClientEngine` as well.  Apart from these, there may be some other custom `AsyncClientHttpEngine` that RestEasy users might have implemented on their own.  I would not want to create a regression for them.

> Are you basically saying that we should really keep separate `submit` / `submitCF` methods in `ClientInvocation`? 

It is not ideal; however, we may have to wait until we can get to a state that we do not need to provide `default` implementations in `AsyncClientHttpEngine`.  The `callback` itself cannot be the factor to decide which `submit` method in `AsyncClientHttpEngine` to call.

asoldano: OK, great, thanks a lot for the clarification. I'll revert the 2 refactoring commits later tonight and see if I can cleanup the code a bit without merging the two execution paths (`Future` and `CompletableFuture` ones). Will tag you for an additional review if you don't mind :-)

I'm really looking forward to see review your future PR(s) for new `AsyncClientHttpEngine` implementations.

anilgursel: Sure!  Looking forward to the update!

asoldano:
@anilgursel , it has taken a while, but I've just force pushed a new refactored version of the changes, taking your comments above in consideration. Perhaps you have a bit of time to review.
If it's ok, I'll probably add some comments and maybe some additional tests.

Something I'm wondering is whether the new executorSubmit method is needed (https://github.com/resteasy/Resteasy/pull/1990/files#diff-404c9c65109fd0e824e42bebbd542f8cR847) and why it does not close the response as the other is doing... any explanation?
Thanks!

anilgursel:
Please feel free to add more comments and tests.  The more the better 😄 

`executorSubmit` is used for the scenario where an async api is used but the `ClientHttpEngine` in use is not an `AsyncClientHttpEngine` instance.  I do not think we can rely on the existing one as it returns a `Future`.

The other `executorSubmit` closes the response only if `callback != null`, and I believe it is probably because of the comment in JAX-RS code:

```java
     * Once this invocation callback method returns, the underlying {@link javax.ws.rs.core.Response}
     * instance will be automatically closed by the runtime.
     * </p>
     *
     * @param response response data.
     */
    public void completed(RESPONSE response);
```

So, if `callback` is not provided, it will not close the response.  That is why I did not close it.  Closing the response will not allow working on the response object on the chained functionalities of `CompletableFuture`.

asoldano:
@anilgursel thanks for the review, I've re-pushed the last commit, I agree on the DRY principle, it should be better now. I've splitted the longest lines. Regarding the executorSubmit method, right, I didn't pay enough attention to the check on the callback where the response is closed; I believe the two methods can be merged into one, using CompletableFuture, WDYT? (I've done that in the commit here, but clearly I can rollback that if needed)

asoldano:
@anilgursel , I've copied one of the existing test for the rx() client feature and made it use the async client engine. The test shows some failures (and his hence excluded for now), maybe you want to have a look and see if it's because some usage scenarios should simply be forbidden or what. It's late here now I might not have time to get back on this before Monday next week, unfortunately.

asoldano:
@anilgursel , I believe I might have solved the problem with the last two commits; the test is now passing, I've basically implemented the new submit methods returning CompletableFuture in the two currently available implementations of AsyncClientEngine. Any feedback, just add comments ;-)

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ApacheHttpAsyncClient4Engine.java
```diff
@@ -190,6 +194,49 @@ public ClientResponse extractResult(ClientResponse response)
...
    }
 
+   @Override
+   public <T> CompletableFuture<T> submit(ClientInvocation request,
+         boolean buffered,
+         ResultExtractor<T> extractor,
+         ExecutorService executorService) {
+      if (buffered) {
+         final CompletableFuture<T> cf = new CompletableFuture<>();
+         final InvocationCallback<T> callback = new InvocationCallback<T>()
```
anilgursel: This would cause `Response` object to be closed.  So, any chaining on `CompletableFuture` will not be able to work on `Response` object (especially if it is happening on a separate thread).  We can use `FutureCallback`  without a `InvocationCallback` in the picture, actually.

anilgursel: Please see Javadoc if `InvocationCallback`.  Sorry, I have to go offline now :(

anilgursel:
@asoldano  Without our changes in the picture, in `public <T> CompletionStage<T> get(Class<T> responseType)` method on `master` branch, if you replace:

* `return CompletableFuture.supplyAsync(() -> builder.get(responseType));` 
* with return `builder.async().get(responseType).get()` 

you will see the same problem.

I would expect `builder.get(responseType)` and `builder.async().get(responseType).get()` behave the same (again, `CompletableFuture` and any of the stuff in the PR is out of the picture here).  This suggests me there is an issue somewhere else.

asoldano:
@anilgursel thanks for your feedback; I'm failing in reproducing what you're saying. Can you share a code snippet with the reproducer of the problem? Btw, buffering (and hence the 'InvocationCallback') should end up being used for the 'rx(...)' methods only, which return 'CompletionStage<T>'. The 'rx(..)' methods, which return 'CompletionStage<Response>' results in 'buggering = false' and no callback for returning the 'CompletableFuture' instance.

asoldano:
I've edited the previous commit, realized that I used wrong quotes for highlighting codes, resulting a in misleading message

anilgursel:
@asoldano Sorry for the delayed response.  Thanks for the clarification on `buffering=false`!  It sounds good.  

Here is a commit to reproduce the problem I was mentioning on `master` https://github.com/anilgursel/Resteasy/commit/082fe986484e4b50526046e546e512563f497ae4.

asoldano:
Thanks @anilgursel ,  I've tried cherry-picking your commit locally on top of master, enabled it by commenting out the `@Ignore` on the class and ran it. I can see the same error I was getting some days ago ("blocking reads inside an async io-handler are not allowed"). I believe that's now addressed in this PR, also considering the test I added is passing and that has `useAsyncHttpEngine()` called on the client. Would you be OK with me merging the current PR (except perhaps some additional comments/javadoc) ?

anilgursel:
@asoldano Yes, I am ok to merge.

asoldano:
@anilgursel cool, thanks a lot for your contribution here, I really appreciated.
I'm looking forward for your next PR(s) on the ReactorNettyClientHttpEngine :-)

anilgursel:
@asoldano Thanks for welcoming contributions!  Yes, I will submit a PR soon for `ReactorNettyClientHttpEngine `.

Between, what is the timeline for RestEasy 4.1?

asoldano:
> @asoldano Thanks for welcoming contributions! Yes, I will submit a PR soon for `ReactorNettyClientHttpEngine `.

Great!


> 
> Between, what is the timeline for RestEasy 4.1?

Not completely decided yet, but likely 4 months from now. Having said this, we might decide to release a 4.0.1 way earlier than that, with a selection of fixes from the current master (I'm trying to wrap up the 4.0.0.Final this week), in which case this PR would definitely be included.

PR #2001, opened 2019-04-26
[RESTEASY-2229] Cs wrapped exception

FroMage:
When CompletionStage completes exceptionally its exceptions are not wrapped, but when they complete exceptionally due to an exception throw in one of its pipeline callbacks (in thenApply() for example) then, CompletionStage wraps the exception before passing it to us. We should unwrap it, otherwise our exception mappers will never match and we get internal errors instead of the expected answers.


FroMage:
Thanks!

PR #2009, opened 2019-04-29
[RESTEASY-2230] check for method private declaration

rsearls:


rsearls:
This is addressing CDI spec, where Section 3.15. "Unproxyable bean types, rule 3
  - classes which have non-static, final methods with public, protected or default visibility,
see https://issues.jboss.org/browse/RESTEASY-1015

Resteasy is not to add @Scope annotation in this case

asoldano:
@rsearls so does this mean this is a proper fix for RESTEASY-1015? What's the relationship with RESTEASY-2230 / Netty then?

rsearls:
RESTEASY-1015 was a requested fix for classes only declared final.  It did not identify any issues with methods declared final.   However I did test with this fix at that time.  It caused the netty tests to fail, so I removed it.  

Ron and I agreed to accept the RESTEASY-1015 and open a new jira against netty for the following reasons.
    * I wanted the RESTEASY-1015 fix to get into the May code freeze.
    * There is no reported issue of classes with methods declared final.
    * I am not familiar with any of the netty code and did not know how long it would take to determine
      if this code change would be appropriate.
    * A change of this sort would be high risk to resteasy in general.  I did not want to risk breakage at 
      this time.



In that fix I also added this code change which caused netty tests to fail, so I removed it.
Not knowing netty or how long it 
might take to understand and any change that should or should not be taken, I asked the fix for RESTEASY-1015 to be accepted and I would create this jira for later investigation.

ronsigal:
@rsearls, I'm a little confused. You said, "I also added this code change which caused netty tests to fail", but the testsuite passes with this PR ... .

rsearls:
I didn't have the check in place for the initial class, so the methods for all super classes was being checked.
One of those methods in class Object is declared final.

ronsigal:
Ah, @rsearls, I see. Thanks. Actually, when the spec says "classes which **have** non-static, final methods with public, protected or default visibility", I would have guessed it meant to include inherited methods as well. But, given that fact about Object, it obviously can't. They should have said, "classes which **declare**. Anyway ... .

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
```diff
@@ -271,7 +272,26 @@ private void setBeanManager(BeanManager beanManager)
...
+      boolean isFinal = Modifier.isFinal(clazz.getModifiers());
 
+      if (!isFinal) {
+         // check methods
+         for (Method m : clazz.getMethods()) {
+            if (clazz == m.getDeclaringClass())
+            {
+               int mod = m.getModifiers();
+               if (Modifier.isFinal(mod) && !Modifier.isStatic(mod)
+                  && (Modifier.isPublic(mod) || Modifier.isProtected(mod)))
```
ronsigal: I think !Modifier.isPrivate(mod) would be better: it would capture the default visibility case as well.

ronsigal: @rsearls, great. Thanks.

I don't know why, but github says you made the change "3 hours from now". We should contact some quantum physicists. Lol.

PR #2012, opened 2019-05-03
[RESTEASY-754] Ported code from 2014 PR into current master.  Add mod…

yogeshdabhi:
…ule to jboss-modules for mime4j


asoldano:
@yogeshdabhi , I don't see the point of this PR, the fix you're proposing is already on master (https://github.com/resteasy/Resteasy/commit/e76ad11deacf99398b0dc3836aa67b5ed8a88e63) as a matter of fact you get conflicts here. Closing this.

PR #2013, opened 2019-05-03
[RESTEASY-2232] Don't ignore MessageBodyWriters without @Produces

ronsigal:
annotation


asoldano:
@ronsigal , there seems to be a checkstyle violation, can you please check / fix ? :-)

PR #2015, opened 2019-05-03
[RESTEASY-2198] Ignore media type "application/signed-exchange"

ronsigal:


ronsigal:
Don't commit yet - I forgot to fix up the javadoc in the test.

PR #2028, opened 2019-05-09
[RESTEASY-2176]:Client-side SSL - HostnameVerificationPolicy.STRICT s…

jimma:
…houldn't accept wildcard certificates


ronsigal:
@jimma, Chapter 50.3. "Apache HTTP Client 4.x and other backends" of the Resteasy doc refers to the HostnameVerifier. Maybe a note about the default behavior?

ronsigal:
> @jimma, Chapter 50.3. "Apache HTTP Client 4.x and other backends" of the Resteasy doc refers to the HostnameVerifier. Maybe a note about the default behavior?

Never mind. There are so many options in ResteasyClientBuilder ...

PR #2029, opened 2019-05-09
[RESTEASY-1407] Prevent InputStreamProvider from writing Content-Length

ronsigal:
when using HttpClient.


ronsigal:
Tests failing with another gzip issue. Happened to me at home, as well.

ronsigal:
I don't like this solution, which sets a JVM wide flag.

PR #2031, opened 2019-05-12
[RESTEASY-2176]:Client-side SSL - HostnameVerificationPolicy.STRICT

ronsigal:
shouldn't accept wildcard certificates


ronsigal:
Copied from @jimma's https://github.com/resteasy/Resteasy/pull/2028.

jimma:
@ronsigal  We don't a docbook section to talk about ssl configuring. I added in https://github.com/resteasy/Resteasy/pull/2032, and  the content about  HostnameVerificationPolicy is  under the ssl configuration section. please help review and correct.

PR #2036, opened 2019-05-22
[RESTEASY-2249] Fixing private PostConstruct method detection

spyrkob:
JIRA: https://issues.jboss.org/browse/RESTEASY-2249

@ronsigal I think there's also a bug in code detecting Interceptor PostConstruct methods in JaxrsInjectionTarget - it only checks for parameters, but the spec allows those methods to have a return type and throw a checked exception as well. I tried fixing that, but I was not able to get your EJB test in PostConstructInjectionTest to work (currently TestPostInjectCdiOn only checks /normal/get path never /ejb/get)


resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/util/Types.java
```diff
@@ -810,7 +810,7 @@ public static boolean hasPostConstruct(Class<?> clazz)
 
    public static boolean hasPostConstruct(Class<?> clazz, Function<Method, Boolean> validateParameterCount)
    {
-      for (Method m : clazz.getMethods())
+      for (Method m : clazz.getDeclaredMethods())
```
ronsigal: Hmmm. You're right, getMethods() returns only public methods. On the other hand, getDeclaredMethods() doesn't look at inherited methods. Probably needs to call getDeclaredMethods() while recursively visiting all superclasses. Thank you.

ronsigal:
Hi @spyrkob,

I'm thinking I might have looked at https://docs.oracle.com/javaee/5/api/javax/annotation/PostConstruct.html, which says "The return type of the method MUST be void", instead of the newer https://docs.oracle.com/javaee/7/api/javax/annotation/PostConstruct.html.

-Ron

ronsigal:
re: "I was not able to get your EJB test in PostConstructInjectionTest to work (currently TestPostInjectCdiOn only checks /normal/get path never /ejb/get)"

Ah, yes. Thanks again. I'll have to look into that.

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/util/Types.java
spyrkob: Thanks @ronsigal I didn't think of that.
Added a recursive check for superclasses, if it looks good, I'll squash the two commits

ronsigal: So far, so good, @spyrkob. Thanks.

There's also the alternative signatures for Interceptors, but I still have to look at the EJB case.

ronsigal:
Well, @spyrkob, I've learned two things about the EJB test.
1. It makes no sense the way it is written. The @PostConstruct method should be in an interceptor.
2. When the test is fixed, it fails because validation called from JaxrsInjectionTarget.inject() occurs before the call to the @PostConstruct method.

I've created RESTEASY-2264 "Validation and @PostConstruct, Part II" with an updated version of PostConstructInjectionTest. @spyrkob, could you update PostConstructInjectionTest and related files in your pull request? Thanks.

spyrkob:
Added the test changes from RESTEASY-2264 with disabled EJB test. Lemme know if it looks all right, and I'll squash the commits.

spyrkob:
@ronsigal I forgot to follow up on this, sorry. Does the PR look good to be merged in?

ronsigal:
Hi @spyrkob,

I also forgot all about it. Thank you for cleaning up the EJB mess I made. It looks good. In the interest of best practices, yes, could you squash the commits?

Also, do you feel like doing a PR for branch 3.10.0.Final?

-Ron

spyrkob:
Thanks @ronsigal 
Ports for 3.10 (#2164) and 3.6.1 (#2161) are up.

PR #2037, opened 2019-05-22
[RESTEASY-2251] Test for DynamicFeature

tterem:
Jira: https://issues.jboss.org/browse/RESTEASY-2251
3.7: https://github.com/resteasy/Resteasy/pull/2038


tterem:
1. see my answer on jira
2. well, RESTEASY-2251 is a task to write a test and RESTEASY-2246 is a bug report
3. You are right about commenting out, I  rather splitted test into two tests and used Ignore annotation on failing test.

PR #2043, opened 2019-05-23
Rest Client MP 1.2 fork from SmallRye

patriot1burke:
Fixed exception mappers
Fixed Async bugs


testsuite/microprofile-tck/pom.xml
```diff
@@ -55,8 +55,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.wildfly</groupId>
-      <artifactId>wildfly-microprofile-config-smallrye</artifactId>
+      <groupId>io.smallrye</groupId>
+      <artifactId>smallrye-config</artifactId>
+      <version>1.3.6</version>
```
asoldano: Out of curiosity, why is this needed?
In any case, we should also fix the resteasy-dependencies-bom/pom.xml accordingly.

patriot1burke: Implementation of rest client depends on config :(

resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/BuilderResolver.java
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright 2015-2017 Red Hat, Inc, and individual contributors.
```
ronsigal: Should we update the date?

rsvoboda:
We should also sync the codebase in master 
e.g. SR Config version in master is 1.3.3 and in the branch it's 1.3.7, there is still wildfly.wildfly-microprofile-config-smallrye and version.microprofile.restclient is 1.0.1 (branch has 1.2.1)

asoldano:
@rsvoboda https://github.com/resteasy/Resteasy/pull/2045

PR #2046, opened 2019-05-24
MP Rest Client SSL support

michalszynkiewicz:


resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/header/ClientHeadersRequestFilter.java
```diff
@@ -92,8 +90,6 @@ public void filter(ClientRequestContext requestContext) {
                 .forEach(
                         (key, values) -> requestContext.getHeaders().put(key, castToListOfObjects(values))
                 );
-
-        ResteasyProviderFactory.getContextDataMap().put(HttpHeaders.class, new HttpHeadersContextProvider(requestContext));
```
michalszynkiewicz: There used to be a test in MP Rest Client TCK that required it but the test is removed now.

michalszynkiewicz:
To be able to get a successful build, I added exclusions for tests that verify features that are not yet implemented.

resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/header/ClientHeadersRequestFilter.java
asoldano: Great one less thing in the map there.

michalszynkiewicz:
The PR for 3.8: https://github.com/resteasy/Resteasy/pull/2047
This one can be closed, right?

michalszynkiewicz:
Turns out this should be done in 3.8 only, closing.

PR #2053, opened 2019-05-29
[RESTEASY-2165]:Add json merge patch support

jimma:


providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
```diff
@@ -32,36 +32,39 @@
...
+   @SuppressWarnings(
+   {"rawtypes", "unchecked"})
    public void filter(ContainerRequestContext requestContext) throws IOException
    {
-      //Strict the filter is only executed for patch method and media type is APPLICATION_JSON_PATCH_JSON_TYPE
       if (requestContext.getMethod().equals("PATCH")
-            && MediaType.APPLICATION_JSON_PATCH_JSON_TYPE.isCompatible(requestContext.getMediaType()))
+            && MediaType.APPLICATION_JSON_PATCH_JSON_TYPE.isCompatible(requestContext.getMediaType())
+            || requestContext.getMethod().equals("PATCH")
+            && APPLICATION_JSON_MERGE_PATCH_JSON_TYPE.isCompatible(requestContext.getMediaType()))
```
asoldano: Can you rework the if conditions so that requestContext.getMethod().equals("PATCH") is executed once?

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
```diff
@@ -80,23 +94,43 @@ public void filter(ContainerRequestContext requestContext) throws IOException
...
          }
+
       }
 
    }
 
+   private Object getTarget() throws ProcessingException
+   {
+      HttpRequest request = ResteasyContext.getContextData(HttpRequest.class);
+      HttpResponse response = ResteasyContext.getContextData(HttpResponse.class);
```
asoldano: You should probably pass 'request' and 'response' objects here from the caller as parameters instead of getting them again from the context here

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
```diff
@@ -80,23 +94,43 @@ public void filter(ContainerRequestContext requestContext) throws IOException
...
+      request.setHttpMethod("GET");
+      Registry methodRegistry = ResteasyContext.getContextData(Registry.class);
```
asoldano: Same as above

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/PatchMethodFilter.java
```diff
@@ -80,23 +94,43 @@ public void filter(ContainerRequestContext requestContext) throws IOException
...
+      ResourceInvoker resourceInovker = methodRegistry.getResourceInvoker(request);
```
asoldano: Same as above; getting the ResourceInvoker from the registry given the request is expensive, you should use the data you already got in the caller.

jimma:
@asoldano  Fixed and refactored.

PR #2056, opened 2019-05-30
 [Resteasy-2246] Dynamic features should be resolved once for each resource method

NicoNes:


NicoNes:
Hi @asoldano and @tterem 

This PR is a fix proposal for the `DynamicFeature` multiple invocations issue.

Its is based on @tterem  PR https://github.com/resteasy/Resteasy/pull/2037 where I just removed the `@Ignore("RESTEASY-2246")` on test `testDynamicFeatureProcessing`.

As you will maybe notice, I try in my first commit to no change the public interface `JaxrsInterceptorRegistryListener` to preserve backward compatibility.
 So I was doing a kind of merge between old and new providers. Unfortunately many others tests was failing because, doing things this way I was loosing the sorting provided by `JaxrsInterceptorRegistryImpl.createArray()` invoked in the `postMatch()` method.

So in my second commit I rollback everything and modify interface `JaxrsInterceptorRegistryListener` to get the new provider registered on the `parentProviderFactory` that trigger the update.
This way no need to rebuild no new `resourceMethodProviderFactory` and reconfigure everything, we just have to update filter and interceptor registry based on `parentProviderFactory`  updates.

Let me know what you think please.

asoldano:
@NicoNes thanks, we'll review soon!

PR #2059, opened 2019-06-03
[RESTEASY-2265] Statistics

rsearls:


resteasy-core/src/main/java/org/jboss/resteasy/statistics/MethodStatisticsLoggerImpl.java
```diff
@@ -0,0 +1,75 @@
...
+   @Override
+   public long getAvgExecutionTime() {
+      long avgExecTime = -1;
+
+      try
+      {
+         synchronized (this)
+         {
+            avgExecTime = totalExecutionTime / invocationCnt;
+         }
```
asoldano: Why is the synchronized block here within the 'try' part of the try-catch block? Shouldn't the 'catch' part also run in the synchronized block ?

asoldano:
can you either use the WFLY jira reference or create a RESTEasy jira (and reference it) for this feature?

PR #2062, opened 2019-06-05
[RESTEASY-2266] Multiple ApplicationPath fix

anilgursel:


anilgursel:
@asoldano @ronsigal Can you please review the change?  You can find more details at https://issues.jboss.org/browse/RESTEASY-2266.

Only openjdk8 SERVER_VERSION=16.0.0.Final failed, all others passed.  I do not think the failure is related to this change though:

```
[INFO] Running org.jboss.resteasy.test.providers.sse.SseTest
[ERROR] Tests run: 12, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 38.314 s <<< FAILURE! - in org.jboss.resteasy.test.providers.sse.SseTest
[ERROR] testReconnect(org.jboss.resteasy.test.providers.sse.SseTest)  Time elapsed: 32.531 s  <<< FAILURE!
java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.jboss.resteasy.test.providers.sse.SseTest.testReconnect(SseTest.java:309)
```

ronsigal:
Hi @anilgursel,

First of all, I want to thank you for working on this problem. So far, I'm not entirely clear why this fix is working. Line 103 in ServletContainerDispatcher.init() is executed only if useGlobal is true, which isn't normally the case. I'll try running your failing test case tomorrow and maybe it'll help me understand.

By the way, I don't know if this is relevant, but this isn't the first time we've run into problems with multiple Applications. There's also https://issues.jboss.org/browse/RESTEASY-1709 "Same JAX-RS Application instance injected across applications".

-Ron

anilgursel:
@ronsigal  Thanks for looking into the issue!

When we debugged with Resteasy 3.6.2, `useGlobal` was also true and multiple `ApplicationPath` scenario was working.  So, we did not look much into `useGlobal`.  But, there may be a core underlying issue which might had been somehow hidden in 3.x series.

ronsigal:
Hi @anilgursel, 

Well, I see why useGlobal is getting set to true.
>org.jboss.resteasy.springboot.ResteasyAutoConfiguration.resteasyBootstrapListener()

creates a ResteasyProviderFactory and stores it in the ServletContext. Then ServletContainerDispatcher.init() finds it and sets useGlobal to true:
```
      ResteasyProviderFactory globalFactory = (ResteasyProviderFactory) servletContext.getAttribute(ResteasyProviderFactory.class.getName());
      if (globalFactory == null && ctxDeployment != null) {
         globalFactory = ctxDeployment.getProviderFactory();
      }
      Dispatcher globalDispatcher = (Dispatcher) servletContext.getAttribute(Dispatcher.class.getName());
      if (globalDispatcher == null && ctxDeployment != null) {
         globalDispatcher = ctxDeployment.getDispatcher();
      }

      String application = bootstrap.getInitParameter("javax.ws.rs.Application");
      String useGlobalStr = bootstrap.getInitParameter("resteasy.servlet.context.deployment");
      boolean useGlobal = globalFactory != null;
```
Also, ResteasyAutoConfiguration.resteasyBootstrapListener() creates a SynchronousDispatcher and stores it in the ServletContext, where ServletContainDispatcher.init() finds it (the same one every time) and calls

>contextDataMap.putAll(dispatcher.getDefaultContextObjects());

I think the history there is that, originally, RESTEasy didn't handle more than a single Application is a web application, and the use of a single global ResteasyProviderFactory and Synchronous dispatcher is meant to replicate the earlier behavior.  Normally, though, that wouldn't happen. Wildfly, for example, doesn't do that.

So, I have two questions now.

1. What changed since 3.6.2?
2. @liweinan, is the use of global objects necessary in the resteasy-spring-boot-starter?

-Ron

anilgursel:
@ronsigal Thanks for the analysis.  However, the multi `ApplicationPath` works with 3.6.2 (without the recent configuration related changes).  Maybe it was never meant to work and the underlying core issue was somehow shadowed.  

Between, we are facing an issue in resteasy-spring-boot starter tests.  Not sure if it is related to this issue or not, but it is related to recent configuration changes.  [This assertion](https://github.com/sharathsrinivasa/resteasy-spring-boot/blob/upgrade-2-resteasy-4/resteasy-spring-boot-starter-test/src/test/java/org/jboss/resteasy/springboot/AsyncJobIT.java#L53) is failing with Resteasy 4.0.0.Final.  Upon debugging, I see that [getParameter("resteasy.async.job.service.enabled")](https://github.com/resteasy/Resteasy/blob/4.0.0.Final/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ConfigurationBootstrap.java#L34) returns `null`.   If I change the line to `getInitParameter("resteasy.async.job.service.enabled”)`, it is successful.  Adding the note here just in case it may help troubleshooting this issue.

ronsigal:
@anilgursel, this is disturbing. I just ran your app with RESTEasy 3.6.2.Final and this test
```
	@Test
	public void testSample() {
		Client client = ClientBuilder.newClient();
		Builder request = client.target("http://localhost:8080/sample-app/echo").request();
		Response response = request.post(Entity.entity("is there anybody out there?", "text/plain"));
		System.out.println("status: " + response.getStatus());
		System.out.println("response: " + response.readEntity(String.class));
	}

	@Test
	public void testSample2() {
		Client client = ClientBuilder.newClient();
		Builder request = client.target("http://localhost:8080/second-app/echo").request();
		Response response = request.post(Entity.entity("is there anybody out there?", "text/plain"));
		System.out.println("status: " + response.getStatus());
		System.out.println("response: " + response.readEntity(String.class));
	}
```
results in
```
status: 404
response: The resource you've requested, has not been found!
status: 200
response: {"timestamp":1559935226546,"echoText":"is there anybody out there?"}
````
testSample2() runs first, btw. The useGlobal stuff looks the same. I don't know what to make of it. Could I be doing something wrong in running your app? I'm just running Application in Eclipse.

The other issue you mention, getParameter() vs getInitParameter(), is also disturbing. I'll look into it.

**Comment by anilgursel** [comment]
I just tested with both 3.6.2 and 3.7.0, and for both I see the two tests you shared passing:

```
status: 200
response: {"timestamp":1559938376104,"echoText":"is there anybody out there?"}
status: 200
response: {"timestamp":1559938376209,"echoText":"is there anybody out there?"}
```

**Comment by anilgursel** [comment]
Between, tests with 3.6.2 and 3.7.0 passes, but `useGlobal` is still `true`.

**Comment by ronsigal** [comment]
Well, I understand the getParameter() vs getInitParameter() problem.

Normally, HttpServletDispatcher.init() stores the ServletConfig and ServletContext in ResteasyContext before ConfigurationBootstrap.createDeployment() is called.

Here, though, ResteasyAutoConfiguration.resteasyBootstrapListener() creates a ListenerBootstrap and calls ListenerBootstrap.createDeployment() before ServletConfig and ServletContext get into ResteasyContext. I see that ResteasyAutoConfiguration.resteasyBootstrapListener() has access to ResteasyContext and could store it in time, but it doesn't have access to ServletConfig. So there's a partial solution.

When I went to store the ServletContext in ResteasyContext, I noticed that ResteasyContext doesn't exists because resteasy-spring-boot-starter uses RESTEasy 3.7.x instead of 4.0. In 3.x, the context map is handled by ResteasyProviderFactory, so that doesn't rule out the partial solution. I'm just wondering if there should be two branches of resteasy-spring-boot-starter. @liweinan?

Anyway, I'm not too worried about the getParameter() vs getInitParameter() problem. It's more or less fixable. I still haven't figured out why I'm not getting the test app to run in 3.6.2, though.

**Comment by sharathsrinivasa** [comment]
@ronsigal w.r.t resteasy-spring-boot-starter, we should have two branches, one for 3.x and master for 4.x work. If you need a test app with RestEasy 4.0.0.Final, you could use [this](https://github.com/sharathsrinivasa/resteasy-spring-boot/tree/upgrade-2-resteasy-4).

**Comment by liweinan** [comment]
@ronsigal Yes we should have two branches for resteasy-springboot-starter. I have talked this with @asoldano , and we need to upgrade resteasy-springboot-starter to use RESTEasy 4.x in master, and maintain a branch for RESTEasy 3.x.

**Comment by anilgursel** [comment]
> I still haven't figured out why I'm not getting the test app to run in 3.6.2, though.

@ronsigal Do you still have this problem?

**Comment by ronsigal** [comment]
@anilgursel, now I've got the test passing for 3.7.0.Final and failing for 4.0.0.Final, so I should be able to track down the problem.

**Comment by ronsigal** [comment]
@anilgursel, I've made some progress in understanding the problem. I see that 

1. Spring Boot creates a distinct servlet for each JAX-RS Application
2. resteasy-spring-boot-starter creates a single SynchronousDispatcher, whch is used by each servlet.

So, what happens is, there are two instances of HttpServletDispatcher, one per servlet, two instances of ServletContainerDispatcher, but one instance of SynchronousDispatcher, which is where context objects are stored. Consequently, there is a single ServletConfig stored in SynchronousDispatcher. So, even though the test passes, if I add
```
String p = ConfigProvider.getConfig().getOptionalValue("resteasy.servlet.mapping.prefix", String.class).orElse("d'oh");
System.out.println("prefix: " + p);
```
to Echo.echo(), I get
```
prefix: /second-app
prefix: /second-app
```
I haven't figured out the right solution yet. I'm thinking that ResteasyAutoConfiguration, rather than creating a SynchronousDispatcher to be used globally, should let ServletContainerDispatcher create a new SynchronousDispatcher for each servlet. I haven't tested anything yet, though.

@liweinan, what do you think about that?

**Comment by ronsigal** [comment]
@anilgursel, by the way, I'll be off tomorrow (Wednesday).

**Comment by ronsigal** [comment]
For the record, when I comment out contextInitialized() in ResteasyAutoConfiguration.resteasyBootstrapListener(), I get
```
prefix: /second-app
prefix: /sample-app
```
However, the injections don't happen. I'll have to learn more about Spring Boot.

ronsigal:
Ok, I think I have a fix, which involves letting ServletContainerDispatcher store the ServletConfig. I just have to figure out how to write a test that doesn't depend on resteasy-spring-boot-starter.

ronsigal:
@anilgursel, I've committed a fix in https://github.com/resteasy/Resteasy/pull/2075. Could you take a look?

ronsigal:
Replaced by https://github.com/resteasy/Resteasy/pull/2077.

sharathsrinivasa:
@ronsigal @asoldano Thanks fo the quick fix!

PR #2063, opened 2019-06-06
[RESTEASY-2268] ResteasyHttpHeaders does not follow JAX-RS spec concerning modifiability

NicoNes:
ResteasyHttpHeaders.getRequestHeaders() and ResteasyHttpHeaders.getRequestHeader() must return not modifiable collection

Signed-off-by: NicoNes <nicolas.nesmon@gmail.com>


NicoNes:
Hey @asoldano, 

Since I never succeed in running all the tests  on my computer I always push first than I let Travis let me know what's wrong...One day I'll have to find out why local tests are not running

Anyway, I did the fix and  everything should be OK now. :+1:

asoldano:
@NicoNes thanks!

PR #2065, opened 2019-06-12
[RESTEASY-2204] spring-reactor RxInvoker (almost exact copy of rxjava2 one)

crankydillo:
[Project Reactor](https://projectreactor.io/)

I'm doing local builds of this code for one of my projects and wondered if you might want this.  Especially, since it's basically your code:)


asoldano:
@crankydillo, interesting, we'll evaluate this, thanks. In the meantime, we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 . Do you agree to these terms?"

crankydillo:
Thanks.  Yes, the ASL 2.0 terms are fine.  This is basically your code anyway:)

resteasy-reactor/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-core-spi</artifactId>
+        <version>${project.version}</version>
+    </dependency>
+    <dependency>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-context-propagation</artifactId>
```
crankydillo: I don't think this is needed as I didn't do anything to support auto-context propagation.  Please let me know what I need to do here.

asoldano: Considering the context propagation does not work, I think we could even remove this dependency. We'll need at least a very brief mention of this spring-reactor module in the documentation, and there we should also have a note on the context propagation being up to the user.

PR #2085, opened 2019-07-02
[RESTEASY-2291] Native image improvements for MP REST Client on Quarkus

kenfinnigan:
Modify MP REST Client implementation to be able to completely disable SSL class usage if it's not enabled. This helps with DCE.

### DIFF
--- a/resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/ProviderFactoryTest.java
+++ b/resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/ProviderFactoryTest.java
@@ -1,0 +1,35 @@
+package org.jboss.resteasy.microprofile.client;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+import org.jboss.resteasy.core.providerfactory.ResteasyProviderFactoryImpl;
+import org.jboss.resteasy.plugins.providers.DefaultTextPlain;
+import org.jboss.resteasy.plugins.providers.IIOImageProvider;
+import org.jboss.resteasy.spi.ResteasyProviderFactory;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ProviderFactoryTest {
+
+    @Test
+    public void testDefaultProvider() {
+        // Reset to ensure we start from clean slate
+        RestClientBuilderImpl.setProviderFactory(null);
+
+        RestClientBuilderImpl builder = (RestClientBuilderImpl) RestClientBuilder.newBuilder();
+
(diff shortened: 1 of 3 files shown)


kenfinnigan:
Once this is approved/merged, it would be great if we could get a 4.1.1.Final released

kenfinnigan:
@asoldano @ronsigal Not sure why my changes result in a failure on JDK 8 on WF 15, given my changes don't impact the RESTEasy Client at all. Any ideas?

kenfinnigan:
@patriot1burke can you take a look?

asoldano:
@kenfinnigan you can ignore the failure, it's unrelated and transient.

asoldano:
No problem for having a 4.1.1 if this all is critical.

kenfinnigan:
Woohoo, CI passed this time!

@asoldano are these tests sufficient?

resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/ProviderFactoryTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+        Assert.assertTrue(builder.getBuilderDelegate().getProviderFactory().getProviderClasses().contains(IIOImageProvider.class));
+        Assert.assertTrue(builder.getBuilderDelegate().getProviderFactory().getProviderClasses().contains(DefaultTextPlain.class));
+    }
+
+    @Test
+    public void testCustomProvider() {
+        ResteasyProviderFactory provider = new ResteasyProviderFactoryImpl(null, true);
+        provider.registerProvider(IIOImageProvider.class);
+        RestClientBuilderImpl.setProviderFactory(provider);
```
asoldano: This actually has a side-effect of leaving the RestClientBuilderImpl at the end of the test in a state that's different from the state it had before the test. This could lead to strange/unexpected failures in case of testsuite execution order changes, etc. Can you restore the original state using a try-finally block? Thanks.

resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/SslTest.java
```diff
@@ -0,0 +1,18 @@
...
+public class SslTest {
+
+    @Test
+    public void testEnabled() {
+        Assert.assertTrue(RestClientBuilderImpl.SSL_ENABLED);
+    }
+
+    @Test
+    public void testDisabled() {
+        RestClientBuilderImpl.setSslEnabled(false);
```
asoldano: Same comment as above regarding the state.

kenfinnigan:
@asoldano made the changes to the tests.

Will do another PR when this is merged to update to latest MP REST Client spec

PR #2086, opened 2019-07-05
[RESTEASY-1886] Extend RxInvoker's with methods that return Response.

ronsigal:


ronsigal:
Needs addition to RESTEasy user guide.

ronsigal:
Hey, @jimma, could you take a look at this, just enough to verify that I haven't screwed up Rx2ObservableExtTest? Thanks.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
```diff
@@ -367,10 +435,13 @@ public void run()
          }
          catch (Throwable e)
          {
+            waitForSubscriber();
```
jimma: @ronsigal, Do we need to notify client error immediately when there is error happens instead of wait for the readLatch to countdown?

jamezp:
Closing this as it seems stale. Please feel free to reopen if I'm wrong :)

PR #2087, opened 2019-07-08
Vertx improvements

vietj:
- upgrade to Vert.x 3.7.1
- implementation of AsyncClientHttpEngine based on Vertx HTTP Client


docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
```diff
@@ -966,6 +971,60 @@ ResteasyClient client = ((ResteasyClientBuilder)ClientBuilder.newBuilder()).clie
         </programlisting>
     </section>
 
+     <section id="vertx_client">
+         <title>Vertx Client Engine</title>
+         <para>
+           Still as a drop in replacement, RESTEasy allows selecting a Vertx based HTTP engine.
+           The Vertx implementation can perform asynchronous client invocations. It can be used with
+           and without Vert.x and it provides the following features:
```
cescoffier: I would remove the `with or without Vert.x`, as it may be confusing here.

docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
```diff
@@ -966,6 +971,60 @@ ResteasyClient client = ((ResteasyClientBuilder)ClientBuilder.newBuilder()).clie
...
+           The Vertx engine is enabled by adding a dependency to the <emphasis>org.jboss.resteasy:resteasy-client-vertx</emphasis>
+           artifact to the Maven project; then the client can be built as follows:
+         </para>
+         <programlisting>
+VertxClientHttpEngine engine = new VertxClientHttpEngine();
+ResteasyClient client = ((ResteasyClientBuilder)ClientBuilder.newBuilder())
+  .clientEngine(engine).build();
+         </programlisting>
+         <para>
+           A Vertx instance can also be provided when creating the client engine, as well as options configuration:
```
cescoffier: `Vert.x`

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
```diff
@@ -0,0 +1,283 @@
...
+            CompletableFuture<T> tmp = new CompletableFuture<>();
+            vertx.executeBlocking(promise -> {
+                try {
+                    T result = extractor.extractResult(response);
+                    tmp.complete(result);
+                } catch (Exception e) {
+                    tmp.completeExceptionally(e);
+                }
+            }, ar -> {
+                //
```
cescoffier: Should you complete in this callback, or do you expect blocking downstream stages?

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        RequestOptions options = new RequestOptions();
+        MultiMap headers = MultiMap.caseInsensitiveMultiMap();
+        request.getHeaders().asMap().forEach(headers::add);
+        options.setHeaders(headers);
+        if (body != null) {
+            headers.set(HttpHeaders.CONTENT_LENGTH, "" + body.length());
+        }
+        options.addHeader(HttpHeaders.USER_AGENT.toString(), "Vertx");
```
cescoffier: Should it be `Resteasy with Eclipse Vert.x` here?

resteasy-client-vertx/src/test/java/org/jboss/resteasy/test/client/vertx/VertxClientEngineTest.java
```diff
@@ -0,0 +1,348 @@
...
+             public void failed(Throwable throwable) {
+                cf.completeExceptionally(throwable);
+             }
+          });
+
+      final String result = cf.get(10, TimeUnit.SECONDS);
+      assertEquals(valuableData, result);
+   }
+/*
+   @Test
```
cescoffier: Should this one be removed?

kenfinnigan:
Not sure if it's easily verifiable, but we'd also want to be able to automatically change the underlying engine used by MP REST Client impl in RESTEasy to use this for Quarkus

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
vietj: we actually complete from the worker thread to avoid blocking the event-loop so we don't care much of the result on the event loop

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
vietj: yes maybe, I must admit I set it to make a test pass :-)

vietj:
I think the MP rest client should build the correct engine directly passing the shared vertx instance and possibly configuration (like SSL configuration build from config)

resteasy-client-vertx/src/test/java/org/jboss/resteasy/test/client/vertx/VertxClientEngineTest.java
vietj: I forgot to resolve this, will try to make it work

vietj: done and pushed

docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
```diff
@@ -966,6 +971,60 @@ ResteasyClient client = ((ResteasyClientBuilder)ClientBuilder.newBuilder()).clie
         </programlisting>
     </section>
 
+     <section id="vertx_client">
+         <title>Vertx Client Engine</title>
+         <para>
+           Still as a drop in replacement, RESTEasy allows selecting a Vertx based HTTP engine.
+           The Vertx implementation can perform asynchronous client invocations. It can be used with
```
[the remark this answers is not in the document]
vietj: in practice Resteasy client is blocking (because it uses `InputStream`), so Resteasy would first need to be changed to achieve it.

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
vietj: I haven't changed this because i think it does not make much a difference

vietj:
Updated after the useful comments

asoldano:
Thanks @vietj @cescoffier and @kenfinnigan , I'll review this too soon.

asoldano:
> Not sure if it's easily verifiable, but we'd also want to be able to automatically change the underlying engine used by MP REST Client impl in RESTEasy to use this for Quarkus

Feel free to propose a change to the resteasy-client-microprofile for this. Are you thinking about something controlled through classpath (so basically scan for available asynch client http engines)?

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
```diff
@@ -0,0 +1,309 @@
...
+
+        URI uri = request.getUri();
+        options.setHost(uri.getHost());
+        options.setPort(uri.getPort());
+        options.setURI(uri.getRawPath());
+
+        CompletableFuture<ClientResponse> future = new CompletableFuture<>();
+
+        HttpClientRequest clientRequest = httpClient.request(method, options, response -> {
+            future.complete(toRestEasyResponse(request.getClientConfiguration(), response));
```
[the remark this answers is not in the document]
vietj: the deprecation advocates to use Vert.x Web Client (which uses this HttpClient under the hood) but I think for this low level client using the Vert.x HttpClient is more appropriate (less dependencies / less overhead). This deprecation is temporary until Vert.x 4 that gets the `HttpClient` methods un-deprecated (but signature changes a bit).

I can add a suppress warnings annotation if you want with a comment saying that this will go away with Vert.x 4

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
```diff
@@ -0,0 +1,309 @@
...
+                    tmp.complete(result);
+                } catch (Exception e) {
+                    tmp.completeExceptionally(e);
+                }
+            });
+            return tmp;
+        });
+    }
+
+    public CompletableFuture<ClientResponse> submit(final ClientInvocation request) {
```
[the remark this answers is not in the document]
vietj: sure

resteasy-dependencies-bom/pom.xml
```diff
@@ -348,6 +348,13 @@
                 <artifactId>vertx-core</artifactId>
                 <version>${version.io.vertx}</version>
             </dependency>
+            <dependency>
+                <groupId>io.vertx</groupId>
+                <artifactId>vertx-core</artifactId>
+                <version>${version.io.vertx}</version>
+                <classifier>tests</classifier>
+                <type>test-jar</type>
```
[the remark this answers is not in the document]
vietj: yes I can add it, although I believe it's not mandatory because it's a dependency management that won't affect the classpath

vietj:
Updated PR with feedback

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/engines/vertx/VertxClientHttpEngine.java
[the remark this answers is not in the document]
asoldano: Thanks

resteasy-dependencies-bom/pom.xml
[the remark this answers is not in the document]
asoldano: Actually the <scope> you just added properly fixes the scope of the dependency in the resteasy-client-vertx dependency tree. Thanks.

asoldano:
Merged, thanks a lot

kenfinnigan:
@asoldano I think we'd want to make it the default in Quarkus. So would need to see if we can do that in Quarkus or whether we need to do it in RESTEasy

PR #2100, opened 2019-07-12
Performance improvements

patriot1burke:
This needs a good review.  I'll squash after approval.


patriot1burke:
I should have elaborated on the improvements:
* Matches are now cached if and only if its a static regular expression i.e. no path params and only if the match is a resource method (not resource locator).  Match is based on path, content-type and accept header.  This avoids all the complex and expensive matching that happens, especially for content negotiation.  Future approach would be to further optimize the matching layer for simple cases.
* Lazy instantiation of ResteasyUriInfo things.  This was a pretty expensive operation
* Improvements to Netty layer.

asoldano:
Replaced by #2104

PR #2114, opened 2019-07-25
[RESTEASY-2312] Introduce support for Spring REST annotation handling

geoand:
cc @asoldano @patriot1burke

### DIFF
--- a/docbook/reference/en/en-US/modules/RESTEasy_Spring_Integration.xml
+++ b/docbook/reference/en/en-US/modules/RESTEasy_Spring_Integration.xml
@@ -332,4 +332,118 @@ server.deploy(deployment);]]></programlisting>
             <![CDATA[modules/system/layers/base/org/jboss/resteasy/resteasy-spring/main/bundled/resteasy-spring-jar]]>.
         </para>
     </section>
+    <section>
+        <title>Processing Spring Web REST annotations in RESTEasy</title>
+        <para>
+            RESTEasy also provides the ability to process Spring Web REST annotations (i.e. Spring classes annotated with @RestController)
+            and handle related REST requests without delegating to Spring MVC. This functionality is currently experimental.
+        </para>
+        <para>
+            In order for RESTEasy to be able to process Spring @RestController, you first need to include the following dependency.
+        </para>
+        <para><programlisting>        &lt;dependency&gt;
+            &lt;groupId&gt;org.jboss.resteasy&lt;/groupId&gt;
+            &lt;artifactId&gt;resteasy-spring-web&lt;/artifactId&gt;
+            &lt;version&gt;4.2.0-SNAPSHOT&lt;/version&gt;
+            &lt;/dependency&gt;
+
+        </programlisting></para>
+        <para>
+            Currently RESTEasy does not auto-scan for @RestController annotated classes,
+            so you need to add all @RestController annotated classes to your web.xml file as shown in the following example.
+        </para>
+        <programlisting><![CDATA[<web-app>
+   <display-name>RESTEasy application using Spring REST annotations</display-name>
+
+    <context-param>
+        <param-name>resteasy.scanned.resource.classes.with.builder</param-name>
+        <param-value>org.jboss.resteasy.spi.metadata.SpringResourceBuilder:org.example.Controller1,org.example.Controller2</param-value>
+    </context-param>
+
+   ...
+</web-app>]]></programlisting>
+    </section>
+    <para>
+        In the example above, Controller1 and Controller2 are registered and are expected to be annotated with @RestController.
+    </para>
+    <para>
+        The list of the currently supported annotations can be found below
+    </para>
+    <table frame="topbot">
+        <tgroup cols="2" rowsep="1" colsep="1">
+            <thead>
+                <row>
+                    <entry>Annotation</entry>
+                    <entry>Comment</entry>
+                </row>
+            </thead>
+            <tbody>
+                <row>
+                    <entry>@RestController</entry>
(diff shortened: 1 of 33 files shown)


resteasy-spring-web/src/main/java/org/jboss/resteasy/web/ResponseEntityReturnTypeHandler.java
```diff
@@ -0,0 +1,46 @@
+package org.jboss.resteasy.web;
```
patriot1burke: package should have spring in it.  org.jboss.resteasy.spring.web or something

resteasy-spring-web/src/main/java/org/jboss/resteasy/web/ResponseStatusBuiltResponseProcessor.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.jboss.resteasy.specimpl.BuiltResponse;
+import org.jboss.resteasy.spi.BuiltResponseProcessor;
+import org.jboss.resteasy.spi.metadata.ResourceMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+import javax.ws.rs.core.Response;
+
+public class ResponseStatusBuiltResponseProcessor implements BuiltResponseProcessor {
```
patriot1burke: I think this could be implemented as a ContainerResponseFilter instead.

patriot1burke:
I didn't see any documentation either.

geoand:
Thanks for the comments @patriotburke1! I'll address them tomorrow 😊

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/BuiltResponseProcessor.java
```diff
@@ -0,0 +1,13 @@
+package org.jboss.resteasy.spi;
+
+import org.jboss.resteasy.spi.metadata.ResourceMethod;
+
+import javax.ws.rs.core.Response;
+
+public interface BuiltResponseProcessor {
```
patriot1burke: Same here, I think this can be implemented as a combination of DynamicFeature and ContainerFilterResponse.  You scan annotatoins and view return types within DynamicFeature.  You add a ContainerFilterResponse for the method if its a Spring one.  This filter handles the transformation between the spring return object and jaxrs Response.

geoand: Thanks for the tip, looking into it

geoand:
@patriot1burke I updated the PR to get rid of the 2 SPI's I had added that could replaced with the `DynamicFeature` and it worked like I charm! Thanks for the great tip!

I also addressed the wrong package issue you mentioned.

As for the documentation, where would you like me to add it and what do you think should be added? 
The thing with the feature as I see it is that it's only really meant to be used by Quarkus. Of course it could be used outside of it, but there are a few things that would be need to be configured manually. Do you think such configuration should be included?

Thanks

geoand:
@rsearls Thanks

I have noticed that too and it's due to the different line-ending characters. I tried to change it but all my attempts failed. 
I will give it another go however and see if I can find something. You wouldn't happen to have any tips would you?

geoand:
It seems like my IDE is doing something weird with the line endings, I'll see how to fix it and then update the PR

geoand:
@rsearls formatting issues should now be fixed and the PR should be much easier to review

patriot1burke:
What's manual?

If you add META-INF/services/javax.ws.rs.ext.Providers with the FQN if your DynamicFeature, it should just work.  Also, you could create a regular Feature to register the param injectors too.

geoand:
@patriot1burke the way I have things, "telling" RESTEasy which classes are Spring Controllers is manual (by setting `resteasy.scanned.resource.classes.with.builder`).
The other parts of course can be registered as you mention.

geoand:
The test failure seems unrelated:

```
The command "travis_wait 60 mvn -B -Ptravis,jboss-repository -fae -Dserver.version=$SERVER_VERSION ${ELYTRON:+-Delytron} install" exited with 1.
```

geoand:
I'd be happy to add whatever docs you think are relevant for this use case

geoand:
@asoldano thanks!

I created a JIRA [here](https://issues.jboss.org/browse/RESTEASY-2312). Should I add any more information to it? I will also update the PR to be against master instead of `4.1`.
Futhermore I will add basic information in docs. Any preference where that should be?
 
As for the tests, I believe they should be separate, since this feature is different from the other Spring stuff in that it doesn't actually involve Spring, only reading Spring metadata. WDYT?

geoand:
@asoldano PR now is against `master`. I also squashed the commits into a single one

geoand:
I also added some basic documentation in the Spring Integration section

asoldano:
OK, thanks @geoand .
Last question I have is about Spring version. I'm locally testing an upgrade of the whole project from 4.3.22.RELEASE to the 5.0.14.RELEASE version you're pulling here... however is that choose for a specific reason? Would 5.1.x be fine as well?

geoand:
@asoldano no particular reason for the version I chose. It should work with anything higher than 4.3.x.
5.1.x should be perfectly fine as well

asoldano:
@geoand ok, thanks

geoand:
I closed and reopened the PR a few times to get restart Travis - the failures seem totally irrelevant

liweinan:
great!

PR #2120, opened 2019-07-30
[RESTEASY-2314] Upgrade to SpringFramework 5.0.14.RELEASE

asoldano:


liweinan:
I downloaded this patch and recompile my local resteasy:

image

And recompile my local distribution:

image

Then I use the patched resteasy in resteasy-spring-boot:

image

And rebuild `resteasy-spring-boot` with the patched resteasy distribution and it passed tests and built correctly:

image

And here is the dependency tree in resteasy-spring-boot:

image

We can see the spring components are compatible with each other.

liweinan:
Waiting for the test result from travis ci...

PR #2122, opened 2019-07-31
[RESTEASY-2179] updated arquillian versions;  Added security manage c…

rsearls:
…heck code blocks where possible.  Add permission classes to test archives as needed

### DIFF
--- a/server-adapters/resteasy-netty4-cdi/pom.xml
+++ b/server-adapters/resteasy-netty4-cdi/pom.xml
@@ -14,7 +14,6 @@
     <description/>
     <properties>
         <dep.arq.weld.se>2.0.0.Final</dep.arq.weld.se>
-       <dep.arq.version>1.4.1.Final</dep.arq.version>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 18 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProxyTest.java
```diff
@@ -28,6 +29,7 @@
 import io.reactivex.Single;
 
 @RunWith(Arquillian.class)
+@RunAsClient
```
asoldano: Why is this needed? I mean, it's not a problem, just wondering why.

rsearls: It is needed when security manager is being used.  Without it arquillian throws exception,
java.util.NoSuchElementException

rsearls:
Rebase complete

server-adapters/resteasy-netty4-cdi/pom.xml
```diff
@@ -14,7 +14,7 @@
     <description/>
     <properties>
         <dep.arq.weld.se>2.0.0.Final</dep.arq.weld.se>
-        <dep.arq.version>1.1.15.Final</dep.arq.version>
+       <dep.arq.version>1.4.1.Final</dep.arq.version>
```
asoldano: uh, just spotted this. Is there a reason why we can't use the 'dep.arquillian-bom.version' from the resteasy-dependencies-bom/pom.xml? (to avoid duplicating the property)

rsearls: rebase .. complete
version ref change .. complete

PR #2123, opened 2019-08-01
[RESTEASY-2244] @Produces not picked up well

spyrkob:
JIRA: https://issues.jboss.org/browse/RESTEASY-2244


ronsigal:
Thanks, @spyrkob. Would you be interested in writing a test in which the return media type comes from the resource class?

spyrkob:
@ronsigal sure, can do that, but without some hackery it won't be possible to reproduce the issue as the correct Producer would be returned by the method.getProduces() already.

ronsigal:
@spyrkob, yeah, you're right. Good point. I suspect that the best practice would be to have a test in place just to prevent any future regressions. With a note explaining that there is currently no way for the test to fail. If you don't mind. ;-)

spyrkob:
@ronsigal added the test, let me know if it looks OK

spyrkob:
@asoldano sorry my IDE must have messed up something... should be fixed now

PR #2124, opened 2019-08-01
[RESTEASY-2315] 3.9-branch check for more unproxyable providers

mattdrees:
This checks for inherited final methods, in addition to final methods declared directly by the provider.

This follows up on work done in #2000 / #2009, which handled declared final methods, but not inherited final methods. (#2009 had some discussion about how in reality every class inherits some final methods from java.lang.Object, but I think spirit of the CDI spec and the implementation in Weld do not include Object's methods.)

[Jira](https://issues.jboss.org/browse/RESTEASY-2315)


mattdrees:
PR for `master` is at https://github.com/resteasy/Resteasy/pull/2125

ronsigal:
Same comment as in https://github.com/resteasy/Resteasy/pull/2125.

mattdrees:
rebased onto the tip of 3.9

mattdrees:
added the check

mattdrees:
CI seemed to fail due to a connection reset from google-maven-central, similar to https://github.com/resteasy/Resteasy/pull/2125#issuecomment-521742090

ronsigal:
Thanks, @mattdrees.

PR #2125, opened 2019-08-01
[RESTEASY-2315] check for more unproxyable providers

mattdrees:
This checks for inherited final methods, in addition to final methods declared directly by the provider.

This follows up on work done in https://github.com/resteasy/Resteasy/pull/2009, which handled declared final methods, but not inherited final methods. (That PR had some discussion about how in reality every class inherits some final methods from java.lang.Object, but I think spirit of the CDI spec and the implementation in Weld do not include Object's methods.)

[Jira](https://issues.jboss.org/browse/RESTEASY-2315)


ronsigal:
Hi @mattdrees,

This looks good. I just have one question. I'm looking at https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#unproxyable, which gives the first category of unproxyable classes as:
```
classes which don’t have a non-private constructor with no parameters,
```
Shouldn't we be checking for that? Or am I forgetting something?

-Ron

mattdrees:
Thanks @ronsigal!
Yeah, we probably should check for such classes. I'll update my commit.

mattdrees:
I added the check, and also rebased onto the tip of master.

mattdrees:
CI seemed to fail due to a connection reset from google-maven-central. I wonder if I can re-request a build.

ronsigal:
@mattdrees, I've restarted the tests. There must be some permissions thing going on - I don't really know how it works.

ronsigal:
Other than that, it looks good. @mattdrees, thanks for adding the constructor test.

mattdrees:
thanks Ron!

PR #2135, opened 2019-08-13
RESTEASY-2324: RESTEasy client should disable Apache Http Client cookie management feature by default

NicoNes:


NicoNes:
I don' t know why but one travis build failed...

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
```diff
@@ -276,7 +277,7 @@ public void testEJBs() throws Exception {
    public void testSessionScope() throws Exception {
       log.info("starting testSessionScope()");
       client.close();
-      client = ClientBuilder.newClient();
+      client = new ResteasyClientBuilderImpl().cookieManagementEnabled(true).build();
```
asoldano: Please do not reference the ResteasyClientBuilderImpl directly; use:

`client = ((ResteasyClientBuilder)(ClientBuilder.newBuilder())).cookieManagementEnabled(true).build();`

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/CookieInjectionTest.java
```diff
@@ -68,7 +68,7 @@
 
    @BeforeClass
    public static void init() {
-      client = ClientBuilder.newClient();
+      client = new ResteasyClientBuilderImpl().cookieManagementEnabled(true).build();
```
asoldano: Same as before.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
NicoNes: Done

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/CookieInjectionTest.java
NicoNes: Done

NicoNes:
Hi @asoldano ,

You're right we should add documentation about this.
I tried to do so adding few words in `RESTEasy_Client_Framework.xml`. Let me know what you think about it. (And do no hesitate to correct my english also if necessary ;)).

Thanks

asoldano:
@NicoNes sorry for the late feedback. Thanks for the addition to the doc. I'm merging this because I don't want to make the process over complicated. However, I've just realized that it would probably be better to have a ``enableCookieManagement()`` method in the builder, similarly to what we have for using async, disabling trustmanager, etc. I'll change that just after merging. Thanks.

NicoNes:
Hey @asoldano,

I thought about it when I did it and  I hesitated because, once `enableCookieManagement()` will be called, it will not longer be possible to  build a `ResteasyClient` instance without cookie management from this builder. 
In other way the builder is no longer usable as we expect it to be.

But as you said it's already the case for other methods so let's keep it consistent.

Thanks !

asoldano:
Follow-up on https://github.com/resteasy/Resteasy/pull/2169

PR #2142, opened 2019-08-23
[RESTEASY-2331] Refactor hardcoded servlet dependencies

patriot1burke:
Resteasy has a few places that are dependent on Servlet.  This PR removes those dependencies


ronsigal:
I look forward to reading this. As soon as 3.9.0.Final is out the door next week. ;-)

PR #2143, opened 2019-08-28
RESTEASY-2330

liweinan:
Adjust the Optional feature implemented for @PathParam according to the discussion in EAP7-1248


liweinan:
waiting for final decision from discussion...

PR #2137, opened 2019-08-15
RESTEASY-2085 index.html (welcome-file in web.xml) is broken on Tomcat (branch 3.9)

liweinan:
…t (#2134)

* - RESTEASY-2085 index.html (welcome-file in web.xml) is broken on Tomcat
- RESTEASY-2319 Resteasy-based application not deploying under Weblogic application server

* fix url value

* Update HttpServlet30Dispatcher.java

PR #2144, opened 2019-08-29
RESTEASY-2330-b3.9

liweinan:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServlet30Dispatcher.java
```diff
@@ -17,7 +17,11 @@
  * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
  * @version $Revision: 1 $
  */
-@WebServlet(asyncSupported = true, value="")
+// Note. Setting value to "/RESTEASY_HttpServlet30Dispatcher" frees up the root "/" context
+//       to serve static content. The path "/RESTEASY_HttpServlet30Dispatcher" will be installed
+//       and it will lead to HttpServlet30Dispatcher, but no resources will be there and status
+//       404 will be returned.
+@WebServlet(asyncSupported = true, value="/RESTEASY_HttpServlet30Dispatcher")
```
asoldano: Does this change really belong to this fix? IOW, why is it needed?

asoldano: Looks like this is already in 3.9 because of https://github.com/resteasy/Resteasy/pull/2137/files , so it will go away when merging.

PR #2145, opened 2019-08-29
RESTEASY-2332

liweinan:
add document for optional typed parameters


liweinan:
Added document:

image

PR #2151, opened 2019-09-02
Extended conneg capabilities by implementing JAX-RS-mod-qs Specification

retog:
This implements the spec improvement as per https://github.com/factsmission/jax-rs-spec/


retog:
The branch from which i made this PR got polluted with another commit, will re-create from other branch.

PR #2152, opened 2019-09-02
Extended conneg capabilities by implementing JAX-RS-mod-qs Specification

retog:
This implements the spec improvement as per https://github.com/factsmission/jax-rs-spec/

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
@@ -551,20 +551,19 @@ public class ServerResponseWriter
             }
             return o.equals(selectMostSpecific(o, this)) ? 1 : -1;
          }
+         if ((o.q * o.qs) < (this.q * this.qs))
+         {
+            return -1;
+         }
+         if ((o.q * o.qs) > (this.q * this.qs))
+         {
+            return 1;
+         }
          if (o.q < this.q)
          {
             return -1;
          }
          if (o.q > this.q)
-         {
-            return 1;
-         }
-         // zzzz.q == this.q
-         if (o.qs < this.qs)
-         {
-            return -1;
-         }
-         if (o.qs > this.qs)
          {
             return 1;
          }


ronsigal:
Hi @retog,

This is interesting, and, based on https://www.w3.org/DesignIssues/Conneg, it looks like Tim Berners-Lee agrees with you, which is a pretty strong recommendation. The problem I have is that you are proposing a change to a fundamental algorithm, the Request Matching algorithm in Section 3.7.2 of the JAX-RS, aka Jakarta RESTful Web Services, specification. I think that rather than changing an implementation, it's best to propose the idea to the specification group at https://projects.eclipse.org/projects/ee4j.jaxrs. Have you communicated with them?

-Ron

retog:
Hi @ronsigal 

Tahanks for your comment. I am aware that this is a change to the JAX-RS spec (in order to better implement the HTTP spec). I tried to reach the spec authors, unfortunately the users@jsr311.dev.java.net mailing list on which I used to be active is no longer operating, neither is jsr370-experts@jax-rs-spec.java.net. As a newer list I found jaxrs-spec@javaee.groups.io so I wrote there, but nobody seems to be managing the list and aproving new members, so I made a Pull Request against the spec (https://github.com/jax-rs/spec/pull/6), which has not been commented so far. I didn't kow about jakarta, is that the place where the spec is being mantaned?

So at the end I was hoping Resteasy as the most active implementation might adopt the change and this would go back to the specification when there is again some active specification effort. 

Cheers,
Reto

ronsigal:
Hi @retog,

Ah, yes, the world is upside down. All of the JEE specs have moved to the Eclipse Foundation, so there's a new group, new mailing list, etc. That's jakarta, and, as you say, that's where everything is happening now. In fact, there's a new name: JAX-RS is now Jakarta RESTFul Web Services. Here's the front page: https://projects.eclipse.org/projects/ee4j.jaxrs. The group is quite active, and they tend to be responsive.

-Gon

asoldano:
https://www.eclipse.org/lists/jaxrs-dev/msg00711.html

jamezp:

PR #2153, opened 2019-09-02
First sorting criterion is specificity independently of compatibility

retog:
Nowhere in the JAX-RS specification compatibility affects sorting.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ServerResponseWriter.java
@@ -543,13 +543,11 @@ public class ServerResponseWriter
       @Override
       public int compareTo(SortableMediaType o)
       {
-         if (o.isCompatible(this))
-         {
-            if (o.equals(this))
-            {
-               return 0;
-            }
-            return o.equals(selectMostSpecific(o, this)) ? 1 : -1;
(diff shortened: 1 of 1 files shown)


ronsigal:
Hi @retog,

My first observation is that ServerResponseWriter$SortableMediaType.compareTo() doesn't seem to get called. I don't know the history of that. Am I missing something?

Second, I'm not sure I understand the new comparison. Aren't you saying that "text/plain" and "application/*" are compatible, based on the wildcard count?

Finally, I would be reluctant to change a fundamental behavior without evidence that it is incorrect.

-Ron

ronsigal:
Hey @retog,

re; "method is called by the sorting invoked"

D'oh! Of course. Thanks.

re: "Sort M in descending order, with a primary key of specificity"

Well, this is really interesting. I don't remember ever noticing that the ordering defined in the Request Matching algorithm (Section 3.7) is different than the ordering in the Media Type algorithm (Section 3.8), and it looks like the compareTo() method is suitable for the former. Now that I think about it, the Request Matching ordering doesn't make sense to me. Consider m1 = text/*;q=1.0 and m2 = application/xml;q=0.5. According to the Request Matching ordering, m1 > m2, and according to the Media Type ordering, m2 > m1. Wow.

@retog, what do you think?

-Ron

jamezp:

PR #2154, opened 2019-09-06
Avoid reflection on String parameters

patriot1burke:
Any String parameters where being unmarshalled by reflectively calling String.valueOf(param).  Should improve performance!

This was found by running Graal and finding that String.class had to be added as a reflective type!


ronsigal:
Line 45 of StringToPrimitive seems redundant, given lines 22-26.

PR #2155, opened 2019-09-11
[RESTEASY-2338] code changes and additions for new module integration…

rsearls:
…-tests-embedded.


ronsigal:
Hi @rsearls,

I think this is really nice, but I feel a little confused.

1. It seems to me that the fundamental point is to organize the various embedded containers under the interface EmbeddedJaxrsServer<T>. Is that the right way to think about this? If that's the case, then maybe you could say a bit about that in the Embedded Container chapter.

2. I'm not sure how to think about the integration-tests-embedded module. Is the point to test the functionality of swapping server implementations? How did you choose the particular four tests? Are they code examples, or is there something deeper there? You're doing things with Applications, interceptors, and readers/writers, which all interact with ResteasyDeployment, but I'm not sure how they are relevant to the embedded containers. What am I missing? Also, if the module is testing the various containers, why not run them as part of the test suite?

I think it's really cool, though.

-Ron

docbook/reference/en/en-US/modules/RESTEasy_Embedded_Container.xml
```diff
@@ -73,7 +73,7 @@ public class UndertowTest
    @Test
    public void testApplicationPath() throws Exception
    {
-      server.deploy(MyApp.class);
+      server.deployOldStyle(MyApp.class);
```
ronsigal: The use of deployOldStyle() seems a little unexplained. Why not just leave it as deploy()? Also, have you changed the semantics of deploy() in a way that would affect anything or surprise anyone?

docbook/reference/en/en-US/modules/RESTEasy_Embedded_Container.xml
```diff
@@ -84,7 +84,7 @@ public class UndertowTest
    @Test
    public void testApplicationContext() throws Exception
    {
-      server.deploy(MyApp.class, "/root");
+      server.deployOldStyle(MyApp.class, "/root");
```
ronsigal: Same question.

resteasy-cache/resteasy-cache-core/src/test/java/org/jboss/resteasy/test/cache/PersistentServerCacheTest.java
```diff
@@ -145,14 +145,15 @@ public void before() throws Exception
...
-      server.start();
       deployment = server.getDeployment();
+      deployment.start();
 
       dispatcher = deployment.getDispatcher();
       deployment.getProviderFactory().property("server.request.cache.infinispan.config.file", "infinispan.xml");
       deployment.getProviderFactory().property("server.request.cache.infinispan.cache.name", "TestCache");
       deployment.getProviderFactory().register(ServerCacheFeature.class);
       deployment.getRegistry().addPerRequestResource(TestResource.class);
+      server.start();
```
ronsigal: Again, have the semantics changed? Would existing code without the server.start() and deployment.start() changes still work?

rsearls: > Is the point to test the functionality of swapping server implementations? 
    + There are 5 objectives for this project.  
        1. Provide a single set of tests that all the containers run.
        2. Expand the set of test cases. It is Alessio's preference that a
           separate module be provided for the common set of tests and that all
           or as many as possible tests in module integration-tests be cloned
           into that module and runnable by each container. 
        3. Maintain the existing interface and semantics of the existing *JaxrsServer 
           classes, because there are unit tests in non-testsuite and non-server-adapter 
           modules that use the containers for testing.  Do not break these.
        4. Provide a standard interface for test configuration and execution using 
           the *JaxrsServer classes as the basis.
        5. Make testing all containers part of the standard build/test process,
           (this is a future task).

> How did you choose the particular four tests? 
      The long term goal is to clone all or as many as possible tests in module 
      integration-tests into module integration-tests-embedded.  The tests selected 
      were an initial sampling of test types to determine where changes might be
      needed in the container configuration code and to provide "proof of concept"
      for cloning a variety of tests from integration-tests.

      ResteasyDeployment has always been used by the *JaxrsServer code to transmit
      application information to the container.  In particular this is how several
      containers gain access to Resteasy's Registry object.  The Registry object 
      at a minimum provides references to the app's resource classes.  Other containers 
      use ResteasyDeployment's methods to get references to the app's resource classes.
      Resteasy's Registry implementation is intertwined with enough other resteasy
      code that it can not be created on its own in a way that is usable by the
      container. ResteasyDeployment.start() initializes all the needed setup and 
      configuration for this object.  I did not investigate how the containers
      might be using other elements of the Registry class.

> Why not run them as part of the test suite.
      As noted in project objectives 2. above, Alessio wanted a separate module 
      in which to maintain an expanded set tests all containers would run.


> The use of deployOldStyle() seems a little unexplained. 
      This method is only present in UndertowJaxrsServer and it is there to 
      allow a pre-existing unit test to run with minimal configuration disruption.
      It is provided to maintain backward compatibility.

> Also, have you changed the semantics of deploy() in a way that would affect 
  anything or surprise anyone?
      I have not.  The semantics for deploy() are unchanged.

> Would existing code without the server.start() and deployment.start() changes still work?
      In all pre-existing code in Resteasy's server.start() has always been required.
      I have changed the order of when deployment configuration is done and when
      server.start is called for several of the containers in order to standardize
      the process for all tests.  The *JaxrsServer classes require the deployment
      configuration object to be created before server.start() is called.

      deployment.start() is needed in many cases because of the container's use of
      the Registry object to get access to the app's resource classes. (See
      description of the use of ResteasyDeployment above.)

ronsigal:
Thanks for clarifying, @rsearls.

PR #2156, opened 2019-09-12
[RESTEASY-2344] Do not include extractResult in requestTimeout

jessicadeery:
Do not include extractResult in the ReactorNettyClient requestTimeout


asoldano:
Thanks, merged.

PR #2160, opened 2019-09-16
Resteasy 2347 json binding provider resteasy configuration null

rsearls:


testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/providers/JsonBindingTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * There are integration tests that check the properly setting when
+ * a ResteasyConfiguration.class is provided.
+ */
+public class JsonBindingTest {
+    private MediaType mediaType = new MediaType("application", "json");
+
+    @Test
+    public void testUseJSONB() throws Exception
+    {
+        System.setProperty(ResteasyContextParameters.RESTEASY_PREFER_JACKSON_OVER_JSONB, "true");
```
asoldano: @rsearls this is fine, however you need to be sure whatever was set for this system property before the test is restored after it. IOW, the test should leave the environment in the same state it was before executing, otherwise all tests running after this would be affected.

rsearls:
Change implemented.

PR #2166, opened 2019-09-19
RESTEASY-2335 RestEasy Microprofile Client has a immutable template p…

liweinan:
…roblem/bug


liweinan:
The fix triggered some microprofile tck test failures, I'll go on working on this.

liweinan:
tck tests passed locally:

image

PR #2168, opened 2019-09-20
RESTEASY-2335 RestEasy Microprofile Client has a immutable template problem/bug

liweinan:
* RESTEASY-2335 RestEasy Microprofile Client has a immutable template problem/bug

* fix NPE


asoldano:
@liweinan merged, thanks for the 3.10 branch backport

PR #2170, opened 2019-09-23
[RESTEASY-2506] Async io

FroMage:
DO NOT MERGE YET

This introduces the `AsyncMessageBodyWriter` interface, to support writing the response as async IO, and `AsyncOutputStream` that the implementors of `HttpResponse.getOutputStream` can return, to support writing async IO.

Json-b and `text/*` now supports async IO, as well as servlet, netty and vertx servers.

I still have to try this out in Quarkus.

All tests pass, but I had to adapt a few tests, due to the fact that:

- For some reason, some sequences of `SseEventSink.send` calls end up with the objects sent in the wrong order
- Calling `SseEventSink.close` does not currently block waiting for pending IO. It could just call `rxFlush().toCompletableFuture().get()` to block, but I wonder if that's the right thing to do.

I guess if I fix those two issues I could pass the tests with no adaptation.


patriot1burke:
FYI: Vertx implementation is current forked within Quarkus.

FroMage:
I've just pushed changes that should queue atomically in the order of the send operations (but don't, for some reason), and force a flush on close so that legacy code doesn't have to chain the completable futures.

FroMage:
Finally fixed the ordering issue which was due to the braindead servlet async IO API.

FroMage:
The CI tests fail in `resteasy-wadl`, but that doesn't seem related to my stuff at all. Lemme rebase. Note that those tests pass on my machine.

FroMage:
Nope, same. @asoldano any idea?

asoldano:
@FroMage I've seen failures in other areas (resteasy-links, resteasy-netty4), mmh. I've re-started the whole travis job, let's see what we get this time and if it's consistent with the previous run.

FroMage:
Well it fails in several places, with what looks like a regression from my code, but I really can't reproduce locally, so that's a bummer. I mean, even `NettyTest.testLarge()` fails on CI, which hints at a truncated output. You'd think I'd notice this locally…

FroMage:
@asoldano well, I've run out of ideas. I'm going to need to debug this on CI.

FroMage:
Ah no, I think I know what I did wrong.

FroMage:
Yes, it's that I've left data in the buffer that we didn't flush, this corresponds to the last chunk size exactly.

FroMage:
And for some reason in this test nobody calls `close()` on the `OutputStream`, which the container is supposed to do, because resteasy doesn't call it.

FroMage:
Damnit. One error left, but only on one CI run.

FroMage:
Latest version. Still missing tests and docs, but those that exist pass on my machine. Let's see what CI says.

FroMage:
It passed CI. I believe that's a first for me :)

asoldano:
@FroMage :-)

asoldano:
@FroMage , I'm thinking to release just after MP REST Client 1.4 is finalised, which should be a matter of few days. Should we try including this too?

FroMage:
I would love that, but I'm still writing tests and docs and busy at the vert.x f2f this week. I should be done before the end of next week. Would that work? You guys can probably start reviewing it, knowing that docs and more tests will come.

asoldano:
@FroMage I'm definitely starting reviewing; end of next week should be doable :-)

FroMage:
Rebased on master, added lots of tests and docs.

FroMage:
Ready for review and merge if CI passes.

FroMage:
Latest developments are:

- I tested the async IO writers and interceptors for the following providers:
  - Atom
  - Jackson
  - JAXB
  - Multipart
  - Jsonp
  - Core providers
- I only tested this on Vertx (by adding deps to those providers), because the other backends already have async IO tests, not much use trying the providers on every backend
- I could not test JSONB, fastinfoset because I would have had to get rid of the other providers from the classpath
- I did not test smime, doseta because they look very complex to test

Every writer/interceptor supports async IO, with the exceptions of:

- GZIP: this is unfortunate, but I didn't find how to do it without buffering, which didn't seem ideal in this case
- `ReaderProvider`: it reads chars (as ints) and writes them as if they were bytes (as ints, very confusing API). I believe this is a bug, I wanted to resolve this before adding async support.
- `Jackson2JsonpInterceptor` is implemented, but it uses the platform encoding instead of the response encoding, again I believe it's a bug.
- `StreamingOutputProvider`: use the new alternative `StreamingAsyncOutputProvider`
- `HtmlRenderableWriter` it seems stuck to an annoying Servlet dependency that looked hard to make work with async IO

Also note that this doesn't support async IO for input, nor on the client. But we can always add this later.

Benchmarking in Quarkus showed no drop in perf in speed or memory.

asoldano:
@FroMage there seems to be a failing test in the vert.x integration. Would you mind having a look?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
```diff
@@ -128,4 +133,55 @@ public static void writeTo(final InputStream in, final OutputStream out) throws
...
+      final byte[] buf = new byte[2048];
+      return writeTo(in, out, buf);
+   }
+
+   private static CompletionStage<Void> writeTo(final InputStream in, final AsyncOutputStream out, byte[] buf)
+   {
+      int read;
+      try
+      {
+         if ((read = in.read(buf)) != -1)
```
stuartwdouglas: InputStream.read is a blocking operation, so if this is called from the IO thread it could block and cause performance problems. It probably won't cause deadlocks, unless the input stream is also backed by the same IO thread (e.g. the Undertow input stream).

The only way to really fix this would be to move this work to an executor if you are on the IO thread.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ReaderProvider.java
```diff
@@ -27,6 +27,8 @@
 @Provider
 @Produces("*/*")
 @Consumes("*/*")
+// FIXME: this does not implement AsyncMessageBodyWriter, but I think the current implementation is flawed as we read chars and write
+// them as if they were bytes, which will never work outside of ASCII
```
stuartwdouglas: yea, this looks really broken on the write side. I don't think it would be common to return a Reader from a resource method, which is likely why nobody has noticed.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
```diff
@@ -182,25 +197,58 @@ private void internalFlushResponseToClient(boolean throwIOException) throws IOEx
...
+                        {
+                           aos = response.getAsyncOutputStream();
+                        } catch (IOException x)
+                        {
+                           close(false);
+                           ret.completeExceptionally(x);
+                           return;
+                        }
+                        // eager composition to guarantee ordering
+                        CompletionStage<Void> a = aos.asyncWrite(SseConstants.EOL);
```
stuartwdouglas: It does not really matter, but we could probably just add a new constant here to avoid the double async write.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
```diff
@@ -182,25 +197,58 @@ private void internalFlushResponseToClient(boolean throwIOException) throws IOEx
...
+                        try
+                        {
+                           aos = response.getAsyncOutputStream();
+                        } catch (IOException x)
+                        {
+                           close(false);
+                           ret.completeExceptionally(x);
+                           return;
+                        }
+                        // eager composition to guarantee ordering
```
stuartwdouglas: I don't understand this comment? Ordering is important everywhere and nowhere else seems to use this pattern?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -56,13 +164,106 @@ public void close() throws IOException
...
+      public synchronized void onWritePossible() throws IOException
+      {
+         flushQueue(response.getOutputStream());
+      }
+
+      @Override
+      public synchronized void onError(Throwable t)
+      {
+         if(lastAsyncOperation != null) {
+            lastAsyncOperation.future.completeExceptionally(t);
```
stuartwdouglas: Do where need to complete the full queue here?

server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/MultiPromise.java
```diff
@@ -0,0 +1,55 @@
...
+      if(!promise.isDone()) {
+         Throwable throwable = cause.get();
+         if(throwable != null)
+            promise.setFailure(throwable);
+         else
+            promise.setSuccess();
+      }
+   }
+
+   public void readyToForward()
```
stuartwdouglas: This needs to be synchronised, as otherwise threads might not see the updated value of 'started'.

server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/MultiPromise.java
```diff
@@ -0,0 +1,55 @@
...
+      this.promise = promise;
+   }
+
+   public ChannelPromise newPromise()
+   {
+      created.getAndIncrement();
+      return ctx.newPromise().addListener(f -> {
+         if(!f.isSuccess())
+            cause.compareAndSet(null, f.cause());
+         if(created.decrementAndGet() == 0)
```
stuartwdouglas: If this happens as 'readyToForward' is being called both this method and 'readyToForward' could see the count as zero and the logic in forward() would be called twice.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
```diff
@@ -71,4 +75,45 @@ public void close() throws IOException {
...
+     protected CompletionStage<Void> asyncWritePart(AsyncOutputStream entityStream, byte[] boundaryBytes, OutputPart part, MultivaluedMap<String, Object> headers)
+     {
+        headers.putAll(part.getHeaders());
+        headers.putSingle(HttpHeaderNames.CONTENT_TYPE, part.getMediaType());
+
+        Object entity = part.getEntity();
+        Class<?> entityType = part.getType();
+        Type entityGenericType = part.getGenericType();
+        AsyncMessageBodyWriter writer = (AsyncMessageBodyWriter) workers.getMessageBodyWriter(entityType, entityGenericType, null, part.getMediaType());
+        LogMessages.LOGGER.debugf("MessageBodyWriter: %s", writer.getClass().getName());
```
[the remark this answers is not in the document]
stuartwdouglas: Looks like there is already heaps of code that uses this pattern, so I would not worry about it.

patriot1burke:
@FroMage IMO, don't worry about smime and doseta.  I don't think they were ever very popular.  IIRC, your changes with with blocking I/O anyways.

stuartwdouglas:
The only other concern I have is making sure that if an AsyncMessageBodyWriter throws an unchecked exception (as opposed to return a CompletableFuture that has completed exceptionally) that it is actually handled correctly, and won't result in the request hanging. This is a super common problem with async and looking at the code I don't think that these guards are present?

server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/MultiPromise.java
FroMage: It can be called twice, because the output promise is checked for `isDone` to not complete it twice.

server-adapters/resteasy-netty4/src/main/java/org/jboss/resteasy/plugins/server/netty/MultiPromise.java
FroMage: OK

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
FroMage: Well that's the limit of my understanding of this brain-dead servlet async IO API. I guessed the error was for a random operation, so I took the last one. No idea if the error applies to all queued async ops. Do you know?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
FroMage: That's special to SSE, which supports CS in the API as a return type, but the JAX-RS spec always gives examples where they call `CS<> writeEvent()` in sequence without chaining them and using their return type. This is super wrong, because if you use `compose` in `writeEvent` then it's possible only the first part of the composition is actually called by the time `writeEvent` returns.

So for example, if `writeEvent` is a composition of `A, B`, and you call `writeEvent(); writeEvent()` it's very possible that the sequence of calls end up being `A, A, B, B`.

This only happens if you ignore the return value of `writeEvent` which is WRONG, but since the JAX-RS spec examples do that, I've had to support the existing tests doing it too.

We can't support this sort of behaviour everywhere though, so for the new async io APIs I don't want to support this broken pattern, and we want users to actually use the returned CS for proper chaining, which is why I haven't supported this behaviour elsewhere.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
FroMage: OK

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
FroMage: So yeah, I know input is broken too because we need async IO support. I plan to do this later though as it wasn't as urgent and this has taken enough time.

It's entirely possible (and very likely) that the input in question is not backed by an IO thread, in which case yeah we're blocking the IO thread, but not deadlocking it.

If it is backed by an IO thread, I don't have the possibility to detect it ATM. I would prefer to delay solving this until later.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/DataSourceProvider.java
```diff
@@ -278,6 +280,22 @@ public void writeTo(DataSource dataSource,
...
 
+   @Override
+   public CompletionStage<Void> asyncWriteTo(DataSource dataSource, Class<?> type, Type genericType, Annotation[] annotations,
+                                             MediaType mediaType, MultivaluedMap<String, Object> httpHeaders,
+                                             AsyncOutputStream entityStream)
+   {
+      LogMessages.LOGGER.debugf("Provider : %s,  Method : writeTo", getClass().getName());
+      try
+      {
+         InputStream in = dataSource.getInputStream();
```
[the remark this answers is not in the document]
FroMage: Same answer as below.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
```diff
@@ -0,0 +1,95 @@
...
+   }
+
+   @Override
+   public void flush() throws IOException {
+      stream.flush();
+   }
+
+   @Override
+   public void close() throws IOException {
+      stream.close();
```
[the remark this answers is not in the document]
FroMage: It doesn't in the blocking version. Didn't want to deviate from it.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
```diff
@@ -0,0 +1,95 @@
...
+
+   @Override
+   public void write(byte[] bytes, int i, int i1) throws IOException {
+      flushHeaders();
+      stream.write(bytes, i, i1);
+   }
+
+   @Override
+   public void flush() throws IOException {
+      stream.flush();
```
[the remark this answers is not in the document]
FroMage: It doesn't in the blocking version. Didn't want to deviate from it.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
```diff
@@ -0,0 +1,95 @@
...
+            } else {
+               value = obj.toString();
+            }
+            ret = stream.asyncWrite(key.getBytes())
+                    .thenCompose(v -> stream.asyncWrite(": ".getBytes()))
+                    .thenCompose(v -> stream.asyncWrite(value.getBytes()))
+                    .thenCompose(v -> stream.asyncWrite("\r\n".getBytes()));
+         }
+      }
+      return ret.thenCompose(v -> stream.asyncWrite("\r\n".getBytes()));
```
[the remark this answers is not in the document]
FroMage: OK

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
```diff
@@ -0,0 +1,95 @@
...
+                  .createHeaderDelegate(obj.getClass());
+            if (headerDelegate != null) {
+               value = headerDelegate.toString(obj);
+            } else {
+               value = obj.toString();
+            }
+            ret = stream.asyncWrite(key.getBytes())
+                    .thenCompose(v -> stream.asyncWrite(": ".getBytes()))
+                    .thenCompose(v -> stream.asyncWrite(value.getBytes()))
+                    .thenCompose(v -> stream.asyncWrite("\r\n".getBytes()));
```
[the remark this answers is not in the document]
FroMage: We can't without chaining promises anyway in many cases, so there will still be underlying costs.

Actually it can be cheaper but only in the case we buffer and flush in chunks, which is the case for vertx/netty backends but not quarkus/servlet, so IMO it's not worth it for us ATM.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
stuartwdouglas: Once there is an error everything is broken, you are not going to get any more notifications as writing is done.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
```diff
@@ -0,0 +1,95 @@
...
+            String value;
+            RuntimeDelegate.HeaderDelegate headerDelegate = delegate
+                  .createHeaderDelegate(obj.getClass());
+            if (headerDelegate != null) {
+               value = headerDelegate.toString(obj);
+            } else {
+               value = obj.toString();
+            }
+            ret = stream.asyncWrite(key.getBytes())
+                    .thenCompose(v -> stream.asyncWrite(": ".getBytes()))
```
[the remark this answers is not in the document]
FroMage: OK

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
```diff
@@ -0,0 +1,95 @@
...
+         for (Object obj : objs) {
+            String value;
+            RuntimeDelegate.HeaderDelegate headerDelegate = delegate
+                  .createHeaderDelegate(obj.getClass());
+            if (headerDelegate != null) {
+               value = headerDelegate.toString(obj);
+            } else {
+               value = obj.toString();
+            }
+            ret = stream.asyncWrite(key.getBytes())
```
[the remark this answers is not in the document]
FroMage: OK

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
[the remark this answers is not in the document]
FroMage: Yeah, this is a copy of the blocking version.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
```diff
@@ -71,4 +75,45 @@ public void close() throws IOException {
...
+        return entityStream.asyncWrite(boundaryBytes)
+                .thenCompose(v -> entityStream.asyncWrite("\r\n".getBytes()))
```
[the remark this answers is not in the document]
FroMage: OK

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
```diff
@@ -71,4 +75,45 @@ public void close() throws IOException {
...
+     {
+        String boundary = mediaType.getParameters().get("boundary");
+        if (boundary == null)
+           boundary = multipartOutput.getBoundary();
+        httpHeaders.putSingle(HttpHeaderNames.CONTENT_TYPE, mediaType.toString() + "; boundary=" + multipartOutput.getBoundary());
+        byte[] boundaryBytes = ("--" + boundary).getBytes();
+
+        return asyncWriteParts(multipartOutput, entityStream, boundaryBytes)
+                .thenCompose(v -> entityStream.asyncWrite(boundaryBytes))
+                .thenCompose(v -> entityStream.asyncWrite("--".getBytes()));
```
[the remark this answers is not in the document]
FroMage: OK

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java
```diff
@@ -71,4 +75,45 @@ public void close() throws IOException {
...
       entityStream.write("\r\n".getBytes());
    }
+
+   protected CompletionStage<Void> asyncWrite(MultipartOutput multipartOutput, MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, AsyncOutputStream entityStream)
+     {
+        String boundary = mediaType.getParameters().get("boundary");
+        if (boundary == null)
+           boundary = multipartOutput.getBoundary();
+        httpHeaders.putSingle(HttpHeaderNames.CONTENT_TYPE, mediaType.toString() + "; boundary=" + multipartOutput.getBoundary());
+        byte[] boundaryBytes = ("--" + boundary).getBytes();
```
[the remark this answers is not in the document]
FroMage: OK

FroMage:
> The only other concern I have is making sure that if an AsyncMessageBodyWriter throws an unchecked exception (as opposed to return a CompletableFuture that has completed exceptionally) that it is actually handled correctly, and won't result in the request hanging. This is a super common problem with async and looking at the code I don't think that these guards are present?

OK lemme test that.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
stuartwdouglas: ok, makes sense

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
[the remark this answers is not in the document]
stuartwdouglas: Servlet always buffers, the content is copied into a direct buffer internally, and it does not actually flush to the wire until the buffer is full.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
stuartwdouglas: I agree that we can ignore this for now. If anyone actually hits it we can tell them not to return a blocking stream from an IO thread, but I don't think there is a real use case where you would run into this.

providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/HeaderFlushedAsyncOutputStream.java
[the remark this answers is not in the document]
FroMage: Servlet buffers, but internally, so I don't see it and still have to call `write(byte[])` and wire the notifications, so I still pay the cost of wiring them.

FroMage:
> OK lemme test that

Tested, it works.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
FroMage: OK done.

FroMage:
> @FroMage there seems to be a failing test in the vert.x integration. Would you mind having a look?

Damnit. This runs locally. Let's see if it fails again?

asoldano:
> > @FroMage there seems to be a failing test in the vert.x integration. Would you mind having a look?
> 
> Damnit. This runs locally. Let's see if it fails again?

Unfortunately, it seems to consistently fail :-/

FroMage:
Well, at least it's consistent. Probably something stupid, but I forgot how I can view the test output?

FroMage:
I've tried with a faster CI and cat the output.

asoldano:
> I've tried with a faster CI and cat the output.

In the past I've used something like this (note the path):

after_script:
 - cat /home/travis/build/resteasy/Resteasy/testsuite/integration-tests/target/surefire-reports/org.jboss.resteasy.test.providers.sse.SseTest-output.txt

FroMage:
OK thanks, found it. I forgot to push a resource file :(

FroMage:
Yay, thanks a lot!

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -1,32 +1,140 @@
...
 
    /**
     * RESTEASY-684 wants to defer access to outputstream until a write happens
     *
     */
-   protected class DeferredOutputStream extends OutputStream
+   protected class DeferredOutputStream extends AsyncOutputStream implements WriteListener
    {
+      private boolean asyncRegistered;
+      private Queue<AsyncOperation> asyncQueue;
```
crankydillo, line 135: I know this is an old-ish PR, but we are working on a possible bug in this area.  I'm trying to understand why a queue is needed.  Aren't clients only supposed to call `asyncWrite` when a previous `asyncWrite` or `asyncFlush` has completed?  I realize there is no guarantee on what callers will do, and the queue could possibly handle weird cases.  Is that why a queue was used?  Trying to understand if it's needed if the caller is chaining them with the `CompletionStage` api (like [here](https://github.com/resteasy/resteasy/blob/6e94de442cd2aac93207c4e1fc06a1bbb214c1d0/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java#L164)).

Anyhow, I know this code has a changed a bit.  If this is better asked in some other way, just let me know.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -56,13 +164,111 @@ public void close() throws IOException
...
+                  // we're not allowed to queue work if the output is ready
+                  lastAsyncOperation = op;
+                  op.work(os);
+               } else {
+                  // just queue
+                  addToQueue(op);
+               }
+            }
+         } else {
+            op.work(null);
```
crankydillo, line 219: How is this ok?  We seem to be getting here and no bytes are happening and the thread hangs waiting for a future that will never complete to happen.  It does appear the bytes are here.

jamezp: @crankydillo What version are you using? I made some changes around here recently. But it should be completed here https://github.com/resteasy/resteasy/pull/2170/files/3c11dd4f389837dc22474e6b4f4d23926a88ccea#diff-a2be1ee2e763d57c738434047303d73e02cf6171a606b8dc34cf3fcc0aa3b3d6R104-R106

crankydillo:
@jamezp We were using 4.7.3.  We are trying 4.7.4, which has your fix.  However, I believe my colleagues think there are some issues with it.  We'll let you know.

PR #2176, opened 2019-10-03
Add support for remote ip/address without requiring HttpServletRequest

FroMage:
New interface ResteasyRequest extends JAX-RS Request with those methods
But it delegates to HttpRequest (perhaps we don't need both?)

This works, but we need to make some decisions:

- Do we keep `ResteasyRequest` as a subtype of JAX-RS `Request`, or do we tell users to inject RESTEasy's `HttpRequest` which has to have those methods anyway?
- Do we keep the `String getRemoteAddress/getRemoteHost` duo or go for something like `InetSocketAddress getRemoteAddress` like Netty has?

WDYT @patriot1burke @asoldano ?

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/HttpRequest.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/HttpRequest.java
@@ -90,4 +90,24 @@ public interface HttpRequest
 
    boolean wasForwarded();
 
+   /**
+    * Returns the Internet Protocol (IP) address of the client
+    * or last proxy that sent the request.
+    *
+    * @return a <code>String</code> containing the
+    * IP address of the client that sent the request
+    */
+   String getRemoteAddress();
+
+   /**
+    * Returns the fully qualified name of the client
+    * or the last proxy that sent the request.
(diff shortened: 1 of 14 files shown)


FroMage:
Hah, saw an SSE test fail, but this is not due to my changes this time. That makes me think that perhaps it wasn't my changes which were guilty in the async io branch…

FroMage:
OK, thanks for the feedback. I will udpate the PR.

asoldano:
Thanks!

FroMage:
Done.

FroMage:
`SseTest.testReconnect:309 expected:<0> but was:<1>` failed again in _one_ build, but this is not related to my PR. Restarted it.

FroMage:
Huh, @asoldano I merged it because I thought you had accepted the PR, but now I checked and it's not clear. Is it alright or did I mess up?

asoldano:
@FroMage no problem for having merged it, I would have done that anyway (just haven't had time yesterday). The only caveat for next time would be that we require all commits to have a reference to the relevant jira, which in this case is missing (btw, if the original commit message does not have the reference, a trick is to add it merging the PR with "Squash and merge" as that allows editing the message)

PR #2181, opened 2019-10-14
Allow UnmodifiableMultivaluedMap to create EntrySet lazily

geoand:
Relates to: https://github.com/quarkusio/quarkus/issues/4345

Basically in a lot of the benchmarks the EntrySet of `UnmodifiableMultivaluedMap` is never queried, however without this patch it is always created.

During CPU profiling (using async-profiler) of Quarkus `0.24.0` which uses RESTEasy `4.3.1.Final`  I discovered entries like the following:

```
--- 220900450 ns (0.13%), 22 samples
  [ 0] ObjectSynchronizer::FastHashCode(Thread*, oopDesc*)
  [ 1] JVM_IHashCode
  [ 2] java.lang.Object.hashCode
  [ 3] java.util.HashMap.hash
  [ 4] java.util.HashMap.put
  [ 5] java.util.HashSet.add
  [ 6] org.jboss.resteasy.specimpl.UnmodifiableMultivaluedMap.buildEntrySet
  [ 7] org.jboss.resteasy.specimpl.UnmodifiableMultivaluedMap.<init>
  [ 8] org.jboss.resteasy.specimpl.ResteasyHttpHeaders.<init>
  [ 9] org.jboss.resteasy.specimpl.ResteasyHttpHeaders.<init>
  [10] org.jboss.resteasy.specimpl.ResteasyHttpHeaders.<init>
  [11] io.quarkus.resteasy.runtime.standalone.VertxUtil.extractHttpHeaders
  [12] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch
  [13] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatchRequestContext
  [14] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.lambda$handle$0
  [15] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$$Lambda$228.2111501191.handle
  [16] io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2
  [17] io.vertx.core.impl.ContextImpl$$Lambda$240.1018596449.run
  [18] java.util.concurrent.ThreadPoolExecutor.runWorker
  [19] java.util.concurrent.ThreadPoolExecutor$Worker.run
  [20] io.netty.util.concurrent.FastThreadLocalRunnable.run
  [21] java.lang.Thread.run
```

With this PR those entries completely go away

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/UnmodifiableMultivaluedMap.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/UnmodifiableMultivaluedMap.java
@@ -20,21 +20,24 @@ public class UnmodifiableMultivaluedMap<K, V> implements MultivaluedMap<K, V>
 
    private final MultivaluedMap<K, V> delegate;
 
-   private final Set<java.util.Map.Entry<K, List<V>>> entrySet;
+   private volatile Set<java.util.Map.Entry<K, List<V>>> entrySet;
 
    public UnmodifiableMultivaluedMap(final MultivaluedMap<K, V> delegate)
    {
+      this(delegate, true);
+   }
+
+   public UnmodifiableMultivaluedMap(final MultivaluedMap<K, V> delegate, final boolean eagerlyInitializeEntrySet)
+   {
       this.delegate = delegate;
-      this.entrySet = buildEntrySet();
+      if (eagerlyInitializeEntrySet) {
+         this.entrySet = buildEntrySet();
+      }
    }
 
    private Set<java.util.Map.Entry<K, List<V>>> buildEntrySet()
    {
       Set<java.util.Map.Entry<K, List<V>>> entrySetDelegator = delegate.entrySet();
-      if (entrySetDelegator == null)
-      {
-         return null;
-      }
       Set<java.util.Map.Entry<K, List<V>>> entrySetToReturn = new HashSet<>();
       for (final java.util.Map.Entry<K, List<V>> entry : entrySetDelegator)
       {
@@ -116,6 +119,9 @@ public class UnmodifiableMultivaluedMap<K, V> implements MultivaluedMap<K, V>
    @Override
    public Set<Entry<K, List<V>>> entrySet()
    {
+      if (this.entrySet == null) {
+         this.entrySet = buildEntrySet();
+      }
       return this.entrySet;
    }
 
--- a/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyHttpHeaders.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyHttpHeaders.java
@@ -34,10 +34,20 @@ public class ResteasyHttpHeaders implements HttpHeaders
       this(requestHeaders, new HashMap<String, Cookie>());
    }
 
+   public ResteasyHttpHeaders(final MultivaluedMap<String, String> requestHeaders, final boolean eagerlyInitializeEntrySet)
+   {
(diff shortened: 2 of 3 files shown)


Sanne:
nice patches :) I'm not at all an expert of RESTEasy so I'll let @asoldano review, but wondering if you could just always initialize it lazily so to not have to change the constructors?
Incidentally it would benefit all clients, no need for adjustments.

geoand:
I was wondering the same, but since I don't know I just kept it optional for now :).

If @asoldano thinks it's safe to always make it lazy, I'll of course go ahead and do it.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/UnmodifiableMultivaluedMap.java
```diff
@@ -116,6 +119,9 @@ public void clear()
    @Override
    public Set<Entry<K, List<V>>> entrySet()
    {
+      if (this.entrySet == null) {
+         this.entrySet = buildEntrySet();
+      }
```
asoldano: I have a concern with this together with the removal of final in entrySet declaration (this is not thread-safe anymore). Do you think we need to care about this and add a synchronized block here?

Sanne: I'd make the field `volatile` and stay away from synchronization. Worst case is you'll have multiple threads building the same set, but that shouldn't be a problem (they are supposed to all build the same set right?).

geoand: I was about to comment about the result being the same even if it's created multiple times as.
I like the volatile solution, so I'll update the PR with it.

geoand:
@asoldano apache commons isn't part of the dependency tree already is it? If not (which should be the case unless I missed something) I don't think it should added just for this case.

geoand:
About the profiling data, I did see a small improvement with this patch but like @Sanne said in another thread, it really depends on the type of benchmark that is being run.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/UnmodifiableMultivaluedMap.java
asoldano: I agree with the optimistic approach here (using ``volatile``).

asoldano:
> nice patches :) I'm not at all an expert of RESTEasy so I'll let @asoldano review, but wondering if you could just always initialize it lazily so to not have to change the constructors?
> Incidentally it would benefit all clients, no need for adjustments.

Well, in theory, I believe whoever uses the UnmodifiableMultivaluedMap would expect a snapshot of the provided map to be taken at creation time. So unless we rename the class, I think it's better to have the lazy approach enabled with a different constructor. But we can go through the list of UnmodifiableMultivaluedMap callers and see if they could use the lazy approach...

geoand:
> > nice patches :) I'm not at all an expert of RESTEasy so I'll let @asoldano review, but wondering if you could just always initialize it lazily so to not have to change the constructors?
> > Incidentally it would benefit all clients, no need for adjustments.
> 
> Well, in theory, I believe whoever uses the UnmodifiableMultivaluedMap would expect a snapshot of the provided map to be taken at creation time. So unless we rename the class, I think it's better to have the lazy approach enabled with a different constructor.

I agree as well. It's more intuitive as a default.

geoand:
I think the CI failure is probably a flake

asoldano:
Merged, thanks!

PR #2183, opened 2019-10-14
Cache result of computation of ResteasyUriInfo

geoand:
Relates to: https://github.com/quarkusio/quarkus/issues/4345

This PR is meant to eliminate with such async-profiles results:

```
--- 2755952637848 bytes (0.82%), 88 samples
  [ 0] int[]
  [ 1] java.util.regex.Matcher.<init>
  [ 2] java.util.regex.Pattern.matcher
  [ 3] org.jboss.resteasy.util.Encode.decodePath
  [ 4] org.jboss.resteasy.specimpl.ResteasyUriInfo.initialize
  [ 5] org.jboss.resteasy.specimpl.ResteasyUriInfo.<init>
  [ 6] io.quarkus.resteasy.runtime.standalone.VertxUtil.extractUriInfo
  [ 7] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch
  [ 8] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatchRequestContext
  [ 9] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.lambda$handle$0
  [10] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$$Lambda$211.2038650648.handle
  [11] io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2
  [12] io.vertx.core.impl.ContextImpl$$Lambda$216.1802137975.run
  [13] java.util.concurrent.ThreadPoolExecutor.runWorker
  [14] java.util.concurrent.ThreadPoolExecutor$Worker.run
  [15] io.netty.util.concurrent.FastThreadLocalRunnable.run
  [16] java.lang.Thread.run
```
,

```
--- 1386496101488 bytes (0.41%), 44 samples
  [ 0] char[]
  [ 1] java.lang.AbstractStringBuilder.<init>
  [ 2] java.lang.StringBuilder.<init>
  [ 3] org.jboss.resteasy.util.Encode.decodePath
  [ 4] org.jboss.resteasy.specimpl.ResteasyUriInfo.initialize
  [ 5] org.jboss.resteasy.specimpl.ResteasyUriInfo.<init>
  [ 6] io.quarkus.resteasy.runtime.standalone.VertxUtil.extractUriInfo
  [ 7] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch
  [ 8] io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatchRequestContext
[...]

PR #2184, opened 2019-10-15
[RESTEASY-2375] Allow Caching of ResteasyUriInfo construction data

geoand:
Follow up of: #2183

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
@@ -48,38 +48,36 @@ public class ResteasyUriInfo implements UriInfo
    private String absoluteString;
    private String contextPath;
    private int queryIdx;
-   private int endPath;
    private int pathStart;
 
 
    public ResteasyUriInfo(final String absoluteUri, final String contextPath) {
-      initialize(absoluteUri, contextPath);
+      this(absoluteUri, contextPath, null);
    }
 
-   protected void initialize(String absoluteUri, String contextPath) {
+   /**
+    * Meant to be called by a layer that know that the absoluteUri will result in the successful
+    * invocation of a JAX-RS endpoint.
+    * The caller is also expected to ensure that InitData is cacheable
+    */
+   public ResteasyUriInfo(final String absoluteUri, final String contextPath, final InitData initData) {
+      initialize(absoluteUri, contextPath, initData != null && InitData.canBeCached(absoluteUri) ? initData : null);
+   }
+
+   protected void initialize(String absoluteUri, String contextPath, InitData initData) {
       this.absoluteString = absoluteUri;
       this.contextPath = contextPath;
 
-      int pathIdx = absoluteUri.indexOf('/');
-      if (pathIdx > 0 && absoluteUri.length() > 3) {
-         if (absoluteUri.charAt(pathIdx - 1) == ':' && absoluteUri.charAt(pathIdx + 1) == '/') {
-            pathIdx = pathIdx + 2;
-            int tmp = absoluteUri.indexOf('/', pathIdx);
-            if (tmp > -1) pathIdx = tmp;
-         }
+      if (initData == null) {
+         initData = new InitData(absoluteUri, contextPath);
       }
-      queryIdx = pathIdx > -1 ? absoluteUri.indexOf('?', pathIdx) : absoluteUri.indexOf('?');
-      endPath = queryIdx > -1 ? queryIdx : absoluteUri.length();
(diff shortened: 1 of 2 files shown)


geoand:
@asoldano I think this is ready for review

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
```diff
@@ -499,4 +539,97 @@ public URI relativize(URI uri)
...
+         return !absoluteUri.contains(";");
+      }
+
+      // ensure that we don't include any of the query params into the cache key
+      // since they don't affect the production of data at all
+      public static String getCacheKey(String absoluteUri, String contextPath) {
+         int queryIdx = absoluteUri.lastIndexOf('?');
+         return (queryIdx > -1 ? absoluteUri.substring(0, queryIdx) : absoluteUri) + "ResteasyUriInfo-Delimiter" + contextPath;
+      }
+   }
```
asoldano: This method is not called anywhere in RESTEasy, so assuming it's needed in Quarkus (or other consumers of RESTEasy), we need a test covering for it. Moreover, line 631 has the same logic of line 587 and I believe it's important for those to stay in synch, which would suggest me to factor that into a single dedicated method.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
```diff
@@ -207,13 +242,18 @@ public void initializeFromRequest(URI requestURI)
     * @param encodedPathSegments list of path segments
     */
    protected void extractMatchingPath(List<PathSegment> encodedPathSegments)
+   {
+      matchingPath = doExtractMatchingPath(encodedPathSegments);
+   }
```
asoldano: Is this method still needed?

asoldano:
@geoand , I've created https://issues.jboss.org/browse/RESTEASY-2375 , please prefix your commit message with "[RESTEASY-2375]" if you're going to rewrite your commit. Thanks.

geoand:
@asoldano Thanks! I will take a look at your comments on Monday.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
```diff
@@ -48,38 +48,36 @@
...
    }
 
-   protected void initialize(String absoluteUri, String contextPath) {
+   /**
+    * Meant to be called by a layer that know that the absoluteUri will result in the successful
+    * invocation of a JAX-RS endpoint.
+    * The caller is also expected to ensure that InitData is cacheable
+    */
+   public ResteasyUriInfo(final String absoluteUri, final String contextPath, final InitData initData) {
+      initialize(absoluteUri, contextPath, InitData.canBeCached(absoluteUri) ? initData : null);
```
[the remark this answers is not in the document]
geoand: Good idea! Done

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
geoand: Nice catch! I got rid of the other part of the code since it wasn't being used anywhere.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriInfo.java
geoand: Not at all :). Removed

geoand:
@asoldano I took care of your comments and added some unit tests. 
Would you like to point me the server adaptors tests you have in mind?

asoldano:
> @asoldano I took care of your comments and added some unit tests.
> Would you like to point me the server adaptors tests you have in mind?

Thanks @geoand . As for the integration tests, I was thinking about something here https://github.com/resteasy/Resteasy/tree/master/server-adapters/resteasy-vertx/src/test for example (but you can choose a different server adapter to use for testing, if you prefer)

geoand:
@asoldano I'm not really seeing how to to use the server-adapters for testing this change.
However I am thinking that since this change essentially affects all requests, if there was a problem it would have manifested in some of the existing tests, no?

asoldano:
> @asoldano I'm not really seeing how to to use the server-adapters for testing this change.
> However I am thinking that since this change essentially affects all requests, if there was a problem it would have manifested in some of the existing tests, no?

Right, and that's why I'm not concerned with regressions atm. What I was really thinking was that the new execution path (with the cache enabled) should have been tested in an integration test too, to avoid future changes from causing regressions. At least we have unit test, though.

geoand:
Thanks @asoldano!

PR #2185, opened 2019-10-15
[RESTEASY-2366] Introduce RESTEasy integration with WildFly Elytr…

Skyllarr:
…on - AuthenticationClient for Authentication / SSL

issue link: [RESTEASY-2366](https://issues.jboss.org/browse/RESTEASY-2366)

RFE Proposal :  https://github.com/wildfly/wildfly-proposals/pull/189

These are proposed changes for the use of new client SPI that wildfly-elytron will implement.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ClientHttpEngineBuilder43.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ClientHttpEngineBuilder43.java
@@ -1,11 +1,16 @@
 package org.jboss.resteasy.client.jaxrs;
 
 import java.io.IOException;
+import java.net.Socket;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.security.SecureRandom;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
+import java.util.ServiceLoader;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SNIHostName;
 import javax.net.ssl.SNIServerName;
@@ -30,10 +35,13 @@ import org.apache.http.conn.util.PublicSuffixMatcherLoader;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.conn.BasicHttpClientConnectionManager;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.apache.http.protocol.HttpContext;
+import org.apache.http.protocol.HttpCoreContext;
 import org.apache.http.ssl.SSLContexts;
 import org.jboss.resteasy.client.jaxrs.engines.ApacheHttpClient43Engine;
 import org.jboss.resteasy.client.jaxrs.engines.PassthroughTrustManager;
 import org.jboss.resteasy.client.jaxrs.engines.factory.ApacheHttpClient4EngineFactory;
+import org.jboss.resteasy.client.jaxrs.spi.ClientConfigProvider;
 
 public class ClientHttpEngineBuilder43 implements ClientHttpEngineBuilder {
 
@@ -75,6 +83,7 @@ public class ClientHttpEngineBuilder43 implements ClientHttpEngineBuilder {
       {
          SSLConnectionSocketFactory sslsf = null;
          SSLContext theContext = that.sslContext;
+         Iterator clientConfigProviderIterator = ServiceLoader.load(ClientConfigProvider.class).iterator();
          if (that.disableTrustManager)
          {
             theContext = SSLContext.getInstance("SSL");
@@ -117,6 +126,26 @@ public class ClientHttpEngineBuilder43 implements ClientHttpEngineBuilder {
                protected void prepareSocket(SSLSocket socket) throws IOException
                {
                   that.prepareSocketForSni(socket);
+               }
+            };
+         } else if (clientConfigProviderIterator.hasNext())
+         {
+            // delegate creation of socket to ClientConfigProvider implementation
+            final ClientConfigProvider configProvider = ((ClientConfigProvider) clientConfigProviderIterator.next());
+            sslsf = new SSLConnectionSocketFactory(SSLContext.getDefault(), verifier) {
+               @Override
+               public Socket createSocket(HttpContext context) throws IOException {
+                  try {
+                     String targetHostUri = context.getAttribute(
+                             HttpCoreContext.HTTP_TARGET_HOST).toString();
+                     if (targetHostUri != null) {
+                        return configProvider.getSSLContext(new URI(targetHostUri)).getSocketFactory().createSocket();
+                     } else {
+                        throw new RuntimeException("URI is not known");
+                     }
+                  } catch (URISyntaxException e) {
+                     throw new RuntimeException(e);
+                  }
                }
             };
          }
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfigProviderFilter.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfigProviderFilter.java
@@ -1,0 +1,48 @@
+package org.jboss.resteasy.client.jaxrs.internal;
+
+import org.jboss.resteasy.client.jaxrs.i18n.LogMessages;
+import org.jboss.resteasy.client.jaxrs.i18n.Messages;
+import org.jboss.resteasy.client.jaxrs.spi.ClientConfigProvider;
+import org.jboss.resteasy.util.BasicAuthHelper;
+
+import javax.ws.rs.client.ClientRequestContext;
+import javax.ws.rs.client.ClientRequestFilter;
+import javax.ws.rs.core.HttpHeaders;
+import java.io.IOException;
+import java.net.URI;
+
+/**
+ * Client filter that will attach authorization header with either HTTP Basic auth or Bearer token auth.
+ * Credentials and token are loaded from ClientConfigProvider implementation.
+ *
+ * @author dvilkola@redhat.com
+ */
+public class ClientConfigProviderFilter implements ClientRequestFilter {
+
+    private final ClientConfigProvider clientConfigProvider;
+
+    public ClientConfigProviderFilter(final ClientConfigProvider clientConfigProvider) {
+        this.clientConfigProvider = clientConfigProvider;
+    }
+
+    @Override
+    public void filter(ClientRequestContext requestContext) throws IOException {
+        if (requestContext.getHeaderString(HttpHeaders.AUTHORIZATION) == null) {
+            URI uri = requestContext.getUri();
+            if (uri == null) {
+                LogMessages.LOGGER.warn(Messages.MESSAGES.unableToLoadClientConfigProviderConfiguration());
+                return;
+            }
+            String token = clientConfigProvider.getBearerToken(uri);
+            if (token != null) {
+                requestContext.getHeaders().putSingle(HttpHeaders.AUTHORIZATION, "Bearer " + token);
+            } else {
+                String username = clientConfigProvider.getUsername(uri);
+                String password = clientConfigProvider.getPassword(uri);
+                if (username != null && password != null) {
+                    requestContext.getHeaders().putSingle(HttpHeaders.AUTHORIZATION, BasicAuthHelper.createHeader(username, password));
+                }
+            }
+        }
+    }
+}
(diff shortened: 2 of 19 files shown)


asoldano:
@Skyllarr , I'll review your PR here soon. In the meantime, please rebase it against master: I believe that will solve the SpecialResourceTest failure.

Skyllarr:
@asoldano Thank you for the info! I am currently on PTO with limited access to the internet. I will be back in 3 weeks and will rebase then.

Skyllarr:
@asoldano I rebased this against master (SpecialResourceTest failure got fixed). Thanks!

Skyllarr:
@asoldano @ronsigal I changed the PR branch to 3.10 as requested. All tests are passing for me locally now. Please review and let me know where the ClientConfigProvider SPI  should be located to avoid circular dependency between wildfly-elytron and resteasy project's testsuite. Thanks a lot

Skyllarr:
@asoldano I added remaining tests. Please let me know how you would prefer to resolve circular dependency between wildfly-elytron and resteasy's testsuite.

Skyllarr:
@ronsigal @asoldano I was not able to reproduce circular dependency. This PR is ready to be reviewed. Thanks!

jimma:
@Skyllarr   Looks this will bring circular dependency (resteasy<->elytron) issue. And spit spi module is not an option for Resteasy 3.10 release. We can't create another spi module and change all the dependency graph in Resteasy.  And this is kind of backwards incompatible change, especially for developer with Resteasy dependency.

To resolve this issue, we may need to make  Elytron client only depends jaxrs api. The idea is creating a new ElytronClientConfigration and pass to create the client :       
 ```
Client myclient = ClientBuilder.newBuilder().build();
      Configuration elytronConfiguraion = new ElytronClientConfiguration(myclient.getConfiguration());
      elytronConfiguraion.register(new ClientConfigProviderFilter(serviceLoaderIterator.next()), Priorities.AUTHENTICATION);
      elytronConfiguraion.register(SecurityContextResolver.class);
      Client client = ClientBuilder.newClient(elytronConfiguraion);
```

The SecurityContextResolve can be something like : 

```
 @Provider
public class SecurityContextResolver implements ContextResolver<Function<String, String>> {
    ...
   @Override
    public Function<String, String> getContext(Class<?> type) {
        if (type == SSLContext.class) {
            return getSslContextFunction;
        }
        return null;
    }
}
```

In ClientHttpEngineBuilder , this can be retrieved by this line check :
```
 if (that.getConfiguration().getRuntimeType()== RuntimeType.CLIENT && that.getProviderFactory().getContextResolver(SslContxt.class, MediaType.WILDF))
```

Or we can simply put this function to configuration property, and get it in ClientHttpEngineBuilder to create the ssl connection.

**Comment by Skyllarr** [comment]
> @Skyllarr Looks this will bring circular dependency (resteasy<->elytron) issue. And spit spi module is not an option for Resteasy 3.10 release. We can't create another spi module and change all the dependency graph in Resteasy. And this is kind of backwards incompatible change, especially for developer with Resteasy dependency.

@jimma Why is this not an option? This would be just 1 simple repo with single interface in it. There is no issue with backwards compatibility because this interface was not used before. Both resteasy and elytron would be dependent on this new spi repo which would resolve circular dep. Btw. circular dependency is only in test scope, which is not transitive.

> To resolve this issue, we may need to make Elytron client only depends jaxrs api...

The desscribed approach is not automatic and would require user to use Elytron classes. This RFE was meant in a way that RESTEasy would load elytron configuration automatically.

**Comment by jimma** [comment]
> > @Skyllarr Looks this will bring circular dependency (resteasy<->elytron) issue. And spit spi module is not an option for Resteasy 3.10 release. We can't create another spi module and change all the dependency graph in Resteasy. And this is kind of backwards incompatible change, especially for developer with Resteasy dependency.
> 
> @jimma Why is this not an option? This would be just 1 simple repo with single interface in it. There is no issue with backwards compatibility because this interface was not used before. Both resteasy and elytron would be dependent on this new spi repo which would resolve circular dep. Btw. circular dependency is only in test scope, which is not transitive.
>
Add another spi maven module or repo is required add another dependency when upgrade resteasy.
User has to add this dependency, jar file to the app when upgrade this minor version.  

> > To resolve this issue, we may need to make Elytron client only depends jaxrs api...
> 
> The desscribed approach is not automatic and would require user to use Elytron classes. This RFE was meant in a way that RESTEasy would load elytron configuration automatically.

Now client request filter and ssl security info has to be configured manually to client. We can look if this can be configured automatically in the 4.x version.

**Comment by Skyllarr** [comment]
@jimma Seems that wildfly-elytron could safely be removed from resteasy dependencies and bom, thus there is no circular dependency anymore. I am reevaluating my tests. Let me know if the interface looks good to you

**Comment by jimma** [comment]
Thanks @Skyllarr . So you are going to fix this by remove wildfly-elytron dependency for test ?

**Comment by Skyllarr** [comment]
@jimma Yes, I have removed elytron dependency from RESTEasy testuite. Please let me know if this approach is alright. I have also added 2 tests that use `mocked ClientConfigProviderImpl` to verify that the new` ClientConfigProvider` interface works for SSLContext and HTTP BASIC auth. Let me know if I should include more tests with this approach. Thanks.

**Comment by Skyllarr** [comment]
@jimma Currently the tests create a separate jar that contains mocked service provider `ClientConfigProviderImplMocked` that is registered in jar's `META-INF/services/` and Main class that uses resteasy client to connect to secured server. Tests verify that client used sslcontext and credentials from mocked provider and connected to secured server successfully. The reason I had to create separate jar is so the other testing classes do not pick up the `META-INF/services/ClientConfigProviderImplMocked`.

We think that tests testing integration with Elytron specifically should be moved to wildfly repository. WDYT?

**Comment by Skyllarr** [comment]
I have moved integration tests to wildfly repository

**Comment by jimma** [comment]
> @jimma Currently the tests create a separate jar that contains mocked service provider `ClientConfigProviderImplMocked` that is registered in jar's `META-INF/services/` and Main class that uses resteasy client to connect to secured server. Tests verify that client used sslcontext and credentials from mocked provider and connected to secured server successfully. The reason I had to create separate jar is so the other testing classes do not pick up the `META-INF/services/ClientConfigProviderImplMocked`. Is this approach OK?
> 
> We think that tests testing integration with Elytron specifically should be moved to wildfly repository. WDYT?

@Skyllarr  This looks good to me.  You must already check the module dependency and maven dependency when both Resteasy and ELytron upgrade in WildFly, there is no issue with this approach, right ?

**Comment by Skyllarr** [comment]
> @Skyllarr This looks good to me. You must already check the module dependency and maven dependency when both Resteasy and ELytron upgrade in WildFly, there is no issue with this approach, right ?

@jimma  I don't know if there is no issue with this approach for sure, but I did manage to locally build WildFly with both RESTEasy and Elytron upgrade built on my feature branches, if that answers your question.

**Comment by Skyllarr** [comment]
@ronsigal @asoldano I rebased this on 3.13. 

The CI is failing on MicroProfile TCK timeout tests, but these test are failing for me locally even without my changes.

**Comment by ronsigal** [comment]
Hi @Skyllarr,

It will take me a little while to get up to speed on this stuff, but I'll start studying. I wouldn't worry about the MP TCK timeouts. I had the same problem, and I was able to eliminate the problem by extending the timeout.

-Ron

File: resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfigProviderFilter.java [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+    public ClientConfigProviderFilter(final ClientConfigProvider clientConfigProvider) {
+        this.clientConfigProvider = clientConfigProvider;
+    }
+
+    @Override
+    public void filter(ClientRequestContext requestContext) throws IOException {
+        if (requestContext.getHeaderString(HttpHeaders.AUTHORIZATION) == null) {
+            URI uri = requestContext.getUri();
+            String token = clientConfigProvider.getBearerToken(uri);
```
ronsigal, line 30: ClientConfigProvider doesn't have any javadoc that guarantees that it will behave nicely if uri is null. Could there be a circumstance in which url == null? Maybe this is a stretch, but what if an application has a ClientRequestFilter that supplies the target URL, and that filter runs after this one? I don't know. But maybe ClientConfigProvider just needs some javadoc that defines its behavior. WDYT, @Skyllarr ?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfigProviderFilter.java
```diff
@@ -0,0 +1,42 @@
...
+            if (token != null) {
+                requestContext.getHeaders().putSingle(HttpHeaders.AUTHORIZATION, "Bearer " + token);
+            } else {
+                String username = clientConfigProvider.getUsername(uri);
+                String password = clientConfigProvider.getPassword(uri);
+                if (username != null && password != null) {
+                    requestContext.getHeaders().putSingle(HttpHeaders.AUTHORIZATION, BasicAuthHelper.createHeader(username, password));
+                }
```
ronsigal, line 38: Should it be considered an error if a ClientConfigProvider exists but it doesn't hold either a token or username/password? Or maybe there whould be a WARNING?

resteasy-dependencies-bom/pom.xml
```diff
@@ -881,11 +881,6 @@
                 <artifactId>wildfly-cli</artifactId>
                 <version>${version.org.wildfly.core.wildfly-cli}</version>
             </dependency>
-            <dependency>
-                <groupId>org.wildfly.security</groupId>
-                <artifactId>wildfly-elytron</artifactId>
-                <version>${version.org.wildfly.security.wildfly-elytron}</version>
-            </dependency>
```
ronsigal, line 888: I guess <version.org.wildfly.security.wildfly-elytron> could be removed ... .

ronsigal:
I haven't read every test, but it looks like a pretty rich set of tests.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfigProviderFilter.java
Skyllarr: @ronsigal Thanks for review! It is not considered error or warning if ClientConfigProvider does not hold token or credentials. It is not necessary to provide those, it could be that it is present only to provide SSL context.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientConfigProviderFilter.java
Skyllarr: @ronsigal Thank you for review, this is a very good point. I thought that requestContext always contains uri that is not null when filters are being evaluated, but that was probably wrong assumption. We could throw an exception of continue silently if uri is null here. 

I think it is best to log it as warning and continue silently to avoid problems in case elytron configuration is present on classpath for other clients that use it.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/ClientHttpEngineBuilder43.java
```diff
@@ -75,6 +83,7 @@ public ClientHttpEngine build()
       {
          SSLConnectionSocketFactory sslsf = null;
          SSLContext theContext = that.sslContext;
+         Iterator clientConfigProviderIterator = ServiceLoader.load(ClientConfigProvider.class).iterator();
```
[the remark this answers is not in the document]
Skyllarr: @asoldano in most scenarios it would work, but it is possible to have keystore configured without truststore as well. For example in elytron client we have a test configurations that use key-store but not truststore, eg. here: https://github.com/wildfly-security/wildfly-elytron/blob/1.x/auth/client/src/test/resources/org/wildfly/security/auth/client/test-wildfly-config-v1_5.xml#L51 .

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/testjar/ClientConfigTestMainClass.java
```diff
@@ -0,0 +1,61 @@
...
+            } catch (Exception e) {
+                if (e.getCause().getMessage().contains("unable to find valid certification path to requested target")) {
+                    result = "SSLHandshakeException";
+                }
+            }
+        }
+        //CHECKSTYLE.OFF: RegexpSinglelineJava
+        System.out.println(result);
+        //CHECKSTYLE.ON: RegexpSinglelineJava
+    }
```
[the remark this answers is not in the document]
Skyllarr: @asoldano Thank you. Fixed

resteasy-dependencies-bom/pom.xml
Skyllarr: @ronsigal Thank you. Fixed

Skyllarr:
@jimma This RFE is now pre-checked. Please let me know if there is anything more needed from me or this can be merged. Thanks

PR #2191, opened 2019-10-21
Multipart new annotations

FroMage:
Allow the multipart resolution to work with the new `@FormParam` annotations


FroMage:
Thanks!

PR #2192, opened 2019-10-21
[RESTEASY-2380] Update jackson-databind to 2.9.10.1

asoldano:


asoldano:
The failure on Travis CI is because of maven issues in their env, merging.

PR #2202, opened 2019-10-28
New jaxb and rest tck

asoldano:


resteasy-dependencies-bom/pom.xml
```diff
@@ -237,8 +237,8 @@
                         <artifactId>javax.activation-api</artifactId>
                     </exclusion>
                     <exclusion>
-                        <groupId>javax.xml.bind</groupId>
-                        <artifactId>jaxb-api</artifactId>
+                        <groupId>jakarta.xml.bind</groupId>
+                        <artifactId>jakarta.xml.bind-api</artifactId>
```
gsmet: But just to be sure, the API dependency is added somehow in the RESTEasy jaxb provider pom?

asoldano: Sure,  https://github.com/resteasy/Resteasy/blob/master/providers/jaxb/pom.xml#L35

gsmet: OK, perfect, then!

PR #2210, opened 2019-11-08
RESTEASY-828 Issue with injected ServletContext in Resteasy/SpringMVC setup

liweinan:
The default Spring `DispatcherServlet` doesn't push the  `ServletContext` into `ResteasyContext` properly, however using `DispatcherServlet` is a convenient way to configure `springframework`, so we need to provided a `ResteasySpringDispatcherServlet` for users to use.

Please note using `DispatcherServlet` is not the only way to configure the RESTEasy/Spring integration, and we also have the other ways such as:

- Using the `resteasy-spring-boot` project to eliminate the manual configuration with Spring/SpringMVC: https://github.com/resteasy/resteasy-spring-boot
- Totally configuring the `web.xml` manually to setup spring container properly and don't use `DispatcherServlet` at all(which is very verbose and not the standard way described in document), and here is the example: https://github.com/resteasy/Resteasy/blob/master/testsuite/integration-tests-spring/inmodule/src/test/resources/org/jboss/resteasy/test/spring/inmodule/web.xml

So the addition of `ResteasySpringDispatcherServlet` is just for the scenario when using Spring/SpringMVC by configuration `DispatcherServlet`, which is described as the standard way to do the configuration in document.

For more detail discussions please check here:

* https://issues.jboss.org/browse/RESTEASY-828

For the document of the `ResteasySpringDispatcherServlet`, it will be tracked by this issue: 

* https://issues.jboss.org/browse/RESTEASY-2392

### DIFF
--- a/resteasy-spring/src/main/java/org/jboss/resteasy/springmvc/ResteasySpringDispatcherServlet.java
+++ b/resteasy-spring/src/main/java/org/jboss/resteasy/springmvc/ResteasySpringDispatcherServlet.java
@@ -1,0 +1,23 @@
+package org.jboss.resteasy.springmvc;
+
+import org.jboss.resteasy.core.ResteasyContext;
+import org.springframework.web.servlet.DispatcherServlet;
+
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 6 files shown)


liweinan:
@asoldano @ronsigal Could you please help to check this? Thanks!

PR #2211, opened 2019-11-08
Fix NPE when ConfigSource is not available

gastaldi:
`source` may be null, causing a NPE in the getName() method.


ronsigal:
Hi @gastaldi,

1. Yeah, the lack of an else clause looks like a bug.

2. In the original code, it looks like the volatile keyword and double check were because two threads could call getName() at the same time. They should get the same result, so it looks like it wouldn't have really mattered, though. But putting it in the constructor should eliminate that problem, so volatile isn't necessary. Making name final seems like just a little best practices thing.

@gastaldi, am I seeing this right? If so, the changes look fine to me.

-Ron

gastaldi:
@ronsigal yes, that's right

ronsigal:
Ok, thanks, @gastaldi.

PR #2216, opened 2019-11-13
MP Rest Client allow async interceptor invocation

michalszynkiewicz:
fixes https://github.com/quarkusio/quarkus/issues/4774


gsmet:
@asoldano could you have a look at that one? It would be nice if we could include it in the Final of Quarkus if it's legit.

Thanks!

asoldano:
> @asoldano could you have a look at that one? It would be nice if we could include it in the Final of Quarkus if it's legit.
> 
> Thanks!

Yep, planning to do that.

michalszynkiewicz:
@asoldano it shouldn't be difficult to add a test for it, I will try to add it when I have a while. Thank you for merging!

PR #2217, opened 2019-11-15
[RESTEASY-2410] added check of ResteasyProviderFactory instance in Bu…

rsearls:
…iltResponse

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/AbstractBuiltResponse.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/AbstractBuiltResponse.java
@@ -313,6 +313,9 @@ public abstract class AbstractBuiltResponse extends Response
    public void close()
    {
       isClosed = true;
+      if (ResteasyProviderFactory.peekInstance() != null) {
+         ResteasyProviderFactory.clearInstanceIfEqual(ResteasyProviderFactory.peekInstance());
+      }
    }
 
    @Override


asoldano:
@rsearls , I'm trying looking again at this. I was thinking that, instead of trying cleaning up the default ResteasyProviderFactory instance in the response as you're proposing, we could see if we can completely avoid setting up a new "default" instance if it's not set already (as, again, I think the singleton instance is a source of problems).
Alternatively, we could deal with this by ensuring the ClientFilterResponseBuilderTest leaves the environment in the same state if sees it before running, iow it does not change the default ResteasyProviderFactory instance.
One way or the other, I would need to reproduce the failure this fix is meant to solve, but I couldn't. Can you clarify how to do that?
Thanks

rsearls:
Running ClientFilterResponseBuilderTest without this fix causes a reference to 
a ResteasyProviderFactory instance (on the clientside) to be held globally and 
used in all tests that runs after it.  Any special code set in that instance of 
ResteasyProviderFactory has the potential to negatively affect any following test.  
This was a discovery made when debugging the failures with 
AllowGzipOnServerAllowGzipOnClientTest and NotAllowGzipOnServerAllowGzipOnClientTest.  
While it was determined that this issue was not a direct cause of those test 
failures, it has the potential to cause havoc in customer code and future tests.
Currently the only way to see this issue is to use the debugger and record the
address of the generated (clientside) ResteasyProviderFactory instance created in 
ClientFilterResponseBuilderTest and check for its reference in the following
test(s).

rsearls:
It is so old as to be obsolete

PR #2223, opened 2019-11-20
[RESTEASY-2411] resteasy-jackson2-provider: default white-list based PolymorphicTypeValidator extension

istudens:
Issue: https://issues.jboss.org/browse/RESTEASY-2411

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/whitelist/WhiteListPolymorphicTypeValidatorTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/whitelist/WhiteListPolymorphicTypeValidatorTest.java
@@ -1,0 +1,150 @@
+package org.jboss.resteasy.test.providers.jackson2.whitelist;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.logging.Logger;
+import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.plugins.providers.jackson.WhiteListPolymorphicTypeValidatorBuilder;
+import org.jboss.resteasy.spi.HttpResponseCodes;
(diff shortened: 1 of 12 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/whitelist/WhiteListPolymorphicTypeValidatorTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+    static ResteasyClient client;
+
+    @Deployment(name = "default")
+    public static Archive<?> deploy() {
+        WebArchive war = TestUtil.prepareArchive(WhiteListPolymorphicTypeValidatorTest.class.getSimpleName());
+        war.addClass(WhiteListPolymorphicTypeValidatorTest.class);
+        Map<String, String> contextParam = new HashMap<>();
+        contextParam.put("jackson.deserialization.whitelist.allowIfBaseType.prefix", Automobile.class.getPackage().getName());
+        contextParam.put("jackson.deserialization.whitelist.allowIfSubType.prefix", Automobile.class.getPackage().getName());
```
asoldano: Actually I believe that in this case the latter context param would be enough.

istudens: Ok, I will add another test. Thanks for a review.

istudens: I've tried to add a test enabled by a system property, it however broke the current `testGood` method. It seems the system property settings override context params, i.e. MP Config. If so, it would need a separate surefire execution to test both. See
https://github.com/resteasy/Resteasy/commit/742cb04ad91956e849b4307dede30b584f3df326

asoldano:
I've merged this (plus fixed the name of the props in the test, which is what was causing the failures on CI). We can provide a test that covers the sysprop configuration too later.

istudens:
@asoldano Would not be enough to have a generic test for system property vs. context param?

asoldano:
Not sure I get exactly what you mean, but I'm open to proposals

PR #2232, opened 2019-11-27
[RESTEASY-2433] Ensure that Providers registered with @RegisterProvid…

geoand:
…er use the InjectorFactory.

Forms part of the solution for: https://github.com/quarkusio/quarkus/issues/5752

### DIFF
--- a/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -450,10 +450,14 @@ public class RestClientBuilderImpl implements RestClientBuilder {
     }
 
     private static Object newInstanceOf(Class<?> clazz) {
-        try {
-            return clazz.newInstance();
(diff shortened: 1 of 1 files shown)


geoand:
@asoldano I'm not really sure how to add a test for this one.

Also I could have made the change to the specific method that registers providers only, but what I actually did seems conceptually more correct to me, but I might be wrong of course.

marcinczeczko:
@geoand - are you going to address also lack of support for injection in ClientHeadersFactory (that can be registered via `@RegisterClientHeaders`) ? I found out that besides my issue quarkusio/quarkus#5752 there is another that asks for client headers too: quarkusio/quarkus/issues/5030

geoand:
@marcinczeczko This PR won't take care of that case unfortunately. But looking at that issue, it's not really certain if we should fix it.
I mean I think we should, but it seems like the spec is unclear.

gsmet:
As for updating RESTEasy for Quarkus 1.1, the idea would be to have the upgrade PR on December 6th latest so that we can include it in the CR1 planned for December 9th.

geoand:
@asoldano ^

geoand:
Thanks @asoldano !

PR #2234, opened 2019-11-27
RESTEASY-2435 finish writing `BasicIntegration` and `CustomizedConfiguration` sections and provide necessary examples

liweinan:


liweinan:
This is part of `RESTEASY-2392 - spring integration document is not complete`, because the task is complex and time consuming. So I divided it into subtasks(finish reviewing it section by section).

PR #2236, opened 2019-12-01
RESTEASY-2439 - update the resteasy github project readme

liweinan:


README.md
```diff
@@ -4,44 +4,41 @@
 
 RESTEasy is a JBoss.org project aimed at providing productivity frameworks for developing client and server RESTful applications and services in Java.  It is mainly a JAX-RS implementation but you'll find some other experimental code in the repository.
 
-The project page can be found at https://resteasy.jboss.org
+The project page can be found at [https://resteasy.github.io/](https://resteasy.github.io/).
```
asoldano: I believe we can leave the resteasy.jboss.org link there (we have a redirect in place).

README.md
```diff
@@ -4,44 +4,41 @@
...
 
 ## JAX-RS
-RESTEasy is a JBoss project that provides various frameworks to help you build RESTful Web Services and RESTful Java applications. It is a portable implementation of the JAX-RS specification. JAX-RS is a JCP specification that provides a Java API for RESTful Web Services over the HTTP protocol.
 
-Read the entire specification [here](https://github.com/jax-rs)
+RESTEasy is a JBoss project that provides various frameworks to help you build RESTful Web Services and RESTful Java applications. It is a portable implementation of the JAX-RS specification. JAX-RS is a JCP specification that provides a Java API for RESTful Web Services over the HTTP protocol.
 
-The RESTEasy code is in the github repository, (https://github.com/resteasy/Resteasy.git).
+Read the entire specification [here](https://github.com/eclipse-ee4j/jaxrs-api).
```
asoldano: I agree this should be improved and link to Jakarta repo provided. In that case, though, you should also mention it in the text above, maybe replacing the reference to JCP or explaining where the spec is today.

README.md
```diff
@@ -4,44 +4,41 @@
...
 
 ## Getting started with RESTEasy
+
 - Read a [book](https://resteasy.github.io/books.html)
 - Check out the [examples](https://github.com/resteasy/resteasy-examples) in the repository.
 - Read the [documentation](https://resteasy.github.io/docs).
+- The new blog is [here](https://github.com/resteasy/resteasy-blog).
```
asoldano: let's keep this change and the new blog changes separate

README.md
```diff
@@ -4,44 +4,41 @@
...
+```
 
 Currently it can be built using JDK 1.8 and 11.
-If you don't have the JBoss Nexus repository configured in your own local Maven settings, you might need to add enable the 'jboss-repository' profile:
-
-       mvn -Pjboss-repository install
-
-
-### Requirements
-- Java Cryptography Extension [JCE](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html)
```
asoldano: Why are you removing these?

README.md
liweinan: yes sir

README.md
liweinan: okay I'll add the notes in the text

README.md
liweinan: okay!

README.md
liweinan: I thought we no longer need these to build, just pure `mvn clean install` can build the project, am I missing anything?

asoldano: The -Pjboss-repository is still needed in some cases (the text says "might need to add", which is true). The JCE, you're right, it's not a strict requirement

liweinan: @asoldano okay!

liweinan: I'll double check it whether it can be removed in my local env.

liweinan:
@asoldano  Could you please help to review this? Thanks!

The JCE extension part is still removed. From my local environment it seems I didn't install it and the tests passed. I used MacOS and use SDKMAN to manage my java installations. However it shouldn't be relative because of these two:

- https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8170157
- https://bugs.openjdk.java.net/browse/JDK-8189377
- https://bugs.openjdk.java.net/browse/JDK-8170157

So in general the openjdk enabled the policy by default.

And from our Travis CI build process we can confirm that it doesn't install extra extensions. Based on above investigation, IMHO we could safely remove that extension installation section.

If anything more need to be modified in this PR please let me know.

PR #2240, opened 2019-12-10
[RESTEASY-2437] Avoid implicit creation of ResteasyProviderFactory in…

asoldano:
…stance when possible; prevent ResteasyProviderFactory instance leakage when doing CDI injection and client response filter processing.


asoldano:
I still need to work on this; this can't be merge as is (for example because the code that checks the current default provider factory and replaces it is not doing that atomically). I'm musing that the implementation of RuntimeDelagate should be split from ResteasyProviderFactory, but really haven't had time to really think about this.

jamezp:
Closing this as it seems stale. Please feel free to reopen if I'm wrong :)

PR #2241, opened 2019-12-12
Update project reactor version to Dysprosium-SR2

sebady:
Update project reactor version to Dysprosium-SR2 (3.3.x)
Minor fix to the utests for netty buffer leak detection


asoldano:
@sebady ,thanks! Before merging, we require that all contributions are made under the terms of the ASL
Do you agree to these terms?

sebady:
Yes.

PR #2244, opened 2019-12-16
RESTEASY-2454: Remove finalize from ClientResponse to improve gc

sebady:
To improve gc remove finalize from ClientResponse and re-add finalize
at the specific client implementation
Make finalize addition configurable with the ReactorNettyClientHttpEngine RestEasyClientReponse.


asoldano:
Merged, thanks!

PR #2246, opened 2019-12-17
[RESTEASY-2456]: JVM Heap leak fix.

artzag:


asoldano:
thanks for the PR @artzag ; I've added a comment to jira, would you mind checking and answering (either there or here). Thanks a lot!

PR #2247, opened 2019-12-18
RESTEASY-2457 Don't close the response upon an extraction error..

crankydillo:
Just call `releaseConnection`, which will prevent connection leaks.


asoldano:
Thanks!

PR #2250, opened 2019-12-25
[RESTEASY-2406] Port use of Microprofile Config to 3.x

ronsigal:


ronsigal:
@asoldano, I noticed that ConfigProvider is replaced by ResteasyConfigProvider almost everywhere on master branch except for a couple of places in resteasy-microprofile-config. Was that an oversight or something subtle?

ronsigal:
Replaced by https://github.com/resteasy/Resteasy/pull/2380.

PR #2252, opened 2020-01-06
RESTEASY-2463 cleanup old org.codehaus.jackson dependencies

vaibhavpatel:


vaibhavpatel:
@asoldano Thanks do much for looking at this! addressed your change request

vaibhavpatel:
Yes, i agree

vaibhavpatel:
@asoldano Thanks for looking at this! addressed your change request

asoldano:
merged, thanks!

PR #2254, opened 2020-01-06
[RESTEASY-2466] removed profile arquillian.remote

rsearls:


rsearls:
README file update complete

PR #2259, opened 2020-01-07
[RESTEASY-2453] Deleted references to resteasy.scan, resteasy.scan.providers, and resteasy.scan.resources in User Guide.

ronsigal:


ronsigal:
Strange, the change to License.xml has already been committed.

PR #2263, opened 2020-01-08
RESTEASY-2467 updated smallrye-config dependency from 1.3.9 to 1.5.1

svenhaag:
I agree with the Apache License 2.0 terms.


asoldano:
Thanks @svenhaag !

PR #2268, opened 2020-01-13
[RESTEASY-2486] Fix for RestEasyClientBuilder http proxy not working

csotiriou:
I observed an issue using Quarkus, which uses the RestEasy REST Client, for which I have opened an issue here:

https://github.com/quarkusio/quarkus/issues/6522

It concerns an issue where the http-proxy setup with a `RestClientBuilder` would not work, i.e. the settings are completely ignored in the following setup:

```
 MyRestClient build = RestClientBuilder.newBuilder()
                .baseUri(UriBuilder.fromUri("http://localhost:3001").build())
                .property(PROPERTY_PROXY_HOST, "localhost") //ignored
                .property(PROPERTY_PROXY_PORT, 3002) //ignored
                .property(PROPERTY_PROXY_SCHEME, "http") //ignored
                .build(MyRestClient.class);
```

The change concerns this part of `ResteasyClientBuilderImpl`:

```
 boolean resetProxy = false;
      if (this.defaultProxy == null) {
         resetProxy = true;
         // check for proxy config parameters
         setProxyIfNeeded(config);
      }
      ClientHttpEngine engine = httpEngine != null ? httpEngine : new ClientHttpEngineBuilder43().resteasyClientBuilder(this).build();
      if (resetProxy) {
         this.defaultProxy = null;
      }
```
This small PR checks if a new proxy has been setup, and if it has, it allows the creation of a new http client engine.

The rationale behind this is that during custom initialization using a `RestClientBuilder`, the httpEngine had already been setup previously, so even if the `setProxyIfNeeded()` was called to set the `defaultProxy` inside the builder instance, the next lines would only take into account if an existing instance of the `httpEngine` is present - which, according to my test cases, was always the case. Therefore, the proxy settings would be discarded since the old instance of the `httpEngine` would always be used.

I have tested it using my scenarios with Quarkus, but I don't have the chance to perform all necessary checks, as I am not familiar with the architecture of the rest of the Resteasy implementation. This part of the code just seemed a bit unintentional so I thought I may as well give it a shot, hoping that my assumptions are correct.

### DIFF
--- a/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -204,21 +204,35 @@ public class RestClientBuilderImpl implements RestClientBuilder {
         T actualClient;
         ResteasyClient client;
 
-        String userProxyHost = Optional.ofNullable(getConfiguration().getProperty(PROPERTY_PROXY_HOST)).filter(String.class::isInstance).map(String.class::cast).orElse(null);
-        Integer userProxyPort = Optional.ofNullable(getConfiguration().getProperty(PROPERTY_PROXY_PORT)).filter(Integer.class::isInstance).map(Integer.class::cast).orElse(null);
-        String userProxyScheme = Optional.ofNullable(getConfiguration().getProperty(PROPERTY_PROXY_SCHEME)).filter(String.class::isInstance).map(String.class::cast).orElse(null);
-
-
         ResteasyClientBuilder resteasyClientBuilder;
         if (envProxyHost != null && !noProxyHosts.contains(baseURI.getHost())) {
-            // Use proxy, if defined
+            // Use proxy, if defined in the env variables
             resteasyClientBuilder = builderDelegate.defaultProxy(
                     envProxyHost,
                     Integer.parseInt(System.getProperty("http.proxyPort", "80")));
-        } else if (userProxyHost != null && userProxyPort != null) {
-            resteasyClientBuilder = builderDelegate.defaultProxy(userProxyHost, userProxyPort, userProxyScheme);
         } else {
-            resteasyClientBuilder = builderDelegate;
+            // Search for proxy settings passed in the client builder, if passed and use them if found
+            String userProxyHost = Optional.ofNullable(getConfiguration().getProperty(PROPERTY_PROXY_HOST))
+                    .filter(String.class::isInstance)
+                    .map(String.class::cast)
+                    .orElse(null);
+
+            Integer userProxyPort = Optional.ofNullable(getConfiguration().getProperty(PROPERTY_PROXY_PORT))
+                    .filter(Integer.class::isInstance)
+                    .map(Integer.class::cast)
+                    .orElse(null);
+
+            String userProxyScheme = Optional.ofNullable(getConfiguration().getProperty(PROPERTY_PROXY_SCHEME))
+                    .filter(String.class::isInstance)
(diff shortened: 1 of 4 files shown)


csotiriou:
@asoldano I understand your point.

Well, to help out with the Quarkus implementation, consider this Microprofile spec for building a customized client:

```
MyRestClient build = RestClientBuilder.newBuilder()
                .baseUri(UriBuilder.fromUri("http://localhost:3001").build())
                .property(PROPERTY_PROXY_HOST, "localhost")
                .property(PROPERTY_PROXY_PORT, 3001)
                .property(PROPERTY_PROXY_SCHEME, "http")
                .build(MyRestClient.class);
```

in this case, the `build() method resolves to `RestClientBuilderImpl` in the resteasy-client-microprofile library.

In this method, the engine along with the initial `engine` is constructed, with the help of the builder of type `URLConnectionClientEngineBuilder` when SSL option is **disabled**. If SSL is **enabled**, then the http engine will be null, and it will be created later with type `ClientHttpEngineBuilder43` (inside the `ResteasyClientBuilderImpl` class, in the location of my PR)

Now, it seems to me that the `URLConnectionClientEngineBuilder` does not have any support for passing proxy options. But the `ClientHttpEngineBuilder43` does have extensive support for proxy options and other types of options. The purpose of having two different builders is not known to me, maybe you have more input about this?

Based on the options I see exposed on the microprofile implementation, all setup paths will actually instantiate an engine through `URLConnectionClientEngineBuilder` (only for MicroProfile) and will produce a specific type of HTTP engine (a URLConnectionEngine), which has no exposed proxy support. Am I right in this? Is this the expected behavior of this builder?

csotiriou:
The more I look at it, the clearer I believe it gets. Here's an alternative approach I am imagining:

1) In the `build(Class<T> aClass)` of the `RestClientBuilderImpl`, use the `getConfiguration().getproperty()` to get the proxy specific settings of the builder settings, and set the default proxy (decide first if priority will be given to the system setting)
2) In the `URLConnectionClientEngineBuilder`, in the `build()` method, check the proxy settings of the `restEasyClientBuilder`.
3) In the `URLConnectionEngine`, after having passed the proxy arguments from step 2), inside the `createConnection()` method, call the `openConnection()` with the proxy settings if they are available.

I believe that those are the dependencies, and I hope they are not invasive. If they are correct, I will try to take a stab at it and update this PR (reverting the old changes), if you still haven't got the time to take a look.

I'm trying to setup my environment properly to test a newly created Quarkus project, with a local Resteasy (to test my changes)

csotiriou:
@asoldano I made some new commits according to my previous comment. Feel free to comment regarding whether I'm on the right track or not this time.

To my eyes, the real problem lies in the fact that `URLConnectionEngine` and `URLConnectionClientEngineBuilder` didn't take into account any proxy settings.

I will continue my tests soon, I would appreciate comments regarding the changes so far (if it still breaks any rule, if it's too invasive, if I am messing with the wrong libraries, etc)

csotiriou:
@asoldano sorry for taking so long to answer, is this new commit what you were looking for?

asoldano:
@csotiriou ops, I forgot this would be your first contribution to RESTEasy, so before merging it I actually need an additional comment from you here confirming your agree your contributions to be under the terms of the ASL 2.0 License (http://www.apache.org/licenses/LICENSE-2.0).
Do you agree to these terms?

csotiriou:
I agree! @asoldano 

So, what's the chance of putting this into the 4.4.x version? Quarkus seems to be using 4.4.2, and I believe migrating to 4.5.0 would seem like a big bump.

asoldano:
@csotiriou thanks (also for the last minute bug fix).
The plan is to include this in next release, whatever it will be.

PR #2270, opened 2020-01-14
[RESTEASY-1510] Implement "resteasy.disable.providers" parameter.

ronsigal:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/plugins/providers/RegisterBuiltin.java
```diff
@@ -52,7 +59,7 @@ public static void register(ResteasyProviderFactory factory)
       }
    }
 
-   public static void registerProviders(ResteasyProviderFactory factory) throws Exception
+   public static void registerProviders(ResteasyProviderFactory factory, Set<String> disabledProviders) throws Exception
```
asoldano: minor thing, but this need the old method to be retained too, as you did for the `register(ResteasyProviderFactory factory)` one above (full backward compatibility needs)

ronsigal: Done

PR #2276, opened 2020-01-27
[RESTEASY-2487] Quarkus fast boot and completion stage unwind

patriot1burke:
This is a big PR.  I'll squash the commits after review as I think looking through each commit might make reviewing easier.  Here's a summary of the changes:

* Improve Resteasy boot time by reducing copying within ResteasyProviderFactoryImpl, Client/ServerHelper, and MediaTypeMap.  Previously a ton of copy-on-write structures and patterns were used.  Even with improvements last summer, WebTargets still did a ton of copying of their underlying ResteasyProviderFactories.  

Code was changed so that at boot time copy-on-write is turned off and after boot is complete, a flag is set to turn ON copy-on-write.
* Next big change was for Quarkus on boot.  I added a ton of new fine-grain registration methods to ResteasyProviderFactoryImpl.  This allows Quarkus to do a lot of boot processing at build time.
* Final big change was that the huge CompletionStage stack that is created for each HTTP invocation was unwound.  In the current codebase, if there is no async injection, a huge CompletionStage chain is created that encompasses: resource object construction, property injection, method argument injection, and method invocation.

ValueInjector, ConstructorInjector, MethodInjector, and classes like it have all the appropriate methods return a Object instead of a CompletionStage.  Request processing, at each step, now checks to see if the returned value is a CompletionStage or not.  If its not, it doesn't have to create a CompletionStage chain.


patriot1burke:
> Look alright to me, but I'm curious what perf changes you've measured for the CS unwind changes?
@FroMage 

In my yourkit profiler analysis, there's no more low-hanging-fruit for boot and runtime optimizations.  The CS unwind show up in a small way in YourKit runs (boot time with construction, run-time with request).  There's no big stuff to optimize anymore.  Only little stuff.

For boot time direct runs, it looks like I shaved off 10% from Quarkus boot time for a resteasy+jackson deployment,  but I can't be sure or exact because the timings swing wildly every time I run it.

For request time direct runs, I couldn't gauge improvement in single-threaded runs other than YourKit not flagging completion stage anymore.  I'm hoping with all these little improvements larger more thorough benchmarks will show the improvements.  I saw a few more little things I could do after this PR is merged.

asoldano:
Speaking of runtime performances, for what is worth here I wanted to re-run some tests I ran at the time of https://github.com/resteasy/Resteasy/pull/2107 , but the test environment seems to have some problems atm. @marekkopecky is likely having a look tomorrow to see if it can be restored.

asoldano:
> Speaking of runtime performances, for what is worth here I wanted to re-run some tests I ran at the time of #2107 , but the test environment seems to have some problems atm. @marekkopecky is likely having a look tomorrow to see if it can be restored.

This is going to take a while, we have a ticket opened to deal with network issues in the lab.

patriot1burke:
@asoldano @ronsigal Ok to merge this?

asoldano:
@patriot1burke merged, thanks!

PR #2277, opened 2020-01-27
[RESTEASY-2481] Support MicroProfile REST Client 1.4.

ronsigal:


ronsigal:
When MicroProfile REST Client goes 1.4-RC1 -> 1.4.Final, I'll update this pull request.

ronsigal:
re: "RC level of the dependency"

Exactly. I gather they're going to do the final soon, but if you want to merge, that's fine.

asoldano:
OK, I've merged this. I think we can have the final upgrade in another simple PR (and perhaps also the @Context injection, if we want). In the mean time, Bill can include your changes too when rebasing his monster PR ... ;-)

PR #2292, opened 2020-02-11
[RESTEASY-2495] ParamConverter not being applied to HeaderParam

Dawuid:
AbstractInvocationCollectionProcessor is processing Array and Collection of elements being ParamConverter aware. But with single values is delegating in final class to process the value. HeaderParamProcessor is not being aware of such transformation.

See https://issues.redhat.com/browse/RESTEASY-2495


Dawuid:
In case of patch being accepted, I agree on the Apache License 2.0 terms.

rsearls:
I've created an new PR which uses this code fix and a testcase.
See PR https://github.com/resteasy/Resteasy/pull/2366
This PR can be withdrawn.

asoldano:
OK, I've merged both PR, so that the fix is recorded as a change from @Dawuid  and the test from @rsearls . Thanks.

PR #2298, opened 2020-02-17
[RESTEASY-2280]:Refactor the execption and log message; Improve the headers change and restore;[RESTEASY-2496]:Add doc

jimma:


jimma:
It fixed both. And the PR message isn't complete,  but the commit message is correct.  I already fixed the PR message.
For the format change, it's not really huge change and It would be better to fix it to align Resteasy's code format and style.
About the test , do you mean we add test for RESTEASY-2496 ?  I'll add one later.

asoldano:
> About the test , do you mean we add test for RESTEASY-2496 ? I'll add one later.

Yes, thanks

PR #2303, opened 2020-02-20
RESTEASY-2502 - revisit Spring Boot starter doc section to keep it up…

liweinan:
…date to current situation


liweinan:
Here is the generated section:

image

PR #2304, opened 2020-02-21
[RESTEASY-2496]:Add test case

jimma:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentPatchTest.java
```diff
@@ -110,5 +126,29 @@ public void testMergePatchStudent() throws Exception {
       Assert.assertEquals("Expected gender is null", null, patchedStudent.getGender());
       client.close();
    }
+   @Test
+   @OperateOnDeployment(DISABLED_PATCH_DEPLOYMENT)
+   public void testPatchDisabled() throws Exception {
+      ResteasyClient client = ((ResteasyClientBuilder)ClientBuilder.newBuilder()).connectionPoolSize(10).build();
```
asoldano: I don't see this client instance being closed before it goes out of scope... can you add at least a client.close() before the end of the test method?

PR #2305, opened 2020-02-21
Fixed TCK test, added passing version of SSE TCK test

FroMage:
Re-resolve the writer after all interceptors are done, making sure we don't get a blocking
one after we started with an async one


server-adapters/resteasy-vertx/src/test/java/org/jboss/resteasy/test/asyncio/actual.txt
```diff
@@ -0,0 +1,45 @@
+0: org.jboss.resteasy.test.cdi.interceptors.resource.InterceptorRequestFilterInterceptor
```
asoldano: Is this actual.txt file used? Same question for the expected.txt

FroMage: OOOPS.

FroMage: force-pushed fixed.

FroMage:
Updated with second fix for TCK as discussed by email.

PR #2312, opened 2020-02-28
Fix Servlet async race

stuartwdouglas:
complete() could be called while there is still data queued.


FroMage:
Closing as not needed anymore.

PR #2313, opened 2020-02-28
[RESTEASY-2512] workaround to handle Container filters

rsearls:


ronsigal:
Hi @rsearls,

So, the NOOPBuiltResponse actually shows up on the client side, right? And then, in the finally block, you still pass the unhandled exception up to the container. Does it also show up on the client? 

What would happen if you wrapped the unhandled exception in the NOOPBuiltResponse and just write that to the client? 

-Ron

rsearls:
NOOPBuiltResponse is strictly a serverside object.  The client should
know nothing about it.  It is a object to fool the container filter
into running in the unlikely scenario there is an unhandled exception.  
NOOPBuiltResponse should NOT be returned to the client.  The normal way 
resteasy responds should be maintained.

ronsigal:
re: "NOOPBuiltResponse is strictly a serverside object."

Ah, yes, I see.

So, the filter chain runs and then the Exception gets thrown, as before. I understand now.

For the record, I've proposed that there always be an ExceptionMapper: https://github.com/eclipse-ee4j/jaxrs-api/issues/855

ronsigal:
1. I think there should be a test.

2. Since this is a behavior change, maybe a note somewhere in the User Guide.

rsearls:
This fix is temporary and a specific smallrye-metrics use case.
Ron has filed a Jaxrs issue with eclipse https://github.com/eclipse-ee4j/jaxrs-api/issues/855
to better address this use case.  We expect this to be accepted in the near 
future at which time this change can be removed.  I don't think this should be 
documented or a test case created because as it is temporary and limited to
a smallrye-metrics situation.

ronsigal:
@rsearls, smallrye metrics went ahead and used the two interceptor trick, didn't they? So maybe this fix isn't even necessary. Does that make sense?

rsearls:
Sure we can skip this then.

ronsigal:
Ok, great. Thanks, @rsearls, I'll close it, but thanks for working on it.

PR #2311, opened 2020-02-25
[RESTEASY-2510] Split RESTUtils to links provider and injector

gytis:
https://issues.redhat.com/browse/RESTEASY-2510

I'm building a Quarkus extension in which I would like to make a use of `resteasy-links` module. In this PR I've introduced a few changes that would make it easier to use this module in the said extension:

1. Added `entityClassName` field to the `@LinkResource` annotation. This will allow to generate JAX-RS resources with `@LinkResource` annotation during the build time in Quarkus.
2. Split `RESTUtils` class into two classes: `ObjectLinksProvider` and `LinksInjector`. This will allow to get links relevant to an object without injecting them into its field.
3. Add `ClassLinksProvider`. This will allow to get links relevant to a class (`list` and `add` operations that work with a given class)

Most of what I've done is reorganise the code. The procedure of finding links is unchanged.

PR #2314, opened 2020-02-29
RESTEASY-2445 getlink(rel) does not work with things like github api pagination due to not splitting headers correctly

liweinan:


liweinan:
This should fix the problem:

image

Reproducer:

https://github.com/liweinan/ghlint

liweinan:
Verified this PR is compatible with https://github.com/resteasy/Resteasy/pull/2311

image

PR #2318, opened 2020-03-06
CDI BeanManager fix for Quarkus 

phillip-kruger:
This is a fix for an issue in Quarkus (https://github.com/quarkusio/quarkus/pull/7639) that is currently  
temporarily fixed inside the Quarkus Extension (https://github.com/quarkusio/quarkus/pull/7639)
 
Signed-off-by:Phillip Kruger <phillip.kruger@gmail.com>

### DIFF
--- a/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientExtension.java
+++ b/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientExtension.java
@@ -8,7 +8,6 @@ import javax.enterprise.inject.spi.AfterBeanDiscovery;
 import javax.enterprise.inject.spi.AfterDeploymentValidation;
 import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.BeanManager;
-import javax.enterprise.inject.spi.BeforeBeanDiscovery;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.ProcessAnnotatedType;
 import javax.enterprise.inject.spi.WithAnnotations;
@@ -19,6 +18,7 @@ import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import javax.enterprise.inject.spi.CDI;
 
 public class RestClientExtension implements Extension {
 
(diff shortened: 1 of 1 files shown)


phillip-kruger:
@asoldano @stuartwdouglas please have a look (I can not assign reviewers :) )

asoldano:
thanks @phillip-kruger ; I believe @ronsigal is the one who should really review this, as he provided the original implementation. Thanks!

ronsigal:
Hi @phillip-kruger,

I'm re-running the failed tests and also running the testsuite on my laptop.

I have a question and a comment:

1. What are you seeing that suggests moving the acquisition of the BeanManager out of observeBeforeBeanDiscovery()? Was it getting a null BeanManager?

2. There are a few places in RestClientExtension.construct() where the manager variable is assumed to have a non-null value. With your changes, manager will get assigned in isCDIActive(). Now the only place where construct() is called is in ClientHeaderProviders.getCustomHeadersFactory():

```
if (RestClientExtension.isCDIActive()) {
    Object factory = RestClientExtension.construct(factoryClass);
```
so it's safe, at least for now. But what if the code changes in the future? Maybe there should be a test at the beginning of construct() that verifies that manager != null or acquires a value.

-Ron

phillip-kruger:
Hi @ronsigal - thanks for getting back to me. 

1) Yes, 
`public void observeBeforeBeanDiscovery(@Observes BeforeBeanDiscovery event, BeanManager beanManager)` does not work in Quarkus. So even if CDI is active, the manager is never set.

2) Agree. This was an issue even before my change. I did not want to do a null check, as some other code might rely on a NullPointerException, but if you agree that we should then I am happy to add that. Let me know.

ronsigal:
Hey @phillip-kruger,

Yeah, I think it's reasonable to do the null check.

The same test (SSETest in resteasy-vertx) is still failing in the jdk 11 runs. I'm running the testsuite with jdk 11 on my laptop to see what happens. It doesn't seem like your change should affect an SSE test, but I'll have to look into it. If you have any insights, they would be welcome. ;-)

phillip-kruger:
Hi @ronsigal . Ok I am adding a null check to this PR. Thanks

phillip-kruger:
@ronsigal - please check again. Now with CDI checking in the `RestClientExtension.construct()` method

ronsigal:
re: "I am adding a null check to this PR."

Thanks, @phillip-kruger.

It looks like the intermittent test failures come from a recent unrelated change. Still looking.

phillip-kruger:
Thanks @ronsigal - Please let me know if there is anything I can help with.

asoldano:
Thanks @ronsigal for the checks and @phillip-kruger for the discussion. The intermittent failure is unrelated, so I'm merging this.

PR #2320, opened 2020-03-11
RESTEASY-2519

liweinan:


resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -755,11 +758,19 @@ else if (valueOf != null)
...
+         throwProcessingException(Messages.MESSAGES.unableToExtractParameter(getParamSignature(), _encode(strVal), target), e);
       }
       return null;
    }
 
+   private String _encode(String strVal) {
+      try {
+         return URLEncoder.encode(strVal, StandardCharsets.UTF_8.toString());
+      } catch (UnsupportedEncodingException e) {
+         return e.getMessage();
```
gastaldi: Should the message be returned here or the original value?

gastaldi: Perhaps it would be better to `log.debug()` this exception and return the original value

liweinan: @gastaldi the original value is unescaped and still unsafe right?

liweinan: adding a debug log is fine, let me add it in this PR. For return value, I'm not sure what to return for an exception, maybe assume this won't happen?

liweinan:
adding @ronsigal  and @asoldano for suggestions.

resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
gastaldi: > @gastaldi the original value is unescaped and still unsafe right?

Yes. I just mentioned that because `e.getMessage()` will return `UTF-8`

resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -755,11 +758,19 @@ else if (valueOf != null)
...
       {
-         throwProcessingException(Messages.MESSAGES.unableToExtractParameter(getParamSignature(), strVal, target), e);
+         throwProcessingException(Messages.MESSAGES.unableToExtractParameter(getParamSignature(), _encode(strVal), target), e);
       }
       return null;
    }
 
+   private String _encode(String strVal) {
+      try {
+         return URLEncoder.encode(strVal, StandardCharsets.UTF_8.toString());
```
gastaldi:
```suggestion
         return URLEncoder.encode(strVal, StandardCharsets.UTF_8.name());

```

resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
gastaldi: Perhaps in this case the best is to just rethrow the exception as a RuntimeException

ronsigal:
Ok, @liweinan, it's on my TODO list.

liweinan:
@ronsigal Thanks!

liweinan:
@gastaldi Nice suggestion! Let me apply your fix :-)

liweinan:
Verification:

I compiled Quarkus locally with this patch:

image

And run the reproducer project with the patched Quarkus:

image

And I can see the exception message is encoded:

image

liweinan:
It seems this PR works. @gastaldi @ronsigal Could you please help to review it? Thanks!

liweinan:
@gastaldi Thanks! I'll wait for @ronsigal and @asoldano input and then take future action.

gastaldi:
I haven't tested, but I wonder if this bug also happens with other `@*Param` annotations?

liweinan:
@gastaldi IMHO potentially all the places that throws out Exception messages directly from user input should be encoded. Maybe we can defer this to further discussion.

liweinan:
Some tests keep failing:

170615A6-527B-48CF-BA86-00B4FD01DB6A

I filtered the test error message:

39CF1C1D-603D-461F-8175-000FCD31DCA0


And the failed tests seem irrelevant:

D2E2BD2E-889A-45A8-8684-8F25BAD1A417

I rerun several times but failed to pass them, I'll go on trying it today.

ronsigal:
Hi @liweinan,

So, if I understand correctly, a string with a script embedded in it is returned to an HTML client, which then executes the script, right? And if the "<" and ">" are encoded then it doesn't look like a script, right? Then this pull request seems fine to me. 

I have no idea how this relates to quarkus, but that doesn't seem to matter, does it?

-Ron

liweinan:
@ronsigal yes your understanding is correct :D

Thanks for checking!

@asoldano Can we merge this?

PR #2321, opened 2020-03-11
SimpleViolationsContainer: adapt to hibernate-validator API changes

tadamski:
This change is needed as a part of
https://issues.redhat.com/browse/WFLY-11566 and requires hibernate-validator update


asoldano:
@tadamski I assume this should be considered WIP till you can include the HV upgrade that allow compiling and passing the testsuite, right?

tadamski:
@asoldano yes, that's correct

tadamski:
As hibernate-validator was released without those API changes this fix won't be currently necessary. Closing for now.

PR #2322, opened 2020-03-14
[RESTEASY-2503] Throw ViolationException for EJBs with just parameter

ronsigal:
violations.


asoldano:
@ronsigal , this is probably one of the cases where having a better integration with jandex would help. Let's keep this in mind...

ronsigal:
@asoldano, Hmmm, jandex. I should look into that ...

PR #2342, opened 2020-03-31
[RESTEASY-2526] refactor code to avoid recursive write method calls

rsearls:


resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
```diff
@@ -155,24 +155,15 @@ public static void writeTo(final InputStream in, final OutputStream out) throws
...
-      try
-      {
-         if ((read = in.read(buf)) != -1)
-         {
-            return out.asyncWrite(buf, 0, read)
-                  .thenCompose(v -> writeTo(in, out, buf));
+      try {
+         while ((read = in.read(buf)) != -1) {
+            out.asyncWrite(buf, 0, read);
+            buf = new byte[2048];
```
asoldano, line 164: Maybe I'm wrong here, but isn't this keeping on allocating new byte arrays? Can't a single one be re-used?

jamezp: +1 we shouldn't need to reinitialize the array.

rsearls:
What if the array was filled by the previous read but the current read is only for 10 bytes say.
Don't the previous bytes that weren't overwritten remain?

jamezp:
> What if the array was filled by the previous read but the current read is only for 10 bytes say.
> Don't the previous bytes that weren't overwritten remain?

They do, but we're not writing them as we use the `read` length so we're only writing the first 10 bytes.

rsearls:
Change made

asoldano:
This is related to https://github.com/resteasy/Resteasy/pull/2170 , so let me pull @FroMage in the discussion, as he might have an opinion on the proposed fix and the impact to Quarkus.

asoldano:
@rsearls see comment above from @FroMage . Any alternative approach? /cc @jamezp

jamezp:
@rsearls @asoldano I think all we need to do is replace the `thenCompose()` with `thenAccept()`. That would at least avoid the stack overflow. I've not analyzed the full call stack though to ensure that's appropriate here.

jamezp:
I think this should work https://github.com/resteasy/Resteasy/compare/4.5...jamezp:RESTEASY-2526.

FroMage:
I don't think so, it should really be `thenCompose` because it's a `flatMap`, where `thenAccept` is a `map` and will not chain the stages.

Now that I read your comment you mention a stack overflow, but this issue only mentions "recursive write method call" which doesn't imply a stack overflow at all.

What error precisely are you observing?

FroMage:
At first glance I can't see anything wrong with the conversion from loop to recursive call. If you have a test to reproduce a stack overflow here, could you print the params to `writeTo(FileInputStream fis, long length, AsyncOutputStream entityStream, byte[] buf)` to see how we could possibly recurse indefinitely?

Even in the imperative version `length` decreases at every iteration. At the very least, `read` should end up being `-1` at some point. Or we're talking about a huge file whose chunks of 2k exceed the number of allowed stack calls?

jamezp:
@FroMage Here's part of the output from the example reproducer on the JIRA:
```
java.lang.StackOverflowError
	at org.eclipse.jetty.util.log.Slf4jLog.isDebugEnabled(Slf4jLog.java:120)
	at org.eclipse.jetty.server.HttpOutput.write(HttpOutput.java:738)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$DeferredOutputStream.write(HttpServletResponseWrapper.java:153)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$WriteOperation.doWork(HttpServletResponseWrapper.java:69)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$AsyncOperation.work(HttpServletResponseWrapper.java:41)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$DeferredOutputStream.queue(HttpServletResponseWrapper.java:217)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$DeferredOutputStream.asyncWrite(HttpServletResponseWrapper.java:180)
	at org.jboss.resteasy.util.CommitHeaderAsyncOutputStream.asyncWrite(CommitHeaderAsyncOutputStream.java:94)
	at org.jboss.resteasy.plugins.providers.ProviderHelper.writeTo(ProviderHelper.java:171)
	at org.jboss.resteasy.plugins.providers.ProviderHelper.lambda$writeTo$1(ProviderHelper.java:172)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1106)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2235)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:143)
	at org.jboss.resteasy.plugins.providers.ProviderHelper.writeTo(ProviderHelper.java:172)
	at org.jboss.resteasy.plugins.providers.ProviderHelper.lambda$writeTo$1(ProviderHelper.java:172)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1106)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2235)
	at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:143)
	at org.jboss.resteasy.plugins.providers.ProviderHelper.writeTo(ProviderHelper.java:172)
	at org.jboss.resteasy.plugins.providers.ProviderHelper.lambda$writeTo$1(ProviderHelper.java:172)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1106)
```

Using `thenCompose()` processes the entire input stream before returning the map AIUI which is the cause of the stack overflow.

FroMage:
Do you have a link to the JIRA?

jamezp:
@FroMage Yes the latter

> At first glance I can't see anything wrong with the conversion from loop to recursive call. If you have a test to reproduce a stack overflow here, could you print the params to `writeTo(FileInputStream fis, long length, AsyncOutputStream entityStream, byte[] buf)` to see how we could possibly recurse indefinitely?
> 

I could probably write a test case for it and we should probably have one to be safe. There is a reproducer attached to the JIRA https://issues.redhat.com/browse/RESTEASY-2526.

> Even in the imperative version `length` decreases at every iteration. At the very least, `read` should end up being `-1` at some point. Or we're talking about a huge file whose chunks of 2k exceed the number of allowed stack calls?

Yes this seems to be the issue. It's large chunks which is definitely possible that seems to be the issue.

FroMage:
OK, this is a pretty annoying issue because we don't know how to turn imperative code to reactive without using recursion. In this case it appears that the output IO is never blocking so we never get back down to the IO thread and cut the size of the stack, so everything is called on the same stack.

I have to see if this is specific to jetty or if it also happens on WF, Undertow, Vertx and Quarkus. This could be impactful.

FroMage:
OK, first good news @cescoffier I could not reproduce in Quarkus. So this probably only happens with blocking IO. I'm guessing Jetty doesn't support servlet blocking IO right?

We've found a library that could help resolve this issue. I just need to be able to reproduce this bug in the RESTEasy test suite.

FroMage:
Actually, it does happen in Quarkus when using undertow. This is because for servlets, I only use async IO if the request is async. Well, at least I can reproduce. Let's try the fix.

cescoffier:
Thanks for the feedback @FroMage! So yes, let's try to see if a trampoline would workaround the deep stack. I believe it will, but there is no better way to prove it than implementing it. 

Then, the question would be whether this should be always used, or only on blocking IO.

FroMage:
It does fix it. I'll try to catch the other such loops and rewrite them and make a PR.

FroMage:
Running tests locally ATM.

FroMage:
I just pushed #2368 which has the fix, the tests all pass locally, let's see what CI has to say.

PR #2343, opened 2020-04-01
[RESTEASY-2542] Excluding tests

asoldano:


asoldano:
@FroMage FYI I'm going to exclude these tests as they now fail quite frequently, basically making the CI results kind of useless... perhaps in the future we might want to try again investigating/addressing the problem...

FroMage:
Nobody has been able to reproduce the failure outside of CI? Or on a CI branch with no modification?

asoldano:
@FroMage I've seen the failure (in resteasy-vertx) once running locally last week on my laptop (but it was really just one case), it was under heavy load at that time (videoconferencing on, plus other testsuite running). As for the CI environment, you can see the frequency has increased by looking at the latest results at https://github.com/resteasy/Resteasy/commits/master . I can create a new "feature-less" PR and see if that can be used for reproducing and playing with this issue...

FroMage:
If you get me a feature-less PR that reproduces it in CI I can toy with it, but last time I tried it passed all the time :(

asoldano:
> If you get me a feature-less PR that reproduces it in CI I can toy with it, but last time I tried it passed all the time :(

sorry, I meant to do this early but remembered just today; anyway, here is it: https://github.com/resteasy/Resteasy/pull/2345 (failed at the first try)
Thanks for any effort

PR #2344, opened 2020-04-03
[RESTEASY-2544] pom updates to arquillian-util and integration-tests

rsearls:


testsuite/integration-tests/pom.xml
```diff
@@ -623,6 +623,50 @@
...
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-install-plugin</artifactId>
+                <version>3.0.0-M1</version>
```
asoldano, line 651: Is specifying the version here really needed ? IOW, does the maven-install-plugin version controlled by current jboss-parent work as well? In that case, I think it would be better not to override the version here.

rsearls:
Change complete

PR #2347, opened 2020-04-09
[DO NOT MERGE] Sse test failure

FroMage:


FroMage:
So this bug is that we're calling `SSEOutputStream.close()`  whose docs say "we close and if it's already closed then it's fine, you can keep closing it forever you will not get an exception", which is the right thing to do IMO.

But the implementation calls `ResteasyAsynchronousResponse.complete()`, which then forwards to a `vertxFlush()` which throws because it's already closed (by the client).

Now, perhaps the right thing to do is to ignore any `IOException` in `SSEOutputStream.close()`  but I do wonder what the behaviour of `ResteasyAsynchronousResponse.complete()` should be. Should it consider that `complete()` should throw if already closed or not?

rsearls:
Just an added note for the others.  The code change of interest is in file
resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java

This proposed fix is to address https://issues.redhat.com/browse/RESTEASY-2542 SSE tests intermittent 
      failures on Travis CI

FroMage:
Hum, why do I not see any CI running?

asoldano:
@FroMage 

> Hum, why do I not see any CI running?

I think that's because the PR was marked as draft. Now, is this still needed, given the discussion going on in #2348  ?

FroMage:
Indeed not.

PR #2348, opened 2020-04-09
[RESTEASY-2542] ignore any IOException in SSEOutputStream.close()

rsearls:
This PR contains the change made in resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
by FroMage in a special test branch , see PR
https://github.com/resteasy/Resteasy/pull/2347

... In the PR 2347 FroMage wrote ....
So this bug is that we're calling SSEOutputStream.close() whose docs say "we close and if it's already closed then it's fine, you can keep closing it forever you will not get an exception", which is the right thing to do IMO.

But the implementation calls ResteasyAsynchronousResponse.complete(), which then forwards to a vertxFlush() which throws because it's already closed (by the client).

Now, perhaps the right thing to do is to ignore any IOException in SSEOutputStream.close() but I do wonder what the behaviour of ResteasyAsynchronousResponse.complete() should be. Should it consider that complete() should throw if already closed or not?

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
@@ -20,6 +20,7 @@ import javax.ws.rs.ext.MessageBodyWriter;
 import javax.ws.rs.sse.OutboundSseEvent;
 import javax.ws.rs.sse.SseEventSink;
 
+import org.jboss.logging.Logger;
 import org.jboss.resteasy.annotations.SseElementType;
 import org.jboss.resteasy.annotations.Stream;
 import org.jboss.resteasy.core.ResourceMethodInvoker;
@@ -40,6 +41,7 @@ import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements SseEventSink
 {
+   private static final Logger LOG = Logger.getLogger(SseEventOutputImpl.class);
    private final MessageBodyWriter<OutboundSseEvent> writer;
 
    private final ResteasyAsynchronousContext asyncContext;
@@ -107,7 +109,19 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
             ResteasyAsynchronousResponse asyncResponse = asyncContext.getAsyncResponse();
             if (asyncResponse != null)
             {
-               asyncResponse.complete();
+               try {
+                  asyncResponse.complete();
+               } catch(RuntimeException x) {
+                  Throwable cause = x;
+                  while(cause.getCause() != null)
+                     cause = cause.getCause();
+                  if(cause instanceof IOException) {
+                     // ignore it, we're closed now
+                  }else {
+                     LOG.debug(cause.getMessage());
+                     return;
+                  }
+               }
             }
          }
          clearContextData();


ronsigal:
Hi @rsearls and @FroMage,

It seems to me that an IOException could indicate one of two things:

1. What's happening here:
```
    synchronized (conn) {
      if (written) {
        throw new IllegalStateException("Response has already been written");
      }
```
This is inconsequential and should be ignored, I think.

2. A real error. For example, what if an attempt to flush fails and throws an IOException? This is a substantive problem, and, I think, should lead to a 500 status.

It seems to me that SseEventOutputImpl. isn't really in a position to distinguish these cases. On the other hand, org.jboss.resteasy.plugins.server.vertx.VertxHttpResponse.finish()
```
   public void finish() throws IOException
   {
      if (ended) return;
      ended = true;
      checkException();
      if (os != null) {
         os.flush();
         if (!isCommitted())
         {
            prepareChunkStream();
         }
      } else {
         prepareEmptyResponse();
      }
      response.end();
   }
```
is in a position to detect the difference between an IOException thrown by checkException() and one thrown by response.end(). 

So I would suggest that the problem gets handled down there. Similarly, I would look at NettyHttpAsyncResponse, etc., for similar situations.

-Ron

FroMage:
Well `finish` is not in a position to know if we want to flush stuff (important) or if we wanted to just close the connection (and it's already been closed, we've nothing to flush). So that's still tricky.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
```diff
@@ -107,7 +107,19 @@ protected void close(boolean flushBeforeClose)
             ResteasyAsynchronousResponse asyncResponse = asyncContext.getAsyncResponse();
             if (asyncResponse != null)
             {
-               asyncResponse.complete();
+               try {
+                  asyncResponse.complete();
+               } catch(RuntimeException x) {
+                  Throwable cause = x;
+                  while(cause.getCause() != null)
```
asoldano, line 114: cause.getCause() != null && cause.getCause() != cause

asoldano:
So we have no way to distinguish between the IOException that would be thrown for the two cases (response already written and real IO issue)?

Maybe stupid question, what's the scenario that leads to the first case intermittently? Is that something that should be prevented and maybe the SseEventOutputImpl (or other sse related impl classes) is missing some locking / synchronization?

jimma:
@rsearls  Before asyncResponse.complete(), there is code to check IOExecption  when flush the response, will it be safe we ignore exceptions for asyncResponse.complete() ?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
FroMage: That really happens?

FroMage:
> So we have no way to distinguish between the IOException that would be thrown for the two cases (response already written and real IO issue)?

We could, by adding an exception subtype.

> Maybe stupid question, what's the scenario that leads to the first case intermittently? Is that something that should be prevented and maybe the SseEventOutputImpl (or other sse related impl classes) is missing some locking / synchronization?

The client closes the connection. In some cases the server has time to notice, in order cases not. That's why it's intermittent.

Now, IMO it's almost always a bad idea to throw an IOException in `close` because it makes it impossible to close (and potentially free up associated resources) when things go bad. When things go bad you want to close and you don't care at all if something had to be flushed and could not.

The intent of `close` can be understood to be several things, but there are valid use-cases where for example the user attempted to flush and got an IOException and then wants to `close`. Or another more important exception forced the user to `close` and at this point it really doesn't matter that something cannot be `flushed`, or that the client already disconnected: we absolutely need to close and free up resources.

If something needs to be flushed and could not, I think it's the `flush` operation that should throw. If the user cares he should always `flush` and treat _that_ exception. The fact that `close` also calls `flush` is a nicety we do for users who forget to flush, but IMO it should never throw at this point.

ronsigal:
re: "The fact that close also calls flush is a nicety we do for users who forget to flush, but IMO it should never throw at this point."

In this case the user is SSE in RESTEasy, and, if I'm not mistaken, it looks like SSE flushes after each message. @jimma , that's what you were saying above, right? If so, then the question of a failed flush() in close() is moot. 

As for other Exceptions that could bubble up to RESTEasy, it seems to me that that's kind of a moving target, since they could come from Vertx all the way down to the operating system. I think it makes sense to throw them for now. If, at some point, we see one that should be suppressed, we can always do that.

jimma:
> 
> In this case the user is SSE in RESTEasy, and, if I'm not mistaken, it looks like SSE flushes after each message. @jimma , that's what you were saying above, right? If so, then the question of a failed flush() in close() is moot.

@ronsigal  yes. Ron. We did it twice when write Sse Event and close to make sure everything is flushed out before close. I meant if there is error in close(), it can be detected in flush first before asyncResponse.complete() ? 

```
         if(flushBeforeClose && responseFlushed) {
            ResteasyContext.pushContextDataMap(contextDataMap);
            try {
               // make sure we flush to await for any queued data being sent
               AsyncOutputStream aos = response.getAsyncOutputStream();
               aos.asyncFlush().toCompletableFuture().get();
            }catch(IOException | InterruptedException | ExecutionException x) {
               // ignore it and let's just close
            }finally {
               ResteasyContext.removeContextDataLevel();
            }
         }
```

> As for other Exceptions that could bubble up to RESTEasy, it seems to me that that's kind of a moving target, since they could come from Vertx all the way down to the operating system. I think it makes sense to throw them for now. If, at some point, we see one that should be suppressed, we can always do that.

For other UncheckedException thrown, it will stop closing all  subscribed SseEventOutput in SseBroadcaster : https://github.com/resteasy/Resteasy/blob/master/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseBroadcasterImpl.java#L64

New version jaxrs api will add something to control this cascading effect, but we have to provide one method to implement: https://github.com/eclipse-ee4j/jaxrs-api/pull/603/files

It seems we should change SseEventOutput.close() api to allow throws IOException like other java OutputStream.close() does . And this wil pop up all IOException from underlying response and can be decided how to handle it ?
```
public void close() throws IOException
```

ronsigal:
re: "I meant if there is error in close(), it can be detected in flush first before asyncResponse.complete() "

Yes, @jimma, good, that's what I thought. So I think that justifies ignoring any IOException in SseEventOutputImpl.close().

re: "For other UncheckedException thrown, it will stop closing all subscribed SseEventOutput in SseBroadcaster"

So, are you saying that the following could throw a RuntimeException before getting to notifyOnCloseListeners():

```
         outputQueue.forEach(eventSink -> {
            eventSink.close();
            notifyOnCloseListeners(eventSink);
         });
```
Maybe notifyOnCloseListeners() should be in finally block?

re: "It seems we should change SseEventOutput.close() api to allow throws IOException like other java OutputStream.close() does"

Well, in this particular case, the IOException coming back from Vertx is what's causing trouble, so I think @rsearls is right to ignore that one. 

I think the way to think about this issue is: Does the Exception happen before or after the real work is done. If flush() throws an Exception right after writing an event, then that exception should get thrown. If the flush() associated with close() throws an IOException, then we know that the first flush succeeded, so the second one can be ignored. Do we all agree on that?

If some other exception is thrown in close(), then it's after all the bytes have been written, and I'm starting to lean towards agreeing with @FroMage that it's not something that has to be communicated to the client. Logging might be reasonable, though.

I'm probably confusing everyone by arguing both sides of the issue. ;-)

jimma:
> So, are you saying that the following could throw a RuntimeException before getting to notifyOnCloseListeners():

         outputQueue.forEach(eventSink -> {
            eventSink.close();
            notifyOnCloseListeners(eventSink);
         });
Maybe notifyOnCloseListeners() should be in finally block?

@ronsigal  I meant if the first subscribed SseEventouput throws unchecked exception, this forEach will stop and other SseEventOutputs won't be closed. 

> I think the way to think about this issue is: Does the Exception happen before or after the real work is done. If flush() throws an Exception right after writing an event, then that exception should get thrown. If the flush() associated with close() throws an IOException, then we know that the first flush succeeded, so the second one can be ignored. Do we all agree on that?

If some other exception is thrown in close(), then it's after all the bytes have been written, and I'm starting to lean towards agreeing with @FroMage that it's not something that has to be communicated to the client. Logging might be reasonable, though.

+1.  RuntimeException thrown breaks SseBroadcaster.close(), we can just log it now.

FroMage:
My main point is really that even if `close` has data to write, if the client has disconnected it should not throw, because otherwise there's no way to recover from the client disconnecting. It's not like the server can reconnect, so the data is lost anyway, but the underlying resources must still be closed.

At the very least, the underlying resources must be closed/deallocated before any throwing. But I still think it doesn't help the user if `close` throws if the client is already disconnected.

Now, perhaps another way to look at it would be that if vertx notifies us that the client closed the socket, we start marking the SSE stream as closed already, rather than schedule an `IOException` for the next action. In that case, we look at the `SseEventSink` docs for `close`:

```java
     /**
     * Close the {@link SseEventSink} instance and release all associated resources.
     * <p>
     * Subsequent calls have no effect and are ignored. Once the {@link SseEventSink} is closed,
     * invoking any method other than this one and {@link #isClosed()} would result in
     * an {@link IllegalStateException} being thrown.
     */
    @Override
    void close();
```

And we fall in the category of "subsequent calls have no effect and are ignored" because it's already closed.

But note that this is purely an implementation detail: detecting the client closed before we called `close` or detecting it during `close` is semantically the same thing.

ronsigal:
re: " I meant if the first subscribed SseEventouput throws unchecked exception, this forEach will stop and other SseEventOutputs won't be closed."

@jimma, so, using twentieth century syntax, lol, maybe it should look like
```
         for (SseEventSink sink : outputQueue) {
            try {
               sink.close();
            } catch (RuntimeException e) {
               log.sinkCloseError(sink);
            } finally {
               notifyOnCloseListeners(sink);
            }
         }
```

ronsigal:
I'll try to summarize:

1. I think we agree that we can ignore an IOException in close(). A failure to flush would be caught earlier. In fact, it looks like SseEventOutputImpl.writeEvent() would have already called close() if it gets an IOException. If it gets a different Exception, it calls
```
SynchronousDispatcher.rethrow(e);
```
2. What about any other Exception thrown during close()? If, indeed, the problem is that the client has closed the connection, then I think it doesn't really matter what we do since nothing is getting back to the client anyway. If there is a different problem, then we can go with @FroMage's suggestion that we swallow the Exception, maybe with a log message.

3. @jimma has raised an issue concerning
```
         outputQueue.forEach(eventSink -> {
            eventSink.close();
            notifyOnCloseListeners(eventSink);
         });
```
It seems like that should be changed so that close() is always called, but that's a different problem which could get a different JIRA.

So, if there are no objections, my summary suggests that the catch clause in
```
                  }else {
                     // is this right?
                     throw x;
                  }
```
could be changed to
``` 
                  }else {
                     log.debug(...);
                     return;
                  }
```
and then we're done. WDYT?

jimma:
FYI, I proposed to allow SseEventSink.close() to throw IOException : https://github.com/eclipse-ee4j/jaxrs-api/issues/859.

FroMage:
@ronsigal your suggestion LGTM.

> FYI, I proposed to allow SseEventSink.close() to throw IOException : eclipse-ee4j/jaxrs-api#859.

I'm not sure how that can help, sorry. I don't see how throwin `IOException` in `close` helps any caller, as I've explained in this thread.

Now, if you really want to fix `SseEventSink.close()` you should consider https://github.com/eclipse-ee4j/jaxrs-api/issues/797 and make it return a `CompletionStage` instead, which is more urgent ;)

jimma:
> @ronsigal your suggestion LGTM.
> 
> > FYI, I proposed to allow SseEventSink.close() to throw IOException : [eclipse-ee4j/jaxrs-api#859](https://github.com/eclipse-ee4j/jaxrs-api/issues/859).
> 
> I'm not sure how that can help, sorry. I don't see how throwin `IOException` in `close` helps any caller, as I've explained in this thread.

Add throw IOException in jaxrs api is not intended to fix/help this. I think if there is IOException , the close() implemenation can decide if this should be logged/swallowed or this is some real unexpected IOException should  notify caller to do more things like the SseBroadcaster.  If Exception is required to thrown from close(), we don't have to wrap in RuntimeException.

ronsigal:
Hey @jimma,

Whether we wrap the IOException or not, we still have to decide what to do with it. I know I've argued both sides of the issue. I'm thinking back to the days when I worked on Remoting, and there was some code where I was checking for some particular network related Exceptions. I think the point was to decide whether to retry an invocation or not, depending on the underlying problem. In any case, I remember that every once in a while, I would learn about a new (to me) condition that would result in an error message I hadn't seen before. That's what I was thinking about when I said, "As for other Exceptions that could bubble up to RESTEasy, it seems to me that that's kind of a moving target". The point is that I feel a **little** uncomfortable swallowing Exceptions when I'm not sure what they might be. But, in this case, given that we're in close() and given that a flush() problem would have already been caught, I think it's reasonable to just log the exception.

-Ron

jimma:
> Hey @jimma,
> 
> Whether we wrap the IOException or not, we still have to decide what to do with it. I know I've argued both sides of the issue. I'm thinking back to the days when I worked on Remoting, and there was some code where I was checking for some particular network related Exceptions. I think the point was to decide whether to retry an invocation or not, depending on the underlying problem. In any case, I remember that every once in a while, I would learn about a new (to me) condition that would result in an error message I hadn't seen before. That's what I was thinking about when I said, "As for other Exceptions that could bubble up to RESTEasy, it seems to me that that's kind of a moving target". The point is that I feel a **little** uncomfortable swallowing Exceptions when I'm not sure what they might be. But, in this case, given that we're in close() and given that a flush() problem would have already been caught, I think it's reasonable to just log the exception.
> 
> -Ron

I agree on just log the exception for our implementation. We think we can create a PR as we all agree on this approach. As for the jaxrs api change/pr, I think this won't conflict or break with this fix in the future.

ronsigal:
Hey @jimma,

re: "we all agree on this approach."

Excellent!! 

1. Mmmm, log at ... WARN? INFO? DEBUG? Any opinions? If an Exception is likely to come from the client side breaking the connection, then I would say DEBUG. Otherwise, maybe INFO. I'm not sure, but I'm guessing that it's possible to check that the connection is broken, based on the reason in the Exception.

2. Jim, what about "I meant if the first subscribed SseEventouput throws unchecked exception, this forEach will stop and other SseEventOutputs won't be closed."? Separate JIRA?

jimma:
> Hey @jimma,
> 
> re: "we all agree on this approach."
> 
> Excellent!!
> 
> 1. Mmmm, log at ... WARN? INFO? DEBUG? Any opinions? If an Exception is likely to come from the client side breaking the connection, then I would say DEBUG. Otherwise, maybe INFO. I'm not sure, but I'm guessing that it's possible to check that the connection is broken, based on the reason in the Exception.

DEBUG looks good as we ignore this exception.

> 2. Jim, what about "I meant if the first subscribed SseEventouput throws unchecked exception, this forEach will stop and other SseEventOutputs won't be closed."? Separate JIRA?

Seperate JIRA is fine. I already created one : https://issues.redhat.com/browse/RESTEASY-2563

ronsigal:
>DEBUG looks good as we ignore this exception.

Ok, cool. @rsearls, if you add the logging I think we can be done here.

rsearls:
Change complete

PR #2351, opened 2020-04-13
[RESTEASY-2543] add regex change

rsearls:


asoldano:
@rsearls , I've merged this, but now that I think more about it I believe a test should be added, perhaps at least what the user reported in the steps to reproduce in jira. Please open up another PR for the test. Thanks

PR #2358, opened 2020-04-20
Set response code to 500 if writing exception fails

stuartwdouglas:
Fix for https://github.com/quarkusio/quarkus/issues/8675

The root cause of the problem was not setting the content type, but it still should not return a 200 response if writing the error fails.


famod:
I am under the impression that even with this fix I wont see a specific message pointing me in the right direction? Meaning: As a user I still won't see the message that is created by `NoMessageBodyWriterFoundFailure`?

asoldano:
@stuartwdouglas this seems to cause a regression (org.jboss.resteasy.test.asyncio.AsyncIOTest failing in resteasy-vertx). Can you please have a look?

PR #2359, opened 2020-04-20
Mp client async interceptors

FroMage:
My proposal for https://github.com/quarkusio/quarkus/issues/8650

This removes the thread-local for `AsyncInvocationInterceptor`, captures context for each invocation, and calls restore/remove for each invocation part.


FroMage:
Note that I haven't ran any mp rest client tests, so @Ladicek should make sure this passes the MP REST Client TCK first.

Ladicek:
I totally don't agree I should make sure this passes MP RestClient TCK :-)

I'll try this with my reproducer for the Quarkus issue, for sure.

FroMage:
> I totally don't agree I should make sure this passes MP RestClient TCK :-)

Well it was worth a try ;)

Who even knows how this sort of craziness is supposed to be tested then? @kenfinnigan ?

Ladicek:
I'm sure there's a Maven module somewhere in the RESTEasy code base that runs the MP RestClient TCK, now that MP RestClient is itself implemented by RESTEasy. So ... well, perhaps it's even part of the CI build! :-)

Ladicek:
I can confirm that this commit indeed fixes the issue I found with Quarkus. (And does so in a much better way than my proposal, which relies on MP ConProp in a somewhat weird way.)

FroMage:
Well that's a good start. Now if we figure out whether it passes the TCK that'd be nice.

kenfinnigan:
MP REST Client TCK is here: https://github.com/resteasy/Resteasy/tree/master/testsuite/microprofile-tck

Pretty sure it's part of regular CI

Ladicek:
It seems the CI build does the entire RESTEasy build, which (I can confirm) does include the MP RestClient TCK. Some of the CI builds have already finished and are green, so it seems good.

FroMage:
If this actually works, it's champaign for everyone. Mandatory procedure for every PR that removes a thread-local.

FroMage:
Similarly, PRs that add a thread-local should have a checkstyle rule that requires sacrifice of a first-born son by the author as prerequisite price to pay.

FroMage:
Note that I'm not suggesting the life of a first-born son equals a few bottles of wine, but entropy being what it is, some effects are harder to reverse than others and we should just avoid adding thread-locals, m'kay?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
```diff
@@ -560,14 +560,18 @@ public ClientResponse invoke()
...
       {
-         return executorSubmit(client.asyncInvocationExecutor(), callback, extractor);
+         return executorSubmit(asyncInvocationExecutor(), callback, extractor);
       }
    }
 
+   public ExecutorService asyncInvocationExecutor() {
+       return client.asyncInvocationExecutor();
+   }
+
```
asoldano, line 570: This is just for convenience here and in ClientInvocationBuilder, isn't it?

asoldano:
Is passing the MP REST Client TCK here enough? (yes, it's part of the build ;-)) IOW, is there a simple way for checking the problem in Quarkus that led to discover the problem here?

Ladicek:
The Quarkus issue contains an automated reproducer.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
FroMage: Well it's used, and this was the previous default. I assume this will be used by the non-MP resteasy client.

FroMage: So this is customised by the MP client.

asoldano: Thanks, just noticed (that's what happens why reviewing from github UI only and not checking out the code...)

asoldano:
@Ladicek thanks for the tests in quarkus.

PR #2362, opened 2020-04-21
Using GitHub Actions for CI

jamezp:
@ronsigal @asoldano What do you guys think about this? It's not testing the two versions of WildFly like Travis currently does, but it does add testing for Windows. We could enable it to test on multiple versions of WildFly too.

One thing I have noticed is on Windows, especially Java 8 and the `resteasy-jdk-http` tests, fail with a "Bind address already in use". I assume this may be some kind of race for the socket actually closing. I can't duplicate it on my Windows VM at all so I'd be tempted to skip it on Windows CI. Thoughts?

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,37 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Java CI with Maven
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    continue-on-error: true
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest ]
+        java: ['1.8', '11']
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/cache@v1
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Build with Maven Java ${{ matrix.java }}
+        run: mvn clean install -fae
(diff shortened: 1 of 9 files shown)


jamezp:
I can't get these tests to fail on my Windows VM. I'm guessing it must be some kind of race though. I'm not sure if it's in Vert.x for Java 11 or the `HttpServer` for Java 8 or if it's just at the OS level and not actually closing the sockets. I'll see what I find on it though.

ronsigal:
Hi @jamezp,

I just have a few random comments.

1. I don't know much about the subject, but  https://knapsackpro.com/ci_comparisons/github-actions/vs/travis-ci doesn't show a big difference between Github Actions and Travis. On the other hand, I think Github is a great gift to humankind, so there's that.

2. Testing in Windows seems like a great idea.
 
3. Something that's come up recently wrt Travis is that @jimma and I have lost the ability to restart jobs. Is there a way to restart a Github Action run? 

4. I notice that the test run output is extremely verbose. Lots of scrolling. Not a big deal, but maybe that can be changed ...

jamezp:
Hi @ronsigal 

> 
> 1. I don't know much about the subject, but  https://knapsackpro.com/ci_comparisons/github-actions/vs/travis-ci doesn't show a big difference between Github Actions and Travis. On the other hand, I think Github is a great gift to humankind, so there's that.
> 2. Testing in Windows seems like a great idea.
> 3. re: "I assume this may be some kind of race for the socket actually closing. I can't duplicate it on my Windows VM at all so I'd be tempted to skip it on Windows CI." Maybe we could wait a little in a loop until the socket closes. If there's still a problem, maybe it's more serious.

I tried a simple sleep and that didn't work. I'm doing some more investigation though.

> 4. Something that's come up recently wrt Travis is that @jimma and I have lost the ability to restart jobs. Is there a way to restart a Github Action run?

It looks like you can restart all the jobs, but not a single job.

> 5. I notice that the test run output is extremely verbose. Lots of scrolling. Not a big deal, but maybe that can be changed ...

Yes that should be reduced on once repository is cached. If you look at one of the [Linux jobs](https://github.com/resteasy/Resteasy/pull/2362/checks?check_run_id=606109895) it's not as much to scroll through. You can also download the files if you'd rather.

jamezp:
I'm going to close this for right now so it doesn't keep running jobs while I try to fix the Windows errors.

jamezp:
I think I have it working now. There are some occasional failures on Windows with the MP TCK, but we can look at those separately or just skip them on Windows if we think that is a better option.

liweinan:
+1 for the PR

PR #2363, opened 2020-04-21
[RESTEASY-2548] add new provider with higher priority than JaxrsFormP…

rsearls:
…rovider

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/JaxrsServerFormUrlEncodedProvider.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/JaxrsServerFormUrlEncodedProvider.java
@@ -1,0 +1,65 @@
+package org.jboss.resteasy.plugins.providers;
+
+import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
+import org.jboss.resteasy.spi.HttpRequest;
+import org.jboss.resteasy.spi.util.FindAnnotation;
+
+import javax.annotation.Priority;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.Encoded;
+import javax.ws.rs.Priorities;
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Form;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.ext.MessageBodyReader;
+import javax.ws.rs.ext.Provider;
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+
+@Priority(Priorities.AUTHORIZATION)
+@Provider
+@Produces("application/x-www-form-urlencoded")
+@Consumes("application/x-www-form-urlencoded")
+public class JaxrsServerFormUrlEncodedProvider  implements MessageBodyReader<Form>
+{
+    protected boolean useContainerParams;
+
+    public JaxrsServerFormUrlEncodedProvider(final boolean useContainerParams)
+    {
+        this.useContainerParams = useContainerParams;
+    }
+
+    @Override
+    public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType)
+    {
+        return (useContainerParams && Form.class.isAssignableFrom(type));
+    }
+
+    @Context
+    HttpRequest request;
+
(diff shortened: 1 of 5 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/JaxrsServerFormUrlEncodedProvider.java
```diff
@@ -0,0 +1,97 @@
...
+    public long getSize(Form form, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType)
+    {
+        return -1;
+    }
+
+    @Override
+    public void writeTo(Form form, Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType, MultivaluedMap<String, Object> httpHeaders, OutputStream entityStream) throws IOException, WebApplicationException
+    {
+        LogMessages.LOGGER.debugf("Provider : %s,  Method : writeTo", getClass().getName());
+        new FormUrlEncodedProvider().writeTo(form.asMap(), null, null, annotations, mediaType, httpHeaders, entityStream);
```
asoldano, line 86: Do we really need to create a new instance of FormUrlEncodedProvider for each invocation?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/JaxrsServerFormUrlEncodedProvider.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public CompletionStage<Void> asyncWriteTo(Form form, Class<?> type, Type genericType, Annotation[] annotations,
+                                              MediaType mediaType, MultivaluedMap<String, Object> httpHeaders,
+                                              AsyncOutputStream entityStream)
+    {
+        LogMessages.LOGGER.debugf("Provider : %s,  Method : writeTo", getClass().getName());
+        return new FormUrlEncodedProvider().asyncWriteTo(form.asMap(), null, null, annotations, mediaType, httpHeaders, entityStream);
```
asoldano, line 95: Same as above

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/JaxrsServerFormUrlEncodedProvider.java
```diff
@@ -0,0 +1,97 @@
...
+        if (useContainerParams) {
+            boolean encoded = FindAnnotation.findAnnotation(annotations, Encoded.class) != null;
+            if (encoded) {
+                map = request.getFormParameters();
+            } else {
+                map = request.getDecodedFormParameters();
+            }
+        } else {
+            map = new FormUrlEncodedProvider().readFrom(null, null,
+                    annotations, mediaType, httpHeaders, entityStream);
```
asoldano, line 65: Do we really need to create a new instance of FormUrlEncodedProvider for each invocation?

rsearls:
Resteasy requires a provider for message reading and writing for "Form"
objects AND supports the REASEasy configuration context parameter, 
"resteasy.use.container.form.params".  This is the support 
JaxrsServerFormUrlEncodedProvider provides.

JaxrsFormProvider provides message reading and writing for "Form" objects.
ServerFormUrlEncodedProvider via inheritance from FormUrlEncodedProvider provides 
message reading and writing for "MultivaluedMap" objects.  More importantly
ServerFormUrlEncodedProvider was created to provide support for REASEasy configration
context parameter, "resteasy.use.container.form.params".  JaxrsFormProvider is
provided as a service by resteasy-core and has no knowledge of this configuration
parameter.
 
    Installation_Configuration.xml line 762-771
    resteasy.use.container.form.params
      Obtain form parameters by using HttpServletRequest.getParameterMap().
      Use this switch if you are calling this method within a servlet filter 
      or eating the input stream within the filter.

Undertow only returns "form" values in a MultivaluedMap<String, String> object.
At sometime in the past this information could be gotten from Undertow in
an input stream; that is no longer the case, so resteasy needs the change to
get the data in the way Undertow provides.

JaxrsServerFormUrlEncodedProvider was modeled on JaxrsFormProvider which
creates a new instance of FormUrlEncodedProvider during reading and writing.

rsearls:
I've simplified JaxrsServerFormUrlEncodedProvider.  Only the message reading is needed with flag
"resteasy.use.container.form.params".  The existing code in JaxrsFormProvider will be used for
message writing.

asoldano:
OK, thanks @rsearls . I'm merging this, anyway I think the information you provided in the comment above should perhaps be added in the relevant classes' header comments in the code. Please consider doing that.

PR #2364, opened 2020-04-22
Using GitHub Actions for CI

jamezp:
This is a replacement for #2362. I guess you can't reopen drafts.


jamezp:
Closing this one too as it looks like further testing is required. I think I have fixes, but I don't want to spam the action results.

PR #2368, opened 2020-04-24
Fix RESTEASY-2526 by using asyncutil lib for async loops

FroMage:
Proper fix for #2342, as described there.

This PR introduces a new (tiny) dependency that is required to properly implement loops using `CompletionStage`. I'm not sure if it's best to add a jboss modules for it, or include it in the jaxrs-all module.

Need guidance from @asoldano about this.


asoldano:
Thanks a lot @FroMage . Yes, we need a new module, to be declared private. I'm reviewing the PR soon.

PR #2372, opened 2020-04-27
[RESTEASY-2472]:@Suspended AsyncResponse timeout still runs, even if …

jimma:
…the response was resumed before


ronsigal:
Probably not necessary, but I changed the wait time in getTimeout() to 2 seconds.

PR #2374, opened 2020-04-28
[RESTEASY-2472] check completetion flag for resume

rsearls:


rsearls:
WIthdrawing this PR.  Use Jim's instead.

PR #2375, opened 2020-04-28
[RESTEASY-2569]  RestClient does not propogate incoming headers

dabloem:
new static helper method:
ClientRequestContextUtils.getDeclaring(ClientRequestContext requestContext);

### DIFF
--- a/resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/ClientHeadersFactoryCDITest.java
+++ b/resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/ClientHeadersFactoryCDITest.java
@@ -36,7 +36,7 @@ public class ClientHeadersFactoryCDITest {
 
       @Inject
       @RestClient
-      private TestResourceIntf service;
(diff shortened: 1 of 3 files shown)


resteasy-client-microprofile/src/test/java/org/jboss/resteasy/microprofile/client/ClientHeadersFactoryCDITest.java
```diff
@@ -36,24 +36,26 @@
...
       @RestClient
-      private TestResourceIntf service;
+      private SubClassResourceIntf service;
 
       public String work() {
          return service.hello("Stefano");
       }
    }
 
    @Path("/")
```
kenfinnigan, line 46: It would be better if the verification that incoming headers are propagated from parent interface methods is added as a separate test.

This changes removes the verification that passing client headers is inherited into a child interface

kenfinnigan: Sorry, ignore this. Not sure why I thought the subclass was already there

PR #2332, opened 2020-03-24
RESTEASY-2524

dabloem:
Additional ProxySelector.
see original smallrye code [here](https://github.com/smallrye/smallrye-rest-client/blob/master/implementation/src/main/java/io/smallrye/restclient/RestClientBuilderImpl.java#L159)

PR #2376, opened 2020-04-28
[RESTEASY-2524] replace existing that searches for defaultProxy with …

rsearls:
…selectHttpProxy method

### DIFF
--- a/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -38,6 +38,8 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.Proxy;
+import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -209,28 +211,9 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
-                    .orElse(null);
-
-            if (userProxyHost != null && userProxyPort != null) {
-                resteasyClientBuilder = builderDelegate.defaultProxy(userProxyHost, userProxyPort, userProxyScheme);
-            } else {
-                //no proxy
-                resteasyClientBuilder = builderDelegate;
-            }
+            selectHttpProxy().ifPresent(proxyAddress -> builderDelegate.defaultProxy(
+                    proxyAddress.getHostString(), proxyAddress.getPort()));
```
asoldano, line 215: Is this really meant to replace the check for user provided proxy properties too? Basically wondering about the difference between this PR and #2332 .

rsearls:
This fix is covered by https://github.com/resteasy/Resteasy/pull/2332.
Withdrawing this one.

PR #2381, opened 2020-05-01
[RESTEASY-2571] Avoid a NullPointerException of the MediaType associa…

jamezp:
…ted with request headers is null.

https://issues.redhat.com/browse/RESTEASY-2571


asoldano:
@jamezp , it might be worth rebasing, so that we test this with the new Github actions

asoldano:
Thanks @jamezp . The only doubt I have here is if the behaviour in case of missing media type in the header is correct/acceptable or what. Most likely better than the NPE, though. @ronsigal , any thought here?

jamezp:
@asoldano @ronsigal The only reason I went with this approach is the [ClientRequestHeaders](https://github.com/resteasy/Resteasy/blob/fb09fcd6e7c3437628ae259da472ce8d930c38c9/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientRequestHeaders.java#L224-L230) allows `null` to be returned and `null` is allowed to be passed to the `HttpEntity`.

asoldano:
@jamezp OK, merging.

PR #2387, opened 2020-05-06
Test intermittent sse failures with TravisCI

asoldano:


asoldano:
I think RESTEASY-2542 was not fixed completely, regardless of the TravisCI->GithubActions transition, as a matter of fact we have a Sse failure here in the integration-testsuite. Before the fix, though, we also used to have a sse test fail in the vertx server adapter module, which seems to have gone.

FroMage:
This test runs:

```java
try (SseEventSink sink = eventSink)
            {
               logger.info("sending 3 events");
               eventSink.send(builder.data("thing1").build());
               eventSink.send(builder.data("thing2").build());
               eventSink.send(builder.data("thing3").build());
               logger.info("sent 3 events");
            }
```

Which is just wrong, not matter what the spec says, because we're never listening to the returned `CompletionStage` which may throw, or just not be completed by the time we close.

So probably we have an exception and we're ignoring.

FroMage:
@asoldano if you allow me to push to your branch I can add logging, but I've no idea how to enable that logging.

asoldano:
Thanks @FroMage . I think you should now be allowed to push here, sorry for the late reply

FroMage:
@asoldano nope:

```
error: failed to push some refs to 'git@github.com:asoldano/Resteasy.git'
```

FroMage:
errr, I added logging commits, but CI doesn't run on the PR anymore…

asoldano:
@FroMage it's because of the conflict, it has to be rebased

asoldano:
I've rebased and force-pushed.... let's see how it goes

asoldano:
Closing this as the investigation moved to https://github.com/resteasy/Resteasy/pull/2391

PR #2389, opened 2020-05-07
[RESTEASY-2576] added client modules

rsearls:


jamezp:
Just an FYI it looks like others are seeing the same issue https://github.community/t5/GitHub-Actions/Getting-Maven-quot-Could-not-transfer-artifact-quot-with-500/m-p/45556# with failing to transfer dependencies.

PR #2399, opened 2020-05-10
[RESTEASY-2581] Add profile to enable testing using Wildfly "standalone-microprofile.xml" configuration file

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2581

Still WIP because some test cases still need for a little investigation.
Basically, this PR introduces a new profile which - like for what happens in Wildfly TS - would use the `standalone-microprofile.xml` config file to start the server up.
Then, when running tests by passing a new property - namely `-Dts.standalone.microprofile` to activate the above mentioned profile, some tests in `integration-tests` and `integration-tests-spring/inmodule` TS modules will fail.
Hence, two new JUint `Category` interfaces have been added:

* `ExpectedFailingBecauseOfSmallRyeMicroprofileOpenApi11`: skips tests that _seem_ to fail because of the SmallRye MP OpenAPI implementation used in Wildfly 19 (further investigation to confirm this is needed)
* `ExpectedFailingWithStandaloneMicroprofileConfiguration`: skips tests that fail because of missing subsystems in Wildfly when `standalone-microprofile.xml` is used.

This PR also bumps  `server.version` to 19.0.0.Final.

Commands to verify that TS is running fine with the new profile:
```
$ export SERVER_VERSION=19.0.0.Final
$ mvn -B -am -pl integration-tests -fae -Dserver.version=$SERVER_VERSION install -Dts.standalone.microprofile
```

PR #2407, opened 2020-05-14
[RESTEASY-2581] Add profile to enable testing using Wildfly "standalone-microprofile.xml" configuration file

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2581, parallel to https://github.com/resteasy/Resteasy/pull/2399 which is for branch `3.11`.

Basically, this PR introduces a new profile which - like for what happens in Wildfly TS - would use the `standalone-microprofile.xml` config file to start the server up.
Then, when running tests by passing a new property - namely `-Dts.standalone.microprofile` to activate the above mentioned profile, some tests in `integration-tests` and `integration-tests-spring/inmodule` TS modules will fail.
Hence, two new JUint `Category` interfaces have been added:

* `ExpectedFailingBecauseOfSmallRyeMicroprofileOpenApi11`: skips tests that _seem_ to fail because of the SmallRye MP OpenAPI implementation used in Wildfly 19 (further investigation to confirm this is needed)
* `ExpectedFailingWithStandaloneMicroprofileConfiguration`: skips tests that fail because of missing subsystems in Wildfly when `standalone-microprofile.xml` is used.

Commands to verify that TS is running fine with the new profile:
```
$ export SERVER_VERSION=19.0.0.Final
$ mvn -B -am -pl integration-tests -fae -Dserver.version=$SERVER_VERSION install -Dts.standalone.microprofile
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 install:
  - mvn -B clean
 script:
- - travis_wait 60 mvn -B -Ptravis -fae -Dserver.version=$SERVER_VERSION ${ELYTRON:+-Delytron} install
+ - travis_wait 60 mvn -B -Ptravis -fae -Dserver.version=$SERVER_VERSION ${ELYTRON:+-Delytron} ${STANDALONE_MICROPROFILE:+-Dts.standalone.microprofile} install
 
 language: java
 jdk:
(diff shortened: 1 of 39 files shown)


fabiobrz:
Hi @ronsigal, as you suggested, this is for 3.12. I can't add the label, though. Would you mind to do that?

fabiobrz:
I closed and reopened the PR in order to trigger Travis CI again as it failed despite the build succeeds both with JDK8 and JDK11 (OpenJDK) locally.

ronsigal:
Looks good. I'll merge it.

PR #2424, opened 2020-05-21
[RESTEASY-2595] Use entity class loader in links provider

gytis:
https://issues.redhat.com/browse/RESTEASY-2595


gytis:
Good question. I haven't used `RolesAllowed` so I haven't seen it fail. But it would make sense to use the same class loader as for the application class. I'll add it to this method as well.

gytis:
Actually no, that shouldn't be needed. `RolesAllowed` comes from one of `resteasy-links` direct dependencies (`jboss-annotations-api_1.3_spec`) so it should be available. The problem that we're facing is that `resteasy-links` tries to load one of the application classes and it's not available in the class loader.

gytis:
Just to be 100% sure that this wouldn't break backwards compatibility I'll rewrite this PR to expose an extra `getLinks` method that takes a class loader as a parameter. And will keep the existing method behaving as before.

gytis:
There's a bunch of changes in this PR but I mainly reshuffled the existing code to the different classes to allow easier use of the provided class loader.

gytis:
@jamezp would you mind taking another look at this once you have a spare minute?

PR #2425, opened 2020-05-22
Add GitHub Actions to 3.12

jamezp:
This adds GitHub Actions for CI to the 3.12 branch. If this looks okay I can cherry-pick it to the 3.13 branch as well.

The second commit for the JavaDoc thing I'm not sure how to validate it still generates the expected JavaDoc. How are those usually built and distributed?

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,71 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Java CI with Maven
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest ]
+        java: ['1.8', '11']
+        wildfly-version: ['18.0.1.Final', '19.0.0.Final']
+
+    steps:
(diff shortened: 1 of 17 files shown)


jamezp:
@asoldano Okay. I'll look for those commits and add them to this PR.

jamezp:
Actually @asoldano were those not included in #2412?

asoldano:
I think the following commits might be neeed (but would have to check):
https://github.com/resteasy/Resteasy/commit/161ae95e97e4ad074b3bd293a94c9bd3cc0725bc
https://github.com/resteasy/Resteasy/commit/566badc79806b5db33ba572c89dd49ce5bca767c
https://github.com/resteasy/Resteasy/commit/e42bc8c9af8e359c33d8a1a420a3202768dadf6f

jamezp:
Okay. I'll give it a better look then. Thanks Alessio.

jamezp:
It does look like we need those three, which don't really apply cleanly so I'll have to sort that out.

We may possibly need these too as all:

- d3b0e5e41a0b028b358f1fbd84f81baffdbd61b9
- ef4aa69fdf9421f5d885c134e10677aa84e5cc85

jamezp:
Okay updated. The one that didn't seem to fit was 566badc79806b5db33ba572c89dd49ce5bca767c. However I need to look at how asyn servlets are processed as maybe it's somewhere other than the `HttpServletResponseWrapper`.

asoldano:
Thanks @jamezp !

> The one that didn't seem to fit was 566badc.

this is probably because of https://issues.redhat.com/browse/RESTEASY-2506, which was implemented in master only. Anyway, I think I'll go ahead and merge this PR.

PR #2430, opened 2020-06-02
[RESTEASY-2597] add methods and test to support setting

rsearls:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/other/DisableAutomaticRetriesTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void testFlag() throws Exception {
+        ResteasyClientBuilder builder = (ResteasyClientBuilder) ClientBuilder.newBuilder();
+        Assert.assertFalse(builder.isDisableAutomaticRetries());
+        builder.disableAutomaticRetries();
+        Client client = builder.build();
+        Assert.assertTrue(builder.isDisableAutomaticRetries());
+        client.close();
+    }
```
asoldano, line 51: This looks like a unit test more than an integration test. If that's what you really meant to do, please move this to the resteasy-client module test directory. If a real integration test is not too hard to write, it should live here, instead.

resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
```diff
@@ -223,6 +223,12 @@
 
    public abstract long getConnectionTimeout(TimeUnit unit);
 
+   /**
+    * boolean, notify apache to disable its automatic retries.
+    */
+   public abstract void disableAutomaticRetries();
```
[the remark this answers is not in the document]
rsearls: Apache's HttpClientBuilder does not offer on/off value switching.
In Apache there is a default value for automaticRetriesDisabled
and one can call its class, HttpClientBuilder.disableAutomaticRetries()
but there is no parameter for switching on/off. Resteasy should reflect that.  
I do not believe boolean parameter should be added to this method.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/other/DisableAutomaticRetriesTest.java
rsearls: done

PR #2432, opened 2020-06-09
Removing annotation to skip tests with OpenAPI issues (3.11)

fabiobrz:
This is to remove the `ExpectedFailingBecauseOfSmallRyeMicroprofileOpenApi11` annotation that was introduced with https://github.com/resteasy/Resteasy/pull/2399 because al the issues reported to the SMallRye OpenAPI upstream project have been solved and component version bump in Wildfly is tracked.

FYI @ronsigal


fabiobrz:
Hi @ronsigal could you please add a "don't merge" label to this one? I am not able to do that and of course this should wait for the next SmallRye OpeanAPI version bump in Wildfly (https://github.com/smallrye/smallrye-open-api/milestone/8 I guess).
BTW this is the reason why tests are failing ATM.

PR #2434, opened 2020-06-10
[RESTEASY-2609] 4.5.4-SNAPSHOT -> 4.5.4.Final

ronsigal:


ronsigal:
The failures come from timeout tests in the MicroProfile REST Client TCK. I was able to configure the timeout and get the tests to pass on my home machine.

PR #2435, opened 2020-06-11
[RESTEASY-2300] Blank Host Name still throws exception

istudens:
https://issues.redhat.com/browse/RESTEASY-2300


asoldano:
Thanks @istudens

PR #2443, opened 2020-06-16
[RESTEASY-2594] added test to show MP-client proxy call

rsearls:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/MPCollectionTest.java
```diff
@@ -0,0 +1,55 @@
...
+        return TestUtil.finishContainerPrepare(war, null, null);
+    }
+
+    private String generateURL(String path) {
+        return PortProviderUtil.generateURL(path, WAR_SERVICE);
+    }
+
+    @Test
+    public void testOne() throws Exception {
+        RestClientBuilderImpl builder = new RestClientBuilderImpl();
```
asoldano, line 48: As this test is serving as an example, it would be better to follow the best practices and use the spec api instead of relying on the implementation class directly:

rsearls: done

rsearls:
Merged into upstream/mp-rest-client-20 rather an master

PR #2450, opened 2020-06-18
[RESTEASY-2625]:Fix @Produces("text/event-stream;element-type=text/xml") throws exception

jimma:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/rx/rxjava2/resource/Rx2FlowableableSSECompatibilityResource.java
```diff
@@ -18,13 +17,11 @@
...
-   @Produces("text/event-stream")
-   @SseElementType("application/json")
+   @Produces("text/event-stream;element-type=application/json")
    void eventStreamThing(@Context SseEventSink eventSink, @Context Sse sse);
 
    @GET
    @Path("flowable/thing")
-   @Produces("text/event-stream")
-   @SseElementType("application/json")
+   @Produces("text/event-stream;element-type=application/json")
```
asoldano, line 25: I'm wondering if it would be better to have both ways of using annotations here, instead of changing all resource methods you could change just one of them and add a comment explaining the reason / referring to the JIRA number.

jimma: Rx2ObservableSSECompatibilityTest is the way to use the @SseElementType("application/json") and we don't have to use @SseElementType here.

ronsigal:
I added a comment about @SseElementType on https://issues.redhat.com/browse/RESTEASY-2625. Basically, I think both forms should be allowed.

FroMage:
I've just verified that this fixes the two test regressions on Quarkus.

PR #2452, opened 2020-06-22
Resteasy 2618 mp rest client query param style

rsearls:


rsearls:
Moved to upstream/mp-rest-client-20 branch

PR #2455, opened 2020-06-22
Resteasy 2617 mp rest client context resolver

rsearls:


rsearls:
Moved to upstream/mp-rest-client-20 branch

PR #2456, opened 2020-06-25
Encode code points that span multiple characters

stuartwdouglas:
Fixes RESTEASY-2633


stuartwdouglas:
Tests are here: https://github.com/resteasy/Resteasy/pull/2480

Is there any reason why this did not make it into the 4.5.6 release for Quarkus?

asoldano:
Thanks Stuart.
As for why this didn't go into 4.5.6, it's because it was decided to make that release as minimal as possible, just fixing the CVE (plus another fix that was there since weeks).
If this and perhaps other fixes are urgent, I can arrange another micro release soon, just let me know.

PR #2463, opened 2020-07-03
[RESTEASY-2943] Support reactive publishers at the engine level.

crankydillo:
Addresses [RESTEASY-2943](https://issues.redhat.com/browse/RESTEASY-2943).  Includes refactoring the reactor-netty-based HttpClientEngine to support the new interface.

This is a PoC that aims to support a more 'direct' line from `rx(Class)`
to a reactive 'provider'.  Some reactive implementations have features
that can be better leveraged with this.  In our case, we really need to
leverage the end user's project-reactor 'context' when we build the
client.  This use case is exercised in the
ReactorTest#testSubscriberContext test.

Some things like code reuse and test coverage have been postponed until
I get some feedback on the concept.

Apologies for not writing up an issue.  We already had this code written
so I'm hoping it better explains what we need than trying to write it up
in an issue.  Just let me know what you'd like me to do that in that
regard.


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/PublisherRxInvokerImpl.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @Override
+    public <T> Publisher<T> get(final GenericType<T> responseType) {
+        return createClientInvocation(HttpMethod.GET, null).submitRx(responseType);
+    }
+
+    @Override
+    public Publisher<Response> put(final Entity<?> entity) {
+        return null;
```
crankydillo, line 45: Of course, these are still TBD.

asoldano:
@FroMage I think this is something that might be interesting for you to review. Do you have time for a quick check?

FroMage:
Yeah, I think it makes sense to support `Publisher` for the client. It's not very useful outside of Reactor because `Publisher` is just bad at dealing with single values, but it doesn't hurt either.

crankydillo:
@FroMage @asoldano Please let me know if you are comfortable with me proceeding.  If so, I'll finish the implementation and add tests.

asoldano:
> @FroMage @asoldano Please let me know if you are comfortable with me proceeding. If so, I'll finish the implementation and add tests.

Yes, please :-)

asoldano:
@crankydillo , I haven't been able to spend much time here lately; is what you had in mind for this PR completed and ready for review? Thanks

crankydillo:
> @crankydillo , I haven't been able to spend much time here lately; is what you had in mind for this PR completed and ready for review? Thanks

Not yet.  I hope to have it done by early next week.  Sorry for the delay.

asoldano:
@crankydillo , no hurry at all, I just wanted to understand if you were blocked on something from me/us or not. Cheers

crankydillo:
@asoldano The PR is done.  Please start your review when you can.  @anilgursel may be interested as well.

rsearls:
I find no issue with this PR, however it presents a situation in which there are 3 contexts in which
Publisher is used.  One is here.  Another is in resteasy-rxjava2.  A third is coming in  resteasy-client-microprofile.  This presents an issue of properly identifying in which context the return type is being
used.   Work is under way to address this.  There will need to be some minor tweeking of this code
once a solution is implemented to address this.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
```diff
@@ -747,7 +782,37 @@ protected ClientResponse filterResponse(ClientRequestContextImpl requestContext,
       }
       catch (Exception ex)
       {
-         exceptionFn.apply(ex);
+         return exceptionFn.apply(ex);
+      }
+
+      return asyncHttpEngineSubmitFn.apply(response -> {
```
ronsigal, line 788: Isn't this line inaccessible now?

ronsigal:
Hi @crankydillo,

First of all, I want to join the rest of the team in thanking you for this contribution. I'm still trying to absorb it, and I probably don't understand everything.

One general comment. As far as I can tell, most reactive frameworks have two kinds of objects, with a single member (single, mono, uni, ...) and a stream of members (flowable, flux, multi, ...), so it would be interesting to unify all of these. I don't know how feasible that is. Currently, for example, SingleRxInvokerImpl and MonoRxInvokerImpl are built on top of CompletionStageRxInvokerImpl, but there's no unifying RxInvoker that they all subclass. Given that Flux and Flowable (but not Observable) both implement Publisher, maybe something can be done there. Don't know.

And a couple of questions:

1. The PublisherRxInvokerImpl methods all return a Publisher, but ReactorNettyClientHttpEngine.submitRx() returns a Mono. What is your plan for returning a Flux? A different submitRx() that returns a Flux would clash.

2. Along similar lines, ReactorTest tests handling a Mono, but I don't see how a Flux would work in the same context. Currently, Flowables, Observables, and Flux's are returned over the SSE infrastructure. and FlowableRxInvokerImpl, ObservableRxInvokerImpl, and FluxRxInvokerImpl understand that. But io.projectreactor.netty:reactor-netty is a whole other world. How would it that work?

Thanks,
Ron

ronsigal:
Oh, another thing I was wondering. ReactorNettyClientHttpEngine depends on the existence of reactor.netty.http.client.HttpClient and the reactor-netty library. Is there anything analogous for the other reactive frameworks? I'm guessing, given the similarity of the other frameworks, that reactor-netty could be adapted to work for rxjava2 and mutiny, but I haven't really looked into it.

crankydillo:
Hi Ron,

Thanks for the kind words!  I will answer your questions inline; however, you should know that I didn't give this stuff 'great' thought.  Our big need is be able to pass information via Reactor's subscriber context from 'end users' down to the Reactor engine we implement.  That was my main goal, and I only worked a few hours on this:)

> Hi @crankydillo,
> 
> First of all, I want to join the rest of the team in thanking you for this contribution. I'm still trying to absorb it, and I probably don't understand everything.
> 
> One general comment. As far as I can tell, most reactive frameworks have two kinds of objects, with a single member (single, mono, uni, ...) and a stream of members (flowable, flux, multi, ...), so it would be interesting to unify all of these. I don't know how feasible that is. Currently, for example, SingleRxInvokerImpl and MonoRxInvokerImpl are built on top of CompletionStageRxInvokerImpl, but there's no unifying RxInvoker that they all subclass. Given that Flux and Flowable (but not Observable) both implement Publisher, maybe something can be done there. Don't know.

With my change, I'm assuming most reactive RxInvokers would be built using composition of `PublisherRxInvokerImpl`, not `CompletionStageRxInvokerImpl`.  I don't know if a base class makes sense.  I probably should have attempted a refactor of the RxJava2 invoker, but I wasn't doing anything like before getting the big picture in.
 
> And a couple of questions:
> 
>     1. The PublisherRxInvokerImpl methods all return a Publisher, but ReactorNettyClientHttpEngine.submitRx() returns a Mono. What is your plan for returning a Flux? A different submitRx() that returns a Flux would clash.

IMO, if you created an engine that returned a `Flux`, that would simply be a mistake and should be caught in code review.  I ASSume it would still work; however, it would be inappropriate.  I don't know much about JAXRS+RestEasy+SSE, but, as you mention below, I'm under the impression that is where `Flux` belongs.

>     2. Along similar lines, ReactorTest tests handling a Mono, but I don't see how a Flux would work in the same context. Currently, Flowables, Observables, and Flux's are returned over the SSE infrastructure. and FlowableRxInvokerImpl, ObservableRxInvokerImpl, and FluxRxInvokerImpl understand that. But io.projectreactor.netty:reactor-netty is a whole other world. How would it that work?

I'm sorry, but I don't understand what you are asking.  `ReactorTest` tests the 'make a request and get a (future) response' path, which is accomplished with `MonoRxInvoker`.  It also tests the 'make a request and get a (future) producer of events' path, which is accomplished with `FluxRxInvoker`.  Apologies if I'm missing something.  I wish I knew more about SSE stuff, but I haven't really looked into it much.

crankydillo:
> Oh, another thing I was wondering. ReactorNettyClientHttpEngine depends on the existence of reactor.netty.http.client.HttpClient and the reactor-netty library. Is there anything analogous for the other reactive frameworks? I'm guessing, given the similarity of the other frameworks, that reactor-netty could be adapted to work for rxjava2 and mutiny, but I haven't really looked into it.

I don't know if there is anything analogous for all reactive frameworks; however, HTTP clients is a pretty common use case for them.  Akka-streams has one.  RxJava had [RxNetty](https://github.com/ReactiveX/RxNetty); however, it's state is uncertain.

AFAIK, it is possible to implement a `RxJava2ClientHttpEngine` on top of reactor-netty (or akka-streams).  In fact, one of the main benefits of the reactive project is the goal to make such things possible.  Each reactive implementation (akka-streams, rxjava2, reactor) is supposed to support interop.

crankydillo:
Hi all!  Any update on this?

crankydillo:
I added another [use case](https://github.com/resteasy/Resteasy/pull/2463/commits/fed453a47dffdde258503033247598c8bb319b8c) our clients have, which is to have an efficient, per request timeout.

ronsigal:
Hi @crankydillo,

Sorry for the radio silence. I'm going to try to take another look here. I'll have to start over and figure out what I was trying to say in July and August. ;-)

-Ron

ronsigal:
Hi @crankydillo,

I should start with my usual apologies for the lateness of my response. To coin a seasonally appropriate metaphor, I feel like I'm trying to wrap my head around a pumpkin where the pumpkin is bigger than my head. ;-)

The last time around I had some half baked questions, and now I have some more. The important difference is that in the meantime I realized that there was a concept here that I didn't even realize that I didn't understand. In particular, I couldn't possibly understand what you're doing without understanding, or even being aware of, the concept of a subscriber context. So my big leap forward is becoming aware of subscriber contexts.

One of the things I was hung up on the last time was how Publisher and Flow fit in. Tell me if I'm wrong, but my impression now is that

 1. you intend to leave FlowInvokerImpl unchanged, and

 2. you're using PublisherRxInvoker in resteasy-client because Publisher is a project independent concept which generalizes Mono, which is specific to the reactor project.
 
So, part of your changes are in resteasy-client, and the rest are in resteasy-reactor and resteasy-client-reactor-netty. One question I have is: Given that the only implementation of the new concepts in resteasy-client are based on netty, maybe they're too specific to go in resteasy-client.

If I understand correctly, using a subscriber context the way you have done is possible because ReactorNettyClientHttpEngine.submitRx() returns a Mono&lt;T&gt; instead of the Future&lt;T&gt; returned by ApacheHttpAsyncClient4Engine.submit(). I'm wondering if you could have ReactorNettyClientHttpEngine.submit() return a Future&lt;Mono&lt;T&gt;&gt; and dispense with ReactiveClientHttpEngine. Maybe it wouldn't be so pretty, I don't know.

Anyway, I'm still working on my comprehension, but I'm making progress.

-Ron

crankydillo:
Hi Ron,

No worries on being late.  Replies inlined.

> The last time around I had some half baked questions, and now I have some more. The important difference is that in the meantime I realized that there was a concept here that I didn't even realize that I didn't understand. In particular, I couldn't possibly understand what you're doing without understanding, or even being aware of, the concept of a subscriber context. So my big leap forward is becoming aware of subscriber contexts.
> 
> One of the things I was hung up on the last time was how Publisher and Flow fit in. Tell me if I'm wrong, but my impression now is that
> 
>     1. you intend to leave FlowInvokerImpl unchanged, and

Yes.

>     2. you're using PublisherRxInvoker in resteasy-client because Publisher is a project independent concept which generalizes Mono, which is specific to the reactor project.

That's correct.

> 
> So, part of your changes are in resteasy-client, and the rest are in resteasy-reactor and resteasy-client-reactor-netty. One question I have is: Given that the only implementation of the new concepts in resteasy-client are based on netty, maybe they're too specific to go in resteasy-client.

I think you could make that case; however, I do wonder how much work it will entail.  My initial thought was to do that; however, there are many internal classes that are needed.  Besides, [Publisher](https://docs.oracle.com/javase/9/docs/api/index.html?java/util/concurrent/Flow.Publisher.html) is in Java 9 and could viewed similar to `CompletionStage` + benefits.  AFAIK, it's all optional to implementers.  Isn't the only negative the addition of a jar file (until you no longer want to support Java 8)?

> If I understand correctly, using a subscriber context the way you have done is possible because ReactorNettyClientHttpEngine.submitRx() returns a Mono<T> instead of the Future<T> returned by ApacheHttpAsyncClient4Engine.submit(). I'm wondering if you could have ReactorNettyClientHttpEngine.submit() return a Future<Mono<T>> and dispense with ReactiveClientHttpEngine. Maybe it wouldn't be so pretty, I don't know.

I will think on this, but my initial impression is it's a bit hacky:)

> Anyway, I'm still working on my comprehension, but I'm making progress.

Thanks!

ronsigal:
Hi @crankydillo,

There are two related Pull Requests that you might be interested in:
 * https://github.com/resteasy/Resteasy/pull/2574
 * https://github.com/resteasy/Resteasy/pull/2580

-Ron

ronsigal:
Hi @crankydillo,

Just checking in again. I still can't help being a little uncomfortable using Publisher as a way of abstracting Mono. It would have been nice if Mono, Single, etc., had an interface (other than Publisher) they could all implement, but it's too late for that.  I feel better, though, now that you've confirmed that you're specifically focused on Mono and not Flux, since they live in completely different worlds inside of RESTEasy. That was throwing me off at first.

Right now I'm hung up on the fact that you're introducing a lot of stuff into resteasy-client to be able to implement something in resteasy-reactor. It feels a little like the tail wagging the dog. I would feel better if the new stuff in resteasy-client could also be used in, for example, resteasy-rxjava2. I'm wondering if other reactive frameworks have the same notion of a subscriber context. Do you know about that?

-Ron

crankydillo:
Hi @ronsigal ,

I'm on vacation now and can't look at this right now.  I'll be back next week and will add some comments.

ronsigal:
re: "on vacation"

Cool.

crankydillo:
Hi Ron,

Well, I'm back at work, but now I'm slammed with that:)  As an aside, we are in the process of writing a reactor-netty server adapter, so be on the lookout for that PR soon.

> 
> Just checking in again. I still can't help being a little uncomfortable using Publisher as a way of abstracting Mono. It would have been nice if Mono, Single, etc., had an interface (other than Publisher) they could all implement, but it's too late for that. I feel better, though, now that you've confirmed that you're specifically focused on Mono and not Flux, since they live in completely different worlds inside of RESTEasy. That was throwing me off at first.
>
> Right now I'm hung up on the fact that you're introducing a lot of stuff into resteasy-client to be able to implement something in resteasy-reactor. It feels a little like the tail wagging the dog. I would feel better if the new stuff in resteasy-client could also be used in, for example, resteasy-rxjava2. I'm wondering if other reactive frameworks have the same notion of a subscriber context. Do you know about that?

Sorry, but I don't know.  I have worked a small amount with akka-streams and almost none with RxJava.  I believe both might be able to leverage the `PublisherRxInvoker` for a more 'direct' line.

Once we get the server adapter hammered out, I may have some time to revisit this.  I can try to get the required pieces isolated from `rest-client`.

crankydillo:
@asoldano @ronsigal Now that the server adapter is coming to a conclusion, I'm trying to tackle some older issues we have.  One of those is tied to some unfortunate `ThreadLocal` coupling we've done.  We hate that coupling, but so far, have not found a way out of it.  Unfortunately, the `Mono#fromCompletionStage` bridge that this PR would make obsolete causes us issues when we try to use the `MonoRxInvoker`.  If we can get this (or something like it), we could avoid this bridge and we'd not only get the original benefit I post about in this PR, but also we could stick with our (unfortunate) threadlocals for a little while longer...

So..  I hope you are OK with me picking this back up.  If not, let me know.  I'm almost certain `resteasy-rxjava2` could use this.  I will think on @ronsigal 's concern with using `Publisher`.  All that comes to mind is that we'd have to introduce an interface here..

asoldano:
> @asoldano @ronsigal Now that the server adapter is coming to a conclusion, I'm trying to tackle some older issues we have. One of those is tied to some unfortunate `ThreadLocal` coupling we've done. We hate that coupling, but so far, have not found a way out of it. Unfortunately, the `Mono#fromCompletionStage` bridge that this PR would make obsolete causes us issues when we try to use the `MonoRxInvoker`. If we can get this (or something like it), we could avoid this bridge and we'd not only get the original benefit I post about in this PR, but also we could stick with our (unfortunate) threadlocals for a little while longer...
> 
> So.. I hope you are OK with me picking this back up. If not, let me know. I'm almost certain `resteasy-rxjava2` could use this. I will think on @ronsigal 's concern with using `Publisher`. All that comes to mind is that we'd have to introduce an interface here..

Hi @crankydillo , I'm definitely OK with you picking this back up. I'd really like to see this reach a state where we all agree and we can merge it :-)
Thanks

crankydillo:
Hi @ronsigal ,

A few more thoughts what you were hung up on:)

>  [snip]
>> Right now I'm hung up on the fact that you're introducing a lot of stuff into resteasy-client to be able to implement something in resteasy-reactor. It feels a little like the tail wagging the dog. I would feel better if the new stuff in resteasy-client could also be used in, for example, resteasy-rxjava2. I'm wondering if other reactive frameworks have the same notion of a subscriber context. Do you know about that?
> 
> Sorry, but I don't know. I have worked a small amount with akka-streams and almost none with RxJava. I believe both might be able to leverage the `PublisherRxInvoker` for a more 'direct' line.
> 
> [snip]

The benefit of what was done here and is very likely going to be true of all the reactive implementations is that you have this 'direct line' from the business logic to the reactive client call.  We currently go through the `CompletionStage` bridge, which removes a lot of features.  To date, these are the benefits we've noted if we can remove that bridge:

1. Reactive-Implementation-specific classes (e.g. `SubscriberContext` can pass from business logic to client).  Original intent of this PR.
1. Threading 'weirdness' with `CompletableFuture` can be avoided.  That is discussed [here](https://github.com/resteasy/Resteasy/pull/2463#issuecomment-814414160).
1. And our latest...Business logic based timeouts (e.g. `webTarget.request().rx(MonoRxInvoker.class).get(String.class).timeout(Duration.ofMillis(50))` should flow naturally to the inner `Publisher`.  This allows for the connection to be freed/closed without waiting for the request/response to finish.  The `CompletionStage` bridge makes that problematic.

It's difficult to see how the above won't be true for most of the reactive implementations.

crankydillo:
It's me again:)  I just pushed a [refactor](https://github.com/resteasy/Resteasy/pull/2463/commits/c1dc6a4d2039b3e0a1faadccf7343de11ef67a43) to make it clearer that reactive support is optional (dependent on the HTTP engine).  I also [refactored](https://github.com/resteasy/Resteasy/pull/2463/commits/24e5bbfe219f048f2abd084251abcc5803a07393) RxJava's SingleRxInvokerImpl to leverage this work.  I think one of the earlier comments was wondering if this could be done to support RxJava2 (and other reactive impls).

Really hoping to get some feedback.  We are desperately trying to avoid forking the repo..

ronsigal:
Hi @crankydillo,

Long time. ;-)

I'll take a look. Remind me if I don't.

By the way, I've been working on something somewhat related, incorporating https://github.com/eclipse/microprofile-reactive-streams-operators to allow something like
```
   @GET
   @Path("get/string")
   @Stream
   public Publisher<String> get() {
      return ReactiveStreams.of(X_ARRAY).buildRs();
   }
```
As of now, it exists only on https://github.com/ronsigal/Resteasy/tree/eap7-1661, in the module at https://github.com/ronsigal/Resteasy/tree/eap7-1661/resteasy-reactive-streams-operators.

-Ron

crankydillo:
> Hi @crankydillo,
> 
[snip]
> 
> By the way, I've been working on something somewhat related, incorporating https://github.com/eclipse/microprofile-reactive-streams-operators to allow something like
> 
> ```
>    @GET
>    @Path("get/string")
>    @Produces(MediaType.TEXT_PLAIN)
>    @Stream
>    public Publisher<String> get() {
>       return ReactiveStreams.of(X_ARRAY).buildRs();
>    }
> ```
> 
> As of now, it exists only on https://github.com/ronsigal/Resteasy/tree/eap7-1661, in the module at https://github.com/ronsigal/Resteasy/tree/eap7-1661/resteasy-reactive-streams-operators.
> 
> -Ron

I'll definitely look at it.  Thanks!

FYI, I'll fix the checkstyle issues shortly.

crankydillo:
Hi Ron,

> I'll take a look. Remind me if I don't.

Gentle reminder:)

> By the way, I've been working on something somewhat related, incorporating https://github.com/eclipse/microprofile-reactive-streams-operators to allow something like
> 
> ```
>    @GET
>    @Path("get/string")
>    @Produces(MediaType.TEXT_PLAIN)
>    @Stream
>    public Publisher<String> get() {
>       return ReactiveStreams.of(X_ARRAY).buildRs();
>    }
> ```
> 
> As of now, it exists only on https://github.com/ronsigal/Resteasy/tree/eap7-1661, in the module at https://github.com/ronsigal/Resteasy/tree/eap7-1661/resteasy-reactive-streams-operators.

I did look at this.  My first thought is we have some name collisions (this PR also has a `PublisherRxInvoker`:).  I do see how you are tying `Publisher` to jaxrs eventsource.  That is a more natural fit to `Publisher`, which I think people should not typically associate only 1 value.  I think someone brought that up above.  Anyhow, looks like some nice work.

> -Ron

ronsigal:
Hi @crankydillo,

re: "Gentle reminder:)"

Well done!! ;-)  I'll look soon.

-Ron

ronsigal:
Hi @crankydillo,

Ok, I'm back. I'll have to spend some time drilling down to do a more careful evaluation, but I have some thoughts on an architectural level. Please tell me if I'm making any sense.

I keep going back to my discomfort with the use of Publisher in this context. I would say that your PublisherRxInvokerImpl is much closer to CompletionStageRxInvokerImpl than it is to PublisherRxInvokerImpl in the resteasy-reactive-streams-operators module. In fact, it seems to me (tell me if I'm wrong) that your PublisherRxInvokerImpl and CompletionStageRxInvokerImpl are almost the same when you factor out the underlying network layer, i.e., apache HttpClient vs reactor netty.

In fact, I would go as far as saying that your PublisherRxInvokerImpl could BE CompletionStageRxInvokerImpl, if it returned CompletionStages instead of Monos. For example, I see that you've reworked MonoRxInvokerImpl to extend PublisherRxInvokerImpl so that it turns a (simple) Publisher into a Mono instead of turning a CompletionStage into a Mono like it did before. 

If that makes sense, then why not just turn it into something like
```
public class CompletionStageRxInvokerReactiveImpl implements CompletionStageRxInvoker { ...
```
Does that make sense? Would it satisfy your requirements? WDYT? 

-Ron

crankydillo:
Thanks Ron.

> I keep going back to my discomfort with the use of Publisher in this context. I would say that your PublisherRxInvokerImpl is much closer to CompletionStageRxInvokerImpl than it is to PublisherRxInvokerImpl in the resteasy-reactive-streams-operators module. In fact, it seems to me (tell me if I'm wrong) that your PublisherRxInvokerImpl and CompletionStageRxInvokerImpl are almost the same when you factor out the underlying network layer, i.e., apache HttpClient vs reactor netty.

They are very similar!  I think your (understandable) hangup is that you associate 'Publisher' with N values.  In this case, N will always be 1.  As such, it is almost identical to `CompletionStage` other than having other useful properties such as being driven by a subscription and having back-pressure (and the benefits I will link to in the next section:).

Honestly, I don't know why [the reactive-streams API](https://www.reactive-streams.org/reactive-streams-1.0.0-javadoc/) doesn't have a first-class 'single-value publisher' concept, but it doesn't seem too.  I'm sure they have a reason.

> In fact, I would go as far as saying that your PublisherRxInvokerImpl could BE CompletionStageRxInvokerImpl, if it returned CompletionStages instead of Monos. For example, I see that you've reworked MonoRxInvokerImpl to extend PublisherRxInvokerImpl so that it turns a (simple) Publisher into a Mono instead of turning a CompletionStage into a Mono like it did before.

It could, but then it would not provide the 3 benefits I list [here](https://github.com/resteasy/Resteasy/pull/2463#issuecomment-823249401).

ronsigal:
re: "would not provide the 3 benefits"

Ok, I thought there might be something I was missing.

Well, what about a MonoRxInvokerNettyImpl? My point is, is there a reason for returning an interface? Could you just return a Mono, or a Single? That would probably involve factoring out what you've added to ClientInvocation and moving it to a separate module, I guess. Or, what about ... creating a RESTEasy specific interface like
```
public interface ResteasyMono<T> {

    public void subscribe(Subscriber<? super T> s);
}
```
that can be converted to a Mono? Would that blow benefit 1?

crankydillo:
Hi @ronsigal ,

Before I reply to your latest comment, I wonder if it fundamentally gets back to one of your first comments, which had this:

> I still can't help being a little uncomfortable using Publisher as a way of abstracting Mono. It would have been nice if Mono, Single, etc., had an interface (other than Publisher) they could all implement, but it's too late for that. 

If this is your core issue, I would like to get to the bottom of that.  You mentioned above that `Publisher` is very much like `CompletionStage`.  I just have to repeat that it is exactly that, just in the reactive context!  Both represent the chosen interface for future values.  `Publisher` just has other characteristics.  In fact, `Publisher` is supposed to be the bridge between other APIs allow for a natural transition.  You can see that in this code where I replaced not only project reactor's integration with `CompletionStage`, but also rxjava2's (`Single.fromPublisher(...)`).  This brings yet another benefit for rxjava, which is deferring the HTTP call until subscription.

Anyhow, I just want to reiterate that the authors of reactive streams did not see a need for an interface that represents a single value at least when you are talking about working at the interface level ([code](https://github.com/openjdk/jdk/blob/b6a729a79b2ca65ec3dc26af1ee8dcd1cd3afa41/jdk/src/java.base/share/classes/java/util/concurrent/Flow.java#L166)).  The engine code here is trying to operate at the interface level.

So if there was an rxjava2 HTTP client engine, that would easily plugin into the MonoRxInvokerImpl as I've defined here, without going through a `CompletionStage`.  Will you get all the benefits I discuss above?  No, but you should get a few.

Basically, it all plays nice in reactive contexts.

ronsigal:
Hey @crankydillo,

Thank you for your patience. It must be frustrating trying to get through to me. ;-)

On the one hand, I can see interpreting Publisher as a general purpose type which can represent 0, 1, ... arbitrary elements. On the other hand, it seems to me that most reactive libraries have a pair of types: Single/Flowable, Mono/Flux, Multi/Uni ... . I do see that reactor has CorePublisher to unify the two. I lean towards associating Publisher with the possibility of multiple items, but maybe I'm flowing against the tide (so to speak). I'll survey some people around here and see what they think. 

In the meantime, what do you think of something like
```
package org.jboss.resteasy.client.jaxrs;

import org.reactivestreams.Subscriber;

public interface Unit<T> {
	
   void subscribe(Subscriber<? super T> s);
   T get();
   CompletableFuture<U> toFuture();
}
```
and
```
package org.jboss.resteasy.client.jaxrs;
...
public interface UnitRxInvoker extends RxInvoker<Unit<?>>
{
   @Override
   Unit<Response> get();
   ...
```
and
```
public interface ReactiveClientHttpEngine extends AsyncClientHttpEngine
{
   /**
    * This is the main bridge from RestEasy to a reactive implementation.
    */
   <T, U> Unit<T, U> submitRx(ClientInvocation request,
                                            boolean buffered,
                                            ResultExtractor<T> extractor);

   <T, U> Unit<T, U> fromCompletionStage(CompletionStage<T> cs);

   <T, U> Unit<T, U> just(T t);

   /**
    * How the reactive implementation handles errors.
    */
   <T, U> Unit<T, U> error(Exception e);
}
```
in resteasy-client-api, and a UnitRxInvokerImpl in resteasy-client. And then
```
public class UnitMono<U> implements Unit<Mono<U>, U> {
	
	private Mono<U> mono;
	
	public UnitMono(Mono<U> mono) {
		this.mono = mono;
	}

	@Override
	public void subscribe(Subscriber<? super U> s) {
		mono.subscribe(s);
	}

	@Override
	public Mono<U> get() {
		return mono;
	}

	@Override
	public CompletableFuture<U> toFuture() {
		return mono.toFuture();
	}
}
```
in resteasy-client-reactor-netty. Then ReactorNettyClientHttpEngine.submitRx() could return
```
...
        return new UnitMono(requestTimeout
                .map(duration -> responseMono.timeout(duration))
                .orElse(responseMono)
                .handle((response, sink) -> {
                    try {
                        sink.next(extractor.extractResult(response));
                    } catch (final Exception e) {
                        try {
                            // We release the connection instead of closing it because the WebApplicationException
                            // *may* make use of the response.  However, since we are releasing the
                            // connection here, handlers of the rethrown exception will not be able
                            // to use the stream (we ASSume buffering happened).  An alternative is
                            // to force handlers to be responsible for closing the response; however,
                            // that does open the possibility of leaks..  But then again, so do
                            // several paths you can take with the Client API.
                            response.releaseConnection();
                        } catch (final IOException ie) {
                            log.warn("There was a problem releasing the connection in an error scenario.", ie);
                        }
                        sink.error(e);
                    }
                }));
```
I haven't worked out every bit of it, but since the original Mono can be extracted from UnitMono, it seems to me that you're not losing anything ...

-Ron

crankydillo:
It feels a little weird to be trying so hard to work around the `Publisher` interface, which at the end of the day, is completely hidden from the end user, but I will give it a go.  However, I can understand the desire for a 'single value' interface in the HTTP call case..

There is some weirdness to bringing `Subscription` into your proposed `Unit` interface because subscriptions ultimately get back to the `request(Int)` problem you are trying to avoid.  I'm als concerned that subscription happens on `Unit` instance.  Typically, code that gets a `Mono` from some method call should assume they pass it up the call stack as far as possible before a subscription takes place.  However, in your proposed design, no one should ever directly call `subscribe` on the `Mono` returned by `get`.

Anyhow, I have dedicated the afternoon to make a go of it, so we'll see where we get:)

ronsigal:
Hey @crankydillo,

I've been wondering if I'm overreacting to the Publisher issue, so I consulted some people in our "Reactive Valhalla", and I got this response: 
```
Mono is a Publisher because for Reactor everything is a stream, even a single response operation or no
response operation. I consider this as an abuse of the Publisher API. While, yes, technically, you can do
that, but... imagine writing Java code only with java.lang.Object because (almost) everything is an Object?
You can quickly see that it's not great. In addition, when APIs, even when implemented using Reactor,
expose Publisher (such as R2DBC, Mongo client, Neo client), the user needs to re-create a Mono or a
Flux to use it. Except that, how do you know the number of items? There is only one way.... looking at the
code, and it may be not simple. 

Anyway, instead of introducing an abstraction, did you consider using CompletionStage? You support them
already, and you can map a Mono to a CompletionStage. The `toFuture` method from Mono does this. Note
that we have developed Reactive Converters [1] which transform Mono into CompletionStage[2].

[1] https://github.com/smallrye/smallrye-reactive-utils/tree/main/reactive-converters
[2] https://github.com/smallrye/smallrye-reactive-utils/blob/main/reactive-converters/converter-reactor/src/main/java/io/smallrye/reactive/converters/reactor/MonoConverter.java
```

So, I feel like I'm not going too far off the rails here. As for any problems with my off the cuff proposal, there's probably nothing there that I would defend too strenuously. I just wanted to convince myself that something of the sort was possible.

In the end, my goal is to incorporate something that works for you and feels right to me. I'm sure we'll get there. I'd like to include this in release 4.7.0.Final. And, again, thanks for your patience.

-Ron

crankydillo:
I was updating my comment when you were writing yours:) Apologies if I came across as harsh.  While I don't fully agree with that quote, I can see both of your points.  I never thought you were off the rails:)

Let me see what I can come up with wrt having a reactive-like single interface that doesn't have the problems you get using `CompletionStage`.

crankydillo:
I hate to sound like a broken record, but I'm still stuck.  It may be because I've gone down a path so far that I can't find my way out. The shortest way to express the situation is we have built _a lot_ around `AsyncHttpClientEngine#submitCF`, which was actually a contribution by our team.  For example, we have many internal `AsyncHttpClientEngine` wrappers that take the `CompletionStage` from the open source implementation and add functionality by making calls to `whenComplete`/`thenApply`/etc. on a `CompletionStage` delegate.  So when I added `Publisher` support, migrating this code was a simple as changing `completionStage.whenApply(..)` to `Mono.fromPublisher(delegate.submitRx()).map(...)`.   I'm probably going to have to do quite a bit of work to migrate this to some kind of non-publisher API.  Ultimately to get anything easy to work with, I likely will need your `Unit#get`, but I don't see how that will work without a bunch of casting. 

Unfortunately, this effort has been de-prioritized.  Ultimately, I picked this back up because we feel it closes a gap in some flaky tests we have.  We'd like to close that gap, but we don't (currently) feel it harms our customers.  The other features I added above also are not technically required at this time.

Setting that aside, my gut tells me that the cleanest, quickest way to rectify this may be leveraging the pattern set forth in the `MpClient`.  So basically, we'd have `MonoClient`, `MonoClientInvocation`, etc and ultimately we could end up casting `ClientInvocation` in the `MonoRxInvokerProvider` to `MonoClientInvocation`.  However, this is an awful lot of work that replaces what I've created above which AFAIK will work for any reactive client engine and any reactive RxInvoker..  I may be able to squeeze out a few more hours to PoC that.  Do you see a reason this is a bad path?

Anyhow, not trying to pressure you.  It's not your fault we have other priorities:).   However, I do want to leave you with a few thoughts.  First, I have colleagues that agree that `Publisher<Response>` is weird-ish:). As I said many times, I always understood that stance.  However, I think something to keep in mind is that I believe `Publisher` is entirely hidden from the end user.  One of the key parts of the reactive RxInvokers would be to choose the appropriate adaption.   I know examples aren't proof of a good idea, but they are better than nothing.  Here are 3 clients that expose `Publisher<Response> Client#execute(Request)` as an API:  [jetty-reactive-client](https://www.baeldung.com/jetty-reactivestreams-http-client), [rxnetty](https://github.com/ReactiveX/RxNetty/blob/a7ca9a9c1d544a79c82f7b17036daf6958bf1cd6/rxnetty-http/src/main/java/io/reactivex/netty/protocol/http/client/HttpClientRequest.java#L67) (ironically, we're trying to replace this with jaxrs2.1+a reactive invoker), and [micronaut HTTP client](https://docs.micronaut.io/latest/api/io/micronaut/http/client/HttpClient.html).

ronsigal:
Hey @crankydillo,

I'm trying to let all this sink in, and I'm feeling torn in two directions.

1. I'm starting to realize that the Publisher issue is at least half theological. I mean, I have a certain point of view in which a "degenerate" PublisherRxInvoker that only ever returns Monos is just "weird-ish" (great word), but it's not incorrect either. It seems to me that one way to go would be to keep PublisherRxInvoker and ReactiveClientHttpEngine in resteasy-client and move ReactiveInvocation and PublisherRxInvokerImpl to resteasy-client-reactor-netty. That way the weird-ish part is factored out of resteasy-client. 

Well, one problem with that is raised by your update to resteasy-rxjava2, where you have SingleRxInvokerImpl depending on PublisherRxInvokerImpl. Well, maybe having PublisherRxInvokerImpl in resteasy-client isn't unreasonable, since the weird-ish part is factored out in ReactorNettyClientHttpEngine. And, now that I've come that far, I guess the same could be said about ReactiveInvocation. So, I think, once starting down that path, I end up where you are. 

Actually, now I'm starting to wonder if ReactorNettyClientHttpEngine could be adapted to return non-unitary Publishers, but that's probably for another time. Well, to project out even further, I'm interested in implementing streaming from client to server, as in gRPC, and maybe, I don't know, ReactorNettyClientHttpEngine could be relevant. If that's something you might be interested in ... .

2. On the other hand, it still feels, deep down, like a PublisherRxInvoker that can't return "real" Publishers is sort of a bait and switch. Adopting something like Unit would make that go away. Have you made any progress in that direction?

Ok, I feel like &lt;monty-python-intonation&gt;my brain hurts&lt;/monty-python-intonation&gt;. On a personal note, I saw your latest remarks last night just before bedtime, and I wished I hadn't. ;-)  But I'm telling you that just to say that I really want to make this work as painlessly as possible, and I guess I'm not doing a great job at it. As always, thanks for your patience.

-Ron

crankydillo:
> Hey @crankydillo,
> 
> I'm trying to let all this sink in, and I'm feeling torn in two directions.
> 
>   I'm starting to realize that the Publisher issue is at least half theological. I mean, I have a certain point of view in which a "degenerate" PublisherRxInvoker that only ever returns Monos is just "weird-ish" (great word), but it's not incorrect either. It seems to me that one way to go would be to keep PublisherRxInvoker and ReactiveClientHttpEngine in resteasy-client and move ReactiveInvocation and PublisherRxInvokerImpl to resteasy-client-reactor-netty. That way the weird-ish part is factored out of resteasy-client.



If it's any consolation, I think this debate may have happened between the teams that wrote the initial reactive interface (`Publisher`).  A teammate suggested that some of those people work only with streams, while others have a concept of 'single' things.  I definitely encourage you to read a bit on the reactive stuff.  `Publisher` is (was?) primarily meant to bridge between the reactive implementations, at least that's what I've been told.  For example, these are perfectly valid:

```
rxjava2.Single<Integer> int = rxjava2.Single.fromPublisher(reactor.Flux.range(0, Integer.MAX_VALUE))
akka.Flow<Integer> ints = akka.Source.fromPublisher(rxjava2.Single.just(42).toFlowable());
```

Along those lines, would it help if I made `MonoRxInvoker` not extend from `PublisherRxInvoker`?  Given what I said above, does it not feel natural to adapt a `Publisher<T>` supplied from a `PublisherRxInvoker` into a `Mono<T>` inside a `MonoRxInvokerImpl`?  This is basically what I did in the `rxjava2.SingleRxInvoker` case..  I didn't do that in the `Mono` case, because `Mono` is a `Publisher`..

This is kind of the key point I've been trying to convey.  Basically, you take `Publisher<T>`, decide what reactive impl you want to use, and then, if that lib has it, decide to work with 'single' things, which basically boils down to 'requesting 1 item' from the publisher..  So to your friend's point about `Object`, if I could define my api like this: `Object myApi()` and then all my clients could do `UsefulThing thing = magicBridge(myApi())`, then I'm not so sure that would be a bad thing:P

This is also why I think you'll find most HTTP client libs are tightly coupled with a reactive impl (i.e. reactor-netty) will choose to use the latter types in its interface.  They know that anyone who wants to use a different reactive impl can just use an adapter.  If those didn't exist, they'd likely return `Publisher<T>`..

Again, I can understand the desire to have the return type indicate a 'single value'.  I would be curious to see debate about why that didn't get into the reactive (bridging) api (`Publisher`).

> Well, one problem with that is raised by your update to resteasy-rxjava2, where you have SingleRxInvokerImpl depending on PublisherRxInvokerImpl. Well, maybe having PublisherRxInvokerImpl in resteasy-client isn't unreasonable, since the weird-ish part is factored out in ReactorNettyClientHttpEngine. And, now that I've come that far, I guess the same could be said about ReactiveInvocation. So, I think, once starting down that path, I end up where you are.
> 
> [Actually, now I'm starting to wonder if ReactorNettyClientHttpEngine could be adapted to return non-unitary Publishers, but that's probably for another time. Well, to project out even further, I'm interested in implementing streaming from client to server, as in gRPC, and maybe, I don't know, ReactorNettyClientHttpEngine could be relevant. If that's something you might be interested in ... .]

If there was a use case, you could certainly add another method and name it such that it hints that `Publisher<T>` could deliver more than 1 value.  Presumably, this would be based off the response.  However, that seems much more the realm of an `RxInvoker`, just like you've done with server side events..
 
>    On the other hand, it still feels, deep down, like a PublisherRxInvoker that can't return "real" Publishers is sort of a bait and switch. Adopting something like Unit would make that go away. Have you made any progress in that direction?

I did make progress and opened the draft PR #2803 ; however, it seems complicated and doesn't 'feel right' to me other than this desire to avoid 'a Publisher of 1 thing'.  I am trying to integrate it with our code base, but I still have lots of failures.

Even with that, I still had to expose `Publisher<T>` from within `Unit<T>`.  Or at least I think I need to do that:)

> Ok, I feel like <monty-python-intonation>my brain hurts</monty-python-intonation>. On a personal note, I saw your latest remarks last night just before bedtime, and I wished I hadn't. ;-) But I'm telling you that just to say that I really want to make this work as painlessly as possible, and I guess I'm not doing a great job at it. As always, thanks for your patience.

If it's any consolation, I saw your comment at bed time as well.  As to patience, it is me who should be thanking you.  I need this feature from _your_ project.  I don't blame you for pushing for what you feel is right!  If I were you, I would be tempted to say 'go away' (but I hope you don't!).   My main issue is getting time to work on this:)

ronsigal:
Hi again, @crankydillo,

I feel like I've just been thinking about all this backwards. I spent some time looking more deeply into my recent parenthetic remark about extending ReactorNettyClientHttpEngine to return "real" Publishers, and now I think that, rather than trying to restrict the interface in resteasy-client to match the implementation in resteasy-client-reactor-netty, the other direction makes more sense. I've written a little test case in which a server method returns a Flux and a client recreates and returns the Flux, and I think this could be an alternative way to implement FluxRxInvokerImpl. Currently, FluxRxInvokerImpl sits on top of our SSE implementation, which happened because the SSE stuff was already there and it works. But the SSE protocol is very general, and there's overhead with field names, comments, etc., so I have a suspicion that a more direct implementation would be more efficient.

In effect, I think we could consider what you've already done as a first step towards a more general treatment. In particular, we could create a JIRA describing a future extension, and we could merge your current work as the first step.

For the first time since we started our long discussion, I feel comfortable. I'm sorry it took me so long to get here, and I hope I haven't sent you too far into the weeds, but it feels good to see the train pull into the station. [Extra points for mixed metaphors?]

1. What do you think?

2. Do you have any interest in working on Phase II? Of course, I understand if you have other priorities.

-Ron

crankydillo:
@ronsigal (hopefully) polite bump:)

resteasy-rxjava2/src/main/java/org/jboss/resteasy/rxjava2/SingleRxInvokerImpl.java
```diff
@@ -1,171 +1,189 @@
...
-      this.completionStageRxInvoker = completionStageRxInvoker;
-      this.singleProvider = new SingleProvider();
+      SinglePublisherInvoker(final ClientInvocationBuilder builder)
+      {
+         super(builder);
+      }
+
+      @Override
+      protected <T> Publisher<T> toPublisher(CompletionStage<T> completable) {
+         return Flowable.fromFuture(completable.toCompletableFuture());
```
ronsigal, line 36: Would it make sense to use Single.fromFuture() here?

ronsigal:
Hey @crankydillo,

I think I see the mountain top. Or at least a mesa.

I appreciate the introduction of UniPublisherRxInvoker and UniPublisherRxInvokerImpl, but the fact is that, as of my last comment, I had come around to thinking that PublisherRxInvoker and PublisherRxImvokerImpl weren't so bad. That is, they promise the return of a Publisher, and that's what they do. It's ReactorNettyClientHttpEngine that "falls short", so to speak. So my goal would be to expand the functionality of ReactorNettyClientHttpEngine at some point. Could you change the names back? [Don't hate me!!]

There are a few things we need to do to wrap this up.

1. We need a JIRA to cover this pull request.. 
2. We need a JIRA to cover future work on ReactorNettyClientHttpEngine, extending it to return a "real" Publisher.
3. Maybe there's room for expanding Section 51.3.8. "Reactor Netty Client Engine" of the RESTEasy User Guide. WDYT?

Do you have time for 1 and 3? If not, I'll take a stab at them, but I think you could probably do a better job. I can take care of 2.

Also, there's a simple conflict in ReactorNettyClientHttpEngine.

Thanks,
Ron

liweinan:
awesome!

crankydillo:
@ronsigal I will take care of 1, 3, and solve the merge conflict within the next few hours.  Thanks!

ronsigal:
Woo hoo!! Thanks, @crankydillo. ;-)

crankydillo:
@ronsigal I believe I did everything.  I hope the windows build failure is unrelated.  Let me know if there's anything else you need me to do.

resteasy-rxjava2/src/main/java/org/jboss/resteasy/rxjava2/SingleRxInvokerImpl.java
crankydillo: Sorry. I missed this one.  `Single#fromFuture` won't get us a `Publisher`.

ronsigal:
Oh, by the way, the test failure is probably nothing. I'll re-start the tests just to be obsessive.

crankydillo:
> Hey @crankydillo,

> Do you think you could add something to Section 51.3.8. "Reactor Netty Client Engine" of the RESTEasy User Guide explaining the benefits of the new feature. I just think that you would do a better job than I would. If the docbook details are a barrier, you could just do it in plain text and I'll translate it.

Ugh, I forgot about that.  Will do!

ronsigal:
Hey @crankydillo,

>a feature that you must support. 

Hmmmm. That's interesting. Ever since RESTEasy 4.0.0.Final, we've distinguished between public and non-public modules. E.g., resteasy-client-api and resteasy-core-spi are public, and resteasy-client and resteasy-core aren't. The idea is that resteasy-client and resteasy-core hold implementation code, which is subject to change. 

I guess my first question is: In what sense do you consider this new feature to be experimental? Given that this is something you intend to use, I was under the impression that it's something a little more than experimental ... . Is it something that you can see going away at some point?

Beyond that, my reaction is manifold:

1. I'm thinking that PublisherRxInvoker really ought to be in resteasy-client-api. I should have thought of that before.
2. PublisherRxInvokerImpl is in resteasy-client, so that's safe.
3. The use of PublisherRxInvokerImpl shows up in MonoRxInvokerImpl. I guess this is somewhat fuzzy, and maybe we (RESTEasy) should think some more about segregating MonoRxInvoker and MonoRxInvokerImpl. But the behavior behind the scenes is in PublisherRxInvokerImpl, so it's an implementation detail. 

Hmmm. Well, I was thinking that the advantages of the new implementation that you've discussed would be worth making public. But then, as you say, we are, in effect, making the behavior of the implementation public.

I guess it comes down to how stable the new behavior is. So that takes me back to my first question: "In what sense do you consider this new feature to be experimental?"

-Ron

crankydillo:
> Hey @crankydillo,
> 
> > a feature that you must support.
> 
> Hmmmm. That's interesting. Ever since RESTEasy 4.0.0.Final, we've distinguished between public and non-public modules. E.g., resteasy-client-api and resteasy-core-spi are public, and resteasy-client and resteasy-core aren't. The idea is that resteasy-client and resteasy-core hold implementation code, which is subject to change.
> 
> I guess my first question is: In what sense do you consider this new feature to be experimental? Given that this is something you intend to use, I was under the impression that it's something a little more than experimental ... . Is it something that you can see going away at some point?
> 
> Beyond that, my reaction is manifold:
> 
>     1. I'm thinking that PublisherRxInvoker really ought to be in resteasy-client-api. I should have thought of that before.
> 
>     2. PublisherRxInvokerImpl is in resteasy-client, so that's safe.
> 
>     3. The use of PublisherRxInvokerImpl shows up in MonoRxInvokerImpl. I guess this is somewhat fuzzy, and maybe we (RESTEasy) should think some more about segregating MonoRxInvoker and MonoRxInvokerImpl. But the behavior behind the scenes is in PublisherRxInvokerImpl, so it's an implementation detail.
> 
> 
> Hmmm. Well, I was thinking that the advantages of the new implementation that you've discussed would be worth making public. But then, as you say, we are, in effect, making the behavior of the implementation public.
> 
> I guess it comes down to how stable the new behavior is. So that takes me back to my first question: "In what sense do you consider this new feature to be experimental?"
> 
> -Ron

RE this being experimental.  I'm not entirely sure what to say.  We have high hopes to use the features this will add.  So I guess we'll know once this is in.  We do have some proof that we need these features.  My initial reactor to your request for documented features is that I hate to do that without tests; however, I forgot that I already had a test that shows how reactor's context can be passed down into an engine.  I have augmented one of the tests to show that a business logic timeout will cause an eager disconnect on the client connection.  So..I've covered 2 of the 3 benefits I described.  I'll add those to the document.

Do you want me to move `PublisherRxInvoker`?

crankydillo:
@ronsigal Please see my last 2 commits.

ronsigal:
Hey @crankydillo,

Ok, I think you've added something nice to RESTEasy, something that is consistent with the general movement towards reactive programming, so I think we can bet that it's more than experimental. And I think it's worth letting people know about it. So, +1 for documentation.

Given that the definition of PublisherRxInvoker is determined by the definition of RxInvoker, I think it's safe to move it to resteasy-client-api.

Cool.

ronsigal:
I've extended the User Guide discussion a bit by borrowing some language from RESTEASY-2943. Let me know if any of it should be changed or could be imporved.

docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
```diff
@@ -1059,8 +1059,25 @@ ResteasyClient client = ((ResteasyClientBuilder)ClientBuilder.newBuilder())
         </programlisting>
         <para>
             When coupled with the MonoRxInvoker, this has several benefits.
-            Most notably, a Reactor-based timeout will eagerly stop in-flight
-            requests.
+            It supports things like this:
```
crankydillo, line 1062: I like this change.

As an aside, I'm not sure how you got this into my PR.  I wasn't aware anyone could push to my repo:P  But.. you are welcome to! :)

crankydillo:
@ronsigal I moved `PublisherRxInvoker`.

docbook/reference/en/en-US/modules/RESTEasy_Client_Framework.xml
ronsigal: > I'm not sure how you got this into my PR

I guess I have some extra administrative powers because I can edit a file and commit it right on the spot. It's very convenient. I'm not sure what would happen if I tried to push to your repo. Sometimes I've been able to do that, and sometimes not.

Old Lodge Skins: Am I still in this world?

Jack Crabb: Yes, Grandfather.

Old Lodge Skins: [groans] I was afraid of that. Well, sometimes the magic works. Sometimes, it doesn't.

ronsigal:
Well, I think we're there. I feel like there should be some kind of button pushing or ribbon cutting ceremony. Possibly on CNN.

@crankydillo, it's been a pleasure working with you. Like Ernie Banks, Chicago Cubs infielder said, "Let's play two." ;-)

crankydillo:
> Well, I think we're there. I feel like there should be some kind of button pushing or ribbon cutting ceremony. Possibly on CNN.
> 
> @crankydillo, it's been a pleasure working with you. Like Ernie Banks, Chicago Cubs infielder said, "Let's play two." ;-)

Likewise.  Really appreciate you taking the time!

PR #2466, opened 2020-07-06
Resteasy 2618 mp rest client query param style 2 nd

rsearls:


asoldano:
@rsearls , please rebase, I've just merged your other pr for MP rest client and this is how conflicting

PR #2467, opened 2020-07-08
[RESTEASY-2642] Add doPrivs to sys prop checks

andymc12:
This PR wraps calls to `System.getProperty(...)` and `Integer.getInteger(...)` in doPriv blocks to ensure that they will work with a Java 2 security manager enabled.

### DIFF
--- a/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -43,7 +43,9 @@ import java.net.ProxySelector;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.security.AccessController;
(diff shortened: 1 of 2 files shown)


asoldano:
Besides that, as this would be your first contribution, we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 . Can you explicitly agree to that here?

andymc12:
>  Now I'm wondering if, for performance sake, you would consider checking the availability of the security manager (System.getSecurityManager() != null) to avoid creating the privileged actions when not needed. WDYT?

No problem.

> Besides that, as this would be your first contribution, we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 . Can you explicitly agree to that here?

I agree to contribute under the terms of the ASL 2.0 license.

Please let me know if there is anything else I need to do in order to contribute.  I'm glad to know that RESTEasy uses ASL 2.0 - I've always been a fan of the Apache license.  Thanks!

PR #2471, opened 2020-07-13
[RESTEASY-2646]:Fix match cache in  RootNode grows infinitely

jimma:


resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
```diff
@@ -26,7 +26,11 @@ public Key(final HttpRequest request, final int start) {
             this.path = start == 0 ? matchingPath : matchingPath.substring(start);
             this.start = start;
             this.method = request.getHttpMethod();
-            this.contentType = request.getHttpHeaders().getMediaType();
+            if (request.getHttpHeaders().getMediaType().getParameters().containsKey("boundary")) {
```
patriot1burke, line 29: You will never get a cache hit with "boundary" as equals() still checks parameters being equal.  Maybe just not cache at all if there is a "boundary"?

jimma: From my local test, equals() doesn't check parameters as Key's contentTypes are all changed with "boundary" removal in this line:
if (request.getHttpHeaders().getMediaType().getParameters().containsKey("boundary")) {
    this.contentType = new MediaType(request.getHttpHeaders().getMediaType().getType(), request.getHttpHeaders().getMediaType().getSubtype());
}

For the following  two request Content-Types, it will create Key with the same contentType : multipart/form-data
```
Content-type: multipart/form-data; boundary=50ccecb8-085c-4869-841f-b40f8d77fe23
Content-type: multipart/form-data; boundary=ad96d758-8668-4657-b8ca-671a5fbb9269
```

ben-manes: I'm sorry, but this is too idealistic. This bug occurs on any request as the media type can be parameterized. A simple example is using [server-async-http](https://github.com/resteasy/resteasy-examples/tree/master/jaxrs-2.0/server-async-http).

```
curl 'http://localhost:8080/chat/listener/ping' --header 'Content-Type: text/plain; 1=1'
curl 'http://localhost:8080/chat/listener/ping' --header 'Content-Type: text/plain; 1=2'
```

A safe fix would be to include all fields in both equals and hashCode. The maximum cache size nicely resolves the concern if there are other unknown exploits, since it limits the worst case cost.

patriot1burke: You definitely do not need to include parameters within the hash code.  Hash codes do not have to be unique and there would only be possible collisions if you had two different resource methods that were dispatched based on media type parameters, i.e. "application/foo+json;version=1".

You DO want equals to match based on parameters, but not any dynamic parameters like boundary.  IMO, the fix is fine.

patriot1burke: Parameter types are compared here:

https://github.com/resteasy/Resteasy/blob/91036412da9ae08c3959d7b4ec33ddc2a584c0d8/resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java#L45

ben-manes: The maximum cache size is certainly the major fix, but without that the simple dynamic parameter of a counter is all that was needed to demonstrate a HashDoS attack. The same resource method was being dispatched to, but since `contentType` isn't in the hashCode mix it results in a collision and degrading to O(n) map operations.

Using a naive JMeter test where every request increments a counter in the dynamic parameter, hitting the same resource, showed latencies degrade as

response_ok
response_dos

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/RootNode.java
```diff
@@ -56,10 +69,12 @@ public ResourceInvoker match(HttpRequest request, int start)
          request.setAttribute(RESTEASY_CHOSEN_ACCEPT, match.chosen);
       } else {
          match = root.match(request, start);
-         if (match.match != null && match.match.expression.getNumGroups() == 0 && match.invoker instanceof ResourceMethodInvoker) {
+         if (cache.size() < CACHE_SIZE && match.match != null && match.match.expression.getNumGroups() == 0 && match.invoker instanceof ResourceMethodInvoker) {
             //System.out.println("*** caching: " + key.method + " " + key.path);
             match.match = null;
-            cache.putIfAbsent(key, match);
+            {
```
rsvoboda, line 75: Why are here these extra braces?

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/RootNode.java
```diff
@@ -56,10 +69,12 @@ public ResourceInvoker match(HttpRequest request, int start)
          request.setAttribute(RESTEASY_CHOSEN_ACCEPT, match.chosen);
       } else {
          match = root.match(request, start);
-         if (match.match != null && match.match.expression.getNumGroups() == 0 && match.invoker instanceof ResourceMethodInvoker) {
+         if (cache.size() < CACHE_SIZE && match.match != null && match.match.expression.getNumGroups() == 0 && match.invoker instanceof ResourceMethodInvoker) {
```
rsvoboda, line 72: So once the cache is full, nothing gets added / old entries removed?

rsvoboda:
@jimma Do you plan to include test case?

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/RootNode.java
jimma: It's ideal that we remove the LRU entry when the cache is full.  Later we can look at [ConcurrentLinkedHashMap](https://github.com/ben-manes/concurrentlinkedhashmap)  which Ben suggested.

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
rsvoboda: @ben-manes can you share sample project and jmeter config for ^^^?

jimma:
@rsvoboda  I didn't figure out how to write a test case for this issue. Do you have any idea ?

rsvoboda:
@jimma nothing simple what could be added into the RESTEasy testsuite.
Maybe just small check on `cache` size or several calls for `public ResourceInvoker match(HttpRequest request, int start)`. But it would probably need to involve Byteman

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/RootNode.java
patriot1burke: Just remember, that 99% of apps will only have a handful of media types (< 5), usually only 1 (json).  Might be a little bigger if character sets are involved, but there's only a handful of those in existence as well.  The issue here was a media type that adds a random parameter.  IMO, if you hit the cache size you're dealing with an app that has another random parameter and the cache will be useless.  Might as well clear it and turn it off in that scenario.

rsvoboda:
You can do this:
```java
    public class MyRootNode extends RootNode {
        public int cacheSize() {
            return cache.size();
        }
    }
```
and run 2050x myRootNode.match(...) and check that myRootNode.cacheSize() returns 2048.

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
rsvoboda, line 29: is `request.getHttpHeaders().getMediaType()` always `not null` ? Just to be sure no NPE can jump from this.

asoldano: Actually the build is failing in resteasy-undertow testsuite because a NPE exactly there. So this has to be fixed @jimma .

ben-manes: @rsvoboda Attached to [RESTEASY-2643](https://issues.redhat.com/browse/RESTEASY-2643) as unsupported type on github. Included the steps, as sent to RedHat security for their verification. This increments a dynamic parameter on every request, `counter=${counter_value}`.

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/RootNode.java
ben-manes: Isn't this per route? The cache size would be hit if an application has more than `CACHE_SIZE` routes, in the expected case.

The `MediaTypeHeaderDelegate` is a cache of media types and it clears itself when full. It doesn't condition for dynamic parameters, so a multipart route would invalidate it regularly.

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
```diff
@@ -26,7 +26,11 @@ public Key(final HttpRequest request, final int start) {
...
+               this.contentType = new MediaType(request.getHttpHeaders().getMediaType().getType(), request.getHttpHeaders().getMediaType().getSubtype());
```
ronsigal, line 30: MediaType.equals() checks the parameters, so I question removing **all** of the parameters when "boundary" is present. Actually, I guess, in principle, that it's possible to use the boundary parameter in a @Consumes annotation, but it doesn't seem likely. So it seems reasonable to me to remove only the "boundary" parameter and leave the others in place.

ronsigal:
As for managing overflow, LRU might be the appropriate strategy, but Bill's idea of just zapping the cache and starting over would have a similar effect. 

1. There should be an entry in the User Guide about RESTEASY_MATCH_CACHE_ENABLED
2. We should add a note about random parameters in the User Guide and the idea of turning off the cache.

ronsigal:
I hate to mention this, but as of WF 20, the jaxrs module allows setting RESTEasy parameters. So ... technically ... we should update jaxrs. I **HOPE** that **doesn't** count as a Feature.

rsvoboda:
@ronsigal WF is using 3.x series - https://github.com/wildfly/wildfly/blob/master/pom.xml#L392
I don't think there is need to do any work on WF side because of this parameter

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
jimma: Fixed.

rsvoboda:
I have prepared RootNodeCacheSizeTest https://github.com/rsvoboda/Resteasy/commit/0406ac99b31d9383db976e33cf37fb7dc5e11362 in my https://github.com/rsvoboda/Resteasy/tree/RootNodeCacheSizeTest branch

If you like it, feel free to cherry pick it and include into this PR.

asoldano:
> I have prepared RootNodeCacheSizeTest [rsvoboda@0406ac9](https://github.com/rsvoboda/Resteasy/commit/0406ac99b31d9383db976e33cf37fb7dc5e11362) in my https://github.com/rsvoboda/Resteasy/tree/RootNodeCacheSizeTest branch
> 
> If you like it, feel free to cherry pick it and include into this PR.

Thanks @rsvoboda . Please @jimma include this.

asoldano:
> I hate to mention this, but as of WF 20, the jaxrs module allows setting RESTEasy parameters. So ... technically ... we should update jaxrs. I **HOPE** that **doesn't** count as a Feature.

As @rsvoboda commented, WildFly is not a concern atm, as this issue affects RESTEasy 4.x series only.

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
jimma: @ronsigal  @patriot1burke The "boundary" parameter check and remove all parameters still can't stop the cache to be invalidated by GET request with ContentType=text/html;version=1~2500.
I think we can directly disable this cache when the request ConteType contains parameter "boundary" or other items to make it less vulnerable. I already sent my change. Please review again.

jimma:
> As for managing overflow, LRU might be the appropriate strategy, but Bill's idea of just zapping the cache and starting over would have a similar effect.

+1. I already changed to clear cache when it's full and start over.

> 
> 1. There should be an entry in the User Guide about RESTEASY_MATCH_CACHE_ENABLED
> 2. We should add a note about random parameters in the User Guide and the idea of turning off the cache.

I'll add it to the documentation later.

jimma:
> I have prepared RootNodeCacheSizeTest [rsvoboda@0406ac9](https://github.com/rsvoboda/Resteasy/commit/0406ac99b31d9383db976e33cf37fb7dc5e11362) in my https://github.com/rsvoboda/Resteasy/tree/RootNodeCacheSizeTest branch
> 
> If you like it, feel free to cherry pick it and include into this PR.
I already cherry-picked, Thanks. I did some change in the test after I improved something as described in this comment : https://github.com/resteasy/Resteasy/pull/2471#discussion_r454886626 .

jimma:
> @jimma Shouldn't be the cache size at this phase equal to 2 (2050-2048)?

oops. Missed one commit change. Fixed.

michalszynkiewicz:
the issue number sems wrong. Is it for RESTEASY-2643 ?

rsvoboda:
> the issue number sems wrong. Is it for RESTEASY-2643 ?

both are correct :)

ronsigal:
re: "WF is using 3.x series"

D'oh!!

resteasy-core/src/main/java/org/jboss/resteasy/core/registry/MatchCache.java
ronsigal: I see - that's the change in RootNode which uses the cache only when there are no parameters. That seems like a reasonable heuristic.

ronsigal:
@asoldano, it's fine with me. I would also add documentation to the TODO list.

ben-manes:
Thank you for the prompt fix. Any timeline on when a release will be made?

PR #2474, opened 2020-07-15
RESTEASY-2650: Use web.xml servlets for app path

andymc12:
Also handle single class that is both an App and a resource class.

Resolves [RESTEASY-2650](https://issues.redhat.com/projects/RESTEASY/issues/RESTEASY-2650)

### DIFF
--- a/resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
+++ b/resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
@@ -26,6 +26,8 @@ import java.util.Set;
 @HandlesTypes({Application.class, Path.class, Provider.class})
 public class ResteasyServletInitializer implements ServletContainerInitializer
 {
+   private static final String RESTEASY_MAPPING_PREFIX = "resteasy.servlet.mapping.prefix";
+   private static final String APPLICATION = "javax.ws.rs.Application";
    static final Set<String> ignoredPackages = new HashSet<String>();
 
    static
@@ -36,41 +38,29 @@ public class ResteasyServletInitializer implements ServletContainerInitializer
    @Override
    public void onStartup(Set<Class<?>> classes, ServletContext servletContext) throws ServletException
    {
-      if (classes == null || classes.size() == 0) return;
-      for (ServletRegistration reg : servletContext.getServletRegistrations().values())
-      {
-         if (reg.getInitParameter("javax.ws.rs.Application") != null)
-         {
-            return; // there's already a servlet mapping, do nothing
-         }
-      }
-
+      if (classes == null || classes.size() == 0)
+         return;
       Set<Class<?>> appClasses = new HashSet<Class<?>>();
       Set<Class<?>> providers = new HashSet<Class<?>>();
       Set<Class<?>> resources = new HashSet<Class<?>>();
 
       for (Class<?> clazz : classes)
       {
-         if (clazz.isInterface() || ignoredPackages.contains(clazz.getPackage().getName())) continue;
+         if (clazz.isInterface() || ignoredPackages.contains(clazz.getPackage().getName()))
+            continue;
          if (clazz.isAnnotationPresent(Path.class))
-         {
             resources.add(clazz);
-         }
-         else if (clazz.isAnnotationPresent(Provider.class))
-         {
+         if (clazz.isAnnotationPresent(Provider.class))
             providers.add(clazz);
-         }
-         else
-         {
+         if (Application.class.isAssignableFrom(clazz))
             appClasses.add(clazz);
-         }
       }
       if (appClasses.size() == 0 && resources.size() == 0) return;
 
       if (appClasses.size() == 0)
       {
          // todo make sure we can do this on all servlet containers
-         //handleNoApplicationClass(providers, resources, servletContext);
+         // handleNoApplicationClass(providers, resources, servletContext);
          return;
       }
 
@@ -95,51 +85,91 @@ public class ResteasyServletInitializer implements ServletContainerInitializer
 
    }
 
+   private Set<ServletRegistration> getServletsForApplication(Class<?> applicationClass, ServletContext servletContext)
+   {
+      Set<ServletRegistration> set = new HashSet<>();
+      ServletRegistration reg = servletContext.getServletRegistration(applicationClass.getName());
+      if (reg != null && reg.getMappings().size() == 1)
+         set.add(reg);
+
+      for (ServletRegistration sr : servletContext.getServletRegistrations().values())
+      {
+         String appClassName = sr.getInitParameter(APPLICATION);
+         if (applicationClass.getName().equals(appClassName) && sr.getMappings().size() == 1)
+            set.add(sr);
+      }
+      return set;
+   }
 
    protected void register(Class<?> applicationClass, Set<Class<?>> providers, Set<Class<?>> resources, ServletContext servletContext)
    {
+      Set<ServletRegistration> servletsForApp = getServletsForApplication(applicationClass, servletContext);
+      // ignore @ApplicationPath if application is already mapped in web.xml
+      if (!servletsForApp.isEmpty())
+      {
+         for (ServletRegistration servletReg : servletsForApp)
+         {
+            String servletClassName = servletReg.getClassName();
+            if (servletClassName == null)
+               servletContext.addServlet(servletReg.getName(), HttpServlet30Dispatcher.class);
+            String prefix = servletReg.getMappings().iterator().next();
+            if (prefix.endsWith("*"))
+               prefix = prefix.substring(0, prefix.length() - 1);
+            if (prefix.length() > 1 && prefix.endsWith("/"))
+               prefix = prefix.substring(0, prefix.length() - 1);
+            registerResourcesAndProviders(servletReg, providers, resources, applicationClass, prefix);
+         }
+         return;
+      }
       ApplicationPath path = applicationClass.getAnnotation(ApplicationPath.class);
       if (path == null)
       {
-         // todo we don't support this yet, i'm not sure if partial deployments are supported in all servlet containers
+         // Application subclass has no @ApplicationPath and no declared mappings to use
+         // TODO: add debug message indicating that an Application was detected but could
+         // not be mapped
          return;
       }
-      ServletRegistration.Dynamic reg = servletContext.addServlet(applicationClass.getName(), HttpServlet30Dispatcher.class);
-      reg.setLoadOnStartup(1);
-      reg.setAsyncSupported(true);
-      reg.setInitParameter("javax.ws.rs.Application", applicationClass.getName());
+      ServletRegistration.Dynamic reg;
+      String mapping = path.value();
+      String prefix;
 
-      if (path != null)
+      if (!mapping.startsWith("/"))
+         mapping = "/" + mapping;
+      prefix = mapping;
+      if (!prefix.equals("/") && prefix.endsWith("/"))
+         prefix = prefix.substring(0, prefix.length() - 1);
+      if (!mapping.endsWith("/*"))
       {
-         String mapping = path.value();
-         if (!mapping.startsWith("/")) mapping = "/" + mapping;
-         String prefix = mapping;
-         if (!prefix.equals("/") && prefix.endsWith("/")) prefix = prefix.substring(0, prefix.length() - 1);
-         if (!mapping.endsWith("/*"))
-         {
-            if (mapping.endsWith("/")) mapping += "*";
-            else mapping += "/*";
-         }
-         // resteasy.servlet.mapping.prefix
-         reg.setInitParameter("resteasy.servlet.mapping.prefix", prefix);
-         reg.addMapping(mapping);
+         if (mapping.endsWith("/"))
+            mapping += "*";
+         else
+            mapping += "/*";
       }
 
+      reg = servletContext.addServlet(applicationClass.getName(), HttpServlet30Dispatcher.class);
+      reg.setLoadOnStartup(1);
+      reg.setAsyncSupported(true);
+      reg.addMapping(mapping);
+
+      registerResourcesAndProviders(reg, providers, resources, applicationClass, prefix);
+   }
+
+   private void registerResourcesAndProviders(ServletRegistration reg, Set<Class<?>> providers, Set<Class<?>> resources,
+         Class<?> appClass, String prefix)
+   {
+      reg.setInitParameter(APPLICATION, appClass.getName());
+      // resteasy.servlet.mapping.prefix
+      reg.setInitParameter(RESTEASY_MAPPING_PREFIX, prefix);
       if (resources.size() > 0)
       {
          StringBuilder builder = new StringBuilder();
          boolean first = true;
-         for (Class resource : resources)
+         for (Class<?> resource : resources)
          {
             if (first)
-            {
                first = false;
-            }
             else
-            {
                builder.append(",");
-            }
-
             builder.append(resource.getName());
          }
          reg.setInitParameter(ResteasyContextParameters.RESTEASY_SCANNED_RESOURCES, builder.toString());
@@ -148,20 +178,15 @@ public class ResteasyServletInitializer implements ServletContainerInitializer
       {
          StringBuilder builder = new StringBuilder();
          boolean first = true;
-         for (Class provider : providers)
+         for (Class<?> provider : providers)
          {
             if (first)
-            {
                first = false;
-            }
             else
-            {
                builder.append(",");
-            }
             builder.append(provider.getName());
          }
          reg.setInitParameter(ResteasyContextParameters.RESTEASY_SCANNED_PROVIDERS, builder.toString());
       }
-
    }
 }
--- a/resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
+++ b/resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
@@ -50,9 +50,9 @@ public class ResteasyServletInitializer implements ServletContainerInitializer
             continue;
          if (clazz.isAnnotationPresent(Path.class))
             resources.add(clazz);
-         if (clazz.isAnnotationPresent(Provider.class))
+         else if (clazz.isAnnotationPresent(Provider.class))
             providers.add(clazz);
-         if (Application.class.isAssignableFrom(clazz))
+         else if (Application.class.isAssignableFrom(clazz))
             appClasses.add(clazz);
       }
       if (appClasses.size() == 0 && resources.size() == 0) return;
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ target
 **/.checkstyle
 **/.factorypath
 
+# VS Code
+###################
+**/.vscode/
 
 # Packages #
 ############
--- a/resteasy-servlet-initializer/pom.xml
+++ b/resteasy-servlet-initializer/pom.xml
@@ -14,11 +14,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>resteasy-core</artifactId>
             <version>${project.version}</version>
@@ -44,6 +39,18 @@
         <dependency>
             <groupId>org.jboss.logging</groupId>
             <artifactId>jboss-logging-processor</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.4.0</version>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/resteasy-servlet-initializer/src/test/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializerTest.java
+++ b/resteasy-servlet-initializer/src/test/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializerTest.java
@@ -16,7 +16,6 @@ import javax.servlet.ServletRegistration;
 
 import org.jboss.resteasy.plugins.server.servlet.HttpServlet30Dispatcher;
 import org.jboss.resteasy.plugins.server.servlet.ResteasyContextParameters;
-import org.jboss.resteasy.plugins.servlet.testapp.AppAndResource;
 import org.jboss.resteasy.plugins.servlet.testapp.AppWithAppPath;
 import org.jboss.resteasy.plugins.servlet.testapp.AppWithNoAppPath;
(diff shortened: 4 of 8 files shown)


andymc12:
@ronsigal @asoldano Could one or both of you review this PR?  If you have any questions or concerns, please let me know.  Thanks!

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
```diff
@@ -99,69 +91,82 @@ protected void handleNoApplicationClass(Set<Class<?>> providers, Set<Class<?>> r
...
+              return;
+          }
+          reg = servletContext.addServlet(applicationClass.getName(), HttpServlet30Dispatcher.class);
+          Collection<String> mappings = reg.getMappings();
+          if (mappings == null || mappings.isEmpty())
+          {
+              // no declared mappings
+              return;
+          }
+          mapping = mappings.iterator().next();
```
ronsigal, line 111: Related to my question on RESTEASY-2650, about using the same Application on two different paths, wouldn't you want to add a servlet for each path? Maybe I'm on the wrong track here.

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
```diff
@@ -99,69 +91,82 @@ protected void handleNoApplicationClass(Set<Class<?>> providers, Set<Class<?>> r
...
       if (path == null)
       {
-         // todo we don't support this yet, i'm not sure if partial deployments are supported in all servlet containers
-         return;
+          if (servletContext.getServletRegistration(applicationClass.getName()) == null)
+          {
+              // Application subclass has no @ApplicationPath and no declared mappings to use
+              return;
+          }
+          reg = servletContext.addServlet(applicationClass.getName(), HttpServlet30Dispatcher.class);
```
ronsigal, line 104: Shouldn't this go after the test "if (mappings == null || mappings.isEmpty())"?

andymc12:
@ronsigal I updated the PR - it fixes the issues you mentioned - and in general should be more complete.  If a web.xml defines a servlet for an application (either by specifying the application class as the servlet name or by specifying the application as an init-parameter with key of `javax.ws.rs.Application`), then that servlet's mapping will be used.  If no servlet is specified for an application in the web.xml, then this code will dynamically add one to handle it.  

Can you take another look and let me know if you have any questions or concerns?  Thanks!

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
```diff
@@ -95,51 +85,91 @@ protected void handleNoApplicationClass(Set<Class<?>> providers, Set<Class<?>> r
 
    }
 
+   private Set<ServletRegistration> getServletsForApplication(Class<?> applicationClass, ServletContext servletContext)
+   {
+      Set<ServletRegistration> set = new HashSet<>();
+      ServletRegistration reg = servletContext.getServletRegistration(applicationClass.getName());
+      if (reg != null && reg.getMappings().size() == 1)
```
ronsigal, line 92: I'm just speculating here, not objecting.

In the JIRA you mentioned having two different servlets for the same Application. I suppose, in general, you would have two explicit <servlet/> declarations, but I gather this is legal:
```
<web-app>
   <servlet-mapping>
      <servlet-name>org.example.MyApplication</servlet-name>
      <url-pattern>/myresources/*</url-pattern>
      <url-pattern>/secure/myresources/*</url-pattern>
   </servlet-mapping>
</web-app>
```
The JAX-RS spec doesn't seem to say anything about that, so I guess it would be a new feature. Is it worth doing for RESTEasy 4.x?

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
```diff
@@ -95,51 +85,91 @@ protected void handleNoApplicationClass(Set<Class<?>> providers, Set<Class<?>> r
...
+   {
+      Set<ServletRegistration> set = new HashSet<>();
+      ServletRegistration reg = servletContext.getServletRegistration(applicationClass.getName());
+      if (reg != null && reg.getMappings().size() == 1)
+         set.add(reg);
+
+      for (ServletRegistration sr : servletContext.getServletRegistrations().values())
+      {
+         String appClassName = sr.getInitParameter(APPLICATION);
+         if (applicationClass.getName().equals(appClassName) && sr.getMappings().size() == 1)
```
ronsigal, line 98: Same comment.

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
```diff
@@ -95,51 +85,91 @@ protected void handleNoApplicationClass(Set<Class<?>> providers, Set<Class<?>> r
...
+            String prefix = servletReg.getMappings().iterator().next();
+            if (prefix.endsWith("*"))
+               prefix = prefix.substring(0, prefix.length() - 1);
+            if (prefix.length() > 1 && prefix.endsWith("/"))
+               prefix = prefix.substring(0, prefix.length() - 1);
+            registerResourcesAndProviders(servletReg, providers, resources, applicationClass, prefix);
+         }
+         return;
+      }
       ApplicationPath path = applicationClass.getAnnotation(ApplicationPath.class);
```
ronsigal, line 124: The spec says
```
If the Application subclass is annotated with @ApplicationPath, implementations are REQUIRED to use
the value of this annotation appended with ”/*” to define a mapping for the added server.  Otherwise,
the application MUST be packaged with a web.xml that specifies a servlet mapping. 
```
Shouldn't this bit go first?

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
andymc12: I've never seen that done before, but I think it is a good idea.  There might need to be more corresponding changes to account for different mapped paths though.  I can open up a new feature JIRA for allowing a single app to be mapped to multiple paths without needing multiple servlets.  Good idea!

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
andymc12: Agreed.  My original reading of that led me to think that `@ApplicationPath` would be overridden if it was specified in the web.xml, but on further reflection I think your interpretation is correct.

In general, I don't like that approach as it means that a deployer has no way of changing the app path without re-coding/re-building the app. But in this case, it looks like that would be required.  I'll re-code this section.

Thanks for catching this!

andymc12: I added a new commit to ensure that the `@ApplicationPath` path is not overridden.

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
andymc12: > The JAX-RS spec doesn't seem to say anything about that, so I guess it would be a new feature. Is it worth doing for RESTEasy 4.x?

yes, we're currently planning to use RESTEasy 4.5.X in Open Liberty.

resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
ronsigal: At times like this I think of the song from Fiddler on the Roof: "Posing questions that would cross a rabbi's eyes". ;-)
  
The first part of register(),
```
      if (!servletsForApp.isEmpty())
      {
```
handles the case where a servlet is declared in web.xml for the Application. The spec says
```
If there is already a servlet that handles this application. That is, a servlet that has an initialization
parameter named  javax.ws.rs.Application whose value is the fully qualified name of the 
Application subclass, then no additional configuration steps are required by the JAX-RS implementation
```
That suggests that it's illegal to have a &lt;servlet&gt; element without a matching &lt;servlet-mapping&gt; element., although I haven't checked that. Also, it's saying, in effect, that @ApplicationPath has no role to play.

My interpretation is that the return statement you took out was correct. That is, if there IS a servlet, just use the servlet and matching &lt;servlet-mapping&gt; and return. Maybe it's worth checking for the existence of a &lt;servlet-mapping&gt;. It sounds like the JAX-RS spec is saying there must be a &lt;servlet-mapping&gt;, so it would be an error if there is none. But maybe that should get caught by the servlet container. I haven't checked that.

On the other hand, the spec says
```
If no servlet handles this  application, JAX-RS  implementations are REQUIRED  to dynamically
add a servlet whose fully qualified name must be that of the Application subclass.  If the Application
subclass is annotated with @ApplicationPath, implementations are REQUIRED to use the value of 
this annotation appended with ”/*” to define a mapping for the added server.  Otherwise, the application
MUST be packaged with a web.xml that specifies a servlet mapping. 
```
That suggests that it's legal to have a &lt;servlet-mapping&gt; element without a matching &lt;servlet&gt; element. So, if there's no declared servlet, and the Application doesn't have an @ApplicationPath, it sounds like we should then look for a &lt;servlet-mapping&gt; instead of giving up.

WDYT?

andymc12: And "if I were a rich man" then I could just retire and leave this issue to somebody else!  :-)

Ok, that spec text aligns with my previous understanding.  Use `@ApplicationPath` path if the path is not specified in the web.xml.  If it is specified in the web.xml, then ignore the `@ApplicationPath` path.  So, first off, I think I need to remove my last commit.

It looks like it is possible for a servlet to be defined in the web.xml without a mapping ( see https://stackoverflow.com/questions/12602805/can-we-invoke-a-servlet-without-servlet-mapping-in-web-xml-entry ).  The spec says that if an Application class is handled by a servlet, then no further configuration is necessary - unfortunately, it doesn't say one way or the other about whether the servlet must be mapped or not...

So, in the case where the servlet is mapped, I think we're fine.  None of my code is destructive if the user has specified things like the servlet class to handle it, etc. It does add RESTEASY init parameters (for scanned classes, etc.) - I don't think this should be a problem, but if we want to absolutely sure we could check that the servlet class name is a RESTEasy servlet before setting those parameters.

In the case where the application is handled by a servlet in the web.xml but does not have a mapping - I tend to think this is a bug - or best case using the jaws of life to remove a staple (using a JAX-RS application just to run something at app startup when a simple servlet would do).  The code as it is now (and as it will be after I remove the latest commit) would map the application to the `@ApplicationPath` path.  I think that this makes sense since we have no other path to use.  Another option would be to ignore the application and issue a warning to the user indicating that they should either delete the unmapped servlet from the web.xml or they should map it in the web.xml.  I'd be fine with either approach.

So:
1) Should we check the servlet class name to be sure it is a RESTEasy servlet before adding RESTEasy init parameters?
2) Should we ignore an application in the unmapped-servlet case and issue a warning?  Or default to the `@ApplicationPath` path?

and whichever way we go on 1 and 2, I will plan to revert the commit that removes the early return.

Thanks again!

ronsigal: re: "I could just retire and leave this issue to somebody else"

Lolll. Don't tempt me!!

I will think about this some more tomorrow. :)

ronsigal: This looks different every time I look at it. Holy Rashoman, Batman!!

I'm going to try to write out all of the possibilities, for my own understanding.

I. No Application

We MUST create a servlet, and there MUST be a &lt;servlet-mapping&gt; in the web.xml.  Technically, the Servlet spec would be ok without the &lt;servlet-mapping&gt;, but JAX-RS considers that an error.

II. Application exists

A. Matching servlet exists

The JAX-RS spec says, "no additional configuration steps are **required** [my emphasis]". However, the Servlet spec allows for the absence of a &lt;servlet-mapping&gt;, and JAX-RS doesn't FORBID using an @ApplicationPath path if it exists. In fact, it doesn't FORBID using "/\*" if @ApplicationPath doesn't exist. So that's different from my previous understanding. Actually, I think assuming we can use "/\*" as a default is probably a bad idea. It's probably better to consider it an error if there's no &lt;servlet-mapping&gt; and no @ApplicationPath.

B. Matching servlet doesn't exist

We MUST create a servlet. If @ApplicationPath exists, we MUST use it. Otherwise, the web.xml MUST have a matching &lt;servlet-mapping&gt;. [I agree that this feels backwards, btw.]

I think that's all consistent with your last comment, so that's good.

As for your questions:

1. If we find a servlet based on an Application name, then the servlet must be intended to be a JAX-RS servlet, right? So I think we're good, unless I'm missing something.

2. I like the idea of using the @ApplicationPath. The developer probably meant it to be used.

And finally, I think I've finally put my finger on something that's been bothering me semi-consciously. Suppose you have 

```
@ApplicationPath("/foo")
public class MyApp extends Application { ... }
```
If web.xml is
```
<web-app>
  <servlet>
    <servlet-name>MyApp</servlet-name>
  </servlet>
  <servlet-mapping>
    <servlet-name>MyApp</servlet-name>
    <url-pattern>/myresources/*</url-pattern>
  </servlet-mapping>
</web-app>
```
Then there **is** a matching servlet, and /myresources overrides /foo. But suppose web.xml is
```
<web-app>
  <servlet-mapping>
    <servlet-name>MyApp</servlet-name>
    <url-pattern>/myresources/*</url-pattern>
  </servlet-mapping>
</web-app>
```
Then, **I guess**, there **isn't** a matching servlet, and /foo overrides /myresources.

If I'm right, then
1. that's weird, and
2. probably &lt;servlet-mapping&gt; should override @ApplicationPath in both cases, for consistency.

If I'm right, maybe we should create an issue.

andymc12: @ronsigal everything looks right there.  

I don't think my changes handle case 1 (No Application subclass), but I don't think the previous code handled it either (looks like it was commented out...).  If we want to handle that case, I'd propose that we create an inner class like:
```
@ApplicationPath("/")
class DefaultApplication extends Application {}
```
and then add that class if there are no other Application subclasses in the web module.

As for the case where there is a servlet-mapping but not a servlet, that seems like a configuration error to me, but I'm checking with members of the web container team here at IBM to verify.  If a user is allowed to specify a servlet-mapping without a servlet, then I agree that if the servlet-mapping/servlet-name matches the Application subclass, then it makes sense to map that app subclass to the associated url-pattern.  That might be something we should clarify in the spec...

Please let me know whether you want me to handle these cases in this PR or a separate issue - either way works for me.  Thanks!

ronsigal: Good morning. ;-)

1. I feel a little uncomfortable putting @ApplicationPath on DefaultApplication just because, if there **were** an Application, its @ApplicationPath would override the &lt;servlet-mapping&gt;.

2. I agree that having a &lt;servlet-mapping&gt; without a &lt;servlet&gt; seems strange, but otherwise I don't know how to make sense of 
```
If no servlet  handles  this  application,  JAX-RS  implementations  are  REQUIRED  to  dynamically
add a servlet ... .  If the Application subclass is annotated with @ApplicationPath, ... .  Otherwise, the application
MUST be packaged with a web.xml that specifies a servlet mapping.
```
3. Given the title of the JIRA, "Use servlet mapping from web.xml when not specified in @ApplicationPath", I think these new things can go here. The only thing I would break out is the code that allows a class to be both an Application and a resource, since that's more of a new feature.

Thanks!

andymc12: Sorry for the slow reply on this - I feel like I've been pulled in a hundred different directions since last week!

> 1. I feel a little uncomfortable putting @ApplicationPath on DefaultApplication just because, if there were an Application, its @ApplicationPath would override the <servlet-mapping>.

No problem.  The behavior of the no Application subclass case is the same in this change as it was before this change.  If we want to address that case further, I'd suggest we open a new JIRA for it.

> 2. I agree that having a <servlet-mapping> without a <servlet> seems strange, but otherwise I don't know how to make sense of

```
If no servlet  handles  this  application,  JAX-RS  implementations  are  REQUIRED  to  dynamically
add a servlet ... .  If the Application subclass is annotated with @ApplicationPath, ... .  Otherwise, the application
MUST be packaged with a web.xml that specifies a servlet mapping.
```

I think the spec clarifies it further using an example and table.  In the example of the web.xml it shows both a servlet and a servlet-mapping element.  The table says:
Screen Shot 2020-09-24 at 8 59 01 AM

That says to me that the same web.xml content is required for an Application subclass without the `@ApplicationPath` as for the case where there is no Application subclass - and in that case too, the servlet and servlet-mapping are present in the web.xml.

My take on the text in question is that if there is no `@ApplicationPath` annotation, then the user MUST provide a web.xml that contains a servlet and servlet-mapping.  If there is an `@ApplicationPath` annotation, then the user does not need to provide a web.xml at all, but they can if they want to override the path specified in the application.

> 3. Given the title of the JIRA, "Use servlet mapping from web.xml when not specified in @ApplicationPath", I think these new things can go here. The only thing I would break out is the code that allows a class to be both an Application and a resource, since that's more of a new feature.

That works for me.  I'll split it out today and open a separate JIRA and PR for that feature.

Thanks again for all the review on this!

andymc12:
I've rebased on master, pulled out the app-and-resource as same class change and unit test.  I think all code review issues have been resolved.  Anything else to do with this one?  Thanks again!

ronsigal:
Hi Andy,

re: "a hundred different directions"

If you can manage a hundred different directions, my red hat is off to you. I max out around three or four. ;-)

Given that neither the current and proposed implementaion handles the case of no Application, that case is theoretical, but we can talk about it anyway. Maybe I'm confused by the phrase "add servlet". I'm assuming it implies that
```
getServletRegistration.getServletRegistration(name)
```
would return null. In fact, now I'm thinking that the paragraph beginning "If no Application subclass is present" is confusing. On the one hand, it's saying "JAX-RS implementations are REQUIRED to dynamically add a servlet" and on the other hand, it's giving an example where web.xml contains a &lt;servlet&gt;. It seems to me that, logically, there should be four cases:

 * no Application, no &lt;servlet&gt;, but a &lt;servlet-mapping&gt; exists
 * no Application, &lt;servlet&gt; exists, and  a &lt;servlet-mapping&gt; exists
 * Application exists, no &lt;servlet&gt;
 * Application exists, &lt;servlet&gt; exists.
 
So there should be two cases in the "If no Application subclass is present" section, and there should be four lines in Table 2.1.

I think two things should probably happen:

 1. RESTEasy should handle the case of no Application (I'm surprised that TCK doesn't catch this), and
 2. The spec should be clarified.

Maybe we could work on a spec change outside of this issue. It would give me a chance to atone for not doing any JAX-RS committee work (due to my multiple directions limit). Based on your recent comment, I don't think we're quite in agreement yet. As for implementing the no Application case, I've created RESTEASY-2710 "ResteasyServletInitializer should handle case of no Application subclass".

Regarding the non-theoretical case where there is an Application, I've added a comment to the ResteasyServletInitializer class.

Thanks for all of this!! You probably thought it was going to be easy. ;-)

asoldano:
@ronsigal , @andymc12 , how do we want to move forward here? Is an agreement reached or further changes are needed?

andymc12:
@asoldano @ronsigal Ultimately, I think the changes as-is does no harm.  IIUC, it handles 3 of the 4 scenarios you mentioned, specifically:
>no Application, \<servlet\> exists, and a \<servlet-mapping\> exists
Application exists, no \<servlet\>
Application exists, \<servlet\> exists.

It does not address the case where there is no Application, no `<servlet>` element, but there is a `<servlet-mapping>` element.  I think this is a very rare case, and might not be allowed under the servlet spec. But even it if it allowed, it doesn't currently work in RESTEasy, so these changes don't break anything.

Worst case, these changes don't completely address all scenarios regarding the web.xml, but they improve things (and match Open Liberty's CXF-based JAX-RS 2.1 behavior).

My proposal would be to merge this as-is and then address the fourth scenario at the JAX-RS and/or Servlet spec project(s) and then make another PR if necessary.  Wdyt?

ronsigal:
Hey @andymc12,

Sorry for letting this slide for so long. I've had to read and re-read our conversation to figure out what we were talking about. My sense of it, at least tonight, is that the section in the spec just doesn't make sense. Or, actually, I think there's a way to make sense of it, but it's pretty twisted. In particular, the text says,
```
* If there is already a servlet that handles this application. That is, a servlet that has an initialization
parameter named javax.ws.rs.Application whose value is the fully qualified name of the Application subclass ...
```
If we take that as the definition of a servlet handling an Application, then
```
<web-app version="3.0" xmlns="http://java.sun.com/xml/ns/javaee"
xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://java.sun.com/xml/ns/javaee
http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd">
  <servlet>
    <servlet-name>org.example.MyApplication</servlet-name>
  </servlet>
  <servlet-mapping>
    <servlet-name>org.example.MyApplication</servlet-name>
    <url-pattern>/myresources/*</url-pattern>
  </servlet-mapping>
</web-app>
```
**doesn't** handle MyApplication!! And then the section

```
If no servlet handles this application, ... For example, if org.example.MyApplication is the name of the
Application subclass, a sample web.xml would be:

<web-app version="3.0" xmlns="http://java.sun.com/xml/ns/javaee"
xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://java.sun.com/xml/ns/javaee
http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd">
  <servlet>
    <servlet-name>org.example.MyApplication</servlet-name>
  </servlet>
  <servlet-mapping>
    <servlet-name>org.example.MyApplication</servlet-name>
    <url-pattern>/myresources/*</url-pattern>
  </servlet-mapping>
</web-app>
```
makes sense. Like you said, they're saying that web.xml has to have &lt;servlet&gt; and &lt;servlet-mapping&gt;. But I think this interpretation is weird, and, based on the way you wrote getServletsForApplication(), I think you would agree with me.

My point is that, without this twisted logic, the spec doesn't make sense. Given that, the best we can do, for now, is to implement what we **think** the spec means to say, and, at least tonight, I think your code does that. I'm going to think about it one more time tomorrow and then, hopefully, merge.

@andymc12, thank you for being so patient!

ronsigal:
Ok, I think we're good. The next step would be to try to clarify the text in the spec, and I'll try to work on that. 

Before I merge this, I think I better run the TCK, just in case.

ronsigal:
TCK passed. Here we go.

PR #2479, opened 2020-07-27
Performance improvements

bcluap:
- Avoid unnecessary use of URI.create - rather do more lazily
- Multivalued maps will 99% of the time have 1 or maybe 2 entries for a key so don't allocate memory for 10
- Avoid unnecessary matcher creation when decoding when a string does not contain % in it - which is the more likely case for URI's to Rest resources

These changes increase throughput by about 2% on a simple Rest service and were found doing some hotspot analysis on load tests on Quarkus.


asoldano:
@bcluap , thanks for your contribution. Since this is your first one, please note that we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 
Do you agree to these terms?

bcluap:
Yes I agree

PR #2482, opened 2020-07-29
[RESTEASY-2641] expanded algorithm checking nonProxyHost parameter

rsearls:


andrebreves:
Shouldn't you need to escape the dots (```\\.```) inside the ```nonProxyHosts``` pattern string to prevent it from matching any character?

rsearls:
I disagree that "\\." is needed.   The code is interested in identifying
an IP address that starts with the characters "127".  The first segment of
an IP is never longer than 3 digits. Valid values are 0 thru 255.   
We do not need to check for string "127." in this case.

benDeMtp:
Hi guys,

Do you know if this fix will be release someday ?
Specifying entire host instead using wildcard is cumbersome.
This fix seem's  do the job.
Thank you

asoldano:
@benDeMtp sorry for the late follow-up.
@rsearls , this has conflicts now and needs to be rebased. Can you please rebase against master? Thanks!

ronsigal:
Hey @rsearls,

Comparing this change to the current RestClientBuilderImpl, it looks like this change is superseded and no longer necessary. Could you verify that?

Thanks,
Ron

rsearls:
This change is no longer needed.

PR #2483, opened 2020-07-30
[RESTEASY-2633] Encode code points that span multiple characters

dkBrazz:
[RESTEASY-2633] Encode code points that span multiple characters - utf-8 mb4, emojis

Backport of #2456 and #2480


dkBrazz:
When is the 3.14 release planned?

PR #2487, opened 2020-08-05
[RESTEASY-2658]:Improve sse client performance

jimma:


resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -66,12 +65,21 @@ public boolean isClosed()
...
    public InboundSseEvent read() throws IOException
    {
-      byte[] chunk = null;
+      Charset charset = StandardCharsets.UTF_8;
+      if (mediaType != null && mediaType.getParameters().get(MediaType.CHARSET_PARAMETER) != null)
+      {
+         charset = Charset.forName(mediaType.getParameters().get(MediaType.CHARSET_PARAMETER));
+      }
+      if (!textLike && escape) {
+         charset = Charset.forName("ISO-8859-1");
```
FroMage, line 74: Why do you use this encoding in this case?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -82,139 +90,51 @@ public InboundSseEvent read() throws IOException
...
-   {
-      int b;
-      while ((b = in.read()) != -1)
-      {
-         if (!textLike && escape && b == '\\')
-         {
-            b = in.read();
-            if (b != '\\' && b != '\n' && b != '\r')
+            int colon = line.indexOf(":");
+            String filedValue = line.substring(colon + 2);
```
FroMage, line 118: This probably breaks if a line contains a colon as the last char. You should check on the colon index before doing a substring.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
```diff
@@ -223,8 +143,55 @@ private void processField(final InboundSseEventImpl.Builder inboundEventBuilder,
...
+         {
             inboundEventBuilder.write(SseConstants.EOL);
-         inboundEventBuilder.write(value);
+         }
+         if (!textLike && escape)
+         {
+            byte[] values = null;
+            try
+            {
+               values = valueString.getBytes("ISO-8859-1");
```
FroMage, line 154: Again this weird encoding. The spec https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events says "Event streams are always decoded as UTF-8. There is no way to specify another character encoding." so I think this applies to data fields too.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventProvider.java
```diff
@@ -146,10 +146,20 @@ public void write(int b) throws IOException
                      }
                      else
                      {
-                        if (escape && (b == '\n' || b == '\r' || b == '\\'))
+                        if (escape && b == '\n')
                         {
                            entityStream.write('\\');
-                           entityStream.write(b);
+                           entityStream.write('n');
```
FroMage, line 152: This used to write `\\` and `\n` while now it writes `\\` and `n` (similarly for `r`), is it really normal?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: This is for only for bytes event stream , like to handle this stream resource method:
```
   @GET
   @Path("get/bytes")
   @Produces(MediaType.APPLICATION_OCTET_STREAM)
   @Stream
   public Flowable<byte[]> getBytes() {
```

FroMage: Well, the bytes can be in latin1, but over the wire they must be converted to UTF8. Is that the case here?

jimma: If this is the bytes stream like I posted above and client side expect byte type, this encoding will be used. For the text date message, all will be encoded with UTF-8 when read the event data.  The data media type controls this flag.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: Thanks for the review. I'll fix this.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventInputImpl.java
jimma: Yes.  UTF-8 encoding applies to data field and text/html like content.  For the client side explicitly wants a byte array with Resteasy created @Stream annotation pair with @Produces(MediaType.APPLICATION_OCTET_STREAM), we can handle it to use  ISO-8859-1 to convert the string to bytes array with 8 bit.  This is special path for Resteasy created @Stream feature for push bytes stream. Resteasy can define the media type for data field with these annotations, so it can know user wants bytes without take all data field as text media type with html UTF-8 encoding by default.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventProvider.java
jimma: Before the change, it actually doesn't escape the data. It still makes the data with "\n"  in two lines like : 
```
data: foo\
bar
```

ronsigal: Hey Jim,

Look at SseEventInputImpl.readLine():
```
         if (!textLike && escape && b == '\\')
         {
            b = in.read();
            if (b != '\\' && b != '\n' && b != '\r')
            {
               throw new RuntimeException(Messages.MESSAGES.expectedExcapedCharacter(b));
            }
            out.write(b);
            continue;
         }
```
If it sees a '\\' followed by a line delimiter, it skips over the '\'. At the end of the data, there is an unescaped line delimiter.

-Ron

jimma: @ronsigal  I changed SseEventProvider to escape the line delimiter: 
```
                        if (escape && b == '\n')
                        {
                           entityStream.write('\\');
                           entityStream.write('n');
                        }
                        else if (escape && b == '\r')
                        {
                           entityStream.write('\\');
                           entityStream.write('r');
                        }
                        else if (escape && b == '\\')
                        {
                           entityStream.write('\\');
                           entityStream.write('\\');
                        }
                        else
                        {
                           entityStream.write(b);
                        }
```
All the line delimiters will be escaped to ASCII characters. this will help to escape  "\" followed by a line delimiter ?

asoldano:
@jimma what's the status of this? Besides the conflict that would need to be resolved, is the fix here what has been provided to the user in https://issues.redhat.com/browse/RESTEASY-2658 for preliminary testing? Asking as he seems to be saying that he has problems with it.
Are the concerns regarding encoding that @FroMage mentioned actually solved?

PR #2488, opened 2020-08-05
RESTEASY-2638

liweinan:
… ParamConverterProvider

### DIFF
--- a/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
+++ b/providers/json-binding/src/main/java/org/jboss/resteasy/plugins/providers/jsonb/AbstractJsonBindingProvider.java
@@ -21,13 +21,14 @@ public class AbstractJsonBindingProvider extends JsonBindingProvider {
 
    @Context
    javax.ws.rs.ext.Providers providers;
+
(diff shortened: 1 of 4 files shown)


liweinan:
proof-of-concept in progress.

liweinan:
reference project: https://github.com/liweinan/RESTEASY2638Reproducer

liweinan:
This POC passed the test here:

* https://github.com/liweinan/RESTEASY2638Reproducer/blob/master/src/test/java/io/weli/resteasy2638/ExampleResourceTest.java#L51

Because JSONB doesn't support Jackson-like interface deserialization:

- * [jakarta ee - How to deserialize an interface with json-b? - Stack Overflow](https://stackoverflow.com/questions/46050845/how-to-deserialize-an-interface-with-json-b)

image

So we have to provide a similar annotation in the PR and deal with it during processing.

@asoldano @rsearls Please help to review if this is acceptable. If so I can split the whole part(especially `DeserializeAs` annotation) into `resteasy-extension` as an optional feature (and detect it during runtime inside our `JsonBindingProvider` to improve performance, if necessary).

liweinan:
I created a separate JIRA and PR for it: https://github.com/resteasy/Resteasy/pull/2496

PR #2496, opened 2020-08-20
RESTEASY-2679 Make JSONB provider to support interface deserialisation

liweinan:


liweinan:
reference: https://github.com/resteasy/Resteasy/pull/2488

ronsigal:
Hey @liweinan,

I get the point of what you're doing, but something feels strange to me. In general terms, the point of using an interface is to allow multiple classes that implement the interface to be suitable in some context. By associating a specific class with the interface, you're losing that, so why not just use the class in the first place?

Here's one use case that I can think of:
```
public class InheritanceTest {

	public interface I {}
	public class C implements I {}
	
	public static class CC1 {
		public void f(I i) {}		
	}
	
	@Retention(RetentionPolicy.RUNTIME)
	@Target(ElementType.PARAMETER)
	public @interface DeserializeAs {
	    Class<?> value();
	}
	
	public static class CC2 extends CC1 {
		
		@Override
		public void f(@DeserializeAs(C.class) I i) {}
	}
}
```
Note that I'm not putting @DeserializeAs on an interface definition, which, as I say, makes less sense to me.

WDYT?

liweinan:
@ronsigal Great suggestion!

Let me refactor it and let you check :D

liweinan:
I'm thinking about injecting annotations via the `readFrom()` method of `MessageBodyReader`:

image

I'm learning about more usages in our tests:

image

After learning more on its usages I'll try to find a solution on this one.

liweinan:
I made a proposed solution:

https://github.com/resteasy/Resteasy/pull/2496/commits/813c26542a51aa615af54b615baa04a52238d692

And here is the sample usage:

https://github.com/liweinan/RESTEASY2638Reproducer/blob/master/src/main/java/io/weli/resteasy2638/ExampleResource.java#L55

However I can't make it work yet because the `JSONBindingProvider` can't fetch the annotations from method directly:

image

The root cause is that under multipart senario, the `JSONBindingProvider` is not directly called but instead called by multipart provider(because just one part of the request data needs `JSONBindingProvider`).

I'll dig more into call chain to see how to solve the problem.

jamezp:

PR #2497, opened 2020-08-24
[RESTEASY-2676] removed PreProcessInterceptor related documentation

rsearls:


resteasy-upgrade-guide/reference/en/en-US/modules/Interceptors.xml
```diff
@@ -8,7 +8,6 @@
    
    <orderedlist>
       <listitem>reader/writer interceptors</listitem>
-      <listitem>server side <classname>PreProcessInterceptor</classname></listitem>
```
asoldano, line 11: The changes to PreProcessInterceptor (deprecated in 3.x and removed in 4.x) applies to the whole RESTEasy 2 interceptor mechanism. So this whole Interceptors.xml chapter has to be revisited.

docbook/reference/en/en-US/modules/Multipart.xml
```diff
@@ -1198,36 +1198,6 @@ String name = proxy.putMultipartForm(new JsonFormResource
...
-    implements PreProcessInterceptor {
-
-    public ServerResponse preProcess(HttpRequest request, ResourceMethod method)
-            throws Failure, WebApplicationException {
-        request.setAttribute(InputPart.DEFAULT_CONTENT_TYPE_PROPERTY, "*/*; charset=UTF-8");
-        return null;
-    }
-}]]></programlisting>
-        </para>
-    </section>
```
asoldano, line 1230: While I see the example making use of PreProcessInterceptors, I'm wondering if the multipart related functionality here should still be documented (perhaps the standard JAX-RS 2.0+ interceptors can be used instead?)

docbook/reference/en/en-US/modules/Multipart.xml
```diff
@@ -1255,45 +1225,4 @@ public Response setMediaType(MultipartInput input) throws IOException
...
-            are set, then the value of
-        </para>
-        <para>
-            <code>InputPart.DEFAULT_CHARSET_PROPERTY</code>
-            will override any charset in the value of
-        </para>
-        <para>
-            <code>InputPart.DEFAULT_CONTENT_TYPE_PROPERTY</code>.
-        </para>
-    </section>
```
asoldano, line 1298: Same as above

asoldano (changes requested):
Thanks @rsearls , I completely forgot the documentation was covering these things and have to be updated. However, I believe we need more than what you're proposing here, as the whole old interceptor framework has gone in 4.x. Can you please check the whole documentation for any reference to that and fix it to refer to the standard JAX-RS interceptors? Thanks.

docbook/reference/en/en-US/modules/Multipart.xml
rsearls: The code you have flagged is being removed from the chapter.

docbook/reference/en/en-US/modules/Multipart.xml
rsearls: The code you have flagged is being removed from the chapter.  I'm confused by your comment.

resteasy-upgrade-guide/reference/en/en-US/modules/Interceptors.xml
rsearls: All references to PreProcessInterceptor are already removed for the book.

docbook/reference/en/en-US/modules/Multipart.xml
asoldano: @rsearls, I think what I meant when I wrote the comment above was that instead of simply removing the documentation section above because the example describing it make use of classes that have been removed in master, we should revisit the example / description of the feature to use the standard JAX-RS interceptors / filters instead. This assumes that the feature here, as well as those in the comments below, is still available on master. If that is not the case, then yes, erasing the section as you're proposing is fine. Do you see what I mean?

PR #2499, opened 2020-08-26
[RESTEASY-2681] merged changes from keycloak into code base

rsearls:


GitHubUser4234:
Hi,

Great that RESTEASY-2681 will be fixed! Currently it looks like the relevant code isn't included yet, in particular the `encodedSignatureInput` variable in JWSInput, `encodeAll` in JWSBuilder and how the sign method passes a StringBuilder to the `encode `/ `encodeAll ` methods. But well, I probably have gotta be patient 😃

rmartinc:
@GitHubUser4234 is right, there are two errors in the current JWS implementation:

* Right now the base64 uses Base64.getURLEncoder/getURLDecoder to perform the Base64 of the parts. This is bad cos in the spec rfc7515 it says is base64 encoding using the URL without the trailing '=' signs. So now the '=' chars are not omitted. You can use the Base64 class in keycloak (as you are doing in the PR) or just remove/add the two possibly trailing equals manually after using the JavaSE URL encoder/decoder.
* The current implementation just sings the `base64_payload` instead of the `base64_header + "." + base64_payload`. This is the part @GitHubUser4234 is missing in the PR.

You can test with https://jwt.io to check everything is right. I used this little test in `JWSTest` that prints the public key used and the final encoded JWS string to use in that page.

```java
   @Test
   public void testJWT() throws Exception {
      KeyPair keyPair = KeyPairGenerator.getInstance("RSA").generateKeyPair();
      StringWriter writer = new StringWriter();
      try (PemWriter pemWriter = new PemWriter(writer)) {
          pemWriter.writeObject(new PemObject("RSA PUBLIC KEY", keyPair.getPublic().getEncoded()));
          pemWriter.flush();
      }
      logger.info(writer.toString());

      String content = "{\"sub\": \"1234567890\",\"name\": \"John Doe\",\"iat\": 1516239022}";
      String encoded = new JWSBuilder()
            .type("JWT")
            .content(content.getBytes(StandardCharsets.UTF_8))
            .rsa256(keyPair.getPrivate());

      logger.info(encoded);
      Assert.assertThat(encoded, CoreMatchers.not(CoreMatchers.containsString("=")));

      JWSInput input = new JWSInput(encoded, ResteasyProviderFactory.getInstance());
      String msg = (String) input.readContent(String.class, null, null, MediaType.TEXT_PLAIN_TYPE);
      Assert.assertEquals(RESPONSE_ERROR_MSG, content, msg);
      Assert.assertTrue(VERIFY_ERROR_MSG, RSAProvider.verify(input, keyPair.getPublic()));
   }
```

rsearls:
I've made the code additions and added the testcase.
Please review code.

rmartinc:
@rsearls I think that now it's complete but do you think that all methods in `JWSBuilder` should be deprecated? In keycloak the idea is using a `SignatureSignerContext` but here nobody is implementing that interface, so it's compulsory to use a deprecated method.

rsearls:
changes complete

rmartinc:
Any update about this?

asoldano (approved):

asoldano:
Thanks @rsearls (and @rmartinc and @GitHubUser4234 for the review)

PR #2501, opened 2020-08-31
[RESTEASY-2683] NPE in ApacheHttpClient43Test because cache is null in MediaTypeMap

rmartinc:
The `cache` can result in null at initialization. I'm changing to always set the `cache` to `classCache` no matter if it was created by the same thread or by any another one. I also used a ConcurrentHashMap instead of normal HashMap.

I'm not protecting the setting part in `mergeEverything`, if you think it's necessary to protect it too just let me know.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
@@ -550,8 +550,8 @@ public class MediaTypeMap<T>
       CachedMediaTypeAndClass cacheEntry = null;
       if (useCache)
       {
+         cacheEntry = new CachedMediaTypeAndClass(type, accept);
          if (classCache != null) {
-            cacheEntry = new CachedMediaTypeAndClass(type, accept);
             cached = classCache.get(cacheEntry);
             if (cached != null) return cached;
          }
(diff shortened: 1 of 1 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
```diff
@@ -550,8 +551,8 @@ private void mergeEverything(Entry<T> entry) {
       CachedMediaTypeAndClass cacheEntry = null;
       if (useCache)
       {
+         cacheEntry = new CachedMediaTypeAndClass(type, accept);
```
jimma, line 554: Why does this line move out of check block ?

resteasy-core/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
```diff
@@ -576,14 +577,14 @@ private void mergeEverything(Entry<T> entry) {
       cached = convert(matches);
       if (useCache) {
          Map<CachedMediaTypeAndClass, List<T>> cache = classCache;
-         if (classCache == null) {
+         if (cache == null) {
             synchronized (this)
             {
                if (classCache == null)
```
jimma, line 583: if (cache== null) check again here ?

resteasy-core/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
rmartinc: This is just because the first time when the classCache is null the cacheEntry is not initialized, it remains null, and therefore the put at the end is done with cacheEntry=null, so nothing is cached the first time (indeed it's cached wrong, with a wrong key). I just fixed this minor issue too. This change is minor, as this is just executed differently the first time when classCache is null, all the rest of times is exactly the same.

resteasy-core/src/main/java/org/jboss/resteasy/core/MediaTypeMap.java
rmartinc: I think that it's not needed to repeat the check for the cache var, cache is local to the method, so if it was null then it's null now. Only classCache can be initialized by other threads.

rmartinc:
Changed to use a HashMap again. :-)

PR #2502, opened 2020-09-02
[RESTEASY-2627] removed CompletionStage element

rsearls:


FroMage (changes requested):
This looks wrong. why would you do that?

asoldano (commented):
How about #2512 instead?

FroMage:
Oh yeah, the other one looks good, let's close this one.

PR #2510, opened 2020-09-10
[RESTEASY-2689] Fix SseEventOutputImpl wrong CompletionStage composition sequence

D01B:


asoldano:
@D01B , thanks for your contribution. Your changes seems to be causing regressions in the testsuite, can you please check and see if it's an issue of your fix or perhaps the tests needs to be updated/improved?
Besides that, we require that all contributions are made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 : do you agree to these terms?

D01B:
I agree with terms of the ASL 2.0 License

D01B:
@asoldano thank you for comment, could you review my changes and merge it Into the master branch? We are really need that Changes in production.

asoldano:
@D01B thanks for the prompt update and the agreement to the license. Now, the changes you're proposing look correct to me; I'm not sure I understand the removal of the try-finally blocks that were closing the sink instances in the test. Can you provide a brief explanation in a comment here? Thanks!

D01B:
@asoldano sure. Write and flush data into the socket is fully asynchronous and might be lazy. That operation expects that socket and internal buffers still capable of writing into it, but try-with-resource might interrupt the lifecycle of the WebSocket before data finally writes into it. 
The socket will be closed automatically when connections closed by the client-side.

asoldano (approved):
OK, thanks

jimma:
should we add a test case for this issue ?

asoldano:
For the records: https://github.com/resteasy/Resteasy/commit/8d3ce03d06046a86791861b78984f076436a3bb8 (thanks Jim)

PR #2512, opened 2020-09-10
[RESTEASY-2627] Chain thenCompose properly

asoldano:


FroMage (approved):
That looks like the right thing to do indeed. Damn that's a sneaky bug you just fixed.

PR #2505, opened 2020-09-04
[RESTEASY-2686] Prepare TS profile for bootable jar testing

marekkopecky:
RESTEasy jira with some additional details: https://issues.redhat.com/browse/RESTEASY-2686

I will also port this to RESTEasy 3.13 once [WFLY-13803](https://issues.redhat.com/browse/WFLY-13803) will be merged

PR #2522, opened 2020-09-14
Use bootable-jar with all modules in TS bootable jar profile

marekkopecky:
jira: https://issues.redhat.com/browse/RESTEASY-2686
related PR: https://github.com/resteasy/Resteasy/pull/2505

I checked bootable jar with latest WF snapshot and I found out, that bootable-jar server is using artifacts from `~/.m2` repo. So if server run is using wf, that was built to maven.repo.local, TS will be executed against wrong artifacts (becaues server won't use artifacts from custom maven repo, but from default repo `~/.m2`). TS execution with `-Dmaven.repo.local` doesn't help as well, because arquillian.xml doesn't use it by default.

This PR removes `<jboss-maven-dist/>` element from wildfly-jar-maven-plugin plugin. Without this changes, plugin builds bootable-jar without modules and modules are used from `~/.m2`. With this change (no jboss-maven-dist xml element), bootable jar is built with all modules, so `~/.m2` is not used.

See some test runs on jira comment.

### DIFF
--- a/testsuite/integration-tests/pom.xml
+++ b/testsuite/integration-tests/pom.xml
@@ -245,7 +245,6 @@
                                     <log-time>true</log-time>
                                     <offline>true</offline>
                                     <plugin-options>
-                                        <jboss-maven-dist/>
                                         <jboss-fork-embedded>true</jboss-fork-embedded>
                                     </plugin-options>
                                     <feature-packs>


marekkopecky:
RESTEasy 3.12 is not used in WF (WF is currently using 3.13.2.Final), so I'm closing this PR.

PR #2526, opened 2020-09-15
(3.13) Adding support for running test suite against bootable JAR

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2694

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,6 +30,7 @@
         <surefire.system.args>-Xms512m -Xmx512m ${modular.jdk.args}</surefire.system.args>
         <!-- Plugins versions -->
         <version.org.jacoco.plugin>0.7.9</version.org.jacoco.plugin>
+        <version.org.wildfly.jar.plugin>2.0.0.Beta8</version.org.wildfly.jar.plugin>
         <version.surefire.plugin>2.19.1</version.surefire.plugin>
         <version.javadoc.plugin>3.0.1</version.javadoc.plugin>
     </properties>
@@ -226,6 +227,11 @@
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>${version.org.jacoco.plugin}</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.wildfly.plugins</groupId>
+                    <artifactId>wildfly-jar-maven-plugin</artifactId>
+                    <version>${version.org.wildfly.jar.plugin}</version>
                 </plugin>
             </plugins>
         </pluginManagement>
--- a/resteasy-dependencies-bom/pom.xml
+++ b/resteasy-dependencies-bom/pom.xml
@@ -16,7 +16,7 @@
     <description>RESTEasy dependencies BOM</description>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dep.arquillian-bom.version>1.1.15.Final</dep.arquillian-bom.version>
+        <dep.arquillian-bom.version>1.6.0.Final</dep.arquillian-bom.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <version.com.fasterxml.classmate>1.5.1</version.com.fasterxml.classmate>
         <version.com.fasterxml.jackson>2.10.5</version.com.fasterxml.jackson>
@@ -80,7 +80,7 @@
         <version.org.wildfly.core.wildfly-cli>7.0.0.Final</version.org.wildfly.core.wildfly-cli>
         <version.org.wildfly.extras.creaper>1.6.1</version.org.wildfly.extras.creaper>
         <version.org.wildfly.wildfly-arquillian-container-managed>2.2.0.Final</version.org.wildfly.wildfly-arquillian-container-managed>
-        <version.org.wildfly.wildfly-arquillian-container-remote>2.1.1.Final</version.org.wildfly.wildfly-arquillian-container-remote>
+        <version.org.wildfly.wildfly-arquillian-container-bootable>3.0.0.Beta3</version.org.wildfly.wildfly-arquillian-container-bootable>
         <version.org.yaml.snakeyaml>1.26</version.org.yaml.snakeyaml>
         <version.weld.api>3.1.SP3 </version.weld.api>
         <version.weld>3.1.5.Final</version.weld>
@@ -893,9 +893,15 @@
             </dependency>
             <dependency>
                 <groupId>org.wildfly.arquillian</groupId>
-                <artifactId>wildfly-arquillian-container-remote</artifactId>
-                <version>${version.org.wildfly.wildfly-arquillian-container-remote}</version>
+                <artifactId>wildfly-arquillian-container-bootable</artifactId>
+                <version>${version.org.wildfly.wildfly-arquillian-container-bootable}</version>
                 <exclusions>
+                    <exclusion>
+                        <groupId>org.jboss.sasl</groupId>
+                        <artifactId>jboss-sasl</artifactId>
+                    </exclusion>
+                    <!-- TODO remove this exclusion once this depends on Jakarta Inject -->
+                    <!-- superseded by jakarta.inject:jakarta.inject-api -->
                     <exclusion>
                         <groupId>javax.inject</groupId>
                         <artifactId>javax.inject</artifactId>
--- a/testsuite/arquillian-utils/pom.xml
+++ b/testsuite/arquillian-utils/pom.xml
@@ -30,6 +30,20 @@
                 </dependency>
             </dependencies>
         </profile>
+        <profile>
+            <id>bootablejar.profile</id>
+            <activation>
+                <property>
+                    <name>ts.bootable</name>
+                </property>
+            </activation>
+            <dependencies>
+                <dependency>
+                    <groupId>org.wildfly.arquillian</groupId>
+                    <artifactId>wildfly-arquillian-container-bootable</artifactId>
+                </dependency>
+            </dependencies>
+        </profile>
     </profiles>
 
     <dependencies>
@@ -38,6 +52,14 @@
             <groupId>org.jboss.shrinkwrap.resolver</groupId>
             <artifactId>shrinkwrap-resolver-depchain</artifactId>
             <type>pom</type>
+            <exclusions>
+                <!-- TODO remove this exclusion once this depends on Jakarta Inject -->
+                <!-- superseded by jakarta.inject:jakarta.inject-api -->
+                <exclusion>
+                    <groupId>javax.inject</groupId>
+                    <artifactId>javax.inject</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
@@ -130,6 +152,12 @@
                 <exclusion>
                     <groupId>org.sonatype.aether</groupId>
                     <artifactId>aether-api</artifactId>
+                </exclusion>
+                <!-- TODO remove this exclusion once this depends on Jakarta Inject -->
+                <!-- superseded by jakarta.inject:jakarta.inject-api -->
+                <exclusion>
+                    <groupId>javax.inject</groupId>
+                    <artifactId>javax.inject</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
--- a/testsuite/integration-tests/pom.xml
+++ b/testsuite/integration-tests/pom.xml
@@ -203,6 +203,152 @@
                 <microprofile.excluded.groups>org.jboss.resteasy.category.ExpectedFailingWithStandaloneMicroprofileConfiguration,org.jboss.resteasy.category.ExpectedFailingBecauseOfSmallRyeMicroprofileOpenApi11</microprofile.excluded.groups>
             </properties>
         </profile>
+        
+        <profile>
+            <id>bootablejar.profile</id>
+            <activation>
+                <property>
+                    <name>ts.bootable</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <!-- Disable the standard copy-based provisioning -->
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-antrun-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>initialize-basedirs</id>
+                                <phase>none</phase>
+                                <goals>
+                                    <goal>run</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.wildfly.plugins</groupId>
+                        <artifactId>wildfly-jar-maven-plugin</artifactId>
+                        <executions>
+                            <!-- Package a jaxrs-server layer with a few other layers -->
+                            <execution>
+                                <id>bootable-jar-observability-packaging</id>
+                                <goals>
+                                    <goal>package</goal>
+                                </goals>
+                                <phase>process-test-resources</phase>
+                                <configuration>
+                                    <output-file-name>jaxrs-server.jar</output-file-name>
+                                    <hollowJar>true</hollowJar>
+                                    <record-state>false</record-state>
+                                    <log-time>true</log-time>
+                                    <plugin-options>
+                                        <jboss-fork-embedded>true</jboss-fork-embedded>
+                                    </plugin-options>
+                                    <feature-packs>
+                                        <feature-pack>
+                                            <groupId>${testsuite.wildfly.galleon.pack.group.id}</groupId>
+                                            <artifactId>wildfly-galleon-pack</artifactId>
+                                            <version>${server.version}</version>
+                                            <!-- Required for rxjava -->
+                                            <includedPackages>
+                                                <package>org.jboss.resteasy.resteasy-rxjava2</package>
+                                            </includedPackages>
+                                        </feature-pack>
+                                    </feature-packs>
+                                    <layers>
+                                        <layer>jaxrs-server</layer>
+                                        <layer>microprofile-config</layer>
+                                        <layer>datasources</layer>
+                                        <layer>h2-default-datasource</layer>
+                                        <layer>undertow-legacy-https</layer>
+                                        <layer>ejb</layer>
+                                        <layer>mail</layer>
+                                    </layers>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+
+                    <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <install.dir>${project.build.directory}/wildfly</install.dir>
+                                <bootable.jar>${project.build.directory}/jaxrs-server.jar</bootable.jar>
+                                <arquillian.xml>arquillian-bootable.xml</arquillian.xml>
+                                <container.management.port.managed>${container.management.port.managed}</container.management.port.managed>
+                                <securityManagerArg>${securityManagerArg}</securityManagerArg>
+                                <additionalJvmArgs>${additionalJvmArgs}</additionalJvmArgs>
+                                <ipv6>false</ipv6>
+                                <ipv6ArquillianSettings></ipv6ArquillianSettings>
+                                <node>127.0.0.1</node>
+                                <project.version>${project.version}</project.version>
+                                <version.resteasy.testsuite>${version.resteasy.testsuite}</version.resteasy.testsuite>
+                                <security.provider>${security.provider}</security.provider>
+                                <modular.jdk.args>${modular.jdk.args}</modular.jdk.args>
+                                <container.base.dir.managed>${container.base.dir.managed}</container.base.dir.managed>
+                                <container.offset.managed>0</container.offset.managed>
+                                <container.qualifier.managed>${container.qualifier.managed}</container.qualifier.managed>
+                                <container.qualifier.managed.encoding>${container.qualifier.managed.encoding}</container.qualifier.managed.encoding>
+                                <container.offset.managed.encoding>${container.offset.managed.encoding}</container.offset.managed.encoding>
+                                <container.management.port.managed.encoding>${container.management.port.managed.encoding}</container.management.port.managed.encoding>
+                            </systemPropertyVariables>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <id>default-test</id>
+                                <phase>none</phase>
+                            </execution>
+                            <execution>
+                                <id>disable-jsonb-client</id>
+                                <phase>none</phase>
+                            </execution>
+
+                            <execution>
+                                <id>default-test-bootable-jar</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>test</goal>
+                                </goals>
+                                <configuration>
+                                    <excludes>
+                                        <!-- Tests requires excluding JsonBindingProvider-->
+                                        <exclude>**/JsonBindingAnnotationsJacksonTest.java</exclude>
+                                        <exclude>**/SseJsonEventTest</exclude>
+                                        <!-- Tests without proper arquillian container in arquillian-bootable.xml -->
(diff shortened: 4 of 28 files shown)


ronsigal:
I added @NotForBootableJar to the deployment to some tests.

resteasy-dependencies-bom/pom.xml
```diff
@@ -15,8 +15,8 @@
     <name>RESTEasy dependencies BOM</name>
     <description>RESTEasy dependencies BOM</description>
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dep.arquillian-bom.version>1.1.15.Final</dep.arquillian-bom.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
marekkopecky, line 18: This additional space is not necessary ...

resteasy-dependencies-bom/pom.xml
```diff
@@ -80,8 +80,8 @@
         <version.org.wildfly.core.wildfly-cli>7.0.0.Final</version.org.wildfly.core.wildfly-cli>
         <version.org.wildfly.extras.creaper>1.6.1</version.org.wildfly.extras.creaper>
         <version.org.wildfly.wildfly-arquillian-container-managed>2.2.0.Final</version.org.wildfly.wildfly-arquillian-container-managed>
-        <version.org.wildfly.wildfly-arquillian-container-remote>2.1.1.Final</version.org.wildfly.wildfly-arquillian-container-remote>
-        <version.org.yaml.snakeyaml>1.26</version.org.yaml.snakeyaml>
+         <version.org.wildfly.wildfly-arquillian-container-bootable>3.0.0.Beta3</version.org.wildfly.wildfly-arquillian-container-bootable>
+         <version.org.yaml.snakeyaml>1.26</version.org.yaml.snakeyaml>
```
marekkopecky, line 84: This additional space is not necessary ...

testsuite/arquillian-utils/pom.xml
```diff
@@ -30,14 +30,36 @@
...
+        </profile>
     </profiles>
 
     <dependencies>
         <!-- arquillian general -->
         <dependency>
             <groupId>org.jboss.shrinkwrap.resolver</groupId>
             <artifactId>shrinkwrap-resolver-depchain</artifactId>
-            <type>pom</type>
+            <type>pom</type>
```
marekkopecky, line 54: Same here (many spaces at the end of a line)

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
```diff
@@ -74,7 +75,8 @@
 @RunWith(Arquillian.class)
 @RunAsClient
 @Category({
-    ExpectedFailingWithStandaloneMicroprofileConfiguration.class // MP is missing jboss.naming.context.java.jboss.exported.jms.RemoteConnectionFactory
+    ExpectedFailingWithStandaloneMicroprofileConfiguration.class, // MP is missing jboss.naming.context.java.jboss.exported.jms.RemoteConnectionFactory
+    NotForBootableJar.class // needs ejb + standalone-full
```
marekkopecky, line 79: Is it possible to use standalone-full for bootable-jar testing here?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DuplicateDeploymentTest.java
```diff
@@ -27,6 +28,7 @@
  */
 @RunWith(Arquillian.class)
 @RunAsClient
+@Category(NotForBootableJar.class)
```
marekkopecky, line 31: Some description with reason would be useful (eg add `// no log check support for bootable-jar in RESTEasy TS so far`)

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProviderPriorityTest.java
```diff
@@ -31,6 +33,7 @@
 
 @RunWith(Arquillian.class)
 @RunAsClient
+@Category(NotForBootableJar.class) // no RX layer so far
```
marekkopecky, line 36: > no RX layer so far

Where is RX classes used here?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProxyTest.java
```diff
@@ -63,6 +64,7 @@
       war.addClass(PortProviderUtil.class);
       war.addClass(Category.class);
       war.addClass(MicroProfileDependent.class);
+      war.addClass(NotForBootableJar.class);
```
marekkopecky, line 67: Why is this annotation added here?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jsonb/basic/JsonBindingDebugLoggingTest.java
```diff
@@ -48,6 +50,7 @@
  */
 @RunWith(Arquillian.class)
 @ServerSetup({DebugLoggingServerSetup.class}) // TBD: remove debug logging activation?
+@Category(NotForBootableJar.class) // no log check support for bootable-jar in RESTEasy TS so fa
```
marekkopecky, line 53: typo, `s/fa/far/g`

pom.xml
```diff
@@ -226,6 +227,11 @@
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>${version.org.jacoco.plugin}</version>
+                </plugin>
```
marekkopecky, line 230: Additional spaces at the end of this line ...

pom.xml
```diff
@@ -30,6 +30,7 @@
         <surefire.system.args>-Xms512m -Xmx512m ${modular.jdk.args}</surefire.system.args>
         <!-- Plugins versions -->
         <version.org.jacoco.plugin>0.7.9</version.org.jacoco.plugin>
+        <version.org.wildfly.jar.plugin>2.0.0.Beta6</version.org.wildfly.jar.plugin>
```
marekkopecky, line 33: Beta7?

testsuite/integration-tests/pom.xml
```diff
@@ -203,6 +203,169 @@
...
+                                <install.dir>${project.build.directory}/wildfly</install.dir>
+                                <bootable.jar>${project.build.directory}/jaxrs-server.jar</bootable.jar>
+                                <arquillian.xml>arquillian-bootable.xml</arquillian.xml>
+                                <container.management.port.managed>${container.management.port.managed}</container.management.port.managed>
+                                <securityManagerArg>${securityManagerArg}</securityManagerArg>
+                                <additionalJvmArgs>${additionalJvmArgs}</additionalJvmArgs>
+                                <ipv6>false</ipv6>
+                                <ipv6ArquillianSettings></ipv6ArquillianSettings>
+                                <node>127.0.0.1</node>
+                                <!--<jboss.home>${jboss.home}</jboss.home>-->
```
marekkopecky, line 302: Can `jboss.home` and `jboss.server.config.file.name` be removed here?

marekkopecky (changes requested):
I'm adding some comments & questions

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProxyTest.java
fabiobrz: @ronsigal added this. I don't have it in my branch. I think it was because he noticed failures in different profiles execution (e.g.: MicroProfile). I am pushing a version that does not have the annotation, so we can see what the CI checks will report.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProviderPriorityTest.java
fabiobrz: You're right. There's no use of rx classes, but I added the annotation because I noticed a rx related error. Maybe I was too quick, I am pushing changes that remove it and we'll see what happens, although this should not be considered at all when running `ts.bootable` profile since I noticed that you always used `-Dts.disable.microprofile...` right?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
fabiobrz: I don't think so. The supported layers combinations never bring up a standalone-full. I discussed this with @jfdenise who could add more input, in case it is needed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProviderPriorityTest.java
marekkopecky: > this should not be considered at all when running ts.bootable profile since I noticed that you always used -Dts.disable.microprofile... right?

There is no `ts.disable.microprofile`. There is `disable.microprofile.tests`, but it affects tests with `MicroProfileDependent`. But this RestClientProviderPriorityTest class is not annotated by `MicroProfileDependent`, so it's not related.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
marekkopecky: Right, sorry, standalone profiles probably doesn't make sense in this context. So what layer is missing for correct execution of this test (from standalone-full)? Messaging?

testsuite/integration-tests/pom.xml
```diff
@@ -203,6 +203,167 @@
...
+                <property>
+                    <name>ts.bootable</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-dependency-plugin</artifactId>
+                        <executions>
+                            <execution>
```
marekkopecky, line 219: I'm thinking about `unpack` and `unpack-resteasy` executions. Why are they needed? According to the Readme file, `-Dserver.home=not-needed-for-bootable.jar` should be used. In that case, download profile won't be activated, so there is no need to replace `unpack` and `unpack-resteasy` executions.

If `-Dserver.home=not-needed-for-bootable.jar` won't be used, download profile will be activated together with `bootablejar.profile`. In this case, there are two active profiles with the executions of the same names. And AFAIR there is no guarantee from maven side which profile will be executed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
fabiobrz: IIRC was messaging but there were several more things missing. In fact I tried and added the new messaging layer but it didin't work.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProviderPriorityTest.java
fabiobrz: > There is no ts.disable.microprofile. There is disable.microprofile.tests

Yeah, I was typing quickly and didn't go look for the name but I see you got it.

> But this RestClientProviderPriorityTest class is not annotated by MicroProfileDependent, so it's not related.

Good to know. Annotations were removed, let's see what CI checks say.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
marekkopecky: It would be great to describe more details about reason for ignoring this testcase to java comment ...

testsuite/integration-tests/pom.xml
fabiobrz: Thanks for this comment, this is interesting. So there's no guarantee that - when explicitly using `-Dts.bootable` the executions here will be used instead of those defined in the download profile?
What would you suggest then? Following up on your reasoning, then we know for sure that those in the download profile will be executed if `-Dserver.home=not-needed-for-bootable.ja` is passed, once these are removed.
And being the precondition the same I think this last one is a worse scenario. WDYT?
Anyway, I agree with you and I'd like to have a clean solution here.

marekkopecky: > So there's no guarantee that - when explicitly using -Dts.bootable the executions here will be used instead of those defined in the download profile?

AFAIK you can define execution in default `<build>` once at most. And you can override it in one profile at most. There is nothing like profile orders in maven AFAIK. So if you don't define execution in default `<build>`, but you define it two times differently in two different active profiles, you don't have any guarantee, which definition will be used (@asoldano @ronsigal feel free to correct me, if I'm wrong).

> those in the download profile will be executed if `-Dserver.home=not-needed-for-bootable.jar` is passed

Actually, download profile is activated like this:

```
          <activation>
            <property>
              <name>!server.home</name>
            </property>
          </activation>
```

So if `-Dserver.home=not-needed-for-bootable.jar` is passed, download profile is not activated (see `!` in `name` element), which is something that we want here.

> I'd like to have a clean solution here.

`-Dserver.home=not-needed-for-bootable.jar` is obviously not super-clean solution. But it is a solution and it is deterministic. I don't see any other simple solution without completely changing the way how maven profiles are constructed now in this repo. @asoldano / @ronsigal Let us know if you have some better idea here ...

fabiobrz: Thanks @marekkopecky - at the moment I am focusing on the fact that the we need those executions to be skipped.
As you say, the current implementation only executes those when `!server.home`, i.e. when the download profile is activated.
And this is deterministic, so I agree with you that this could be the most robust solution.
I am going to push changes, let's see how CI checks go.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
fabiobrz: Ok, adding more detail about the fact that a 250 MB bootable JAR is basically useless.

testsuite/integration-tests/pom.xml
fabiobrz: Can we resolve this now, @marekkopecky? WDYT?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
marekkopecky: > which is useless

Hmm, ok, but what is missing in bootable jar to have this working?

testsuite/integration-tests/pom.xml
marekkopecky: Yes, thanks!

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/InjectionTest.java
fabiobrz: Messaging was missing for sure, and I added it. Then it failed again, and I discussed the thing with the feature developer, who made his experiments as well and he got to the conclusion that a full server was needed to make this work.
Maybe @jfdenise can elaborate a bit if what you need to know is the full list of subsystems that should in place.

marekkopecky: I don't think that this blocks this PR. But I'm wondering whether this is some issue that could affects users. @jfdenise can you please provide more details on this?

jfdenise: I don't have the list of needed subsystems, I realized that a standalone-full.xml was used in the test (embedded broker for example). I made the bootable JAR to provision a full config and the test passed. BTW, I also noticed that I had to remove one of the 2 executions in the arquilian-bootable.xml. We are starting 2 server that unzip there content in the same location (install-dir). This breaks messaging journal that doesn't seem to like the WildFly directory to be re-written.

marekkopecky: >  We are starting 2 server that unzip there content in the same location (install-dir)

+1, this looks like a TS bug. @fabiobrz we can for example add some suffix to the second arquillian container

fabiobrz: Thanks @marekkopecky and @jfdenise, done.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProviderPriorityTest.java
fabiobrz: @marekkopecky this test is passing now, can we resolve this conversation?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/RestClientProxyTest.java
fabiobrz: @marekkopecky this test is passing now, can we resolve this conversation?

fabiobrz:
Hi @ronsigal, @marekkopecky finally CI checks are passing for JDK 8 too :)

ronsigal:
re: "@ronsigal added this."

Sorry I didn't notice this comment before. At some point, I was seeing eight (IIRC) test failures, and they were resolved when I added that annotation to the war. Strangely, six of the tests actually referenced the annotation, and two of them didn't, but even so were fixed when I added it. I couldn't explain it, but it worked ... :-\

fabiobrz:
Hi @ronsigal, it’s fine to me, I was just explaining what happened. Anyway I think everything is fixed ATM.

testsuite/integration-tests/pom.xml
```diff
@@ -203,6 +203,148 @@
...
+                                <goals>
+                                    <goal>package</goal>
+                                </goals>
+                                <phase>process-test-resources</phase>
+                                <configuration>
+                                    <output-file-name>jaxrs-server.jar</output-file-name>
+                                    <hollowJar>true</hollowJar>
+                                    <record-state>false</record-state>
+                                    <log-time>true</log-time>
+                                    <offline>true</offline>
```
marekkopecky, line 246: As discussed on google chat, this offline element needs to be removed ...

fabiobrz: +1 many thanks.

asoldano:
@ronsigal , shall we merge this?

marekkopecky:
@asoldano I run some CI runs and it seems that there is 108 stable errors with bootable profile :( Please wait a little bit with merging ...

Btw, there is also 20 stable test errors without bootable profile but with WF 21.0.0.Beta1. Resteasy 3.13 TS is currently using wf 20.0.1.Final by default, it it's not visible in upstream CI

marekkopecky:
See also https://github.com/resteasy/Resteasy/pull/2505#discussion_r496456065

asoldano:
> @asoldano I run some CI runs and it seems that there is 108 stable errors with bootable profile :( Please wait a little bit with merging ...
> 
> Btw, there is also 20 stable test errors without bootable profile but with WF 21.0.0.Beta1. Resteasy 3.13 TS is currently using wf 20.0.1.Final by default, so it it's not visible in upstream CI

Thanks @marekkopecky . As for the failures with WFLY 21 Beta1, most of them are because of https://issues.redhat.com/browse/WFCORE-5143 . There were also a couple of failures due to https://issues.redhat.com/browse/RESTEASY-2711 which should be fixed now.

fabiobrz:
> > @asoldano I run some CI runs and it seems that there is 108 stable errors with bootable profile :( Please wait a little bit with merging ...
> > Btw, there is also 20 stable test errors without bootable profile but with WF 21.0.0.Beta1. Resteasy 3.13 TS is currently using wf 20.0.1.Final by default, so it it's not visible in upstream CI
> 
> Thanks @marekkopecky . As for the failures with WFLY 21 Beta1, most of them are because of https://issues.redhat.com/browse/WFCORE-5143 . There were also a couple of failures due to https://issues.redhat.com/browse/RESTEASY-2711 which should be fixed now.

Hi @asoldano - there is also an issue related to GAL-325 and I am actually also removing ignores for rx Java related tests because those could work as well when including the package explicitly.
Finally I am looking for a way to have also those tests inspecting the logs included.
I hope to be done by tomorrow.

fabiobrz:
retest this please

fabiobrz:
Hi @asoldano and @ronsigal - I've pushed last changes to this PR and I see one test failing on ubuntu/JDK 1.8/WF19.1.0.Final.
The same test is not failing on same OS/WFLY version but JDK 11.
Locally the same test is passing with RHEL7/JDK 1.8/WFLY 19.1.0.FInal:
```
$ mvn --version
Apache Maven 3.6.1 (d66c9c0b3152b2e69ee9bac180bb8fcc8e6af555; 2019-04-04T21:00:29+02:00)
Maven home: /usr/local/apache-maven/apache-maven-3.6.1
Java version: 1.8.0_232, vendor: Oracle Corporation, runtime: /home/fburzigo/.sdkman/candidates/java/8.0.232-open/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "3.10.0-957.12.2.el7.x86_64", arch: "amd64", family: "unix"

$ cd testsuite

$mvn clean verify -am -pl integration-tests -Dmaven.repo.local=/home/fburzigo/projects/git/marekkopecky/Resteasy/local-repo -Dserver.version=19.1.0.Final -Dtest=org.jboss.resteasy.test.microprofile.restclient.RestClientProxyRedeployTest
...
[INFO] --- maven-surefire-plugin:2.19.1:test (default-test) @ resteasy-integration-tests ---

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.jboss.resteasy.test.microprofile.restclient.RestClientProxyRedeployTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.539 sec - in org.jboss.resteasy.test.microprofile.restclient.RestClientProxyRedeployTest

Results :

Tests run: 2, Failures: 0, Errors: 0, Skipped: 0

[INFO] 
[INFO] --- maven-surefire-plugin:2.19.1:test (disable-jsonb-client) @ resteasy-integration-tests ---

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.jboss.resteasy.test.microprofile.restclient.RestClientProxyRedeployTest
Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.443 sec - in org.jboss.resteasy.test.microprofile.restclient.RestClientProxyRedeployTest

Results :

Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
...
```
So I asked to retest the PR but I don't think this error should block. WDYT?

Regarding WFLY 21.0.0.Beta1, we have QE internal pipelines runs against several JDKs/RHELs and there the only failing tests (18) are due to https://issues.redhat.com/browse/WFCORE-5143, as Alessio noticed as well.
Also new `ts.bootable` executions are showing the same outcome.

I am not sure what the plans are for https://github.com/wildfly/wildfly-core/pull/4345 to be merged and a new WFCORE release + WFLY 21 Beta2 are bit maybe this could turn all the tests green for WFLY 21.
WDYT?
CC @ehsavoie

fabiobrz:
Closed and re-opened to trigger CI checks again.

PR #2540, opened 2020-09-23
[WFLY-13790] added sync block

rsearls:
https://issues.redhat.com/browse/WFLY-13790 Memory leak caused by 
            org.eclipse.yasson.internal.JsonBinding


asoldano:
Replaced by #2542

PR #2560, opened 2020-10-03
RESTEASY-2722 - Fix check on wildcard subtype

rsvoboda:
Fix repeated check	on wildcard subtype against the same instance

https://issues.redhat.com/browse/RESTEASY-2722


asoldano (commented):
Thanks @rsvoboda , nice catch!

PR #2562, opened 2020-10-05
[RESTEASY-2725] Remove dependency from jackson to the jaxb provider

Postremus:
Some applications only need jackon, and where previously forced to exclude the jaxb-provider on their own. This compile dependency is now removed.

I also added jaxb-provider back to modules, which before only had jackson provider. This is to keep the same behavior for jose-jwt and resteasy-links.


Postremus:

asoldano (approved):
Thanks and sorry for the late follow-up

PR #2563, opened 2020-10-10
[RESTEASY-2728] Clients running in a resource method throw safer WebA…

ronsigal:
…pplicationExceptions


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
```diff
@@ -191,7 +208,19 @@ public ClientInvocation(final ResteasyClient client, final URI uri, final Client
             }
          }
          if (status >= 300 && status < 400)
-            throw new RedirectionException(response);
+         {
+            Config config = ConfigProvider.getConfig();
+            boolean originalBehavior = config.getOptionalValue(ResteasyContextParameters.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR, boolean.class).orElse(false);
```
asoldano, line 213: Can this be moved to a private method to be invoked from the "if" below (in order to avoid duplication below and benefit from the short-circuit behaviour of the || operator)?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
```diff
@@ -214,36 +243,75 @@ public ClientInvocation(final ResteasyClient client, final URI uri, final Client
...
-         case 415 :
-            throw new NotSupportedException(response);
-         case 500 :
-            throw new InternalServerErrorException(response);
-         case 503 :
-            throw new ServiceUnavailableException(response);
-         default :
-            break;
+      Config config = ConfigProvider.getConfig();
+      boolean originalBehavior = config.getOptionalValue(ResteasyContextParameters.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR, boolean.class).orElse(false);
```
asoldano, line 247: Same as above

resteasy-jaxrs/src/main/java/org/jboss/resteasy/client/exception/ResteasyBadRequestException.java
```diff
@@ -0,0 +1,103 @@
...
+
+import javax.ws.rs.core.Response;
+
+/**
+ * A runtime exception indicating a {@link javax.ws.rs.core.Response.Status#BAD_REQUEST
+ * bad client request}.
+ *
+ * @author Sergey Beryozkin
+ * @author Marek Potociar
+ * @since 2.0
```
asoldano, line 11: These should be removed, here and in all the other exception classes being added in this commit

resteasy-jaxrs/src/main/java/org/jboss/resteasy/client/exception/ResteasyClientErrorException.java
```diff
@@ -0,0 +1,194 @@
...
+ * A base runtime application exception indicating a client request error
+ * (HTTP {@code 4xx} status codes).
+ *
+ * @author Marek Potociar
+ * @since 2.0
+ */
+public class ResteasyClientErrorException extends ResteasyWebApplicationException {
+
+    private static final long serialVersionUID = -4101970664444907990L;
+
```
asoldano, line 31: Is the UID here actually properly generated? It seems to match the one of JAX-RS ClientErrorException. Please double check the UIDs for all other exceptions being added in this PR.

asoldano (changes requested):

FroMage (changes requested):
LGTM, besides the changes that @asoldano asked for.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
ronsigal: Nice. Done.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
ronsigal: Done.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/client/exception/ResteasyBadRequestException.java
ronsigal: Hey @asoldano,

I wanted to ask you about that. I copied these exceptions almost bit by bit from the originals, except for the name change and some fiddling with the response field. That's why I left them in. Do you feel confident about removing them?

Similarly, the original classes have an Oracle EULA. It looks like I removed most but not all of them. What about that?

-Ron

resteasy-jaxrs/src/main/java/org/jboss/resteasy/client/exception/ResteasyClientErrorException.java
ronsigal: Good catch - I didn't think about that. It turns out that most of them remain unchanged, but a few are different. Done.

ronsigal:
Thanks for looking at this, @FroMage.

ronsigal:
I've added some material to the User Guide, so, except for the question about javadoc and EULAs, this pull request should be done.

ronsigal:
I'm going to port all of this to master branch and create a Feature Request to include this in 3.14.0.FInal.

jamezp:
Hi @ronsigal. Looking at this I had an idea I'd like to run past you. What do you think of more of a wrapper approach. Something like https://github.com/resteasy/Resteasy/compare/3.14...jamezp:RESTEASY-2728. Not all tests pass with that as I know it's not 100% correct yet, but I wanted to run the concept past you.

melloware (approved):

ronsigal:
Closing. Replaced by https://github.com/resteasy/Resteasy/pull/2595.

PR #2567, opened 2020-10-15
[RESTEASY-2733]:Fix SseEventInputImpl wrongly split the event from st…

jimma:
…ream


jimma:
Close this PR as this is included in Mp rest client 2.0 support patch.

PR #2569, opened 2020-10-16
To fix all mp rest client 20 tck failures 

jimma:


asoldano:
I believe this is not needed anymore at this point.

PR #2570, opened 2020-10-16
[RESTEASY-2694] (3.11) Adding support for running test suite against bootable JAR

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2694 for the 3.11 branch, currently used by EAP 7.3.x

There's still one Bouncycastle related integration test failure and only when WildFly 21.0.0.Final is being used. 
The test passes with WildFly 20 Final. Besides it seems to be affecting just bootable JAR executions + WFLY 21, see https://issues.redhat.com/browse/WFLY-13903?focusedCommentId=15335605&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15335605.
Anyway the test is not ignored or skipped because the destination (3.11) branch is not used by WildFly any more, hence not an issue here.

Draft because there are failures in MP REST Client TCK, @asoldano, @ronsigal, @marekkopecky do you have any thought?

*UPDATE*: same TCK failures when running the TS from an unchanged `3.11` branch, so this should not block this PR. 
Removing the `Draft` status.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,6 +30,7 @@
         <surefire.system.args>-Xms512m -Xmx512m ${modular.jdk.args}</surefire.system.args>
         <!-- Plugins versions -->
         <version.org.jacoco.plugin>0.7.9</version.org.jacoco.plugin>
+        <version.org.wildfly.jar.plugin>2.0.0.Final</version.org.wildfly.jar.plugin>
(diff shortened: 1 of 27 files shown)


fabiobrz:
Now the Bouncycastle test case is passing after updating the deployment-structure for the case.

PR #2481, opened 2020-07-29
[RESTEASY-2620] added SSE support and testcases

rsearls:

PR #2574, opened 2020-10-20
[RESTEASY-2640] MicroProfile REST Client 2.0

asoldano:
https://issues.redhat.com/browse/RESTEASY-2640


asoldano:
FYI @radcortez @kenfinnigan

radcortez:
Great! Thank you!

We also need to update Config to 2.0. I've already did the changes for RESTEasy, but we also need to wait until we have a SR Config 2.0 release.

radcortez:
We need this:
https://github.com/radcortez/Resteasy/commit/ca3cd6020b05b9509867c824783e039172c3c982

MP Config introduced a breaking change. I'll open the PR once SR Config 2.0 is released.

asoldano:
@radcortez , ok, fine, do you know when the Config SR 2.0 is planned for? Speaking of your proposed changes, I'm wondering if it would be possible to include them already, without the MP Config dependency bump and the @Overrides on the additional methods.

@kenfinnigan @radcortez is the new dependency on rxjava2 OK to you / from a Quarkus perspective? should anyone else review that?

radcortez:
> @radcortez , ok, fine, do you know when the Config SR 2.0 is planned for? Speaking of your proposed changes, I'm wondering if it would be possible to include them already, without the MP Config dependency bump and the `@Overrides` on the additional methods.

Yes, we should be able to proceed with the changes without the bump. I'll propose a PR for it.

Unfortunately, there are still a few open issues on MP Config. I was hoping to move these forward, to release SR Config 2.0 with the correct dependency update. On the SR side, we are ready.

> @kenfinnigan @radcortez is the new dependency on rxjava2 OK to you / from a Quarkus perspective? should anyone else review that?

Quarkus is already including rxjava2:
https://github.com/quarkusio/quarkus/blob/957069411cd1d823a3a2bf68f50735ae05ffb41d/bom/application/pom.xml#L97

I guess we just need to make sure that the versions align.

radcortez:
Here: 
https://github.com/resteasy/Resteasy/pull/2577

kenfinnigan:
Not sure about rxjava2.

@cescoffier, @FroMage what are your thoughts?

FroMage:
I think we don't want it by default, since we prefer mutiny.

asoldano:
> I think we don't want it by default, since we prefer mutiny.

In that case, I think we need to make the MpPublisherMessageBodyReader implementation pluggable. Actually, @jimma previously provided an implementation based on project reactor. We would need one based on mutiny and we should then resolve it through service loader or equivalent.

cescoffier:
Yes, we don't want Rx Java 2. First, Rx Java 2 is EOL, then in Quarkus we prefer Mutiny and we would like to avoid having multiple reactive programming libraries in the classpath if we can avoid it. 

@asoldano I can provide an implementation for Mutiny.

jimma:
> In that case, I think we need to make the MpPublisherMessageBodyReader implementation pluggable. Actually, @jimma previously provided an implementation based on project reactor. We would need one based on mutiny and we should then resolve it through service loader or equivalent.

There is no Munity shipped in WildFly, and it only includes rxjava2 and rxjava3.  If we can include Munity in Wildfly ,then the implementation with Munity by default will be enough. @kabir @bstansberry  Is there any issue we include Munity in WildFly now ?

cescoffier:
@jimma I know that @kabir is already working with Mutiny but in the context of an extension pack. Note that Mutiny has no dependency (except Reactive Streams which is already a dependency from RX)

kabir:
@jimma - @cescoffier is correct that I have a feature pack with the reactive specs, which include Mutiny. That FP lives here https://github.com/wildfly-extras/wildfly-mp-reactive-feature-pack.

However, I am working on adding Mutiny along with the needed things for WildFly 22 (hopefully). This currently lives in my branch at https://github.com/kabir/wildfly/tree/reactive-port. More specifically, the module including Mutiny is currently (I might move things around a bit to be more grouped to keep it a bit cleaner, but looking again I think this one is fine) [here](https://github.com/kabir/wildfly/blob/reactive-port/galleon-pack/src/main/resources/modules/system/layers/base/io/smallrye/reactive/mutiny/main/module.xml)

kabir:
@jimma in case I wasn't clear, use the same module name and it should be ok :)

FroMage:
Originally, RESTEasy only depended on `CompletionStage` for singles, and ReactiveStreams's `Publisher` for streams, and did not require any implementation. We even have an SPI for turning those to/from RxJava1, 2, Mutiny… Why can't that SPI be used here as well?
Did you require an implementation of `Publisher`?

jimma:
> However, I am working on adding Mutiny along with the needed things for Reactive Messaging for Kafka to WildFly 22 (hopefully). This currently lives in my branch at https://github.com/kabir/wildfly/tree/reactive-port. More specifically, the module including Mutiny is currently (I might move things around a bit to be more grouped to keep it a bit cleaner, but looking again I think this one is fine) here

Thanks for update @kabir

jimma:
> Originally, RESTEasy only depended on CompletionStage for singles, and ReactiveStreams's Publisher for streams, and did not require any implementation. We even have an SPI for turning those to/from RxJava1, 2, Mutiny… Why can't that SPI be used here as well?
Did you require an implementation of Publisher?

Yes. We need an implementation of Publisher.

cescoffier:
@jimma can you point me where you need an implementation?

jimma:
> @jimma can you point me where you need an implementation?

@cescoffier  We need create an Publisher for client :  https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/RsSseClient.java#L38 

This is the test to create this client : https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/BasicReactiveStreamsTest.java#L80

cescoffier:
Ok, but here you just return a `Publisher` (interface), that does not mean you need to ship an implementation. Do you know where do you create the instance of these `Publisher` created? Is it in some kind of SPI (I guess I already have implemented that SPI for Mutiny, or at least part of it) or is it build-in?

jimma:
@cescoffier  It is created in MpPublisherMessageBodyReader:  https://github.com/jimma/Resteasy/blob/mp-rest-client-20-multiny/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/publisher/MpPublisherMessageBodyReader.java#L52.  ( I just changed to munity to implement this Reader).

This reader is registered in Resteasy to convert the SseEventInput to Publisher.

cescoffier:
@jimma Thanks! That's exactly what I was looking for. 

How do you want to handle the back-pressure? In other words, what happens (or what's expected to happen) if you have to emit too many items and the downstream can't keep up? Do you want to buffer? How large is your buffer? Do you want to fail? Drop? 

Also, if you fail, the final `complete` is useless. Failures are terminal. Not a big deal with this implementation, as it will just do nothing in the `complete()` method. 

I can propose to you to implement a specific implementation of `Publisher` - not requiring any reactive programming library. Of course, using Mutiny is fine (ok, I may be a bit biased). Still, depending on your answer to the back-pressure question, I can create a dedicated implementation of `Publisher` doing exactly what you are doing, without requiring any dependency.

asoldano:
> I can propose to you to implement a specific implementation of `Publisher` - not requiring any reactive programming library. Of course, using Mutiny is fine (ok, I may be a bit biased). Still, depending on your answer to the back-pressure question, I can create a dedicated implementation of `Publisher` doing exactly what you are doing, without requiring any dependency.

The `Publisher` implementation not requiring any reactive programming library here looks like the way to go, imho. Ideally, we could then provide alternative implementations using Mutiny, RxJava or anything else, letting the user / integrating project decide what to rely on.

jimma:
>  How do you want to handle the back-pressure? In other words, what happens (or what's expected to happen) if you have to emit too many items and the downstream can't keep up? Do you want to buffer? How large is your buffer? Do you want to fail? Drop?

@cescoffier  I think we need buffer here, but with limited buffer size like 512 and make it configurable. If there are too many items, we remove the oldest items to make sure the latest server sent events are not dropped.

cescoffier:
@asoldano Ok, how do you want to proceed? I can work on this Today (during my afternoon). 
@jimma That sounds good, a configured size buffer (512 by default) with a drop oldest when full (do we want to log when that happens?).

jimma:
> @jimma That sounds good, a configured size buffer (512 by default) with a drop oldest when full (do we want to log when that happens?).
Yes. Log this is good.

asoldano:
> @asoldano Ok, how do you want to proceed? I can work on this Today (during my afternoon).

Feel free to go ahead, thanks :-)

cescoffier:
@asoldano @jimma https://github.com/resteasy/Resteasy/pull/2578.
I have 2 questions:

* How can I configure the `provider`
* How can I add more tests, especially to verify the overflow behavior

ronsigal:
Sorry I'm a little late to the party, but my reaction was similar to @FroMage's. The JAX-RS client framework has a mechanism for returning arbitrary reactive types by way of RxInvokers, and we already have several of them. I haven't looked into this, but would it be possible to connect MP REST Clients to RxInvokers?

ronsigal:
To elaborate, this
```
	@Test
	public void testFlowable() throws Exception {
		TestResourceIntf client = builder.baseUrl(new URL("http://localhost:8081")).build(TestResourceIntf.class);
		Flowable<String> flowable = client.getFlowable();
		flowable.subscribe((String s) -> System.out.println(s),
				           (Throwable t) -> t.printStackTrace());
	}

	@Path("/")
	@RegisterRestClient(baseUri="http://localhost:8081")
	public interface TestResourceIntf {
		@Path("flow")
		@GET
		public Flowable<String> getFlowable();
	}

	@Path("/")
	public static class TestResource {
		@Path("flow")
		@GET
		public Flowable<String> getFlowable() {
			return Flowable.fromArray("one", "two");
		}
	}
```
works on the current master branch up to the point that a FlowableRxInvokerImpl is found and it subscribes to an SseEventSource, which seems promising. There's a mismatch concerning ClientRequestFilters, though. I haven't tracked that down yet.

Anyway, just a thought.

ronsigal:
By the way, the pure RESTEasy version works:
```
	@Test
	public void testResteasyFlowable() throws Exception {
		ResteasyClient resteasyClient = (ResteasyClient) ResteasyClientBuilder.newClient();
		TestResourceIntf client = resteasyClient.target("http://localhost:8081").proxy(TestResourceIntf.class);
		Flowable<String> flowable = client.getFlowable();
		flowable.subscribe((String s) -> System.out.println(s),
				(Throwable t) -> t.printStackTrace());
	}

	@Path("/")
	public interface TestResourceIntf {
		@Path("flow")
		@Stream
		@GET
		public Flowable<String> getFlowable();
	}

	@Path("/")
	public static class TestResource {
		@Path("flow")
		@Stream
		@GET
		public Flowable<String> getFlowable() {
			return Flowable.fromArray("one", "two");
		}
	}
```
Note that I added @Stream to the getFlowable() method, which tells RESTEasy to use SSE to send elements of the Flowable back to the client.

cescoffier:
@ronsigal with your version, how is the `Publisher` created? 
Basically, what's there: https://github.com/jimma/Resteasy/blob/mp-rest-client-20-multiny/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/publisher/MpPublisherMessageBodyReader.java?rgh-link-date=2020-10-22T08%3A21%3A42Z#L52 ?

jimma:
> @asoldano @jimma #2578.
> I have 2 questions:
> 
> * How can I configure the `provider`

@cescoffier  There is service provider file to register this to Resteasy runtime : https://github.com/resteasy/Resteasy/blob/mp-rest-client-20/resteasy-client-microprofile-base/src/main/resources/META-INF/services/javax.ws.rs.ext.Providers

> * How can I add more tests, especially to verify the overflow behavior

There is tck test for this Publisher implementation : https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/sse/BasicReactiveStreamsTest.java

We can reuse this to create a test to verify our implementation.  I can help add a test for this to check the overflow behavior later today or Monday.

jimma:
> Sorry I'm a little late to the party, but my reaction was similar to @FroMage's. The JAX-RS client framework has a mechanism for returning arbitrary reactive types by way of RxInvokers, and we already have several of them. I haven't looked into this, but would it be possible to connect MP REST Clients to RxInvokers?

@rsearls  did some work with this approach : https://github.com/resteasy/Resteasy/pull/2481/files.  From the change set , we have to create RxInvoker<Publisher> and RxInvokerProvider etc.  And we also need to think about this still works for  client api : 
rx(), and I am not sure if this new added thing will impact the the functionality we already have.  Now the MPPublisherMessageReader approach is a bit simple, we only need on provider to get this work.

ronsigal:
Hi everyone,

I'm not pushing hard on my suggestion because @jimma and @rsearls know the resteasy-client-microprofile code better than I do. But I'll add a couple of remarks.

re: "how is the Publisher created?"

In JAX-RS, there's a hook to allow reactive extensions: 
```
Client client = ClientBuilder.newClient();
FlowableRxInvoker invoker = client.target(generateURL("/get/thing")).request().rx(FlowableRxInvoker.class);
```
The implementation of FlowableRxInvoker in RESTEasy creates a Flowable by subscribing to an SseEventSource. Similarly, ObservableRxInvokerImpl creates an Observable, FluxRxInvokerImpl creates a Flux, etc. Currently, the implemtation of rx() looks for an exact match, but that could be modified to look for a subclass of Publisher. I.e., 
```
PublisherRxInvoker invoker = client.target(generateURL("/get/thing")).request().rx(PublisherRxInvoker.class);
```
or, for a specific Publisher implementation
```
PublisherRxInvoker invoker = client.target(generateURL("/get/thing")).request().rx(FlowableRxInvoker.class);
```

re: "From the change set , we have to create RxInvoker and RxInvokerProvider etc."

I'm just wondering if the existing FlowableRxInvoker, etc., are enough. They all produces Publishers

ronsigal:
Just as an aside, the @Stream annotation in RESTEasy says that when a resource method returns a Publisher, the contents should be streamed (via SSE in RESTEasy). Otherwise, the conents are accumulated and sent in a single message. I'm wondering if that might be a reasonable addition to MP REST Client.

ronsigal:
Changing ClientHelper.getRxInvokerProviderFromReactiveClass() to
```
   protected RxInvokerProvider<?> getRxInvokerProviderFromReactiveClass(final Class<?> clazz)
   {
	   if (getReactiveClassesForWrite() == null) return null;
	   Map<Class<?>, Class<? extends RxInvokerProvider<?>>> map = getReactiveClassesForWrite();
	   for (Entry<Class<?>, Class<? extends RxInvokerProvider<?>>> entry : map.entrySet()) {
		   if (clazz.isAssignableFrom(entry.getKey())) {
			   RxInvokerProvider rip = rpf.createProviderInstance(entry.getValue());
			   if (rip != null) {
				   return rip;
			   }
		   }
	   }
	   return null;
   }
```
the following works:
```
	@Path("/")
	@RegisterRestClient(baseUri="http://localhost:8081")
	public interface TestResourceIntf {
		@Path("flow")
		@Stream
		@GET
		public Publisher<String> getFlowable();
	}

	@Path("/")
	public static class TestResource {
		@Path("flow")
		@Stream
		@Produces("application/json")
		@GET
		public Publisher<String> getFlowable() {
			return Flowable.fromArray("one", "two");
		}
	}

	@Test
	public void testResteasyPublisher() throws Exception {
		ResteasyClient resteasyClient = (ResteasyClient) ResteasyClientBuilder.newClient();
		TestResourceIntf client = resteasyClient.target("http://localhost:8081").proxy(TestResourceIntf.class);
		Publisher<String> publisher = client.getFlowable();
		publisher.subscribe(new Subscriber<String>() {
			int i;
			Subscription sub;
			@Override
			public void onSubscribe(Subscription s) {
				System.out.println("onSubscribe()");
				sub = s;
				sub.request(2);
			}
			@Override
			public void onNext(String t) {
				System.out.println(t);
				i++;
			}
			@Override
			public void onError(Throwable t) {
				t.printStackTrace();
			}
			@Override
			public void onComplete() {
				// TODO Auto-generated method stub
				System.out.println("i: " + i);
			}
		});
		Thread.sleep(11111111);
		System.out.println("testResteasyPublisher() done");
	}
```
works. In other words, the FlowableRxInvokerImpl serves to produce a Publisher. This is JAX-RS, of course, not MP REST Client.

ronsigal:
With a couple of little changes in resteasy-client-microprofile,
```
	@Path("/")
	@RegisterRestClient(baseUri="http://localhost:8081")
	public interface TestResourceIntf {
		@Path("flow")
		@Stream
		@GET
		public Publisher<String> getFlowable();
	}

	@Test
	public void testFlowable() throws Exception {
		TestResourceIntf client = builder.baseUrl(new URL("http://localhost:8081")).build(TestResourceIntf.class);
		Publisher<String> publisher = client.getFlowable();
		publisher.subscribe(new Subscriber<String>() {
			@Override
			public void onSubscribe(Subscription s) {
				s.request(2);
			}
			@Override
			public void onNext(String t) {
				System.out.println(t);
			}
			@Override
			public void onError(Throwable t) {
				t.printStackTrace();
			}
			@Override
			public void onComplete() {
			}
		});
	}
```
works. The changes probably aren't very nice. I'm doing some kludgy stuff to prevent NPEs because the relationship between resteasy-client-microprofile and the SSE stuff isn't quite right. Consider this a POC.

ronsigal:
I've created an experimental pull request, https://github.com/resteasy/Resteasy/pull/2580, with an implementation of SSE in MP REST Client without a Publisher MBW. There are just a few changes:
1. The filters in MP REST Client depend on a ClientInvoker, so I've made the ClientInvoker created by MP REST Client available to SseEventSourceImpl. Note that it only works for FlowRxInvokerImpl right now.
2. ClientHelper.getRxInvokerProviderFromReactiveClass() will look for a subclass of the reactive class rather than an exact match.
3. FlowableRxInvokerProvider.isProviderFor() checks for isAssignable() instead of an exact match.

ronsigal:
A couple more remarks.

1. I think Clement's idea of having a built-in Publisher implementation is elegant with respect to implementing the MicroProfile REST Client 2.0. Maybe it could go into resteasy-core so that it could be available to users? A "lightweight" Publisher ... ?

2. Having a MessageBodyReader for Publisher satisfies the MicroProfile REST Client 2.0 requirements. Given that someone might want the richness of a Flowable or a Flux or whatever, it would be a nice extension to be able to specify a particular implementation of a Publisher. Also given that the heart of MpPublisherMessageBodyReader is very similar to FlowableRxInvokerImpl.eventSourceToObservable() [should have been eventSourceToFlowable, but never mind], I think it might be reasonable to make it possible to plug one of FlowableRxInvokerImpl, FluxRxInvokerImpl, etc. into MpPublisherMessageBodyReader. But that's an implementation detail.

So, basically, I'm in agreement with the current direction.

cescoffier:
It would be great to be able to transform the producer Publisher into the type desired by the user (Flowable, or Mutiny's Multi...). We had this request a few times in Quarkus.

asoldano:
Thanks everybody. I've included the latest changes by Clement and Jim here and I'm merging this PR. @ronsigal , please create a jira for the enhancements you've been describing above.

ronsigal:
Hi @cescoffier,

re: "It would be great to be able to transform the producer Publisher into the type desired by the user (Flowable, or Mutiny's Multi...). We had this request a few times in Quarkus."

Just to clarify, 
```
Client client = ClientBuilder.newClient();
FlowableRxInvoker invoker = client.target(...).request().rx(FlowableRxInvoker.class);
Flowable<String> flowable = (Flowable<String>) invoker.get(String.class);
```
and
```
ResteasyClient resteasyClient = (ResteasyClient) ResteasyClientBuilder.newClient();
TestResourceIntf client = resteasyClient.target(...).proxy(TestResourceIntf.class);
Flowable<String> flowable = client.getFlowable();
```
work in core RESTEasy. It's just the MP REST Client that needs work.

ronsigal:
re: "please create a jira for the enhancements you've been describing above."

https://issues.redhat.com/browse/RESTEASY-2745

PR #2575, opened 2020-10-20
Ensure that multiple impls of UriBuilder don't cause a CCE

geoand:
This is needed because the client should be usable
regardless of the server implementation of JAX-RS
that might be in use


geoand:
It would be great to get this in 4.x. I assume I need to do another PR for that? If so, what branch would it target?

geoand:
cc @asoldano

asoldano (approved):

asoldano:
@geoand can you clarify which branch you're actually referring to? Ideally, next version of RESTEasy from master would be 4.6.0 and it would be meant for Quarkus.

geoand:
And would that be created from `master`? If so, then I'm all good :)

asoldano:
> And would that be created from `master`? If so, then I'm all good :)

Yes it will

geoand:
Thanks 👍

PR #2582, opened 2020-10-26
Add test for https://github.com/resteasy/Resteasy/pull/2578

jimma:
This PR is rebased with https://github.com/resteasy/Resteasy/pull/2578 and changes  the following things: 
1. Add executor to SSEPublisher to avoid block the Publisher emit
2. Add system property "resteasy.microprofile.sseclient.buffersize" to configure the Subscription queue size.
3. Add test for https://github.com/resteasy/Resteasy/pull/2578 and configure the Subscription queue size to 5 to check the overflow behavior.

### DIFF
--- a/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -20,6 +20,8 @@ import org.jboss.resteasy.microprofile.client.header.ClientHeaderProviders;
 import org.jboss.resteasy.microprofile.client.header.ClientHeadersRequestFilter;
 import org.jboss.resteasy.microprofile.client.impl.MpClient;
 import org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl;
+import org.jboss.resteasy.microprofile.client.publisher.MpPublisherMessageBodyReader;
+//import org.jboss.resteasy.microprofile.client.publisher.MpPublisherMessageBodyReader;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.spi.ResteasyUriBuilder;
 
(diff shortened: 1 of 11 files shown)


jimma:
@cescoffier  I added one test for SSEPublisher to check the overflow behavior for https://github.com/resteasy/Resteasy/pull/2578, please review.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -20,6 +20,8 @@
 import org.jboss.resteasy.microprofile.client.header.ClientHeadersRequestFilter;
 import org.jboss.resteasy.microprofile.client.impl.MpClient;
 import org.jboss.resteasy.microprofile.client.impl.MpClientBuilderImpl;
+import org.jboss.resteasy.microprofile.client.publisher.MpPublisherMessageBodyReader;
+//import org.jboss.resteasy.microprofile.client.publisher.MpPublisherMessageBodyReader;
```
cescoffier, line 24: Shouldn't we remove this commented line?

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/publisher/SSEPublisher.java
```diff
@@ -0,0 +1,290 @@
...
+     * Reads the server-sent event from the {@code input} and pass them to the processor.
+     * The processor handles the downstream requests and buffer/drop items according to them.
+     *
+     * @param processor the stream
+     * @param input     the SSE input
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private void pump(final SSEProcessor processor, final SseEventInputImpl input) {
+        Map<Class<?>, Object> contextDataMap = ResteasyContext.getContextDataMap();
+        executor.execute(() -> {
```
cescoffier, line 66: should it handle rejection?

cescoffier (changes requested):
LGTM. I've made 2 minor comments.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
jimma: Thanks. Fixed.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/publisher/SSEPublisher.java
jimma: Do you mean handle executor injection ?

cescoffier: if the executor is stopped (for whatever reason), you cannot submit another task. You will receive a `RejectedExecutionException`. What should we do in this case? Log and return a failed publisher? Log and return a `Publisher` just emitting the completion signal?

jimma: @cescoffier Thanks for pointing out this issue.  I modified to this to create a new thread to run this if there is RejectedExecutionException thrown.

cescoffier (approved):

asoldano (approved):

PR #2583, opened 2020-10-27
Fix bootable jar profile for branch 3.11

tommaso-borgato:
At present, because of the use of property `forward.compatibility` for mutual exclusive profiles `default` and `forward.compatibility` we cannot exclude test modules `integration-tests`, `integration-tests-spring`, `microprofile-tck` from being executed for bootable jar

### DIFF
--- a/testsuite/pom.xml
+++ b/testsuite/pom.xml
@@ -94,7 +94,7 @@
             <id>default</id>
             <activation>
                 <property>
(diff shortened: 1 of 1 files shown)


tommaso-borgato:
@asoldano, @fabiobrz this prevent unwanted modules from being executed when testing bootable jar

asoldano (approved):

PR #2585, opened 2020-10-30
[RESTEASY-2749] (3.11) Remove mail layer tests when executing bootable JAR test profile

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2749

This is to remove mail layer tests when executing bootable JAR test profile because so that it can be tested against EAP XP 2 feature pack, which currently does not provide such layer.

### DIFF
--- a/testsuite/integration-tests/pom.xml
+++ b/testsuite/integration-tests/pom.xml
@@ -264,7 +264,6 @@
                                         <layer>h2-default-datasource</layer>
                                         <layer>undertow-legacy-https</layer>
                                         <layer>ejb</layer>
-                                        <layer>mail</layer>
                                     </layers>
                                 </configuration>
                             </execution>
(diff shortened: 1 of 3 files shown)


fabiobrz:
Hi @asoldano - I've checked the the CI checks detail and it seems to me that this only fails:

* because of https://issues.redhat.com/browse/RESTEASY-2748, which is fixed by https://github.com/resteasy/Resteasy/pull/2586

* because of errors in the TCK execution which where already there and are not affected by this PR.

We have this PR running in internal pipelines and the mail layer issue is solved, hence I think this could be merged. WDYT?

asoldano (commented):
Thanks @fabiobrz , the changes look ok. However, I think we should have a corresponding PR for 3.14 before merging this.

fabiobrz:
Hi @asoldano - I filed a PR (https://github.com/resteasy/Resteasy/pull/2592) against the `3.14` branch, however those changes are in advance and are tested just on WildFly 21 Final, as current XP 2 base EAP (7.3.4) depends on RESTEasy 3.11 and the tests code base would not be aligned with it.

asoldano (approved):

PR #2586, opened 2020-10-30
[RESTEASY-2748] (3.11) Fixing Bouncycastle test deployment structure

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2748

This is to comply with the EAP 7.3.x Bouncycastle module structure and to prevent the test deployment which use such configuration to fail.

### DIFF
--- a/testsuite/integration-tests/src/test/resources/jboss-deployment-structure-bouncycastle.xml
+++ b/testsuite/integration-tests/src/test/resources/jboss-deployment-structure-bouncycastle.xml
@@ -2,10 +2,7 @@
 <jboss-deployment-structure>
     <deployment>
         <dependencies>
-            <module name="org.bouncycastle.bcmail" services="import"/>
-            <module name="org.bouncycastle.bcpg" services="import"/>
-            <module name="org.bouncycastle.bcpkix" services="import"/>
-            <module name="org.bouncycastle.bcprov" services="import"/>
+            <module name="org.bouncycastle" services="import"/>
         </dependencies>
     </deployment>
 </jboss-deployment-structure>


fabiobrz:
Hi @asoldano - I checked the CI checks details and it seems that those failed because of known TCK runner failures. 
The issue with the Bouncycastle module configuration when deploying which affected a couple of tests (`BlacklistedMediaTypeTest`, `CryptoTest`) is solved by this PR hence I'd say it could be merged. WDYT?

CC @ronsigal

asoldano (commented):
Thanks @fabiobrz . I'd rather see a PR for 3.14 and/or master before merging this; in particular, I'm wondering about the results with both latest and previous WildFly target containers, which have different Bouncycastle libraries. I guess you might need to have different deployments / tests for different target containers.

fabiobrz:
> Thanks @fabiobrz . I'd rather see a PR for 3.14 and/or master before merging this; in particular, I'm wondering about the results with both latest and previous WildFly target containers, which have different Bouncycastle libraries. I guess you might need to have different deployments / tests for different target containers.

Thanks @asoldano - what you suggest seems reasonable to me and deserves more investigation.
In the meantime I filed this: https://issues.redhat.com/browse/WFLY-14065 so let's see how it goes there.

fabiobrz:
So @asoldano - regarding the `3.14` branch - this seems to be affecting the configuration of the Galleon provisioning when running the Bootable JAR test execution.
The test in that branch runs fine against WildFly 20 and 21, so I filed a different RESTEasy issue and PR, which should fix the `3.14` branch for executions against the latest WildFly server versions:

* https://issues.redhat.com/browse/RESTEASY-2769

* https://github.com/resteasy/Resteasy/pull/2593

asoldano (approved):

PR #2587, opened 2020-10-30
[RESTEASY-2747] (3.11) Adding documentation about mutually exclusive usage of -Delytron and -Dts.standalone.microprofile properties

fabiobrz:
Fixes https://issues.redhat.com/browse/RESTEASY-2747

The -Delytron and -Dts.standalone.microprofile properties are mutually exclusive because both the related profiles need to set the server configuration file property value to the one of a custom file.
Besides, elytron is the default when dealing with microprofile. Hence there's no need for any configuration to be added, which is what the elytron profile does.
This PR is for updating documentation accordingly.

### DIFF
--- a/testsuite/README.MD
+++ b/testsuite/README.MD
@@ -97,6 +97,9 @@ the usage of `standalone-microprofile.xml` configuration file:
 
 > mvn clean verify -Dserver.home=PATH_TO_WIDLFLY_HOME -Dts.standalone.microprofile
 
+Please notice that the `-Dts.standalone.microprofile` property usage is not compatible with `-Delytron` property since both will try to set 
+a custom value for the server configuration file property, leading to unexpected results.
+
 ### Bootable jar testing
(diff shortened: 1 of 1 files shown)


asoldano (commented):
Is this an issue on 3.14 and master as well? If that's the case, please create PRs and link them here

fabiobrz:
> Is this an issue on 3.14 and master as well? If that's the case, please create PRs and link them here

Thanks @asoldano. I'll verify the cases for `3.14` and `master` and will file one or more PRs accordingly.

fabiobrz:
Hi @asoldano - I checked the `master` and there's no `standalone.microprofile.profile` profile there which could conflict with the `elytron` one, hence no PR was filed for the change to be back-ported there.
Instead the `3.14` is affected by the issue and I've cherry-picked this commit there: https://github.com/resteasy/Resteasy/pull/2591

asoldano (approved):

PR #2588, opened 2020-11-01
RESTEASY-2734 The current WADL GRAMMAR generation logic incorrectly clean the schema types

liweinan:


ronsigal:
Hey @liweinan,

Could you give me a quick hint about what problem you are solving?

Thanks,
Ron

liweinan:
@ronsigal This PR is to address two issues:

- The previous `ResteasyWadlDefaultResource ` is not `abstract`, so its path can't be customized by users.(and the document is updated to address this: https://github.com/resteasy/Resteasy/pull/2589)
- The previous implementation will cleanup the `GRAMMAR` section incorrectly which is bug that has to be fixed.

asoldano (approved):

PR #2590, opened 2020-11-10
[RESTEASY-2767] Make MP Config optional

andymc12:
Resolves https://issues.redhat.com/browse/RESTEASY-2767

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ target
 **/.checkstyle
 **/.factorypath
 
+# VS Code
+###################
+.vscode
 
 # Packages #
 ############
--- a/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/WhiteListPolymorphicTypeValidatorBuilder.java
+++ b/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/WhiteListPolymorphicTypeValidatorBuilder.java
@@ -2,8 +2,7 @@ package org.jboss.resteasy.plugins.providers.jackson;
 
 import java.util.StringTokenizer;
 
-import org.eclipse.microprofile.config.Config;
-import org.jboss.resteasy.microprofile.config.ResteasyConfigProvider;
+import org.jboss.resteasy.config.ResteasyConfig;
 
 import com.fasterxml.jackson.databind.jsontype.BasicPolymorphicTypeValidator;
 
@@ -15,7 +14,7 @@ public class WhiteListPolymorphicTypeValidatorBuilder extends BasicPolymorphicTy
 
    public WhiteListPolymorphicTypeValidatorBuilder() {
       super();
-      Config c = ResteasyConfigProvider.getConfig();
+      ResteasyConfig c = ResteasyConfig.instance;
       String allowIfBaseType = c.getOptionalValue(BASE_TYPE_PROP, String.class).orElse(null);
       if (allowIfBaseType != null) {
          StringTokenizer st = new StringTokenizer(allowIfBaseType, ",", false);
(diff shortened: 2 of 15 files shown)


rsearls (approved):

asoldano (commented):
Thanks, the PR here looks OK, though I'm not sure yet this is the solution we should be going with; I've commented at https://issues.redhat.com/browse/RESTEASY-2767?focusedCommentId=15521287&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15521287 , can you please have a look? /cc @ronsigal

andymc12:
Hi @asoldano - sorry for the delay, was out on an extended Thanksgiving vacation.  

I looked at the 3.X code you pointed out in the JIRA and I wonder if it would actually work without MP Config - at least the APIs.  Do you know if it has been tested that way?

Here are some faults I see:
```
public class ResteasyConfig
{
   private Config config;
...
   public ResteasyConfig()
   {
      try
      {
         Class.forName("org.eclipse.microprofile.config.spi.ConfigSource");
         Class.forName("org.jboss.resteasy.microprofile.config.ServletConfigSourceImpl");
         config = ResteasyConfigProvider.getConfig();
      }
      catch (Throwable e)
      {
         // Leave config == null.
      }
   }
...
```
I think that having a `Config` field (that is a MP Config API class) will prevent the `ResteasyConfig` class from loading unless MP Config APIs are on the classpath.  Even though it sets it to null in the constructor, I think the JDK will still try to load the `Config` class when initializing the `ResteasyConfig` class.

Next, (assuming that the class can still be loaded) I see things like:
```
   public String getValue(String propertyName)
   {
      return config == null ? null : config.getOptionalValue(propertyName, String.class).orElse(null);
   }
...
   public Iterable<String> getPropertyNames()
   {
      return config == null ? null : getPropertyNames0();
   }
```
which will always return null if there is no MP Config.

The approach that I have here is much more extensible.  It would allow containers (like Open Liberty) to plug-in their own config mechanism.  If we want both 3.X and 4.X to use the same config mechanism, I could attempt to backport what I have here to the 3.X stream.

PS - please let me know if I should respond in the JIRA too - I'm still learning the RESTEasy workflow.

asoldano:
@ronsigal please see above ^

ronsigal:
Hey @andymc12,

Thank you for this. To your comments on the RESTEasy 3.x version, two words: "Umm, yep". I'm pretty sure I wrote the code to make MP Config optional, so ... that's embarrassing. I didn't think about the classloading issue.

As for "which will always return null if there is no MP Config", actually, there's more to it. A typical parameter access looks like
```
      ResteasyConfiguration context = ResteasyProviderFactory.getContextData(ResteasyConfiguration.class);
      if (context != null)
      {
         String s = context.getParameter(SUPPRESS_VIOLATION_PATH);
         if (s != null)
         {
            suppressPath = Boolean.parseBoolean(s);
         }
      }
```
The implementations of ResteasyConfiguration (e.g., org.jboss.resteasy.plugins.server.servlet.ServletBootstrap) inherit
```
  protected ResteasyConfig resteasyConfig = ResteasyConfigFactory.getConfig();
```
from ConfigurationBootstrap and then (ServletBootstrap) do this:
```
   public String getParameter(String name)
   {
      String val = resteasyConfig.getValue(name);
      if (val == null) val = config.getInitParameter(name); // ServletConfig
      if (val == null) val = super.getParameter(name);      // ServletContext
      return val;
   }
```
So, if MP Config isn't available (or so I intended), RESTEasy falls back to the web.xml parameters.

After my initial panic, I talked to James Perkins about the issue, and he ended up submitting an alternative  pull request (https://github.com/resteasy/Resteasy/pull/2639), which includes the "fallback" mechanism. In any case, I appreciate that you brought up the issue, and especially appreciate that you discovered the classloading problem.

Also, if you figure out the RESTEasy workflow, please let me know. ;-)

ronsigal:
I've merged https://github.com/resteasy/Resteasy/pull/2639, so I'll close this one. Thanks, @andymc12, for pushing this forward.

andymc12:
Sounds good - thanks @ronsigal!

PR #2595, opened 2020-11-16
[3.14] [RESTEASY-2728] Clients running in a resource method throw safer WebApplicationException

jamezp:

This may replace #2563 if we decide to go with this approach.


resteasy-jaxrs/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * @param response the response to sanitize.
+     *
+     * @return the new response
+     */
+    static Response sanitize(final Response response) {
+        return Response.status(response.getStatus())
+                .allow(response.getAllowedMethods())
+                // TODO (jrp) should we do this?
+                .entity(response.getEntity())
```
ronsigal, line 139: Originally, I wanted to just eliminate the Response by default, but Sergey argued that, once the exception got thrown up to the container, the Response would just be random bits. I was a little concerned that even so, the bits were escaping, but I didn't feel strongly enough about it to push back on the point.

jamezp: I feel that at least the status is important. The media type seemed fine too and the `javax.ws.rs.NotAllowedException` throws an `IllegalArgumentException` if the headers don't contain an `Allow` header. The entity I question a bit.

ronsigal: Hmmm, I guess there are two scenarios.

1. If the WAE escapes to the container, then i think that the fewer bits escape, the better. Some day I want to learn to be a hacker and figure out how to get to those bits, but, as Aragorn said, "It is not this day". ;-)

2. If the developer catches the WAE, they might want to have access to the whole Response.

Well, I guess if they want to access the whole Response, they can turn on the old behavior. That could be an argument for omitting the entity.

jamezp: Yeah that is my concern with it. From the JIRA the headers seemed to be the biggest issue, but in theory this could leak something like class names. However that could be considered a development error allowing that to escape as well.

ronsigal: Well, our "philosophy", such as it is, has been to impose safety checks. I mean, we could just say, "catch the damn exception", but the new behavior is meant to head off problems if someone hasn't done that. So, I think it's best to err on the side of safety.

By the way, the discussion on the JAX-RS issue https://github.com/eclipse-ee4j/jaxrs-api/issues/917 has leaned the other way. They seem to be more concerned about not breaking anything, at least until JAX-RS 4.0.

jamezp: Yes that makes sense to me. I'll make that change and see how it goes.

I saw that. I can see where they're coming from. The concern being breaking running applications.

ronsigal: Yeah, I agree. So far I got three thumbs up when I suggested changing the behavior in JAX-RS 4.0. Oh, one of them is you!! ;-)

ronsigal:
Replaced by https://github.com/resteasy/Resteasy/pull/2623.

PR #2596, opened 2020-11-17
[RESTEASY-2771] Decode App Path portion of RequestURL

andymc12:
This resolves JIRA [RESTEASY-2771](https://issues.redhat.com/browse/RESTEASY-2771) by decoding the incoming servlet request URL.  Thus requests like `GET /app%21/resource` will match app/resource combinations like:

```
@ApplicationPath("app%21")
public MyApp1 extends Application {}
```

```
@ApplicationPath("app!")
public MyApp2 extends Application {}
```

```
@Path("resource")
public MyResource {

   @GET
   public String get() {
      return "Hello World!";
   }
}
```

### DIFF
--- a/resteasy-core/pom.xml
+++ b/resteasy-core/pom.xml
@@ -62,6 +62,19 @@
         </dependency>
 
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.9.5</version>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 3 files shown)


andymc12:
Ugh... the failures from this certainly indicate that the fix is not as simple as I had hoped.  I'll move it to draft state for now.

rsearls:
This PR look fine to me.

asoldano (commented):
Thanks @andymc12

PR #2599, opened 2020-11-25
[3.11] [RESTEASY-2728] Clients running in a resource method throw safer WebApplicationException

jamezp:


resteasy-jaxrs/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @param e the exception to possibly wrapped
+     *
+     * @return the wrapped exception or the original exception if the exception has already been wrapped the the
+     * wrapping feature is turned off
+     */
+    static WebApplicationException wrap(final WebApplicationException e) {
+        final ResteasyConfig config = ResteasyConfigFactory.getConfig();
+        final boolean originalBehavior = Boolean.parseBoolean(config.getValue(ResteasyContextParameters.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR,
+                ResteasyConfig.SOURCE.SERVLET_CONTEXT, "false"));
```
ronsigal, line 65: It looks like 3.11 is before we started using MicroProfile Config. Maybe

       ResteasyConfiguration config = ResteasyProviderFactory.getContextData(ResteasyConfiguration.class);
       boolean originalBehavior = false;
       if (config != null)
       {
          try
          {
             String s = config.getParameter(ResteasyContextParameters.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR);
             originalBehavior = (s == null ? true : Boolean.parseBoolean(s));
          }
          catch (Exception ex)
          {
             //
          }
       }

jamezp: Ah okay. Nice catch. I'll have a look at that. I suppose I should have at least tried to compile it :) My apologies.

jamezp: @ronsigal Should we also check the `getInitParameter()` as well?

ronsigal:
Oh, man, it's been a while since I looked at this. I think that the implementation of ResteasyConfiguration is ServletBootstrap, with

   public String getParameter(String name)
   {
      String val = config.getInitParameter(name);
      if (val == null) val = super.getParameter(name);
      return val;
   }

where config is a ServletConfig, and super.getParameter() looks at the ServletContext. So we should be OK.

jamezp:
@ronsigal Cool. Updated!

ronsigal:
I've started running the TCK with this PR.

ronsigal:
The TCK passed. I think there's going to be a problem with the RESTEasy tests since I used a system property to turn on the old behavior. It'll need a web.xml instead. I can work on it after dinner.

jamezp:
I don't think it will be an issue since a `WebApplicationException` would need to be caught and the response from the exception to be queried for a test to fail. I don't see any WildFly tests that catch a `WebApplicationException`. When I ran the test suite against nearly the same commit it passed for me.

ronsigal:
Oh, OK. I wish these checks would run, though.

jamezp:
This is something we can look into possibly doing. Kabir as done some work on testing third party stuff against WildFly. Worst case we could probably create a job on ci.wildfly.org.

ronsigal:
I'm seeing failures like

Tests run: 10, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.971 sec <<< FAILURE! - in org.jboss.resteasy.test.client.exception.ClientWebApplicationExceptionTest
testNoCatchNewBehaviorOldExceptions(org.jboss.resteasy.test.client.exception.ClientWebApplicationExceptionTest)  Time elapsed: 0.008 sec  <<< FAILURE!
java.lang.AssertionError: expected null, but was:<bar>

ronsigal:
java.lang.AssertionError: expected null, but was:<bar>
	at org.junit.Assert.fail(Assert.java:88)

jamezp:
Shoot okay. I will run the testsuite on this branch. I just ran it on the 3.14 branch. I should have done it on this one, sorry.

ronsigal:
Ah, a couple of the tests have finished, and they have that failure also. I was expecting trouble with requesting the old behavior, but these are new behavior tests ...

ronsigal:
I'm getting:
```
ClientWebApplicationExceptionMicroProfileProxyTest.testNoCatchNewBehaviorOldExceptions:233 expected null, but was:<bar>
ClientWebApplicationExceptionResteasyProxyTest.testNoCatchNewBehaviorOldExceptions:233 expected null, but was:<bar>
ClientWebApplicationExceptionTest.testNoCatchNewBehaviorOldExceptions:355 expected null, but was:<bar>
```
New behavior, old exceptions.

jamezp:
> I'm getting:
> 
> ```
> ClientWebApplicationExceptionMicroProfileProxyTest.testNoCatchNewBehaviorOldExceptions:233 expected null, but was:<bar>
> ClientWebApplicationExceptionResteasyProxyTest.testNoCatchNewBehaviorOldExceptions:233 expected null, but was:<bar>
> ClientWebApplicationExceptionTest.testNoCatchNewBehaviorOldExceptions:355 expected null, but was:<bar>
> ```
> 
> New behavior, old exceptions.

Weird. That means something may be working differently which seems odd. I'll have a look.

ronsigal:
It looks like sanitizing the Response would remove the headers ...

ronsigal:
Closing. Replaced by https://github.com/resteasy/Resteasy/pull/2602

PR #2601, opened 2020-11-25
[3.13] [RESTEASY-2728] Clients running in a resource method throw safer WebApplicationException

jamezp:


jamezp:
This appears to no longer be needed.

PR #2606, opened 2020-11-30
[RESTEASY-2710] added new class and test.

rsearls:


resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java
```diff
@@ -67,11 +64,15 @@ else if (clazz.isAnnotationPresent(Provider.class))
...
-      if (appClasses.size() == 0)
-      {
-         // todo make sure we can do this on all servlet containers
-         //handleNoApplicationClass(providers, resources, servletContext);
-         return;
+      if (appClasses.size() > 1) {
+         for (Class<?> app : appClasses) {
+            // App has provided an Application impl.  The local default one
+            // isn't needed.  Remove it.
+            if (app.getSimpleName().equals("DefaultApplicationClazz")) {
```
asoldano, line 71: I suggest creating a String constant using something like DefaultApplicationClazz.getClass().getName() and then use that to run the comparison here, so that a future refactoring of the class does not risk breaking this.

asoldano (changes requested):

ronsigal:
Hey @rsearls,

Note that Andy has made some changes to ServletContainerInitializer in https://github.com/resteasy/Resteasy/pull/2474:

https://github.com/resteasy/Resteasy/blob/1bf0e2396d460932b47637a6dd673bf83b2fa8d6/resteasy-servlet-initializer/src/main/java/org/jboss/resteasy/plugins/servlet/ResteasyServletInitializer.java

so you should update on top of that.

-Ron

ronsigal:
Ok, I merged https://github.com/resteasy/Resteasy/pull/2474.

emretanriverdi (approved):

jamezp:

PR #2624, opened 2020-12-07
[RESTEASY-2641] ported nonProxyHost code from master

rsearls:


resteasy-client-microprofile/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -173,21 +181,59 @@ public RestClientBuilder executorService(ExecutorService executor) {
         ClassLoader classLoader = aClass.getClassLoader();
 
         List<String> noProxyHosts = Arrays.asList(
-           ConfigProvider.getConfig().getOptionalValue("http.nonProxyHosts", String.class).orElse("localhost|127.*|[::1]").split("|"));
-        String proxyHost = ConfigProvider.getConfig().getOptionalValue("http.proxyHost", String.class).orElse(null);
+                getSystemProperty("http.nonProxyHosts", "localhost|127.*|[::1]").split("\\|"));
+        String envProxyHost = getSystemProperty("http.proxyHost", null);
```
ronsigal, line 185: @rsearls, 3.14 incorporates MicroProfile Config. Is there a reason to not use it here?

PR #2632, opened 2020-12-15
[RESTEASY-2781] Clients running in a resource method throw safer WebA…

ronsigal:
…pplicationException


ronsigal:
Hey @jamezp,

If you have time, could you give this a quick look? The only thing new is that I put the new Exceptions in resteasy-client-api, and added a ResteasyDeployment.onServer() method to resteasy-core-spi.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/DefaultResponseExceptionMapper.java
```diff
@@ -13,12 +18,15 @@ public Throwable toThrowable(Response response) {
...
+        return WebApplicationExceptionWrapper.wrap(new WebApplicationException("Unknown error, status code " + response.getStatus(), response));
     }
 
     @Override
     public boolean handles(int status, MultivaluedMap headers) {
-        return status >= 400;
+       final Config config = ConfigProvider.getConfig();
+       final boolean originalBehavior = config.getOptionalValue(ResteasyContextParameters.RESTEASY_ORIGINAL_WEBAPPLICATIONEXCEPTION_BEHAVIOR, boolean.class).orElse(false);
+       final boolean serverSide = ResteasyDeployment.onServer();
+       return status >= (originalBehavior || !serverSide ? 400 : 300);
```
jamezp, line 29: This looks different from how we did it in the 3.x branches.

resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
```diff
@@ -0,0 +1,151 @@
...
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.jboss.resteasy.spi.ResteasyDeployment;
+
+/**
+ * An interface which allows a {@link WebApplicationException} to be unwrapped.
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+public interface WebApplicationExceptionWrapper<T extends WebApplicationException> {
```
jamezp, line 47: Should this just be in the implementation and not the API?

resteasy-client-api/pom.xml
```diff
@@ -45,6 +45,21 @@
...
+        <dependency>
+            <groupId>io.smallrye.config</groupId>
+            <artifactId>smallrye-config</artifactId>
+            <exclusions>
+               <exclusion>
+                  <groupId>javax.annotation</groupId>
+                  <artifactId>javax.annotation-api</artifactId>
+               </exclusion>
+            </exclusions>
+        </dependency>
```
jamezp, line 62: This seems a little weird adding it to the API.

jamezp (commented):
@ronsigal Just some general comments.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/DefaultResponseExceptionMapper.java
ronsigal: I realized as I was releasing 3.11.3.Final that I had changed the test to include 3xx statuses, which makes sense, I think, since the same leakage problem applies. There was a deadline for 3.11.3.Final and I didn't have time to change that, but I did it for 3.14.0.Final. So asking for the original behavior affects this class too.

resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
ronsigal: Hmmm. The new Exceptions depend on it:
```
public class ResteasyBadRequestException extends BadRequestException implements WebApplicationExceptionWrapper<BadRequestException> {
```

ronsigal: Maybe there's a better way ...

resteasy-client-api/pom.xml
ronsigal: Well, it's related to your previous comment, since WebApplicationExceptionWrapper uses MP Config:
```
    static WebApplicationException wrap(final WebApplicationException e) {
        final Config config = ConfigProvider.getConfig();
        final boolean originalBehavior = config.getOptionalValue("resteasy.original.webapplicationexception.behavior", boolean.class).orElse(false);
```

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/DefaultResponseExceptionMapper.java
jamezp: Okay. Maybe not a huge deal, but that's going to be different for EAP 7.3 and EAP 7.4 then. I'll have to back-port that fix to the 3.12 and 3.13 PR's I guess too.

resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
jamezp: Well the more I think about it the example I'd done originally should work too. I forgot in general that the RESTEasy 4.x is not backwards compatible with 3.x. I should have designed this a little better because it does seem weird to have this in the API. It makes the API know too much about the implementation.

resteasy-client-api/pom.xml
jamezp: Yeah. I think we should find a way to keep this part in the implementation. It feels like implementation details are leaking into the API here.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/DefaultResponseExceptionMapper.java
ronsigal: Yeah, my bad.

resteasy-client-api/pom.xml
ronsigal: I'm not sure I'm following. "resteasy.original.webapplicationexception.behavior" is meant to be available to the user ...

ronsigal: In fact, maybe all of those user facing parameter names should be in resteasy-core-spi ... ?

resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
ronsigal: > the example I'd done originally should work too

What example is that?

jamezp: https://github.com/jamezp/Resteasy/commit/2a19fc5144a5f452b74cc43f4c159a30077b3024

resteasy-client-api/pom.xml
jamezp: The parameter is fine, but the way we retrieve is the implementation detail right? For example using the Config API is an implementation detail of where the property is stored.

ronsigal: Ah, I didn't see this when I just responded to your email. I guess in my mind, MP Config is like the US dollar ... the global standard. But, you're right, maybe next year will be using NanoProfile Config. ;-)

jamezp: Ha these days you never know :)

resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/WebApplicationExceptionWrapper.java
```diff
@@ -0,0 +1,151 @@
...
+     * the response is {@linkplain #sanitize(Response) sanitized}.
+     *
+     * @param e the exception to possibly wrapped
+     *
+     * @return the wrapped exception or the original exception if the exception has already been wrapped the the
+     * wrapping feature is turned off
+     */
+    static WebApplicationException wrap(final WebApplicationException e) {
+        final Config config = ConfigProvider.getConfig();
+        final boolean originalBehavior = config.getOptionalValue("resteasy.original.webapplicationexception.behavior", boolean.class).orElse(false);
```
jamezp, line 62: Thinking about this, would this take into account context parameters from a `web.xml`?

jamezp (approved):

PR #2633, opened 2020-12-19
[RESTEASY-629] Round down the datetime in seconds in evaluatePreconditions

essobedo:
Fix for https://issues.redhat.com/browse/RESTEASY-629

## Motivation

The method `evaluatePreconditions(Date)` is not useable as it is now when the input date has milliseconds because it will be compared with a date without milliseconds so it will always return `null` indicating that the content has changed which is incorrect.

## Modifications:

* Trims the milliseconds from the input date before comparing the dates in order to compare dates with the same precision
* Adds the corresponding integration tests


essobedo:
Please note that the build failures don't seem to be related to this change since they occur on other PRs

essobedo:
@ronsigal @asoldano Happy new year! 

Could you please tell me what do you think of this PR proposal?

asoldano (approved):

PR #2638, opened 2020-12-29
[RESTEASY-2798] Save accesses to volatile member

essobedo:
Fix for https://issues.redhat.com/browse/RESTEASY-2798

## Motivation

Accessing to a volatile member has a cost in term of performances so it could be interesting to reduce the total amount of accesses to volatile member when it is possible.

## Modifications:

* Lazy initializes by relying on a local variable to reduce volatile reads
* Use a local variable instead of doing several volatile reads within the same method
* Remove the modifier `volatile` from member fields that are only read and write in synchronized blocks
* Replace the modifier `volatile` by `final` when the fields are not modified
* Remove the modifier `volatile` from member fields that are read and write in pair with another `volatile` member field by relying on piggybacking


essobedo:
@asoldano What do you thing of this PR? Is it clear enough?

ronsigal:
Looks good to me. Thanks, @essobedo!

ronsigal (approved):

PR #2639, opened 2021-01-05
[RESTEASY-2767] Make the Eclipse MicroProfile Config API optional.

jamezp:
https://issues.redhat.com/browse/RESTEASY-2767

This is an alternate to the approach  in #2590 which fully makes the Eclipse MicroProfile Config optional for RESTEasy.

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/config/ConfigurationFactory.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/config/ConfigurationFactory.java
@@ -1,0 +1,96 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2021 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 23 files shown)


jamezp:
I've temporarily moved this to a draft because I think we need a few further changes.

jamezp:
I'm not sure why the resteasy-client-vertx tests are failing. I'm not able to reproduce this locally on my Windows VM.

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/config/ConfigurationFactory.java
```diff
@@ -0,0 +1,96 @@
...
+public interface ConfigurationFactory {
+
+    /**
+     * Returns a new factory. The factory with the lowest {@linkplain #priority() priority} will be selected.
+     *
+     * @return the new factory
+     *
+     * @throws RuntimeException if the service loader could not find a factory
+     */
+    static ConfigurationFactory getInstance() {
```
ronsigal, line 42: I'm stepping through to be sure I understand it all, and I've noticed that getInstance() keeps reloading the ServiceLoader. Shouldn't we store the ConfigurationFactory once it's loaded?

ronsigal: Well, the ServiceLoader javadoc says, " A service loader maintains a cache of the providers that have been loaded so far", so I guess it's ok.

jamezp: I thought a cache would be good as well TBH. However I wasn't sure how to push/pop this each time in the `ResteasyContext`. If we think there is a performance issue we should look more into this though.

ronsigal: I guess getting parameters is something that doesn't happen a lot, so maybe it's not a problem. In any case, we could look into it later.

I want to update the User Guide a little, but I can do that in a separate PR. I'll merge this now so I can run the TCK tonight.

jamezp: Excellent. This was thought as well. If we see performance issue we can fix it then.

PR #2652, opened 2021-01-14
[RESTEASY-2812] Fix MediaType cache poisoning with trailing semi-colon

SteveNewson:
The `MediaTypeHeaderDelegate` holds a static cache mapping string
representations of Media Types (e.g. application/json) to their
`javax.ws.rs.core.MediaType` equivalent. The reduces repeated parsing of
the same, or similar, Media Types; increasing performance of most
service calls by a small amount.

The class boils down to two methods: `parse(String)` and
`toString(MediaType)`. These two methods are not symmetric since the
parse method may receive additional sections in the Media Type string
(e.g. "application/json; charset=utf-8"). However, it is expected
that when passing a bare Media Type within the additional sections, it
should be symmetric.

Hence: `parse("application/json") == MediaType("application", "json");`

The cache introduces a bug, however, if the `parse` method is called
with a Media Type with a trailing semi-colon. This is technically not
valid for a `Content-Type` but it still can exist. When this happens,
the cache is effectively poisoned as `toString(MediaType("application",
"json"))` 
will now return "application/json;".

The fix here is to use `internalToString` to normalise the type being
put into the `reverseMap`. The `map` has to stay the same.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
@@ -69,11 +69,11 @@ public class MediaTypeHeaderDelegate implements RuntimeDelegate.HeaderDelegate<M
           result = internalParse(type);
           final int size = map.size();
           if (size >= MAX_MT_CACHE_SIZE) {
-              map.clear();
-              reverseMap.clear();
+              clearCache();
           }
+          final String normalisedType = internalToString(result);
           map.put(type, result);
-          reverseMap.put(result, type);
+          reverseMap.put(result, normalisedType);
       }
       return result;
    }
@@ -159,8 +159,7 @@ public class MediaTypeHeaderDelegate implements RuntimeDelegate.HeaderDelegate<M
           result = internalToString(type);
           final int size = reverseMap.size();
           if (size >= MAX_MT_CACHE_SIZE) {
-             reverseMap.clear();
-             map.clear();
+             clearCache();
           }
           reverseMap.put(type, result);
           map.put(result, type);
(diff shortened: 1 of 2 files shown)


SteveNewson:
This fixes the underlying problem for us, but I'm not necessarily happy with the implementation. I'm not hugely familiar with this part of the code so please forgive me if there is something I'm missing. However, there are some things about this area which bug me.

Firstly, I feel like `MediaType` should be a POJO which renders the `toString` literally. I'm not clear on what the value of having an override method for the `toString` in `MediaType` would be. A large part of this problem would be eliminated by just having `toString` behave in an obvious way (rather than redirecting through a number of layers back to the delegate).

I'm also not hugely convinced by the need to cache here at all. Without the history it's hard to see, but it feels like an optimisation with little benefit. I suspect a better implementation would be to just have this create `MediaType` objects on every call to `parse` rather than trying to cache them. There'd be a lot of garbage being created by this, but I'd expect that to be short-lived and pretty much what the JVM GC is good at cleaning.

If the cache should remain, then I'd also argue that separating the caching concern from the mapping concern would be sensible here. Apart from anything, it would make this class eminently simpler to test :).

SteveNewson:
I think the failed builds here are a red-herring. Since there's no way in GH Actions for me to trigger a rebuild of the failing job, I'll just trigger a dummy change.

SteveNewson:

rsearls (approved):

ronsigal:
Hi @SteveNewson,

Thank you for this fix. Nice work!!

The only think I would change is to move the tests to testsuite/unit-tests. 

> I'm not clear on what the value of having an override method for the toString in MediaType would be.

Well, it's defined in the JAX-RS, so the issue is out of scope here. Personally, I think it's nice that toString() creates a String representation of the MediaType.

> I'm also not hugely convinced by the need to cache here at all.

I wasn't involved in the introduction of the cache, but I believe that it was in response to some profiling.

ronsigal:
Well, clearCache() should be public.

ronsigal:
Replaced by https://github.com/resteasy/Resteasy/pull/2653.

PR #2654, opened 2021-01-25
[RESTEASY-2817] Use a shared thread pool for async request timeout ha…

asoldano:
…ndling and set min pool size to 1 to workaround https://bugs.openjdk.java.net/browse/JDK-8129861

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/Filter30Dispatcher.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/Filter30Dispatcher.java
@@ -10,8 +10,6 @@ import javax.servlet.annotation.WebFilter;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.NewCookie;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
 
 /**
  * @author <a href="mailto:bill@burkecentral.com">Bill Burke</a>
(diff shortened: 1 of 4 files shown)


asoldano:
@ronsigal (/cc @jimma , @liweinan , @rsearls ), please review this. Do you see any problem with the shared thread pool (to avoid having a new thread created for each deployment)?
Any alternative idea to deal with the issue here?

ronsigal:
I see that the javadoc for `ScheduledThreadPoolExecutor` says
```
While this class inherits from ThreadPoolExecutor, a few of the inherited tuning methods are not useful for it.
In particular, because it acts as a fixed-sized pool using corePoolSize threads and an unbounded queue,
adjustments to maximumPoolSize have no useful effect.
```
A fixed pool size of 1 doesn't seem right for a shared thread pool ...

asoldano:
@ronsigal I see your point given the javadoc above. Perhaps it would be better to still have a pool for each deployment, even if it means allocating more resources. WDYT?

ronsigal:
Hey @asoldano,

> Do you have any suggestion for the default size? the number of CPU cores?

It's really frustrating that I couldn't find a thread pool with scheduling that can also grow. 

Anyway, I've never thought about this before, but I see a lot of discussion out there. It looks like you can do this

> Runtime.getRuntime().availableProcessors();

which gives, not exactly the number of physical or, in the case of hyperthreading cpus, logical processors. It's the number of **available** cores, which can change if some processors are dedicated to some particular thread. But it seems like a reasonable starting point - give the program the ability to use as much resources as it can. If it's a problem, clever users can tune further.  So I give it +1. Which is not to say I know what I'm talking about here.

Also, I would change the parameter's name to RESTEASY_ASYNC_TIMEOUT_SCHEDULER_POOL_SIZE, since it's both MIN and MAX.

jamezp:
Resolved by #2881

PR #2663, opened 2021-02-01
[RESTEASY-2750] MicroProfile-Client classes and tests back ported fro…

rsearls:
…m master


rsearls:
resolved file merge issue in different branch

PR #2666, opened 2021-02-02
Resteasy 2750 micro profile rest 2.0 to 3.15 port merged

rsearls:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/config/MicroProfileConfigFilterTest.java
```diff
@@ -134,28 +129,12 @@ public void testContextInject() throws Exception {
...
       Assert.assertEquals(200, response.getStatus());
       Assert.assertEquals("text/plain", response.getHeaderString("Content-Type"));
    }
-
-   /**
-    * @tpTestDetails Verify that ConfigSource names are retrieved properly.
-    * @tpSince RESTEasy 3.12.0
-    */
-   @Test
-   public void testConfigSourceNames() throws Exception {
```
ronsigal, line 151: Does this not work any more?

ronsigal:
Hi @rsearls,

Wow, this is a lot. Thank you for handling it.

I guess I have one concern. There are some method signature changes in resteasy-client-microprofile-base and resteasy-client. Given that we're implementing a new major version MP REST Client, I think that's ok, although I'm not sure about documentation. As for the changes in resteasy-client, they're all in org.jboss.resteasy.client.jaxrs.internal.proxy, which "sounds" like it should be ok, but, technically, they're public classes ... .

-Ron

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/config/MicroProfileConfigFilterTest.java
rsearls: The tests in testsuite/integration-tests were updated to those that reside in master.  That test does not reside in master.

rsearls:
These are the minimum changes required to get mp-rest-client to run.

ronsigal:
> These are the minimum changes required

I don't doubt that. I'm wondering if it's possible to leave the original methods in place and just add the new versions.

ronsigal:
Hi @rsearls,

I added a section to the User Guide.

-Ron

ronsigal:
I also fixed the @tpSince versions.

I still have two quesions:

1. I don't know why you removed "<para><emphasis role="bold">Note.</emphasis> Related to the change described in section ..." from MicroProfile_Rest_Client.xml.   DefaultResponseExceptionMapper still has the code
> return status >= (originalBehavior || !serverSide ? 400 : 300);

2. Is there any problem restoring the original versions of the methods whose signature you changed?

Thanks,
Ron

rsearls:
I put the public methods back for proxy dir.

rsearls:
> I don't know why you removed "Note. Related to the change described in section ..." from 
> MicroProfile_Rest_Client.xml. DefaultResponseExceptionMapper still has the code

I was back porting from master.  Since 2.0 changes were additions assumed docs were additions only as well.

ronsigal:
> I put the public methods back for proxy dir.

Great, thanks.

> I was back porting from master. Since 2.0 changes were additions assumed docs were additions only as well.

Ah, I see. That bit came in with a different set of changes. I can put that back in.

Ok, this is great. As soon as the Feature Team agrees, I can merge this.

ronsigal:
In my last push to https://github.com/rsearls/Resteasy/tree/RESTEASY-2750-MicroProfileREST-2.0-to-3.15-port-MERGED, I accidentally wrote over the classes in resteasy-client with the restored methods. I've re-restored them.

ronsigal:
This work in this pull request is included in https://github.com/resteasy/Resteasy/pull/2677. Thanks, @rsearls.

PR #2669, opened 2021-02-05
[RESTEASY-2823]:Remove AnnotationArrayKey equals check from ResteasyJackson2Provider

jimma:


asoldano (commented):
Hi @jimma , besides the comment below, I just want to make sure that you thought carefully about the implications of changing / removing the AnnotationArrayKey concept. Please check this https://github.com/resteasy/Resteasy/commit/6866e3ff2cd74402fb7250c50690ab31c39271a2 if you didn't already and ensure the changes you're proposing here won't restore the problem solved in that commit nor what led Jackson having the AnnotationBundleKey abstraction. If you already checked this all and believe we're safe ... great ;-)

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java
```diff
@@ -72,16 +72,15 @@ public boolean isWriteable(Class<?> aClass, Type type, Annotation[] annotations,
...
       private int hash;
 
       private ClassAnnotationKey(final Class<?> clazz, final Annotation[] annotations)
       {
-         this.annotations = new AnnotationArrayKey(annotations);
          this.classKey = new ClassKey(clazz);
-         hash = this.annotations.hashCode();
-         hash = 31 * hash + classKey.hashCode();
+         this.annotations = annotations;
+         hash = Arrays.hashCode(this.annotations) + classKey.hashCode();
```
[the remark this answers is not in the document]
jimma: > Hi @jimma , besides the comment below, I just want to make sure that you thought carefully about the implications of changing / removing the AnnotationArrayKey concept. Please check this [6866e3f](https://github.com/resteasy/Resteasy/commit/6866e3ff2cd74402fb7250c50690ab31c39271a2) if you didn't already and ensure the changes you're proposing here won't restore the problem solved in that commit nor what led Jackson having the AnnotationBundleKey abstraction. If you already checked this all and believe we're safe ... great ;-)

I believe we don't have this problem because we completely remove the AnnotationArrayKey, but we still keep the code to 
check annotation equals with:
```
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;

        ClassAnnotationKey that = (ClassAnnotationKey) o;

        if (!annotations.equals(that.annotations)) return false;
        if (!classKey.equals(that.classKey)) return false;

        return true;
    }
```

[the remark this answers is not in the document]
jimma: > @jimma are you removing the multiply for 31 on purpose (if yes, why) ?

 Arrays.hashCode(this.annotations) is better than multiply with 31. Arrays.hashCode(annotations) simply does this:
```
    public static int hashCode(Object a[]) {
        if (a == null)
            return 0;

        int result = 1;

        for (Object element : a)
            result = 31 * result + (element == null ? 0 : element.hashCode());

        return result;
    }
```

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java
```diff
@@ -72,16 +72,15 @@ public boolean isWriteable(Class<?> aClass, Type type, Annotation[] annotations,
 
    private static class ClassAnnotationKey
    {
-      private AnnotationArrayKey annotations;
+      private  Annotation[] annotations;
       private ClassKey classKey;
       private int hash;
```
jamezp, line 77: If we're making a change we should make these all final :)

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java
```diff
@@ -72,16 +72,15 @@ public boolean isWriteable(Class<?> aClass, Type type, Annotation[] annotations,
...
 
       private ClassAnnotationKey(final Class<?> clazz, final Annotation[] annotations)
       {
-         this.annotations = new AnnotationArrayKey(annotations);
          this.classKey = new ClassKey(clazz);
-         hash = this.annotations.hashCode();
-         hash = 31 * hash + classKey.hashCode();
+         this.annotations = annotations;
```
jamezp, line 82: We should copy the array here since arrays are mutable. 
```
this.annotations = Arrays.copyOf(annotations, annotations.length);
```

There should be some handling for a `null` array too.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/AbstractBuiltResponse.java
```diff
@@ -212,19 +212,21 @@ public void setMetadata(MultivaluedMap<String, Object> metadata)
...
          {
             ann.add(annotation);
          }
+         for (Annotation annotation : methodAnnotations)
+         {
+            ann.add(annotation);
+         }
+         annotations = ann.toArray(new Annotation[ann.size()]);
+      } else {
+         annotations = methodAnnotations;
```
jamezp, line 228: This could be simplified to:
```
public void addMethodAnnotations(Annotation[] methodAnnotations)
{
   if (annotations != null)
   {
      if (methodAnnotations != null) {
         final int currentLen = annotations.length;
         annotations = Arrays.copyOf(annotations, currentLen + methodAnnotations.length);
         System.arraycopy(methodAnnotations, 0, annotations, currentLen, methodAnnotations.length);
      }
   } else if (methodAnnotations != null) {
      annotations = Arrays.copyOf(methodAnnotations, methodAnnotations.length);
   }
  ```

jamezp (commented):
I share @asoldano concerns about why the `AnnotationArrayKey` was created. I'll try to give it some thought too, but if you've already done that Jim then I won't :)

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java
jimma: +1.  Fixed.

providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java
jimma: The annotation arrays won't be changed after it calls into ResteasyJackson2Provider. This annotaitons arrays are actually fixed and filled with method annotaitons, so we don't need to do another array copy and null checks.

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/AbstractBuiltResponse.java
jimma: >       if (methodAnnotations != null) {
         final int currentLen = annotations.length;
         annotations = Arrays.copyOf(annotations, currentLen + methodAnnotations.length);
         System.arraycopy(methodAnnotations, 0, annotations, currentLen, methodAnnotations.length);
      }

If I understand correctly, the  Arrays.copyof() will create another array and copy , and System.arrayCopy will do another array copy when annotations and methodAnnotations are not null.   Will it be slower than only copy once with `annotations = ann.toArray(new Annotation[ann.size()])` ?

jamezp: `System.arraycopy()` should be fast as it's a native call. It's not really critical we do it this way, but it should allocate less resources as we directly use a new array instead of using a new array in the collection, then creating another new array in the `toArray()`.

jamezp:
@jimma Do we still want to do this? If so we should rebase this and get a new test run on it.

jamezp:
Closing is for now as it needs a rebase. If we still want to do this, a new PR should be opened.

PR #2675, opened 2021-02-08
[RESTEASY-2805] Remove dependency on MicroProfile Config.

ronsigal:
[RESTEASY-2805] Added uncommitted file.

### DIFF
--- a/resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
+++ b/resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
@@ -1,18 +1,24 @@
 package org.jboss.resteasy.microprofile.config;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Optional;
 
 import javax.servlet.ServletContext;
 
-import org.eclipse.microprofile.config.Config;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
 import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 public class ResteasyConfig
 {
-   private Config config;
+   private static final Method GET_CONFIG;
+   private static final Method GET_OPTIONAL_VALUE;
+   private static final Method GET_PROPERTY_NAMES;
+   private Object config;
 
    public enum SOURCE
    {
@@ -21,23 +27,59 @@ public class ResteasyConfig
       SERVLET_CONTEXT
    }
 
+   static {
+      Method getConfig;
+      Method getOptionalValue;
+      Method getPropertyNames;
+      try {
+          final ClassLoader classLoader = getClassLoader();
+          Class.forName("org.jboss.resteasy.microprofile.config.ServletConfigSourceImpl", false, classLoader);
+          final Class<?> configProvider = Class.forName("org.eclipse.microprofile.config.ConfigProvider", false, classLoader);
+          getConfig = configProvider.getDeclaredMethod("getConfig", ClassLoader.class);
+          final Class<?> clazz = Class.forName("org.eclipse.microprofile.config.Config", false, classLoader);
+          getOptionalValue = clazz.getDeclaredMethod("getOptionalValue", String.class, Class.class);
+          getPropertyNames = clazz.getDeclaredMethod("getPropertyNames");
+      } catch (Throwable ignore) {
+          getConfig = null;
+          getOptionalValue = null;
+          getPropertyNames = null;
+      }
+      GET_CONFIG = getConfig;
+      GET_OPTIONAL_VALUE = getOptionalValue;
+      GET_PROPERTY_NAMES = getPropertyNames;
+   }
+
    public ResteasyConfig()
    {
-      try
+      if (GET_CONFIG != null)
       {
-         Class.forName("org.eclipse.microprofile.config.spi.ConfigSource");
-         Class.forName("org.jboss.resteasy.microprofile.config.ServletConfigSourceImpl");
-         config = ResteasyConfigProvider.getConfig();
-      }
-      catch (Throwable e)
-      {
-         // Leave config == null.
+         try
+         {
+            config = GET_CONFIG.invoke(null, getClassLoader());
+         }
+         catch (Exception e)
+         {
+            // Leave config == null.
+         }
       }
    }
 
+   @SuppressWarnings("unchecked")
    public String getValue(String propertyName)
    {
-      return config == null ? null : config.getOptionalValue(propertyName, String.class).orElse(null);
+      if (config == null)
+      {
+         return null;
+      }
+      try
+      {
+         Optional<String> opt = (Optional<String>) GET_OPTIONAL_VALUE.invoke(config, propertyName, String.class);
+         return opt.orElse(null);
+      }
+      catch (IllegalAccessException | InvocationTargetException e) {
+         LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_OPTIONAL_VALUE);
+         return null;
+     }
    }
 
    public String getValue(String propertyName, SOURCE source)
@@ -72,6 +114,7 @@ public class ResteasyConfig
       }
    }
 
+   @SuppressWarnings("unchecked")
    public String getValue0(String propertyName, SOURCE source, String defaultValue)
    {
       if (config == null)
@@ -102,7 +145,15 @@ public class ResteasyConfig
       }
       else
       {
-         return config.getOptionalValue(propertyName, String.class).orElse(defaultValue);
+         try
+         {
+            Optional<String> opt = (Optional<String>) GET_OPTIONAL_VALUE.invoke(config, propertyName, String.class);
+            return opt.orElse(defaultValue);
+         }
+         catch (IllegalAccessException | InvocationTargetException e) {
+            LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_OPTIONAL_VALUE);
+            return defaultValue;
+        }
       }
    }
 
@@ -111,11 +162,19 @@ public class ResteasyConfig
       return config == null ? null : getPropertyNames0();
    }
 
+   @SuppressWarnings("unchecked")
    public Iterable<String> getPropertyNames0()
    {
       if (System.getSecurityManager() == null)
       {
-         return config.getPropertyNames();
+         try
+         {
+            return (Iterable<String>) GET_PROPERTY_NAMES.invoke(config);
+         }
+         catch (IllegalAccessException | InvocationTargetException e) {
+            LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_PROPERTY_NAMES);
+            return null;
+         }
       }
       else
       {
@@ -126,7 +185,14 @@ public class ResteasyConfig
                @Override
                public Iterable<String> run() throws Exception
                {
-                  return config.getPropertyNames();
+                  try
+                  {
+                     return (Iterable<String>) GET_PROPERTY_NAMES.invoke(config);
+                  }
+                  catch (IllegalAccessException | InvocationTargetException e) {
+                     LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_PROPERTY_NAMES);
+                     return null;
+                  }
                }
             });
          }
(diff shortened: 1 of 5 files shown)


ronsigal:
Hey @jamezp,

If you have a minute, could you take a look at this? I borrowed your reflexive idea, but left everything else alone, so there should be no behavior changes.

-Ron

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -1,18 +1,25 @@
...
 import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 public class ResteasyConfig
 {
-   private Config config;
+   private static boolean MP_CONFIG_PRESENT;
+   private static final Method GET_CONFIG;
+   private static final Method GET_OPTIONAL_VALUE;
+   private static final Method GET_VALUE;
```
jamezp, line 21: This should probably be deleted as it doesn't look like it's used.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -21,23 +28,50 @@
...
+         Optional<String> opt = (Optional<String>) GET_OPTIONAL_VALUE.invoke(config, propertyName, String.class);
+         return opt.orElse(null);
       }
-   }
-
-   public String getValue(String propertyName)
-   {
-      return config == null ? null : config.getOptionalValue(propertyName, String.class).orElse(null);
+      catch (IllegalAccessException | InvocationTargetException e) {
+         LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_VALUE);
```
jamezp, line 72: The argument should be `GET_OPTIONAL_VALUE`.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -102,20 +137,38 @@ public String getValue0(String propertyName, SOURCE source, String defaultValue)
...
       {
-         return config.getOptionalValue(propertyName, String.class).orElse(defaultValue);
+         try
+         {
+            final Object config = GET_CONFIG.invoke(null, getClassLoader());
+            Optional<String> opt = (Optional<String>) GET_OPTIONAL_VALUE.invoke(config, propertyName, String.class);
+            return opt.orElse(defaultValue);
+         }
+         catch (IllegalAccessException | InvocationTargetException e) {
+            LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_VALUE);
```
jamezp, line 147: The argument should be `GET_OPTIONAL_VALUE`.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -102,20 +137,38 @@ public String getValue0(String propertyName, SOURCE source, String defaultValue)
...
       if (System.getSecurityManager() == null)
       {
-         return config.getPropertyNames();
+         try
+         {
+            final Object config = GET_CONFIG.invoke(null, getClassLoader());
+            return (Iterable<String>) GET_PROPERTY_NAMES.invoke(config);
+         }
+         catch (IllegalAccessException | InvocationTargetException e) {
+            LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_VALUE);
```
jamezp, line 169: The argument should be `GET_PROPERTY_NAMES`.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -126,7 +179,15 @@ public String getValue0(String propertyName, SOURCE source, String defaultValue)
...
                public Iterable<String> run() throws Exception
                {
-                  return config.getPropertyNames();
+                  try
+                  {
+                     final Object config = GET_CONFIG.invoke(null, getClassLoader());
+                     return (Iterable<String>) GET_PROPERTY_NAMES.invoke(config);
+                  }
+                  catch (IllegalAccessException | InvocationTargetException e) {
+                     LogMessages.LOGGER.debugf(e, "Failed to invoke the configuration API method %s.", GET_VALUE);
```
jamezp, line 188: The argument should be `GET_PROPERTY_NAMES`.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -102,20 +137,38 @@ public String getValue0(String propertyName, SOURCE source, String defaultValue)
...
    public Iterable<String> getPropertyNames()
    {
-      return config == null ? null : getPropertyNames0();
+      return !MP_CONFIG_PRESENT ? null : getPropertyNames0();
    }
 
+   @SuppressWarnings("unchecked")
    public Iterable<String> getPropertyNames0()
    {
       if (System.getSecurityManager() == null)
```
jamezp, line 161: While we already do this, I'm not sure we should be wrapping this in a privileged action as this is publicly accessible.

jamezp (changes requested):
Is this something you asked me to port @ronsigal and I totally spaced it? If so I apologize.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfigProvider.java
```diff
@@ -1,22 +0,0 @@
-package org.jboss.resteasy.microprofile.config;
-
-import org.eclipse.microprofile.config.Config;
-import org.eclipse.microprofile.config.ConfigProvider;
-import org.eclipse.microprofile.config.spi.ConfigBuilder;
-import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
-
-public final class ResteasyConfigProvider
```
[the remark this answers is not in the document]
ronsigal: Hey @jamezp,

> Is this something you asked me to port

Not at all. You already did the hard part. :)

> This is a public class 

Good point. Thanks.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: You're right.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: CopyAndPasteException ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: Same ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: Same ...

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: Hmmm. But if I take that out, the method won't work if there's a SecurityManager ... .

ronsigal:
Well, that's a lot of mistakes for so little code. D'oh.

ronsigal:
I think it comes down to how classloading works in WildFly, and I can't find any documentation about eager vs lazy. @jamezp, how familiar are you with WF classloading? Or, who can we ask?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
jamezp: Well it would delegate the work to the config API. By doing it here we are allowing unchecked access to the property names. Currently for SmallRye this isn't an issue, but if they ever decide to add a check it could be an issue.

jamezp:
> I think it comes down to how classloading works in WildFly, and I can't find any documentation about eager vs lazy. @jamezp, how familiar are you with WF classloading? Or, who can we ask?

I'm not sure I follow. What is coming down to class loading? I'm fairly familiar with JBoss Modules and how it works in WildFly.

ronsigal:
> What is coming down to class loading?

Ah, sorry. I'm worried about ResteasyConfigProvider. I realize now that I built org.eclipse.microprofile.config.Config into the RESTEasy API:

>    public static Config getConfig() {
      return ConfigProvider.getConfig(ResteasyConfigProvider.class.getClassLoader());
   }

As of now, there is no reference to ResteasyConfigProvider outside of the class itself, so I'm wondering if WildFly is going to look for ResteasyConfigProvider in the course of loading RESTEasy. The Java Language Specification talks about resolving fields, but if there's no field or parameter with ResteasyConfigProvider type, will WildFly just ignore it?

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: > it would delegate the work to the config API

I'm not following. Isn't the call to GET_PROPERTY_NAMES wrapped in a PrivilegedExceptionAction because it's a reflective call?

jamezp:
Ah I see. Yes it will just be ignored unless it's explicitly referenced. In other words since it's not used in RESTEasy it shouldn't be a problem at all.

ronsigal:
>it will just be ignored unless it's explicitly referenced

Ah, perfect. That's a relief. I'll put the class back. Thanks!!

What about the PrivilegedExceptionAction? I'm still not clear about that.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
jamezp: `Method.invoke()` does not require special permissions so we'd override any permission checking in the `Confg.getPropertyNames()`.

jamezp:
> What about the PrivilegedExceptionAction? I'm still not clear about that.

I don't think we should be wrapping that in a privileged action. We don't for retrieving values.

ronsigal:
>I don't think we should be wrapping that in a privileged action. We don't for retrieving values.

My concern has been that, in the presence of a SecurityManager, we would be changing the previous behavior. But then I tried

         try {
            SecurityManager manager = new SecurityManager();
            System.setSecurityManager(manager);
            Config config = ConfigProvider.getConfig();
            System.out.println("done");
        } catch (Exception e) {
           e.printStackTrace();
        }
    }

and I got

    java.lang.IllegalStateException: No ConfigProviderResolver implementation found!
	 at org.eclipse.microprofile.config.spi.ConfigProviderResolver.loadSpi(ConfigProviderResolver.java:138)
	 at org.eclipse.microprofile.config.spi.ConfigProviderResolver.instance(ConfigProviderResolver.java:124)
	 at org.eclipse.microprofile.config.ConfigProvider.getConfig(ConfigProvider.java:85)
	 at org.jboss.resteasy.microprofile.client.SecurityTest2.main(SecurityTest2.java:20

Apparently, it couldn't access the file system. When I did the ConfigProvider.getConfig() __before__ setting the SecurityManager, it then failed on System.getEnv(). So, it seems that the issue goes beyond RESTEasy. I'm curious how smallrye deals with that, but it's not my concern right now.

One other thing. Given that the reflexive method calls are to __public__ methods, I don't think that there's even a need for PrivilegedExceptionActions. I'm going to take them out.

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
```diff
@@ -1,18 +1,24 @@
...
 import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 public class ResteasyConfig
 {
-   private Config config;
+   private static boolean MP_CONFIG_PRESENT;
+   private static final Method GET_OPTIONAL_VALUE;
+   private static final Method GET_PROPERTY_NAMES;
+   private static Object config;
```
jamezp, line 21: I don't think we want this to be static. There is a change we want a different one per deployment.

jamezp (changes requested):

resteasy-jaxrs/src/main/java/org/jboss/resteasy/microprofile/config/ResteasyConfig.java
ronsigal: Well, RESTEasy's servlet ConfigSources should find the appropriate ServletContext, etc., but it couldn't hurt. Ok, thanks.

jamezp (approved):
Looks good! Thanks @ronsigal

ronsigal:
Thanks for your help, @jamezp!!

PR #2678, opened 2021-02-15
Update CorsFilter.java

randymodo:
use getters instead of direct variable to get values


randymodo:
I am working on a project where we extended this class and have some extra logic including in the getters, the logic in our getters was being ignored and found that they were not being used

asoldano (approved):

asoldano:
@randymodo sorry for the late feedback. The changes are fine and straightforward, though as you might contribute more in the future, please note we require all contributions to be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 
Do you agree to these terms?

randymodo:
Yes I agree to these terms, is just saying that enough or is there something more official I need to do?

asoldano:
> Yes I agree to these terms, is just saying that enough or is there something more official I need to do?

It's fine like this. I'm merging, thanks.

PR #2680, opened 2021-02-22
[RESTEASY-2826] resteasy-vertx: Response Headers with String value of same type are overwritten

petrberan:
https://issues.redhat.com/browse/RESTEASY-2826

### DIFF
--- a/server-adapters/resteasy-vertx/src/main/java/org/jboss/resteasy/plugins/server/vertx/VertxHttpResponse.java
+++ b/server-adapters/resteasy-vertx/src/main/java/org/jboss/resteasy/plugins/server/vertx/VertxHttpResponse.java
@@ -146,7 +146,7 @@ public class VertxHttpResponse implements HttpResponse
                response.headers().add(key, delegate.toString(value));
             } else
             {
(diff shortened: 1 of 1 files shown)


petrberan:
Hi, I'm having problem with GH actions in this PR. I have multiple fails in Windows environment, but I believe that my change should not affect those tests in any way.

By looking at other PRs, they have exactly the same fail exceptions. Is this GH actions problem or it's caused by my changes and as such should I fix them?

asoldano:
> Hi, I'm having problem with GH actions in this PR. I have multiple fails in Windows environment, but I believe that my change should not affect those tests in any way.
> 
> By looking at other PRs, they have exactly the same fail exceptions. Is this GH actions problem or it's caused by my changes and as such should I fix them?

Hi @petrberan ,  sorry for the late feedback and thanks for the PR. You can disregard these Windows failures, they're being investigated and are unrelated to this PR.
Now, the change you're proposing is fine and straightforward, though as you might contribute more in the future, please note we require all contributions to be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0
Do you agree to these terms?

PR #2682, opened 2021-03-01
[RESTEASY-2842] DoPriv for file access in Mime4JWorkaround

andymc12:
This change resolves an issue when a Java 2 security manager is present.


asoldano (changes requested):
@andymc12 Thanks for the PR. It looks like CI is failing because of a checkstyle error, would you mind checking/fixing? Thanks!

PR #2688, opened 2021-03-09
[RESTEASY-2834] Do not use a static BeanManager. Use the one, where a…

jamezp:
…pplicable, from the client builder.

https://issues.redhat.com/browse/RESTEASY-2834


jamezp:
@ronsigal Let me know what other branches we need this in.

ronsigal:
Really, master is the only other active branch. I don't know why I said "other branches" on the JIRA.

PR #2689, opened 2021-03-09
[RESTEASY-2851]Fix Java 2 Security Permission in SseEventSourceScheduler.shutdownNow()

jim-krueger:


asoldano:
@jim-krueger , we require all contributions to be made under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 
Do you agree to these terms?

asoldano (approved):

ronsigal:
@jim-krueger, do you agree to the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0?

jim-krueger:
@ronsigal Wow, missed seeing this for a while sorry.  Yes, I agree to the terms of the ASL 2.0 License.

resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java
```diff
@@ -142,7 +145,13 @@ void shutdownNow()
...
          if (!scheduledExecutorService.isManaged())
          {
-            this.scheduledExecutorService.shutdownNow();
+            AccessController.doPrivileged(new PrivilegedAction<Void>() {
+              @Override
+              public Void run() {
+                scheduledExecutorService.shutdownNow();
+                return null;
+              }
+           });
```
jamezp, line 154: For performance sake we should do something like:
```java
if (System.getSecurityManager() == null) {
    scheduledExecutorService.shutdownNow();
} else {
    AccessController.doPrivileged(new new PrivilegedAction<Void>() {
        @Override
        public Void run() {
            scheduledExecutorService.shutdownNow();
            return null;
        }
    });
}
```

jamezp (changes requested):

jamezp:
Replaced by #3273 3273

PR #2711, opened 2021-03-31
[RESTEASY-2854] corrected when onError is called

rsearls:


asoldano:
@rsearls , what the jira for this PR? I can't find it in the title / commit message

PR #2717, opened 2021-04-02
[RESTEASY-2866] Ensure proxies implement all intfs

andymc12:
This PR resolves JIRA [RESTEASY-2866](https://issues.redhat.com/browse/RESTEASY-2866) by adding all implemented interfaces of the delegate class to the array of interfaces used when building the proxy class.


rsearls:
Please provide a integration test as well in order to confirm this change has not broken any other code.

rsearls (changes requested):
Please provide a integration test as well in order to confirm this change has not broken any other code.

andymc12:
Hi @rsearls - thanks for the review.  I added a new integration test that casts the proxy injected into `@Context Configuration config` into a RESTEasy-specific interface, `HeaderValueProcessor`.

asoldano (approved):

PR #2722, opened 2021-04-07
Only set the status of response with @ResponseStatus if it's not already set

geoand:
Fixes: https://github.com/quarkusio/quarkus/issues/16321

### DIFF
--- a/resteasy-spring-web/src/main/java/org/jboss/resteasy/spring/web/ResponseStatusContainerResponseFilter.java
+++ b/resteasy-spring-web/src/main/java/org/jboss/resteasy/spring/web/ResponseStatusContainerResponseFilter.java
@@ -1,5 +1,7 @@
 package org.jboss.resteasy.spring.web;
 
+import org.jboss.resteasy.spi.HttpResponseCodes;
+
 import javax.ws.rs.container.ContainerRequestContext;
 import javax.ws.rs.container.ContainerResponseContext;
 import javax.ws.rs.container.ContainerResponseFilter;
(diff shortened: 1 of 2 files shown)


geoand:
cc @asoldano

geoand:
The CI failure seems unrelated

geoand:
@asoldano would it possible to have a `4.5.10.Final` (or maybe `4.5.9.1.Final`?)  with this included?

jamezp:
I think a test for this would be nice. @liweinan can you have a look at this? I'm not too familiar with Spring.

resteasy-spring-web/src/main/java/org/jboss/resteasy/spring/web/ResponseStatusContainerResponseFilter.java
```diff
@@ -7,14 +7,18 @@
 
 public class ResponseStatusContainerResponseFilter implements ContainerResponseFilter {
 
-    private final int status;
+    private final int defaultStatusCode;
+    private final int newStatusCode;
 
-    public ResponseStatusContainerResponseFilter(final int status) {
-        this.status = status;
+    public ResponseStatusContainerResponseFilter(final int defaultStatusCode, final int newStatusCode) {
```
jamezp, line 13: This is a public type so we should keep the original constructor otherwise there is a potential migration issue. I don't know if we should just use 200 as the default or not, but I think we should use something like that.

geoand:
Unfortunately the problem that this fixes can be surfaced with RESTEasy on its own because the feature that causes it is provided by Quarkus.

resteasy-spring-web/src/main/java/org/jboss/resteasy/spring/web/ResponseStatusContainerResponseFilter.java
geoand: We can do that - I'll update the PR tomorrow

geoand: Fixed

liweinan:
I applied this PR in my local env:

image

And use the patched local build as dependency of `resteasy-spring-boot` project:

image

And I'll the tests passed:

image

So I approve this PR.

liweinan (approved):

geoand:
Thanks 👍

jamezp:
Part of this is up to the Quarkus team, but we may need to back port this to other branches if Quarkus wants to consume it.

geoand:
It would be good to have some kind of `4.5` release we can pick up in the `1.13` stream

PR #2723, opened 2021-04-08
[RESTEASY-1721] ResourceMethodInvoker checks if @Context HttpServletR…

ronsigal:
…esponse was written to.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
@@ -250,8 +250,9 @@ public class ContextParameterInjector implements ValueInjector
          this.delegate = delegate;
       }
 
-      public void write(byte b[]) throws IOException {
-          delegate.write(b;
+      public void write(byte b[]) throws IOException
+      {
+          delegate.write(b);
           outputStreamWasWritten = true;
       }
      
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
@@ -9,11 +9,15 @@ import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.spi.ValueInjector;
 import org.jboss.resteasy.spi.util.Types;
 
+import javax.servlet.ServletOutputStream;
+import javax.servlet.WriteListener;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.ext.Providers;
 import javax.ws.rs.sse.Sse;
 import javax.ws.rs.sse.SseEventSink;
+
+import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
@@ -37,6 +41,7 @@ public class ContextParameterInjector implements ValueInjector
    private ResteasyProviderFactory factory;
    private Type genericType;
    private Annotation[] annotations;
+   private boolean outputStreamWasWritten = false;
 
    public ContextParameterInjector(final Class proxy, final Class rawType, final Type genericType, final Annotation[] annotations, final ResteasyProviderFactory factory)
    {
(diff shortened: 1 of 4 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -195,4 +209,50 @@ public ClassLoader run() {
...
+      {
+         this.delegate = delegate;
+      }
+
+      @Override
+      public void write(int b) throws IOException
+      {
+         delegate.write(b);
+         outputStreamWasWritten = true;
+      }
```
asoldano, line 232: Is there a reason why you're not overriding `public void write(byte b[], int off, int len)` method as well?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/context/HttpServletResponseContextTest.java
```diff
@@ -0,0 +1,54 @@
...
+    * is bundled in the server.
+    * @tpSince RESTEasy 3.1.0
+    */
+   @Test
+   public void test() {
+      Client client = ClientBuilder.newClient();
+      Builder request = client.target(generateURL("/test")).request();
+      Response response = request.get();
+      Assert.assertEquals(200, response.getStatus());
+      Assert.assertEquals("context", response.readEntity(String.class));
```
asoldano, line 52: Better adding a client.close()

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
ronsigal: Well, write(byte b[]) and write(byte b[], int off, int len) both end up calling write():
```
    public void write(byte b[]) throws IOException {
        write(b, 0, b.length);
    }
   ...
    public void write(byte b[], int off, int len) throws IOException {
        Objects.checkFromIndexSize(off, len, b.length);
        // len == 0 condition implicitly handled by loop bounds
        for (int i = 0 ; i < len ; i++) {
            write(b[off + i]);
        }
    }
```

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/context/HttpServletResponseContextTest.java
ronsigal: Ah, yes, ok, good.

ronsigal: Done.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
jamezp: In the default implementation they do, but we can't know for sure that all delegates will do that. It could be buffered as well meaning `outputStreamWasWritten` won't be changed to `true` until a flush happens.

ronsigal: Ah, so someone could extend ServletOutputStream and overwrite the other write()'s. Ok, thanks.

jamezp: Potentially yes. IMO any delegating class should override and delegate all methods. Assumptions can't be made about the delegate.

ronsigal: Mmmm, yeah, ok. By "all methods" ... if someone overrides ServletOutputStream, it's really unknowable what other methods they could have added ... . I'm not sure what to do about that ... .

ronsigal: Well, not exactly unknowable ...

jamezp: As long as we override each method we don't have to care what they do :) We just simply delegate.

ronsigal: But suppose they add a new method that doesn't override some method in ServletOutputStream or OutputStream? I could find it by reflection, but then what? Maybe ContextOutputStream should itself be a proxy, which could catch all methods.

This is a job for Python. ;-)

jamezp: Well we can't really do anything about that. We only care about the base class we're wrapping :) And IMO no, we don't want a proxy for performance reasons. If a container adds it's own methods that aren't accessible from the base type that's on them :D

ronsigal: Yeah, a proxy in a proxy. 

Ok, ServletOutputStream has 17 methods. :)

Here I go.

ronsigal:
Replaced by https://github.com/resteasy/Resteasy/pull/2751.

PR #2727, opened 2021-04-09
[RESTEASY-2612] Support CompletableFuture resource method return type

spyrkob:
Issue: https://issues.redhat.com/browse/RESTEASY-2612

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
@@ -1025,9 +1025,7 @@ public class ResteasyProviderFactoryImpl extends ResteasyProviderFactory
         while (mapper == null) {
             if (asyncType == null)
                 break;
-            Map<Class<?>, AsyncResponseProvider> asyncResponseProviders = getAsyncResponseProviders();
-            if (asyncResponseProviders != null)
-                mapper = asyncResponseProviders.get(asyncType);
+            mapper = findProvider(asyncType, getAsyncResponseProviders());
             if (mapper == null)
                 asyncType = asyncType.getSuperclass();
         }
@@ -1040,9 +1038,7 @@ public class ResteasyProviderFactoryImpl extends ResteasyProviderFactory
         while (mapper == null) {
             if (asyncType == null)
                 break;
-            Map<Class<?>, AsyncClientResponseProvider> asyncClientResponseProviders = getAsyncClientResponseProviders();
-            if (asyncClientResponseProviders != null)
-                mapper = asyncClientResponseProviders.get(asyncType);
+            mapper = findProvider(asyncType, getAsyncClientResponseProviders());
             if (mapper == null)
                 asyncType = asyncType.getSuperclass();
         }
@@ -1056,9 +1052,7 @@ public class ResteasyProviderFactoryImpl extends ResteasyProviderFactory
         while (mapper == null) {
             if (asyncType == null)
                 break;
-            Map<Class<?>, AsyncStreamProvider> asyncStreamProviders = getAsyncStreamProviders();
-            if (asyncStreamProviders != null)
-                mapper = asyncStreamProviders.get(asyncType);
+            mapper = findProvider(asyncType, getAsyncStreamProviders());
             if (mapper == null)
                 asyncType = asyncType.getSuperclass();
         }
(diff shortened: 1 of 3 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
```diff
@@ -1115,17 +1115,18 @@ public <T extends Throwable> ExceptionMapper<T> getExceptionMapperForClass(Class
    public <T> AsyncResponseProvider<T> getAsyncResponseProvider(Class<T> type)
    {
       Class asyncType = type;
-      AsyncResponseProvider<T> mapper = null;
-      while (mapper == null)
+      if (getAsyncResponseProviders() == null)
```
stevenschlansker, line 1118: Should the three calls to `getAsyncResponseProviders()` be stored in a local and re-used?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/response/CompletionStageResponseTest.java
```diff
@@ -95,6 +95,26 @@ public void testText() throws Exception
       response.close();
    }
 
+   /**
+    * @tpTestDetails Resource method returns CompletableFuture<String>.
+    * @tpSince RESTEasy 4.7
+    */
+   @Test
+   public void testCompletabeFutureText() throws Exception
```
stevenschlansker, line 103: Typo, `Completabe`

spyrkob:
Thanks for the review @stevenschlansker, I updated and rebased the PR

resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
```diff
@@ -1021,17 +1021,16 @@ public <T extends Throwable> ExceptionMapper<T> getExceptionMapperForClass(Class
     //   @Override
     public <T> AsyncResponseProvider<T> getAsyncResponseProvider(Class<T> type) {
         Class asyncType = type;
```
jamezp, line 1023: This could be removed and the usage below could just be replaced by the `type` argument.

resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
```diff
@@ -1021,17 +1021,16 @@ public <T extends Throwable> ExceptionMapper<T> getExceptionMapperForClass(Class
...
+        for (Class<?> aClass : asyncResponseProviders.keySet()) {
+            if (aClass.isAssignableFrom(asyncType)) {
+                return asyncResponseProviders.get(aClass);
+            }
         }
-        return mapper;
+        return null;
     }
 
     public <T> AsyncClientResponseProvider<T> getAsyncClientResponseProvider(Class<T> type) {
```
jamezp, line 1036: We should probably do the same thing from above here. Possibly with the `getAsyncStreamProvider()` too.

resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
```diff
@@ -1021,17 +1021,16 @@ public <T extends Throwable> ExceptionMapper<T> getExceptionMapperForClass(Class
...
-            if (mapper == null)
-                asyncType = asyncType.getSuperclass();
+
+        final Map<Class<?>, AsyncResponseProvider> asyncResponseProviders = getAsyncResponseProviders();
+        if (asyncResponseProviders == null)
+            return null;
+        for (Class<?> aClass : asyncResponseProviders.keySet()) {
+            if (aClass.isAssignableFrom(asyncType)) {
+                return asyncResponseProviders.get(aClass);
+            }
```
jamezp, line 1031: I'm not sure we should remove the `type.getSuperClass()` checking. That's a change in behavior. My assumption is this is done so we don't check the `CompletionStatge`, but maybe I'm wrong.

jamezp (changes requested):

florian-signoret-ibm:
Hello,
Is there a plan to complete this PR?
Developers often forget to use `CompletionStage` instead of `CompletableFuture` and hit the issue.
Thank you!

jamezp:
> Hello, Is there a plan to complete this PR? Developers often forget to use `CompletionStage` instead of `CompletableFuture` and hit the issue. Thank you!

Hi @florian-signoret-ibm,
I plan on looking at this by the end of the week. I'll back port this to 6.2 and likely 5.0 as well.

resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/ResteasyProviderFactoryImpl.java
spyrkob: @jamezp sorry for delay, I must have missed the notification

I added back the type.getSuperClass check. I assumed it was an attempt to check if the return type extends  `CompletionStage` but maybe there's another case that it was meant to handle

jamezp (approved):

PR #2735, opened 2021-04-09
[RESTEASY-2868] Skip the Vert.x client tests on Windows or by system …

jamezp:
…property until RESTEASY-2868 is resolved.

https://issues.redhat.com/browse/RESTEASY-2868


jamezp:
This doesn't need to me merged if #2736 is merged.

PR #2622, opened 2020-12-05
[RESTEASY-2766] Upgrade Resteasy to support vert.x 4

mschorsch:
This PR upgrades Resteasy to support vert.x 4
https://issues.redhat.com/browse/RESTEASY-2766 and https://issues.redhat.com/browse/RESTEASY-2665

Note: Vert.x 4 has no final release yet. This PR makes the relevant changes in preperation for the final release.

PR #2736, opened 2021-04-10
[RESTEASY-2868] Use a handler to pause the client response while

jamezp:
the delegating input stream is created.

https://issues.redhat.com/browse/RESTEASY-2868

This should be reviewed by someone that understands this better than I do. I think we want to do something along these lines though.

Also #2735 does need to be merged if this is merged.


ronsigal:
re: "This should be reviewed by someone that understands this better than I do."

That wouldn't be me. @rsearls, it looks like you were involved in the related pull request https://github.com/resteasy/Resteasy/pull/2622.

jamezp:
I've ran the tests several times so far and see no errors.

asoldano (approved):
Thanks @jamezp !

PR #2741, opened 2021-04-13
[RESTEASY-2843] Do not add the target type to the thrown exception. I…

jamezp:
…nstead log it as a debug message.

https://issues.redhat.com/browse/RESTEASY-2843

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/Messages.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/Messages.java
@@ -652,9 +652,8 @@ public interface Messages
    @Message(id = BASE + 865, value = "Unable to determine base class from Type")
    String unableToDetermineBaseClass();
 
-   @Message(id = BASE
-         + 870, value = "Unable to extract parameter from http request: {0} value is \''{1}\'' for {2}", format = Format.MESSAGE_FORMAT)
-   String unableToExtractParameter(String paramSignature, String strVal, AccessibleObject target);
+   @Message(id = BASE + 870, value = "Unable to extract parameter from http request: %s value is '%s'")
+   String unableToExtractParameter(String paramSignature, String strVal);
 
    @Message(id = BASE
          + 875, value = "Unable to find a constructor that takes a String param or a valueOf() or fromString() method for {0} on {1} for basetype: {2}", format = Format.MESSAGE_FORMAT)
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.resteasy_jaxrs.i18n;
 
+import java.lang.reflect.AccessibleObject;
 import java.net.URL;
 
 import javax.ws.rs.core.MediaType;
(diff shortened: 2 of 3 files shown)


resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/Messages.java
```diff
@@ -652,9 +652,8 @@
    @Message(id = BASE + 865, value = "Unable to determine base class from Type")
    String unableToDetermineBaseClass();
 
-   @Message(id = BASE
-         + 870, value = "Unable to extract parameter from http request: {0} value is \''{1}\'' for {2}", format = Format.MESSAGE_FORMAT)
-   String unableToExtractParameter(String paramSignature, String strVal, AccessibleObject target);
+   @Message(id = 3870, value = "Unable to extract parameter from http request: %s value is '%s'")
```
asoldano, line 655: Did you change BASE+870 into 3870 on purpose?

asoldano (commented):
Thanks for the PR. Besides for the minor comment regarding the use of BASE, I think this is fine and you can merge it.

melloware (approved):

resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/Messages.java
jamezp: Yes, but I can revert it. IMO it makes it difficult to find messages when the ID is added. When I tried to find "3870" with grep or my IDE it didn't find anything.

I can change it back for sure though. I was going to make a separate proposal to change all id's :) Email first, not just change it all.

ronsigal: Yeah, I recognize that there's a little problem searching for a message. I usually grep on a piece of the message itself, which I think is pretty effective. I think that using the BASE is somewhat preferable in that it eliminates a minor (probably "very minor") source of an error. In any case, I think it's better to have everything consistent, so I would suggest reverting the change and then discussing the whole issue with the team. I'm about a -0.2 on it, but maybe there's a general sense going the other way.

jamezp: Okay. I'll change this part back. We have a more in depth discussion elsewhere as this was not the place to make a change like this :)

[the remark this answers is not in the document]
jamezp: We can make them explicit with the with positional arguments. Generally though it shouldn't be a problem unless some does some huge refactor.

I did run the tests which is something I want to look at because it breaks the checkstyle when you run them :) But yes, if I skip the checkstyle they pass.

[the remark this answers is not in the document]
ronsigal: Ok, cool. I'll merge and port to 4.5 and 4.6.

jamezp:
I don't mind back porting it. It's a simple cherry-pick. I'll need to port it to 3.15 and 3.13 as well.

ronsigal:
I've already started. What about you? Do we have a race condition? ;-)

jamezp:
I haven't, but for 4.6 and 4.5 it would take me just a minute with a cherry-pick. The other two branches can't be cherry-picked so I'd need to manually port that.

ronsigal:
I just submitted one for 4.5. How about I do 4.6 and you do the other two?

ronsigal:
Also, I was thinking we don't need 3.16 ... hopefully there won't be another 3.16 release ...

jamezp:
Okay that works. I have a 4.6 ready.

ronsigal:
Oh, OK. I'll ditch my 4.6. ;-)

PR #2742, opened 2021-04-13
[RESTEASY-2869]:Cache the hasDecorator result in Jackson2Provider

jimma:


asoldano (commented):
Thanks @jimma . This is nice, though I'm wondering if we can actually make a bit more. Am I right in saying that the annotations being looked at for hasDecorator are defined at deployment time (basically they're the annotations attached to the resource methods). If that's the case, can the information be computed once and provided to the ResteasyJackson2Provider through context injection or the ResteasyProviderFactory?

asoldano:
Actually, nevermind, I'm thinking we might want to discuss a general optimization here in another discussion thread.

PR #2751, opened 2021-04-14
[RESTEASY-1721] ResourceMethodInvoker checks if @Context HttpServletR…

ronsigal:
…esponse was written to.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
@@ -9,11 +9,15 @@ import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.spi.ValueInjector;
 import org.jboss.resteasy.spi.util.Types;
 
+import javax.servlet.ServletOutputStream;
+import javax.servlet.WriteListener;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.Application;
 import javax.ws.rs.ext.Providers;
 import javax.ws.rs.sse.Sse;
 import javax.ws.rs.sse.SseEventSink;
+
+import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
@@ -39,6 +43,7 @@ public class ContextParameterInjector implements ValueInjector
    private ResteasyProviderFactory factory;
    private Type genericType;
    private Annotation[] annotations;
+   private volatile boolean outputStreamWasWritten = false;
 
    public ContextParameterInjector(final Class<?> proxy, final Class<?> rawType, final Type genericType, final Annotation[] annotations, final ResteasyProviderFactory factory)
    {
@@ -125,6 +130,15 @@ public class ContextParameterInjector implements ValueInjector
                }
                throw new LoggableFailure(Messages.MESSAGES.unableToFindContextualData(rawType.getName()));
             }
+            // Fix for RESTEASY-1721
+            if ("javax.servlet.http.HttpServletResponse".equals(rawType.getName()))
+            {
+               if ("getOutputStream".equals(method.getName()))
+               {
+                  ServletOutputStream sos = (ServletOutputStream) method.invoke(delegate, objects);
+                  return new ContextOutputStream(sos);
+               }
+            }
             return method.invoke(delegate, objects);
          }
          catch (IllegalAccessException e)
@@ -164,7 +178,7 @@ public class ContextParameterInjector implements ValueInjector
       }
 
       return createProxy();
-  }
+   }
 
    protected Object createProxy()
    {
@@ -199,6 +213,11 @@ public class ContextParameterInjector implements ValueInjector
       }
    }
 
+   boolean outputStreamWasWrittenTo()
+   {
+      return outputStreamWasWritten;
+   }
+
    protected Class<?>[] computeInterfaces(Object delegate, Class<?> cls)
    {
       Set<Class<?>> set = new HashSet<>();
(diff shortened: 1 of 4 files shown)


ronsigal:
Replaces https://github.com/resteasy/Resteasy/pull/2723.

ronsigal:
Working on adding tests for the various methods.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -221,4 +240,150 @@ public ClassLoader run() {
       }
       return set.toArray(new Class<?>[]{});
    }
+
+   public class ContextOutputStream extends ServletOutputStream
+   {
+      private ServletOutputStream delegate;
```
jamezp, line 246: This should be final.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -221,4 +240,150 @@ public ClassLoader run() {
       }
       return set.toArray(new Class<?>[]{});
    }
+
+   public class ContextOutputStream extends ServletOutputStream
```
jamezp, line 244: If we're going to make this an inner type it should be private.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -39,6 +43,7 @@
    private ResteasyProviderFactory factory;
    private Type genericType;
    private Annotation[] annotations;
+   private boolean outputStreamWasWritten = false;
```
jamezp, line 46: This should at least be `volatile` if not even an `AtomicBoolean` or protected by something. Unless of course we know for sure the `ContextParameterInjector` instance is used by a single thread at a time. I do not know that :)

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -125,6 +130,15 @@ public Object invoke(Object o, Method method, Object[] objects) throws Throwable
                }
                throw new LoggableFailure(Messages.MESSAGES.unableToFindContextualData(rawType.getName()));
             }
+            // Fix for RESTEASY-1721
+            if ("javax.servlet.http.HttpServletResponse".equals(rawType.getName()))
```
jamezp, line 134: This question is for my own personal knowledge. Is there a reason we use a proxy here?

As to this line, we need to think this through because this would not work in Jakarta EE 9.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
```diff
@@ -199,6 +213,11 @@ public ClassLoader run() {
       }
    }
 
+   public boolean outputStreamWasWrittenTo()
```
jamezp, line 216: This should be package private or `protected` if we feel this type will be extended.

resteasy-core/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
```diff
@@ -808,7 +808,22 @@ protected MediaType resolveContentTypeByAccept(List<MediaType> accepts, Object e
       return MediaType.WILDCARD_TYPE;
    }
 
-
+   /**
+    * Checks if any bytes were written to a @Context HttpServletResponse
+    * @see ContextParameterInjector for details
+    * Fix for RESTEASY-1721
+    */
+   protected boolean contextOutputStreamWrittenTo()
```
jamezp, line 816: This should be private.

jamezp (commented):
Just a few minor things and it could have to do with some of my current lack of knowledge with how some of the project works :)

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
ronsigal: Done.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
ronsigal: Done.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
ronsigal: Yep, there's no reason that ContextParameterInjector couldn't be shared among threads.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
ronsigal: 1. As to why these are proxies, that probably goes back to before my time, but I suppose it was just for convenience.

2. As to "javax", yeah, that's a good point. I don't want to assume the presence of servlet classes. Maybe
```
rawType.getName().contains("servlet.http.HttpServletResponse")
```
Well, actually, 
```
"javax.servlet.http.HttpServletResponse".equals(rawType.getName() || "jakarta.servlet.http.HttpServletResponse".equals(rawType.getName()
```
would be better, I think.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
ronsigal: Done.

resteasy-core/src/main/java/org/jboss/resteasy/core/ResourceMethodInvoker.java
ronsigal: Done

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
jamezp: Actually thinking about it, it probably doesn't matter. It's not like we can run with the namespace without changes anyway. I say leave it as is.

ronsigal: That's true. And the new test will fail if there's no "javax.servlet.http.HttpServletResponse" on the classpath.

resteasy-core/src/main/java/org/jboss/resteasy/core/ContextParameterInjector.java
jamezp: Sorry, I meant the field not that class :) It shouldn't be mutable for the sake of "self documenting code" IMO.

ronsigal:
@jamezp, you're satisfied?

jamezp:
Yes. Nothing I see is critical.

jamezp (approved):

ronsigal:
Cool. Thank you for your eagle eye.

PR #2754, opened 2021-04-15
[RESTEASY-2845] ResteasyWebTarget.proxy(Class) problem with questionmark in @Path with Regex

petrberan:
https://issues.redhat.com/browse/RESTEASY-2845

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
@@ -16,6 +16,7 @@ import org.jboss.resteasy.util.FeatureContextDelegate;
 import org.jboss.resteasy.util.MediaTypeHelper;
 
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
 import jakarta.ws.rs.client.Entity;
 import jakarta.ws.rs.client.RxInvoker;
 import jakarta.ws.rs.client.RxInvokerProvider;
@@ -26,9 +27,11 @@ import jakarta.ws.rs.container.ResourceInfo;
 import jakarta.ws.rs.core.GenericType;
 import jakarta.ws.rs.core.MediaType;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.util.HashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


petrberan:
Hi, there seems to be a problem with CI dependencies. I'm not aware of any changes in dependencies in my code, is that an error I should fix?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -157,13 +160,31 @@ protected Object invokeSync(Object[] args)
    protected ClientInvocation createRequest(Object[] args)
    {
       WebTarget target = this.webTarget;
-      for (int i = 0; i < processors.length; i++)
+
+      Annotation[] pathParams = new Annotation[method.getParameterTypes().length];
+      HashMap<String, Object> pathParamsMap = new HashMap<>();
+
+      for (int i = 0; i < pathParams.length; i++)
```
jamezp, line 167: This could just be `method.getParameterTypes().length` and the `pathParams` could be deleted as it's only used here.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -157,13 +160,31 @@ protected Object invokeSync(Object[] args)
...
+            }
+         }
+      }
+
+      if (pathParamsMap.size() > 1) {
+         target = target.resolveTemplates(pathParamsMap);
+      }
 
+      for (int i = 0; i < processors.length; i++) {
+         if (processors != null && processors[i] instanceof WebTargetProcessor) {
```
jamezp, line 185: If the `processors` can be `null` we need to do the `null` check sooner otherwise `processors.length` will throw an NPE.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
petrberan: Thank you for your advice, you are correct. I changed the code, tests run just fine and will push the changes once we resolve the second problem

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
petrberan: Thank you for notifying me about this if statement, though this code should not be marked as changed - I may have unknowingly done some format changes and Github decided to highlight them. Similar if statement can be seen on line 185 in the committed file.

Despite that, I believe this check is not necessary, as `processors` is an object created by `ProcessorFactory.createProcessors`which returns an array of objects, no matter if it's empty or not. In my point of view, in order to get null `Processors` , the `ProcessorFactory` should throw an exception itself and thus such case can't happen.

I think that the original idea for the part of this line should be `processors[i]` since it's part of a for loop, but changing the part or completely removing it does not effect tests at all.

Should I fix the code so this check happens before the for loop?

jamezp: Ah right I see. It doesn't matter much, but if you're up for I say we do in this case as it looks odd.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
petrberan: Fixed in new commit

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
petrberan: No problem, fixed in new commit

jamezp (approved):

jamezp:

petrberan:
Hi @jamezp can I ask for a review please? There is a one test failure, but I believe it's not caused by my changes as I don't see it in my local build

jamezp (approved):

PR #2765, opened 2021-04-22
[RESTEASY-1859] switched to using getDeclaredConstructor

rsearls:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/constructor/PrivateNoArgConstructorTest.java
```diff
@@ -0,0 +1,63 @@
...
+        WebArchive war = TestUtil.prepareArchive(
+                PrivateNoArgConstructorTest.class.getSimpleName());
+        war.addClass(PrivateNoArgConstructorResource.class);
+        war.addClass(HappinessParams.class);
+        return TestUtil.finishContainerPrepare(war, null);
+    }
+
+    @Before
+    public void init() {
+        client = (ResteasyClient) ClientBuilder.newClient();
```
jamezp, line 42: More curious than anything, is there a reason not to just use the `Client`?

jamezp (approved):

PR #2776, opened 2021-05-03
[RESTEASY-2880] Config options for in-memory limit

andymc12:
This change addresses issue [RESTEASY-2880](https://issues.redhat.com/browse/RESTEASY-2880) by allowing users to configure the in-memory limit of the server (default 1024 bytes) and the client (default 1MB) by setting the following config properties, respectively:
org.jboss.resteasy.plugins.providers.multipart.memoryThreshold=<newThresholdInBytes>
org.jboss.resteasy.client.jaxrs.engines.fileUploadInMemoryThreshold=<newThresholdInMegaBytes>

### DIFF
--- a/providers/multipart/src/test/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaroundTest.java
+++ b/providers/multipart/src/test/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaroundTest.java
@@ -2,15 +2,47 @@ package org.jboss.resteasy.plugins.providers.multipart;
 
 import static org.junit.Assert.assertEquals;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 public class Mime4JWorkaroundTest {
 
+   private static final Map<String, String> preTestProperties = new HashMap<>();
+
+   @BeforeClass
+   public static void saveCurrentStateOfMemThresholdProperty()
+   {
+      if (System.getProperties().containsKey(Mime4JWorkaround.MEM_THRESHOLD_PROPERTY))
+      {
+         String value = System.getProperty(Mime4JWorkaround.MEM_THRESHOLD_PROPERTY);
+         preTestProperties.put(Mime4JWorkaround.MEM_THRESHOLD_PROPERTY, value);
+      }
+   }
+
+   @AfterClass
+   public static void resetPropertiesToPreTestValues()
+   {
+      if (!preTestProperties.containsKey(Mime4JWorkaround.MEM_THRESHOLD_PROPERTY))
+      {
+         System.clearProperty(Mime4JWorkaround.MEM_THRESHOLD_PROPERTY);
+      }
+      else
+      {
+         preTestProperties.forEach(System::setProperty);
+      }
+   }
+
    @Before
-   public void unsetMemThresholdProperty() {
+   public void unsetMemThresholdProperty()
+   {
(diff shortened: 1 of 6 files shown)


providers/multipart/src/test/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaroundTest.java
```diff
@@ -0,0 +1,47 @@
+package org.jboss.resteasy.plugins.providers.multipart;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class Mime4JWorkaroundTest {
+
+   @Before
```
asoldano, line 10: Ideally, the test should leave the environment in the same state it was before the run. So we should probably store the value of the property before the execution and restore it afterwards.

resteasy-client/src/test/java/org/jboss/resteasy/client/jaxrs/engines/ManualClosingApacheHttpClient43EngineTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class ManualClosingApacheHttpClient43EngineTest {
+
+   @Before
+   public void unsetMemThresholdProperty() {
+      System.clearProperty(ManualClosingApacheHttpClient43Engine.FILE_UPLOAD_IN_MEMORY_THRESHOLD_PROPERTY);
+   }
```
asoldano, line 13: Ideally, the test should leave the environment in the same state it was before the run. So we should probably store the value of the property before the execution and restore it afterwards.

asoldano (changes requested):
Thanks @andymc12 , the implementation looks fine to me! I have a couple of (identical) comments on the tests, please check.

providers/multipart/src/test/java/org/jboss/resteasy/plugins/providers/multipart/Mime4JWorkaroundTest.java
andymc12: Good catch. -I'll add `@BeforeClass` and `@AfterClass` methods to ensure the state of the system property is the after the test as it was before the test. And the same for other test too.  Thanks!

andymc12: @asoldano I've added the before/afterClass methods to ensure the system properties are same before and after the test.  Can you give this PR another look?  Thanks!

andymc12:
@asoldano @ronsigal would you mind taking another look at these changes?

andymc12:
@asoldano @ronsigal would you mind taking another look at this PR and merging if you approve?

asoldano (approved):

asoldano:
Thanks @andymc12 and sorry for the late reply. I've approved this, thanks for having applied the changes I asked for previously. I'll leave this to @ronsigal and @jamezp (who is now leading the project with Ron) to merge.

jamezp:
I'll have a look hopefully today. We've kind of paused merging ATM as we're making some moves towards Jakarta EE 9.1. However we can't keep merging paused as that will create more issues :) I'll have a solution today for the merging part.

andymc12:
Thanks @jamezp - and please let me know if I can help with Jakarta 9.1 effort.

jamezp:
Thanks @andymc12. I've pretty much got it done already https://github.com/jamezp/Resteasy/tree/jakarta-ee91. We're just working on pulling some things out first, then I'll submit a PR once we can determine how we want to transform dependencies. Hopefully by the end it's only Infinispan that needs to be transformed.

PR #2789, opened 2021-05-13
[RESTEASY-2888] Change sub-modules to use their direct parent module …

jamezp:
…as it's parent. Clean up the pom's and dependency tree as much as possible.

https://issues.redhat.com/browse/RESTEASY-2888

While I did clean the POM's up a bit during this, that part is less critical to me at this point. If we'd prefer a separate change for that I understand and will change this to only move the parent POM's.


jamezp:
@asoldano @ronsigal What do you think about this?

jamezp:
Given the approval I'm merging this :)

PR #2790, opened 2021-05-13
[RESTEASY-2892] Make sure the ResteasyConfiguration also checks the S…

jamezp:
…ervletContext for both the ServletBootstrap and FilterBootstrap.

https://issues.redhat.com/browse/RESTEASY-2892
Upstream: #2799

Please double check the checking order for the property value. Current for the `ServletBootstrap` it's:

1. System properties
2. Environment variables
3. `ServletConfig.getInitParameter()`
4. `ServletContext.getInitParameter()`

For the `FilterBootstrap it's:
1. System properties
2. Environment variables
3. `FilterConfig.getInitParameter()`
4. `ServletContext.getInitParameter()`

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/config/DefaultConfiguration.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/config/DefaultConfiguration.java
@@ -32,9 +32,13 @@ import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 import org.jboss.resteasy.spi.ResteasyConfiguration;
 
 /**
- * A default configuration which first attempts to use the Eclipse MicroProfile Config API. If not present on the class
- * path the {@linkplain ResteasyConfiguration configuration} is used to resolve the value, followed by system properties
- * and then environment variables if not found in the previous search.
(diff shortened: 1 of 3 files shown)


ronsigal:
Hmmm, I must have missed this the last time around. When I first introduced the MP Config stuff, I meant for RESTEasy to fall back on this behavior. In fact, that's what the User Guide describes.

ronsigal:
@jamezp, it would be nice to be able to test this stuff, but I don't see how ... .

jamezp:
@ronsigal Yes. I do want to get some testing for it. I think I can probably come up with a way to do it. It would be easy if/when we switch to using Galleon for building the test server.

That said one option is to remove the config subsystem in a test. I could look at doing something like this.

ronsigal:
re: "remove the config subsystem in a test"

That would be interesting. I didn't see how to do that, but maybe write over the resteasy-jaxrs/main/module.xml and take out
```
<module name="org.eclipse.microprofile.config.api"/>

jamezp:
What we would do is use an Arquillian `ServerSetupTask` and execute operations to remove the subsystems we don't want. I can probably look at it tomorrow TBH.

ronsigal:
Oh, I didn't know about that. Great.

PR #2793, opened 2021-05-14
[RESTEASY-2893]:Move the tests under arquillian to testsuite

jimma:


jimma:
Rebased. Please review.

resteasy-dependencies-bom/pom.xml
```diff
@@ -973,6 +974,12 @@
                 <artifactId>asyncutil</artifactId>
                 <version>${version.asyncutil}</version>
             </dependency>
+            <dependency>
+                <groupId>org.jboss.arquillian.container</groupId>
+                <artifactId>arquillian-jetty-embedded-9</artifactId>
+                <version>${version.arquillian.jetty}</version>
+                <scope>test</scope>
+            </dependency>
```
jamezp, line 982: We might not want this here. We had to move the other Arquillian dependencies out of here for EAP reasons. The issue was that since the dependency does not end up in the product repository it causes issues for users not being able to resolve the dependency.

This should probably just be part of the main pom.

jamezp (commented):
Overall I think this is okay. There's the one minor concern with the arquillian-jetty being in the BOM.

While I know you just copied these over, I do think we can get rid of the `.gitignore` files in these directories. I don't think we should block merging based on that though.

resteasy-dependencies-bom/pom.xml
jimma: @jamezp  What do you mean by part of the main pom ? There is no dependency management in main pom (Restasy/pom.xml). We  put all dependency versions in resetasy-dependencies-bom/pom.xml to have one place to manage them. Will this be a problem for Arquillian dependencies ?

jamezp: Sorry I guess just the version under the main pom, https://github.com/resteasy/Resteasy/blob/5664008250924fc4f22e62a5e5dfe83897b41e99/pom.xml#L23-L27 and the dependencies defined in the testsuite pom https://github.com/resteasy/Resteasy/blob/5664008250924fc4f22e62a5e5dfe83897b41e99/testsuite/pom.xml#L32-L81.

The problem was very EAP specific, but it would be more consistent with what we're doing with the other Arquillian dependencies. I can give you the full details if you would like offline.

jamezp:

PR #2794, opened 2021-05-19
[RESTEASY-2874] Web Target Proxy not overriding content type header

petrberan:
https://issues.redhat.com/browse/RESTEASY-2874


liweinan:
@ronsigal @jamezp this looks good. Can it be merged?

jamezp:

jamezp:
@crankydillo Let me think about this. That example seems explicit to me that the user would want to set the `Content-Type` header. However, the entity value should win IMO. It does seem like we should file an issue for this.

crankydillo:
I went ahead and wrote it up [here](https://issues.redhat.com/browse/RESTEASY-3303), primarily so we'll have something to track.  I probably should have done that initially, but I always have to dig out my JIRA account info:P

PR #2795, opened 2021-05-20
[RESTEASY-2901] Clean up possibly-garbage Response before calling ExceptionHandler

bsmelo:
**EDIT:** (Although this was originally discovered running 3.13.2.Final through WildFly 21.0.0.Final; it was reproduced on `main` branch as well - see #2804)

Hello all,

First attempt at contributing to fix https://issues.redhat.com/browse/RESTEASY-2901

I'll wait for a feedback from someone more knowledgable of the code in question, but I'm starting to think a more adequate/elegant fix could be made around the [definition of the `entityStream` used](https://github.com/resteasy/Resteasy/blob/3.13.2.Final/providers/jackson2/src/main/java/org/jboss/resteasy/plugins/providers/jackson/ResteasyJackson2Provider.java#L220) by ResteasyJackson2Provider, since the problem doesn't seem to happen with JSONB.

Cheers!


jamezp:

PR #2801, opened 2021-05-28
Require a minimum of Java 9 for compiling

jamezp:
JIRA's this resolves:
- https://issues.redhat.com/browse/RESTEASY-2907
- https://issues.redhat.com/browse/RESTEASY-2905
- https://issues.redhat.com/browse/RESTEASY-2910

This reduces to only running on WildFly 23 for now. Once WildFly 24 is released we can add that. It also executes tests for Java 11, Java 16 and Java 8. Note the testsuite requires a separate run as any files generated during the initial Java 9+ run may cause issues when running under Java 8. By files I'm referring to things like generated JKS files.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -15,13 +15,13 @@ jobs:
   build:
 
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 60
+    timeout-minutes: 90
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest ]
-        java: ['1.8', '11']
-        wildfly-version: ['22.0.1.Final', '23.0.2.Final']
+        java: ['11', '16']
+        wildfly-version: ['23.0.2.Final']
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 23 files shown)


asoldano (commented):
@jamezp thanks!
I think you should update the .travis.yml too, shouldn't you?
Besides that, I think the PR is ready to be merged.

jamezp:
@asoldano Oh yes good call. I keep forgetting about that file. Should we remove it or do we want to keep using it as well?

jamezp:
I updated the travis.yml. However I'm not sure how to execute the tests with Java 8. I'll have to look into that.

jamezp:
The tests failing in Travis CI are passing for me locally. I also noticed a banner that says travis-ci.org is going away in May of 2021 so this could just stop working at some point. I think we can migrate to travis-ci.com though. We need to configure the project if we want do that.

That said it does look like we should probably add another CI job to run with Elytron enabled. I can look into doing that once this PR is merged.

jamezp:
@asoldano I'm leaning towards removing the travis-ci since travis-ci.org is closing down. The other option is to migrate to travis-ci.com. Do you have a preference?

jamezp:
Superseded by #2820

PR #2803, opened 2021-05-28
Publisher support that allows subscriber context.single value

crankydillo:
@ronsigal 

This is draft work of what you have proposed [here](https://github.com/resteasy/Resteasy/pull/2463#issuecomment-845398776).  I'd rather not have a ton of discussion here until we shutdown #2463.  As it stand today, about half of our test fail when using this:) However, that part may just be changes we need to make.  I had to go super fast as I'm about to be on vacation:)  It also makes us couple with `MonoHttpClientEngine` as opposed to `ReactiveHttpClientEngine`.  I still had to expose `Publisher` from `Unit`.  I'm assuming you would not like that for the same reason you don't like it in the other place:P

While I would like to kill #2463 before going too far, I am curious if you know how to avoid `Publisher<T> get()` in `Unit`.  That is what we came up with to still use our code in our `HttpClientEngine` wrappers.  If I could do `Unit<T, M<T>>`, I don't think I'd have to couple with `MonoHttpClientEngine`, but Java doesn't allow you to do it.


crankydillo:
Hoping that #2463 will work instead of this.

PR #2806, opened 2021-06-01
Fixing the ResteasyViolationException#toString concurrency

rhuan080:
Signed-off-by: Rhuan Rocha <rhuan080@gmail.com>

This issue was identified in a scenario that the exception' object is delegated to a task that works in other threads. As it uses the iterator on ResteasyViolationException#toString and it calls the convertViolations method (that modifies the violationLists) it can generate an issue if called concurrently. I have seen it on 3.9.3.SP1-redhat-00001. It is why I'm sending a PR to this branch. However, the main branch has the same problem. I'll send a PR to the main soon.

PR #2810, opened 2021-06-01
Fixing the ResteasyViolationException#toString concurrency

rhuan080:
This issue was identified in a scenario that the exception' object is delegated to a task that works in other threads. As it uses the iterator on ResteasyViolationException#toString and it calls the convertViolations method (that modifies the violationLists) it can generate an issue if called concurrently.

JIRA: https://issues.redhat.com/browse/RESTEASY-2914
PR to 3.9: https://github.com/resteasy/Resteasy/pull/2806

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/api/validation/ResteasyViolationException.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/api/validation/ResteasyViolationException.java
@@ -12,6 +12,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import javax.validation.ConstraintViolation;
 import javax.validation.ConstraintViolationException;
@@ -37,21 +38,21 @@ public abstract class ResteasyViolationException extends ConstraintViolationExce
 
    public static final String SUPPRESS_VIOLATION_PATH = "resteasy.validation.suppress.path";
 
-   private List<CloneableMediaType> accept;
+   private volatile List<CloneableMediaType> accept;
 
-   private Exception exception;
+   private volatile Exception exception;
 
-   private List<ResteasyConstraintViolation> propertyViolations;
+   private volatile List<ResteasyConstraintViolation> propertyViolations;
 
-   private List<ResteasyConstraintViolation> classViolations;
+   private volatile List<ResteasyConstraintViolation> classViolations;
 
-   private List<ResteasyConstraintViolation> parameterViolations;
+   private volatile List<ResteasyConstraintViolation> parameterViolations;
 
-   private List<ResteasyConstraintViolation> returnValueViolations;
+   private volatile List<ResteasyConstraintViolation> returnValueViolations;
 
-   private List<ResteasyConstraintViolation> allViolations;
+   private volatile List<ResteasyConstraintViolation> allViolations;
 
-   private List<List<ResteasyConstraintViolation>> violationLists;
+   private  volatile List<List<ResteasyConstraintViolation>> violationLists;
 
    private transient ConstraintTypeUtil util = getConstraintTypeUtil();
 
@@ -141,7 +142,7 @@ public abstract class ResteasyViolationException extends ConstraintViolationExce
       convertViolations();
       if (allViolations == null)
       {
-         allViolations = new ArrayList<ResteasyConstraintViolation>();
+         allViolations = new CopyOnWriteArrayList<>();
          allViolations.addAll(propertyViolations);
          allViolations.addAll(classViolations);
          allViolations.addAll(parameterViolations);
@@ -189,14 +190,12 @@ public abstract class ResteasyViolationException extends ConstraintViolationExce
    {
       convertViolations();
       StringBuffer sb = new StringBuffer();
-      for (Iterator<List<ResteasyConstraintViolation>> it = violationLists.iterator(); it.hasNext();)
-      {
-         List<ResteasyConstraintViolation> violations = it.next();
-         for (Iterator<ResteasyConstraintViolation> it2 = violations.iterator(); it2.hasNext();)
-         {
-            sb.append(it2.next().toString()).append('\r');
+      for (List<ResteasyConstraintViolation> violations : violationLists) {
+         for (ResteasyConstraintViolation violation: violations ) {
+            sb.append(violation.toString()).append('\r');
          }
       }
+
       return sb.toString();
    }
 
@@ -250,11 +249,12 @@ public abstract class ResteasyViolationException extends ConstraintViolationExce
          throw new RuntimeException(Messages.MESSAGES.unableToParseException());
       }
 
-      violationLists = new ArrayList<List<ResteasyConstraintViolation>>();
+      List<List<ResteasyConstraintViolation>> violationLists = new ArrayList<List<ResteasyConstraintViolation>>();
       violationLists.add(propertyViolations);
       violationLists.add(classViolations);
       violationLists.add(parameterViolations);
       violationLists.add(returnValueViolations);
+      this.violationLists = new CopyOnWriteArrayList<>(violationLists);
    }
 
    protected int getField(int start, String line)
@@ -311,10 +311,10 @@ public abstract class ResteasyViolationException extends ConstraintViolationExce
          return;
       }
 
-      propertyViolations = new ArrayList<ResteasyConstraintViolation>();
-      classViolations = new ArrayList<ResteasyConstraintViolation>();
-      parameterViolations = new ArrayList<ResteasyConstraintViolation>();
-      returnValueViolations = new ArrayList<ResteasyConstraintViolation>();
+      propertyViolations = new CopyOnWriteArrayList<>();
+      classViolations = new CopyOnWriteArrayList<>();
+      parameterViolations = new CopyOnWriteArrayList<>();
+      returnValueViolations = new CopyOnWriteArrayList<>();
 
       if (getConstraintViolations() != null)
       {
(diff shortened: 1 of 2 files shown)


soul2zimate:
Can you also backport to 3.15 (EAP 7.4.z) and 3.11 (EAP 7.3.z) branches ?

rhuan080:
Okay. :-)

rhuan080:
3.15: https://github.com/resteasy/Resteasy/pull/2816
3.11: https://github.com/resteasy/Resteasy/pull/2817

jamezp:
I apologize for not getting to this earlier before you've opened all the PR's. From the JIRA's it looks like `toString()` is be invoked from possibly multiple threads. While this seems a bit odd to me, if we're concerned about concurrency issues there are a few things I think we need to fix here. I'm happy to help with this or show an example of what I think we need to do.

resteasy-core-spi/src/main/java/org/jboss/resteasy/api/validation/ResteasyViolationException.java
```diff
@@ -189,14 +190,16 @@ public String toString()
...
       StringBuffer sb = new StringBuffer();
-      for (Iterator<List<ResteasyConstraintViolation>> it = violationLists.iterator(); it.hasNext();)
-      {
-         List<ResteasyConstraintViolation> violations = it.next();
-         for (Iterator<ResteasyConstraintViolation> it2 = violations.iterator(); it2.hasNext();)
-         {
-            sb.append(it2.next().toString()).append('\r');
+
+      CopyOnWriteArrayList<List<ResteasyConstraintViolation>> imutableViolations =
+              new CopyOnWriteArrayList<>(violationLists);
```
jamezp, line 195: As mentioned I think we need a larger change here. However this don't fix the issue as the `violationLists` is still iterated when creating this new list. We also don't really need a concurrent list in the `toString()` itself as once it's initialized it won't need to be thread-safe.

rhuan080: Hi @jamezp, thank for sharing your thought.

At the first invocation, the list is created and the second invocation recreates the same list (with the same elements). I'm not avoiding the recreation here as I want to avoid a side effect applying a big change, but I think the better solution is to create this list at the constructor. Thus it does not need to be created on the toString method. This solution solves the java.util.ConcurrentModificationException. In my tests, it has worked fine. I think to solve the issue immediately we can use the  CopyOnWriteArrayList (thus the recreation of the list will not generate the issue) and a second time I can send a PR to solve the recreation. However, I think the solution about the recreation needs more tests. Let me know if it makes sense for you.

rhuan080: Hi @jamezp,
I'm sorry, but I think I didn't explain well my thought. 

The method convertViolations() called by toString initiate the ArrayList. I think the final solution is to call this method inside the ResteasyConstraintViolation' constructor. I don't think the state of this ArrayList (violationList) should be changed after the ResteasyConstraintViolation construction. About the CopyOnWriteArrayList it solves the question about the java.util.ConcurrentModificationException, as at the iteration time it creates an immutable copy of the list. However, I think the final solution is to call the convertViolations() inside all ResteasyConstraintViolation' constructors. The main issue is the toString is updating the internal state. But, we need to check if it will generate a side effect. Thus, I think we can apply this solution and after apply the solution calling the convertViolations() inside the ResteasyConstraintViolation' constructors. What do you think about the final solution?

jamezp: Hi @rhuan080 I don't think that would quite work. The `convertViolations()` would initialize itself with empty lists. With the current design it needs to be invoked several times. It's one of the reason I think we should look at redesigning it. TBH i see other issues with it as well like `line = br.readLine()` executing twice which looks like it would skip a line with each read.

I've got some time to look at this unless you'd rather have a go at it.

rhuan080: Hi @jamezp, thank you. Yes, I have checked you talked about the `convertViolations()`. About the `line = br.readLine()`, I think it is not a problem as it is a stream of stringRep. As it is a string it is thread-safe. I cannot see what is the problem here. If possible can you clarify for me? 

Before using the CopyOnWriteArrayList I have created a solution using `synchronized` and I was able to solve it, but I have applied updates in many methods. Do you want to suggest some solution like it?

jamezp: Hi @rhuan080 I just meant the double `br.readLine()` skips a line each time so something like:
```
line 1
line 2
line 3
line 4
```

All we'd see is line1 and line 3 with line 2 and line 4 being skipped.

If we're just trying to fix this one issue then even using an instance of a new `ArrayList`, e.g. swap your `CopyOnWriteArrayList` with `ArrayList`, would work. However the fact that we're seeing a `ConcurrentModificationException` sounds like we've got some currency issues. If we don't care about missing violations, then we can just do what you're doing, but use an `ArrayList` instead.

rhuan080: Hi @jamezp, I agreed this class should be reviewed, however, I think it is better to open an issue and solve it in another PR. The PR was opened to just solve the ConcurrentModificationException. I agreed it should be reviewed, but we can open an issue discuss what is a better design for this class. Looking at this class I think it is overly complex. What do you think about the issue? Do you agree to open an issue to it and discuss it there?

jamezp: Yes I agree. We should probably do different fixes for different branches IMO. We don't want to touch the 3.x branches too much so leaving those with a fix like this might be okay. I'll need to look at the source a bit more, but we likely have potential issues exposed on the other methods as well.

I will say I don't think we need to use a `CopyOnWriteArrayList` for the local instance though. If we use it for the global instance that is okay though.

rhuan080: Hi @jamezp, I'm sorry for my delay. I have opened this JIRA --> https://issues.redhat.com/browse/RESTEASY-2927
If possible, join this JIRA to start the discussion about the review.

jamezp: @rhuan080 this would be my suggestion for the minimal change https://github.com/resteasy/Resteasy/compare/main...jamezp:RESTEASY-2914.

rhuan080: @jamezp I'll update the branches.

rhuan080: Node!

jamezp:
I'll pull this down and run it locally, but the error looks very strange to me.
```
2021-06-16T22:04:25.4415143Z java.lang.ClassCastException: class java.util.concurrent.CopyOnWriteArrayList cannot be cast to class java.util.ArrayList (java.util.concurrent.CopyOnWriteArrayList and java.util.ArrayList are in module java.base of loader 'bootstrap')
```

jamezp:
We need to do something like https://github.com/resteasy/Resteasy/commit/c58d33992be5ba31d95b731aef4d0a4cd94451e6

jamezp:
Once that is added the commits should be squashed into a single commit.

rhuan080:
Done!

jamezp:
@rhuan080 did you forget to push your changes? I'm not seeing something like https://github.com/resteasy/Resteasy/commit/c58d33992be5ba31d95b731aef4d0a4cd94451e6

rhuan080:
Hi @jamezp, I reviewed my PR and it is correct! I just updated the org/jboss/resteasy/api/validation/ResteasyViolationException.java. I have tested the changes. 

The class you said in main branch is this --> https://github.com/resteasy/Resteasy/blob/main/resteasy-core-spi/src/main/java/org/jboss/resteasy/api/validation/ViolationReport.java.

I could not see the changes you said. I think the changes of c58d339 were no merged to main. Thus it is not available to us.

My PR is correct according to the main forked and the updates done by me.

jamezp:
@rhuan080 The changes need to be in the PR otherwise the build fails :) I can add a commit to your PR and squash it, but we can't merge this PR as is because it breaks the assumptions on the `ViolationReport`.

rhuan080:
Are these changes in a commit in your fork? I can update it here, however, the commit you said is not available to the main branch in the upstream. It is why the change is not available to my fork. My branch is the main + my updates. I have not forgot any push. 

Do you want I update the ViolationReport.java and include it in my commit?

jamezp:
Yes the commits are in my fork only. I've not done any changes upstream. If you could include the changes to the `ViolationReport` in your commit that would be great :)

rhuan080:
Sure! Thank you for clarifying!

rhuan080:
Hi @jamezp I have squashed the commits but now it appears 23 files changed. Can you help me to check if I did something wrong?

rhuan080:
Done! I'll apply it to 3.11 and 3.15 as well.

rhuan080:
Hi @jamezp, I applied it to 3.11 and 3.15. Thank you a lot for helping me here :-)

jamezp:
I've add a small commit on top of this and rebased it. I've force pushed to your branch which I'd normally not do, but I'd like to get this in.

rhuan080:
Hi, I have been applied these updates as suggested in https://github.com/resteasy/Resteasy/pull/2810#issuecomment-862792997.
Let me know if I need to do something to follows up with this task.

jamezp:
Awesome! Thanks @rhuan080! I'll get them merged!

PR #2818, opened 2021-06-02
[RESTEASY-2891] Do not depend on Yasson for the Jakarta EE JSON Bindi…

jamezp:
…ng implementation. Also ensure the Jsonb type is closed after invocation.

https://issues.redhat.com/browse/RESTEASY-2891

One thing I'm not sure how to note is that now it would be required for a user to supply a JSON-P and JSON Binding  implementation. For containers like WildFly this is not an issue.


jamezp:
Replaced by https://github.com/resteasy/resteasy/pull/2978

PR #2820, opened 2021-06-08
Require a minimum of Java 11 for compiling

jamezp:
JIRA's this resolves:
- https://issues.redhat.com/browse/RESTEASY-2924
- https://issues.redhat.com/browse/RESTEASY-2907
- https://issues.redhat.com/browse/RESTEASY-2905
- https://issues.redhat.com/browse/RESTEASY-2910

This reduces to only running on WildFly 23 for now. Once WildFly 24 is released we can add that. It also executes tests for Java 11, Java 16 and Java 8. Note the testsuite requires a separate run as any files generated during the initial Java 11+ run may cause issues when running under Java 8. By files I'm referring to things like generated JKS files.

This replaces #2801


rsearls (approved):

asoldano (approved):
Thanks!

PR #2821, opened 2021-06-09
[RESTEASY-2925]: Inconsitent NewCookie parsing when unknown attributes is found

NicoNes:
https://issues.redhat.com/browse/RESTEASY-2925

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/NewCookieHeaderDelegate.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/NewCookieHeaderDelegate.java
@@ -10,7 +10,6 @@ import javax.ws.rs.ext.RuntimeDelegate;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
 
@@ -37,7 +36,7 @@ public class NewCookieHeaderDelegate implements RuntimeDelegate.HeaderDelegate<N
       Date expiry = null;
 
       ParameterParser parser = new ParameterParser();
-      LinkedHashMap<String, String> map = parser.parse(newCookie, ';');
+      Map<String, String> map = parser.parse(newCookie, ';');
 
       for (Map.Entry<String, String> entry : map.entrySet()) {
          String name = entry.getKey();
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/NewCookieHeaderDelegate.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/NewCookieHeaderDelegate.java
@@ -10,6 +10,7 @@ import javax.ws.rs.ext.RuntimeDelegate;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.LinkedHashMap;
 import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/NewCookieHeaderDelegate.java
```diff
@@ -35,12 +37,16 @@ public NewCookie fromString(String newCookie) throws IllegalArgumentException {
       Date expiry = null;
 
       ParameterParser parser = new ParameterParser();
-      Map<String, String> map = parser.parse(newCookie, ';');
+      LinkedHashMap<String, String> map = parser.parse(newCookie, ';');
```
jamezp, line 40: There is not need for this to be a `LinkedHashMap`. This line can remain unchanged.

resteasy-core/src/main/java/org/jboss/resteasy/util/ParameterParser.java
```diff
@@ -223,11 +222,11 @@ public void setLowerCaseNames(boolean b)
     * @param separator the name/value pairs separator
     * @return a map of name/value pairs
     */
-   public Map<String, String> parse(final String str, char separator)
+   public LinkedHashMap<String, String> parse(final String str, char separator)
```
jamezp, line 225: We shouldn't change the return type here.

resteasy-core/src/main/java/org/jboss/resteasy/util/ParameterParser.java
```diff
@@ -241,11 +240,11 @@ public void setLowerCaseNames(boolean b)
     * @param separator the name/value pairs separator
     * @return a map of name/value pairs
     */
-   public Map<String, String> parse(final char[] chars, char separator)
+   public LinkedHashMap<String, String> parse(final char[] chars, char separator)
```
jamezp, line 243: We shouldn't change the return type here either.

resteasy-core/src/main/java/org/jboss/resteasy/util/ParameterParser.java
```diff
@@ -261,7 +260,7 @@ public void setLowerCaseNames(boolean b)
     * @param separator the name/value pairs separator
     * @return a map of name/value pairs
     */
-   public Map<String, String> parse(
+   public LinkedHashMap<String, String> parse(
```
jamezp, line 263: We shouldn't change the return type here either.

resteasy-core/src/main/java/org/jboss/resteasy/util/ParameterParser.java
```diff
@@ -270,9 +269,9 @@ public void setLowerCaseNames(boolean b)
 
       if (chars == null)
       {
-         return new HashMap<String, String>();
+         return new LinkedHashMap<>();
       }
-      HashMap<String, String> params = new HashMap<String, String>();
+      LinkedHashMap<String, String> params = new LinkedHashMap<>();
```
jamezp, line 274: This should just be `Map<String, String> params = new LinkedHashMap<>()`.

NicoNes:
In this PR I changed the `HashMap<String,String>` used in the parser by a `LinkedHashMap<String,String>`.

I could have let the `parse(...)` methods signatures unchanged but I decided to change them to replace the `Map<String,String>` interface by the `LinkedHashMap<String,String>`  to make it clear that order matter.

Let me know what you think.

jamezp (changes requested):

jamezp:
> In this PR I changed the `HashMap<String,String>` used in the parser by a `LinkedHashMap<String,String>`.
> 
> I could have let the `parse(...)` methods signatures unchanged but I decided to change them to replace the `Map<String,String>` interface by the `LinkedHashMap<String,String>` to make it clear that order matter.
> 
> Let me know what you think.

I don't think we should change the signatures. That can break backwards compatibility.

NicoNes:
@jamezp Even with a `LinkedHashMap` in the signature, backwards compatibility is still OK since a `LinkedHashMap`  is a `Map` and also a `HashMap` right ?
Actually I did the change because it can be be weird to say that cookie name is always the first attribute and then retrieve it iterating over a `Map` interface that does not enforce order.
What bother me more with using `LinkedHasMap` is that it is an implementation not an interface...
I will change it.

Thanks

jamezp (approved):
Looks good thanks @NicoNes!

PR #2826, opened 2021-06-10
[RESTEASY-2928]: SseEventSource keeps on calling onEvent callback after OnComplete callback has been invoked

NicoNes:
Back in day when I introduced `org.jboss.resteasy.plugins.providers.sse.client.SseEventSourceScheduler` I already proposed this change but with @jimma we believed that calling `onComplete` callback after the close status has been set was enough (https://github.com/resteasy/Resteasy/pull/1677#discussion_r214810725). 
Actually it's not the case that's why I add it again.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
@@ -133,6 +133,8 @@ public class SseEventSourceImpl implements SseEventSource
       this.reconnectDelay = reconnectDelay;
       this.alwaysReconnect = alwaysReconnect;
 
+      Runnable onCompleteCallBack = () -> onCompleteConsumers.forEach(Runnable::run);
+
       if (executor == null)
       {
          ScheduledExecutorService scheduledExecutor = null;
@@ -141,17 +143,17 @@ public class SseEventSourceImpl implements SseEventSource
             scheduledExecutor = ((ResteasyWebTarget) target).getResteasyClient().getScheduledExecutor();
          }
          if (name != null) {
-            this.sseEventSourceScheduler = new SseEventSourceScheduler(scheduledExecutor, name);
+            this.sseEventSourceScheduler = new SseEventSourceScheduler(scheduledExecutor, name, onCompleteCallBack);
          } else {
-            this.sseEventSourceScheduler = new SseEventSourceScheduler(scheduledExecutor, String.format("sse-event-source(%s)", target.getUri()));
+            this.sseEventSourceScheduler = new SseEventSourceScheduler(scheduledExecutor, String.format("sse-event-source(%s)", target.getUri()), onCompleteCallBack);
          }
       }
       else
       {
          if (name != null) {
-            this.sseEventSourceScheduler = new SseEventSourceScheduler(executor, name);
+            this.sseEventSourceScheduler = new SseEventSourceScheduler(executor, name, onCompleteCallBack);
          } else {
-            this.sseEventSourceScheduler = new SseEventSourceScheduler(executor, String.format("sse-event-source(%s)", target.getUri()));
+            this.sseEventSourceScheduler = new SseEventSourceScheduler(executor, String.format("sse-event-source(%s)", target.getUri()), onCompleteCallBack);
          }
       }
 
@@ -273,7 +275,6 @@ public class SseEventSourceImpl implements SseEventSource
          }
       }
       sseEventSourceScheduler.shutdownNow();
-      onCompleteConsumers.forEach(Runnable::run);
    }
 
    private class EventHandler implements Runnable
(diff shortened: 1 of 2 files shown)


jamezp:
@jimma can you please review this when you've got a moment?

jimma:
Hi @NicoNes, Thanks for the patch. 
If I understand the change set correctly, it is intended to fix these things:
- when there is event is processing(onEvent), EventSource shouldn't onComplete callback handlers. 
- After all events are processed by onEvent() and SseEventSourceScheduler.shutdownNow() is called, the onComplete() will be executed(cause there is Phaser(1) defined)

The problem here I see is when the  SseEventSourceScheduler.shutdownNow() is called, it might stop and halt the event processing thread. It still can't guarantee the event processing is done.

NicoNes:
Hi @jimma ,

Yes, the goal of this PR is to guarantee that
- no new invocation of  `onEvent` callback will never be done after `onComplete` callback has been invoked
-  `onComplete` callback won't be invoked while  `onEvent` callbacks are still running.

This PR goal is not to guarantee that already running `onEvent` callbacks started before `onComplete` callback invocation won't be stopped. However the existing code already guarantees us that behavior.

`SseEventSourceScheduler.shutdownNow()` implementation stops threads by sending and interrupt signal to them. So if  the tasks running in the thread do not respond to interrupt signal they won't be stopped at all and will continue there execution to the end.
So  all the `onEvent` callbacks started and already running before `SseEventSourceScheduler.shutdownNow()` has been invoked won't be stopped unless they respond to thread interrupt signal. If user choose to respond to the interrupt signal in his callback code, well there is no problem, thats' his choice.

And for all events sent by the server and that will never be processed because the event processing thread has been stopped reacting to the thread interrupt signal (https://github.com/resteasy/Resteasy/blob/3a5c45f1b507a9673b93fa0469e2a33adb542b11/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java#L378), I do not see no problem this is the expected behavior.

jimma:
Thanks for the details. @NicoNes 
> - After all events are processed by onEvent() and SseEventSourceScheduler.shutdownNow() is called, the onComplete() will be executed(cause there is Phaser(1) defined)

> The problem here I see is when the SseEventSourceScheduler.shutdownNow() is called, it might stop and halt the event processing thread. It still can't guarantee the event processing is done.

Here what I mean is there is still incoming event(read from response stream), but it won't process after `SseEventSourceScheduler.shutdownNow()`. As you pointed out , there is  `while (!Thread.currentThread().isInterrupted() && state.get() == State.OPEN)'. So it doesn't fix this as spec says :  
> the `onComplete` callback is invoked "when there are no further events to be received".

 I get what you are trying to do is make sure the onComplete() executes after onEvent() handlers. Now I  think of this : should we make Phaser responds to thread interrupt or wait for amount of time before onComplete()/internalClose() ?

NicoNes:
Hey @jimma,

Thanks, I think I better understand your point now. 

After `SseEventSourceScheduler.shutdownNow()` has been invoked from `SseEventSourceImpl.internalClose()`, the event processing thread  will stop because of `while (!Thread.currentThread().isInterrupted() && state.get() == State.OPEN)` and this line `InboundSseEvent event = eventInput.read(providers)` won't be called. 
So, no more events sent by the server that are on the inputstream won't be read from that inputstream after invocation of  `SseEventSourceImpl.internalClose()`. So I think we can consider those events as sent by the server but not received already since they are not read from the stream yet.

More even if  `InboundSseEvent event = eventInput.read(providers)` is invoked, it will throw an exception because the underlying input stream has already been released by `SseEventSourceImpl.internalClose()`. So even in this case, events sent by the server that remain on the inputstream won't be read.

Let me know if I'm missing something please and give me more details about what you mean:

> should we make Phaser responds to thread interrupt or wait for amount of time before onComplete()/internalClose() ?

Thanks

NicoNes:
Hey @jimma,,


Any update on this one ?

Thanks

jimma:
Hi @NicoNes, don't know why I didn't get the previous comment update notification email and sorry for the late response.
> should we make Phaser responds to thread interrupt or wait for amount of time before onComplete()/internalClose() ?

I mean is it possible when there is error happens and it will wait for Phaser forever. Should we add something to only wait for amount of time and the onComplete handler eventually will be called ?

NicoNes:
Hey @jimma,

Can you please give me the case you have in mind where you think it can wait for phaser forever ?

In my opinion once `SseEventSourceImpl.internalClose()` is invoked,  the event processing thread will exit the `while (!Thread.currentThread().isInterrupted() && state.get() == State.OPEN)` loop, the current partie registered in Phaser for the `while` loop at line https://github.com/resteasy/Resteasy/blob/3a5c45f1b507a9673b93fa0469e2a33adb542b11/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java#L96 will be deregistered in the `finally` block here https://github.com/resteasy/Resteasy/blob/3a5c45f1b507a9673b93fa0469e2a33adb542b11/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceScheduler.java#L114
so at this moment the phaser will be terminated for sure.

More if for any reason, an exception is thrown during the event processing, the partie registered in Phaser will always  be deregistred no matter what happens thanks to the `finally` block. 
So I do not see no risk of blocking forever. Please Correct me if I'm wrong.

> Should we add something to only wait for amount of time and the onComplete handler eventually will be called ?

Well, yes I think it can work. But personally I'm not a fan of this kind of mechanism that rely on some arbitraty amount of time. It give me the impression that we don't really know what we are doing. But if you want we can add this as a kind of double check just in case.

jamezp:

PR #2832, opened 2021-06-13
[RESTEASY-2942] upgraded dependencies to match quarkus/bom/application

rsearls:


jamezp:
We need to think about what to do here. The MP upgrades are higher than the versions currently in WildFly which _may_ be a problem. The tests did pass so maybe it's not a huge deal.

ronsigal:
>We need to think about what to do here.

For now, I'm thinking of RESTEasy 4.6.1.Final, which is for Quarkus. Ideally, when we get to 4.7.0.Final, we can find a compromise that works for both WildFly and Quarkus. It would be great if they both can use 4.7.0.Final.

Anyway, I'll create a pull request for 4.6.1.Final.

ronsigal:
I cloned RESTEASY-2931 to RESTEASY-2942 for main branch.

jamezp:
@ronsigal If we do it in 4.6 we have to do it in 4.7. We can't have a newer version released with older dependencies. If this is required for 4.6 we need to come up with a solution because 4.7 needs to go into WildFly otherwise we're back to 3.x and we'll have to port Jakarta EE 9 changes into 3.x.

resteasy-dependencies-bom/pom.xml
```diff
@@ -66,12 +66,12 @@
         <version.org.jboss.spec.javax.ejb.jboss-ejb-api_3.2_spec>2.0.0.Final</version.org.jboss.spec.javax.ejb.jboss-ejb-api_3.2_spec>
         <version.org.jboss.spec.javax.interceptor.jboss-interceptors-api_1.2_spec>2.0.0.Final</version.org.jboss.spec.javax.interceptor.jboss-interceptors-api_1.2_spec>
         <version.org.jboss.spec.javax.jms.jboss-jms-api_2.0_spec>2.0.0.Final</version.org.jboss.spec.javax.jms.jboss-jms-api_2.0_spec>
-        <version.org.jboss.spec.javax.xml.bind-api_2.3_spec>1.0.1.Final</version.org.jboss.spec.javax.xml.bind-api_2.3_spec>
+        <version.org.jboss.spec.javax.xml.bind-api_2.3_spec>2.0.0.Final</version.org.jboss.spec.javax.xml.bind-api_2.3_spec>
```
jamezp, line 69: This should probably be 2.0.1.Final.

jamezp (commented):
Firstly I need to apologize for the false alarm here. I really should have taken the few minutes to look what the config change meant. I think it's fine and again sorry for raising concern.

That said this does have a conflict. Once resolved I think we can merge this.

jamezp:
Replaced by #2856

PR #2837, opened 2021-06-14
[RESTEASY-2936] Resolve runtime deprecation warnings for Integer, Flo…

rsearls:
…at, Byte, Short, Double, Long

This resolves runtime warnings.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/util/TypeConverter.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/util/TypeConverter.java
@@ -75,14 +75,14 @@ public final class TypeConverter
       }
       if (Character.class.equals(targetType))
       {
-         if (source.length() == 0) return targetType.cast(new Character('\0'));
-         return targetType.cast(new Character(source.charAt(0)));
+         if (source.length() == 0) return targetType.cast('\0');
+         return targetType.cast(source.charAt(0));
       }
       if (char.class.equals(targetType))
       {
          Character c = null;
-         if (source.length() == 0)c = new Character('\0');
-         else c = new Character(source.charAt(0));
+         if (source.length() == 0)c = Character.valueOf('\0');
+         else c = Character.valueOf(source.charAt(0));
          try
          {
             return (T)Character.class.getMethod("charValue").invoke(c);
--- a/resteasy-links/src/test/java/org/jboss/resteasy/test/i18n/TestMessagesAbstract.java
+++ b/resteasy-links/src/test/java/org/jboss/resteasy/test/i18n/TestMessagesAbstract.java
@@ -33,7 +33,7 @@ public abstract class TestMessagesAbstract extends TestMessagesParent
 
       Assert.assertEquals(getExpected(BASE + "00", "cannotGuessCollectionType"), Messages.MESSAGES.cannotGuessCollectionType());
       Assert.assertEquals(getExpected(BASE + "20", "discoveryFailedForMethod", "class", "method", "s"), Messages.MESSAGES.discoveryFailedForMethod("class", "method", "s"));
-      Assert.assertEquals(getExpected(BASE + "35", "failedToInjectLinks", new Integer(17)), Messages.MESSAGES.failedToInjectLinks(new Integer(17)));
+      Assert.assertEquals(getExpected(BASE + "35", "failedToInjectLinks", 17), Messages.MESSAGES.failedToInjectLinks(17));
       Assert.assertEquals(getExpected(BASE + "60", "notEnoughtUriParameters", 3, 5), Messages.MESSAGES.notEnoughtUriParameters(3, 5));
    }
 
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/multipart/ContextProvidersTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/multipart/ContextProvidersTest.java
@@ -1,7 +1,6 @@
 package org.jboss.resteasy.test.providers.multipart;
 
 import static org.hamcrest.CoreMatchers.hasItems;
-import static org.hamcrest.CoreMatchers.is;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ReflectPermission;
@@ -194,7 +193,7 @@ public class ContextProvidersTest {
             InputPart inputPart = it.next();
             customers.add(inputPart.getBody(ContextProvidersCustomer.class, null).getName());
          }
-         Assert.assertThat("Wrong count of customers from response", new Integer(customers.size()), is(2));
+         Assert.assertEquals("Wrong count of customers from response", 2, customers.size());
          Assert.assertThat("Received customers list do not contain all items", customers, hasItems("Bill"));
          Assert.assertThat("Received customers list do not contain all items", customers, hasItems("Bob"));
       } catch (Exception e) {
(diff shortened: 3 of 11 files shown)


resteasy-links/src/test/java/org/jboss/resteasy/test/i18n/TestMessagesAbstract.java
```diff
@@ -33,7 +33,7 @@ public void testLocale() throws Exception
 
       Assert.assertEquals(getExpected(BASE + "00", "cannotGuessCollectionType"), Messages.MESSAGES.cannotGuessCollectionType());
       Assert.assertEquals(getExpected(BASE + "20", "discoveryFailedForMethod", "class", "method", "s"), Messages.MESSAGES.discoveryFailedForMethod("class", "method", "s"));
-      Assert.assertEquals(getExpected(BASE + "35", "failedToInjectLinks", new Integer(17)), Messages.MESSAGES.failedToInjectLinks(new Integer(17)));
+      Assert.assertEquals(getExpected(BASE + "35", "failedToInjectLinks", Integer.valueOf(17)), Messages.MESSAGES.failedToInjectLinks(Integer.valueOf(17)));
```
jamezp, line 36: These could just be `17`, there is no reason to wrap them in an object.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/multipart/ContextProvidersTest.java
```diff
@@ -194,7 +194,7 @@ void doTestGetList() throws Exception {
             InputPart inputPart = it.next();
             customers.add(inputPart.getBody(ContextProvidersCustomer.class, null).getName());
          }
-         Assert.assertThat("Wrong count of customers from response", new Integer(customers.size()), is(2));
+         Assert.assertThat("Wrong count of customers from response", Integer.valueOf(customers.size()), is(2));
```
jamezp, line 197: This is minor and applies to the rest of the changes in this file, but we shouldn't need to wrap these in an object. Also `Assert.assertThat` is deprecated as well so IMO we could switch to `Assert.assertEquals("Wrong count of customers from response", 2, customers.size())`.

resteasy-core/src/main/java/org/jboss/resteasy/util/TypeConverter.java
```diff
@@ -75,14 +75,14 @@ private TypeConverter()
       }
       if (Character.class.equals(targetType))
       {
-         if (source.length() == 0) return targetType.cast(new Character('\0'));
-         return targetType.cast(new Character(source.charAt(0)));
+         if (source.length() == 0) return targetType.cast(Character.valueOf('\0'));
```
jamezp, line 78: The boxing here is not required. It can just be `targetType.case('\0')`

resteasy-core/src/main/java/org/jboss/resteasy/util/TypeConverter.java
```diff
@@ -75,14 +75,14 @@ private TypeConverter()
...
+         return targetType.cast(Character.valueOf(source.charAt(0)));
```
jamezp, line 79: The boxing here is not required. It can just be `targetType.case(source.charAt(0))`

resteasy-core/src/main/java/org/jboss/resteasy/util/TypeConverter.java
```diff
@@ -75,14 +75,14 @@ private TypeConverter()
...
       }
       if (char.class.equals(targetType))
       {
          Character c = null;
-         if (source.length() == 0)c = new Character('\0');
-         else c = new Character(source.charAt(0));
+         if (source.length() == 0)c = Character.valueOf('\0');
+         else c = Character.valueOf(source.charAt(0));
```
jamezp, line 85: Same comments as above here.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/plain/resource/DefaultNumberWriterResource.java
```diff
@@ -12,27 +12,27 @@
    @GET
    public Response get(@PathParam("type") String type) throws Exception {
       if ("Byte".equals(type)) {
-         return Response.ok().entity(new Byte((byte) 123)).build();
+         return Response.ok().entity(Byte.valueOf((byte) 123)).build();
```
jamezp, line 15: This comment is meant for all the primitives here, they don't need to be wrapped in the `*.valueOf()`. They can just be cast and they'll be autoboxed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/basic/resource/DefaultMediaTypeResource.java
```diff
@@ -81,14 +81,14 @@ public Response postFoo(DataSource source) throws Exception {
    @Produces(MediaType.TEXT_PLAIN)
    @POST
    public Response postIntProduce(DataSource source) throws Exception {
-      return Response.ok().entity(new Integer(8)).build();
+      return Response.ok().entity(Integer.valueOf(8)).build();
```
jamezp, line 84: This can just be `8` and autoboxed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/basic/resource/DefaultMediaTypeResource.java
```diff
@@ -81,14 +81,14 @@ public Response postFoo(DataSource source) throws Exception {
...
    }
 
    @Path("postInt")
    @Consumes(MediaType.APPLICATION_OCTET_STREAM)
    @POST
    public Response postInt(DataSource source) throws Exception {
-      return Response.ok().entity(new Integer(8)).build();
+      return Response.ok().entity(Integer.valueOf(8)).build();
```
jamezp, line 91: This can just be `8` and autoboxed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/HeaderParamsAsPrimitivesTest.java
```diff
@@ -398,11 +398,11 @@ public void testGetByte() {
          e.printStackTrace(new PrintWriter(errors));
          fail("resourceHeaderPrimitives.doGet() failed:\n" + errors.toString());
       }
-      resourceHeaderPrimitiveWrappers.doGet(new Byte((byte) 127));
+      resourceHeaderPrimitiveWrappers.doGet(Byte.valueOf((byte) 127));
```
jamezp, line 401: This applies to the changes in this test, I think they should just all be their actual primitive values. The test name indicates "Primitive" yet we're testing the object not a primitive.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/param/QueryParamAsPrimitiveTest.java
```diff
@@ -294,9 +294,9 @@ public void testGetByte() {
       resourceQueryPrimitives.doGet((byte) 127);
       resourceQueryPrimitiveWrappers.doGet((byte) 127);
       List<Byte> list = new ArrayList<Byte>();
-      list.add(new Byte((byte) 127));
-      list.add(new Byte((byte) 127));
-      list.add(new Byte((byte) 127));
+      list.add(Byte.valueOf((byte) 127));
+      list.add(Byte.valueOf((byte) 127));
+      list.add(Byte.valueOf((byte) 127));
```
jamezp, line 299: These should just be `(byte) 127`

jamezp (commented):
Overall I think this is pretty good. I do think we can drop a lot of the boxing into the types.

rsearls:
Changes complete

jamezp (approved):
This is excellent. Thanks @rsearls.

PR #2850, opened 2021-06-21
[RESTEASY-2882] Changed the processing of ResponseExceptionMapper to …

rsearls:
…meet spec requirements

### DIFF
--- a/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
+++ b/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.microprofile.client;
 
+import org.jboss.logging.Logger;
 import org.eclipse.microprofile.rest.client.ext.ResponseExceptionMapper;
 import org.jboss.resteasy.client.jaxrs.internal.ClientResponse;
 import org.jboss.resteasy.client.jaxrs.internal.ClientResponseContextImpl;
@@ -22,6 +23,7 @@ import java.util.Set;
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class ExceptionMapping implements ClientResponseFilter {
     public static class HandlerException extends ResponseProcessingException {
+        private static final Logger LOGGER = Logger.getLogger(ExceptionMapping.HandlerException.class);
         protected ClientResponse handled;
         protected List<ResponseExceptionMapper> candidates;
 
@@ -40,9 +42,16 @@ public class ExceptionMapping implements ClientResponseFilter {
                 if (exception instanceof RuntimeException) throw (RuntimeException) exception;
                 if (exception instanceof Error) throw (Error) exception;
                 for (Class exc : method.getExceptionTypes()) {
-                    if (exc.isAssignableFrom(exception.getClass())) throw (Exception) exception;
+                    if (exception != null && exc.isAssignableFrom(exception.getClass())) {
+                        throw (Exception) exception;
+                    }
                 }
             }
+            // falling through to here means no applicable exception mapper found
+            // or applicable mapper returned null
+            LOGGER.warnf("No default ResponseExceptionMapper found or user's ResponseExceptionMapper returned null."
+                + "  Response status: %s  messge: %s", handled.getStatus(), handled.getReasonPhrase());
+
         }
     }
 
--- a/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
+++ b/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
@@ -151,6 +151,8 @@ public class ProxyInvocationHandler implements InvocationHandler {
                 }
                 if (cause instanceof ExceptionMapping.HandlerException) {
                     ((ExceptionMapping.HandlerException)cause).mapException(method);
+                    // no applicable exception mapper found or applicable mapper returned null
+                    return null;
                 }
                 if (cause instanceof ResponseProcessingException) {
                     ResponseProcessingException rpe = (ResponseProcessingException) cause;
--- a/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -451,6 +451,9 @@ public class RestClientBuilderImpl implements RestClientBuilder {
                 // ignore cast exception
             }
         }
+        if (disabled) {
+            LOGGER.warnf("The default ResponseExceptionMapper has been disabled");
+        }
         return disabled;
     }
 
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/ExceptionMapperReturnsNullTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/ExceptionMapperReturnsNullTest.java
@@ -1,0 +1,73 @@
+package org.jboss.resteasy.test.microprofile.restclient;
+
+import org.eclipse.microprofile.rest.client.RestClientBuilder;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
+import org.jboss.resteasy.test.microprofile.restclient.resource.HealthCheckData;
+import org.jboss.resteasy.test.microprofile.restclient.resource.HealthService;
+import org.jboss.resteasy.test.microprofile.restclient.resource.Ignore404ExceptionMapper;
+import org.jboss.resteasy.utils.PortProviderUtil;
+import org.jboss.resteasy.utils.TestUtil;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.net.URL;
+
+/**
+ * @tpSubChapter MicroProfile rest client
+ * @tpChapter Integration tests
+ * @tpTestCaseDetails The microprofile-rest-client 2.0 specification describes use of a custom
+ * ResponseExceptionMapper implementation and a default ResponseExceptionMapper.
+ * There is a scenario in which the default ResponseExceptionMapper has been disabled
+ * and the custom ResponseExceptionMapper returns null. This test verifies the proper
+ * behavior for this scenario and when the default ResponseExceptionMapper is present.
+ * @tpSince RESTEasy 4.7.0
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class ExceptionMapperReturnsNullTest {
+
+    @Deployment
+    public static Archive<?> deploy()
+    {
+        WebArchive war = TestUtil.prepareArchive(ExceptionMapperReturnsNullTest.class.getSimpleName());
+        return TestUtil.finishContainerPrepare(war, null);
+    }
+
+    private String generateURL(String path)
+    {
+        return PortProviderUtil.generateURL(path,
+                ExceptionMapperReturnsNullTest.class.getSimpleName());
+    }
+
+    @Test
+    public void testNoApplicableExceptionMapper() throws Exception
+    {
+        HealthCheckData data = RestClientBuilder.newBuilder()
+                .property("microprofile.rest.client.disable.default.mapper",true)
+                .baseUrl(new URL(generateURL("/")))
+                .register(new Ignore404ExceptionMapper())
+                .build(HealthService.class)
+                .getHealthData();
+        Assert.assertNull(data);
+    }
+
+    @Test
+    public void testDefaultExceptionMapper()
+    {
+        try {
+            RestClientBuilder.newBuilder()
+                    .baseUrl(new URL(generateURL("/")))
+                    .register(new Ignore404ExceptionMapper())
+                    .build(HealthService.class)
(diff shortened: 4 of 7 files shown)


resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
```diff
@@ -151,6 +151,8 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 }
                 if (cause instanceof ExceptionMapping.HandlerException) {
                     ((ExceptionMapping.HandlerException)cause).mapException(method);
+                    // no applicable exception mapper found or applicable mapper returned null
+                    return null;
```
jamezp, line 155: I'm not sure yet what the right answer is, but I'm not sure `null` should be returned here. Since this is invoking a proxy on an interface returning `null` here may break the interface contract. I don't think we can make assumptions on the implementation.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
```diff
@@ -40,9 +40,13 @@ public void mapException(final Method method) throws Exception {
                 if (exception instanceof RuntimeException) throw (RuntimeException) exception;
                 if (exception instanceof Error) throw (Error) exception;
                 for (Class exc : method.getExceptionTypes()) {
-                    if (exc.isAssignableFrom(exception.getClass())) throw (Exception) exception;
+                    if (exception != null && exc.isAssignableFrom(exception.getClass())) {
```
jamezp, line 43: Does this ever return an array with a null value? I'm just curious because that seems weird if it does.

jamezp (changes requested):
The test doesn't look quite right to me. I made some comments on it, but it also needs some kind of resource to invoke too. Something like:
```
@Path("/")
public class HealthResource {

    @GET
    @Path("/health")
    public Response status() {
        return Response.status(Response.Status.SERVICE_UNAVAILABLE).entity(new HealthCheckData("Down")).build();
    }
}
```

jamezp:
Actually maybe you meant the 404 to be thrown which is why there wasn't a resource. If that is the case we should probably change the name of the test and the exception mapper to `Ingore404*` instead.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/Ignore503ExceptionMapperTest.java
```diff
@@ -0,0 +1,85 @@
...
+ * @tpTestCaseDetails The microprofile-rest-client 2.0 specification describes use of a custom
+ * ResponseExceptionMapper implementation and a default ResponseExceptionMapper.
+ * There is a scenario in which the default ResponseExceptionMapper has been disabled
+ * and the custom ResponseExceptionMapper returns null. This test verifies the proper
+ * behavior for this scenario and when the default ResponseExceptionMapper is present.
+ * @tpSince RESTEasy 4.7.0
+ */
+@RunWith(Arquillian.class)
+@RunAsClient
+public class Ignore503ExceptionMapperTest {
```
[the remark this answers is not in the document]
rsearls: MicroProfileDependent.class does not yet exist in main

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/Ignore503ExceptionMapperTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @Deployment
+    public static Archive<?> deploy()
+    {
+        WebArchive war = TestUtil.prepareArchive(Ignore503ExceptionMapperTest.class.getSimpleName());
+        war.addClass(Ignore503ExceptionMapperTest.class);
+        war.addClass(Ignore503ExceptionMapper.class);
+        war.addClass(HealthService.class);
+        war.addClass(HealthCheckData.class);
+        war.addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+        war.addAsManifestResource(new StringAsset("Dependencies: org.eclipse.microprofile.restclient,org.jboss.resteasy.resteasy-rxjava2 services\n"), "MANIFEST.MF");
```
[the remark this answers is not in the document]
rsearls: We don't need a HealthResource class because we are forcing an error for Ignore503ExceptionMapper to process.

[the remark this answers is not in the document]
jamezp: Okay. We should probably rename it `Ignore404ExceptionMapper` then as that is what we're testing. Or just make it a generic `IgnoreExceptionMapper`. To me 503 implies we're ignoring 503 which we are not :)

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
rsearls: This PR is to properly respond to an edge case inferred by the microprofile-rest-client 2.0 specification.  The spec defines interface ResponseExceptionMapper,
`This mapper will take a Response object retrieved via an invocation of a client 
  and convert it to a Throwable, if applicable.`

The ResponseExceptionMapper method toThrowable(Response),
`... may return null if no throwable should be raised.`

The user may provide custom implementations of this interface to handle exceptions.
A provider must provide an implementation of the interface
`Each implementation will provide out of the box a ResponseExceptionMapper implementation`

The default mapper can be disabled.
`This behavior can be disabled by adding a configuration property 
  microprofile.rest.client.disable.default.mapper with value true`

There can be a scenario in which the implementor's default mapper has been disabled
and the user's mapper implementation returns NULL.  The spec makes no statement about
handling of this scenario.  I can only infer that NULL is to be returned to the caller.

We must allow all the mapper "candidates" to be processed lines 38-47 in ExceptionMapping,
thus providing every opportunity for a mapper to handle the issue.  
A return value of NULL must be checked for at line 43 in ExceptionMapping before
exc.isAssignableFrom(exception.getClass()) is executed.  If this is not done
exc.isAssignableFrom throws an invocation exception because NULL.getClass()
is not valid; that stops further processing of "candidates" and the user's
mapper response is superseded by a misleading exception.

The sole function of ExceptionMapping.HandlerException is to process and throw
an exception.  It is only used in an exception handling block.  It is only
used for microprofile-rest-client proxy handling.  I think it is safe to
assume if the call to method mapException returns normally a user's custom
mapper has return NULL.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
rsearls: Please see response in following section.

rsearls:
Changes complete

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
jamezp: I apologize this `null` check is absolutely correct otherwise `exception.getClass()` could throw an NPE.

It's the returning `null` that really concerns me. Without an `ResponseExceptionMapper` the error may be swallowed so someone getting an NPE because it returns `null` would have a hard time finding out what happened. We should probably at least log a debug message.

TBH though I don't have a better idea than `null` so it might be the best solution.

At least for the default though we should be throwing a `WebApplicationException` if we don't find a mapper.
> Each implementation will provide out of the box a ResponseExceptionMapper implementation that will map the response into a WebApplicationException whenever the response status code is >= 400.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/ExceptionMapperReturnsNullTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void testDefaultExceptionMapper()
+    {
+        try {
+            RestClientBuilder.newBuilder()
+                    .baseUrl(new URL(generateURL("/")))
+                    .register(new Ignore503ExceptionMapper())
+                    .build(HealthService.class)
+                    .getHealthData();
+            Assert.fail("Exception should have been returned");
+        } catch (Exception e) {
```
jamezp, line 69: We should probably specifically look for a `WebApplicationException` and maybe ensure the response is a 404.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
rsearls: > We should probably at least log a debug message.

That seems like a good idea.

> At least for the default though we should be throwing a WebApplicationException if we don't find a mapper.

I don't think that would be valid because the spec also states that the user can
disable the default mapper.

> This behavior can be disabled by adding a configuration property microprofile.rest.client.disable.default.mapper with value true

Perhaps adding a log message that the default mapper has been disabled will suffice.

rsearls: I've added the proposed log messages

jamezp: I am once again an idiot, sorry @rsearls. I was thinking the `ExceptionMapping` was the default mapper. It definitely is not, I'm sorry.

Maybe what we do is log warning a message before returning `null` that the exception wasn't handled and see if we can log some kind of information about it. There really isn't anything better than returning `null` with the exception of raising an exception, but it would take effort to disable raising an exception so it seems it would be intentional.

jamezp: lol you already do that, but could we do something like `cause.getCause()` and if not `null` log either the full exception or just the `cause.getCause().getLocalizedMessage();`?

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -451,6 +451,9 @@ private boolean isMapperDisabled() {
                 // ignore cast exception
             }
         }
+        if (disabled) {
+            LOGGER.warnf("The default ResponseExceptionMapper has been disabled");
```
jamezp, line 455: Do we not internationalize messages in the MP client?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/resource/Ignore503ExceptionMapper.java
```diff
@@ -0,0 +1,21 @@
+package org.jboss.resteasy.test.microprofile.restclient.resource;
+
+import org.eclipse.microprofile.rest.client.ext.ResponseExceptionMapper;
+
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.Response;
+
+public class Ignore503ExceptionMapper implements ResponseExceptionMapper<WebApplicationException> {
```
jamezp, line 9: I still think we should rename this :)

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
rsearls: I moved the log msg in ExceptionMapper so I could get access to the response object status code and message.

jamezp (approved):

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ExceptionMapping.java
```diff
@@ -40,9 +42,16 @@ public void mapException(final Method method) throws Exception {
...
+                        throw (Exception) exception;
+                    }
                 }
             }
+            // falling through to here means no applicable exception mapper found
+            // or applicable mapper returned null
+            LOGGER.warnf("No default ResponseExceptionMapper found or user's ResponseExceptionMapper returned null."
+                + "  Response status: %s  messge: %s", handled.getStatus(), handled.getReasonPhrase());
```
svenhaag, line 53: There is a typo in "%s message: %s"

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
```diff
@@ -151,6 +151,8 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 }
                 if (cause instanceof ExceptionMapping.HandlerException) {
                     ((ExceptionMapping.HandlerException)cause).mapException(method);
+                    // no applicable exception mapper found or applicable mapper returned null
```
svenhaag, line 154: couldn't this be a debug log message too?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/microprofile/restclient/ExceptionMapperReturnsNullTest.java
```diff
@@ -0,0 +1,73 @@
...
+    {
+        try {
+            RestClientBuilder.newBuilder()
+                    .baseUrl(new URL(generateURL("/")))
+                    .register(new Ignore404ExceptionMapper())
+                    .build(HealthService.class)
+                    .getHealthData();
+            Assert.fail("Exception should have been returned");
+        } catch (Exception e) {
+            // success exception was returned
```
svenhaag, line 70: Assert the expected exception class would be cool.

PR #2854, opened 2021-06-25
Exclude transitive dependencies from dependencies.

jamezp:


asoldano (commented):
@jamezp (/cc @ronsigal ), while this would prevent the dependencies' version clash we are having here, I don't think it would be OK. The idea of the "distribution" is to provide an archive containing all the needed dependencies to run RESTEasy from a flat classpath.
I believe the actual problem has to be fixed as proposed in https://github.com/resteasy/Resteasy/pull/2860 .
In the future, we can actually think about dropping the "distribution" completely, as it's likely soemthing obsolete in 2021... ;-)

jamezp:
Yes that would definitely fix it. I don't think we use JBoss Threads though, but it does get rid of the convergence.

I would actually like to get rid of the distribution and switch to creating a galleon pack.

liweinan:
For reference in `resteasy-spring-boot` it maintains a exhaustive exclusion list: 

- https://github.com/resteasy/resteasy-spring-boot/blob/main/resteasy-spring-boot-starter/pom.xml#L77
- https://github.com/resteasy/resteasy-spring-boot/blob/main/resteasy-spring-boot-starter/pom.xml#L357

jamezp:
Replaced with #2862

PR #2858, opened 2021-06-30
RESTEASY-2950 Exceptions thrown from onEvent processing should trigge…

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/RESTEASY-2950


jimma:
@RanabirChakraborty  Can you please add a test case for this change ?

NicoNes:
Hey @RanabirChakraborty, @jimma 

I was also wondering what is the expected behavior when an error is thrown from on an `onEvent` listener and I opened an issue for clarification to the expert group. https://github.com/eclipse-ee4j/jaxrs-api/issues/992.

As  I said to them, if the inspiration of Java `SseEventSource` is the JavaScript's `EventSource` then `onError` must not be invoked when `onEvent` listener throws an exception. JavaScript's `EventSource` does not behave like that.

Feel free to participate in the discussion.

Thanks

jamezp:
Given the issue was closed because the `onError()` should not be invoked. Therefore I'm closing this.

PR #2860, opened 2021-07-01
Resolve version convergence issue with jboss-threads dependency

asoldano:


asoldano:
@ronsigal , this is what I was discussing with you earlier today, please review.

PR #2869, opened 2021-07-20
[RESTEASY-1821] Allow JAX-RS annotations in CDI @Stereotype

petrberan:
https://issues.redhat.com/browse/RESTEASY-1821


jamezp:
I think we need a spec change for this. I've filed https://issues.redhat.com/browse/WFLY-16545 for WildFly and it broke some things. Jakarta REST 4.0 is moving towards requiring CDI and I think the annotations themselves should be bean defining annotations.

PR #2879, opened 2021-07-30
[RESTEASY-2974] ClientConfiguration.toString(null, type, annotations) causes NPE

petrberan:
https://issues.redhat.com/browse/RESTEASY-2974

### DIFF
--- a/resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/i18n/Messages.java
+++ b/resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/i18n/Messages.java
@@ -149,9 +149,6 @@ public interface Messages
    @Message(id = BASE + 185, value = "Unable to set http proxy")
    String unableToSetHttpProxy();
 
-   @Message(id = BASE + 190, value = "Parameter annotated with %s cannot be null")
-   String nullParameter(String annotation);
-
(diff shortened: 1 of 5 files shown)


jamezp:

jamezp (commented):
Overall I'm not too sure about this. The `ParamConverter` is pretty explicit that an `IllegalArgumentException` should be thrown if a parameter is null, https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/ext/paramconverter. I'm tempted to reject the JIRA in general based on that. Unfortunately, the reporter is not active on JIRA any longer.

petrberan:
@jamezp  I tried to reproduce the issue today and I cannot get the same error as described in the issue, since parent methods for toString() are checking for null. I have no idea how I managed to reproduce it last time, but since the behavior is more or less according to specs, I'd vote for resolving this one as Won't do. Or have you by any chance managed to get a reproducer?

That being said, the ParamConverter specs do not restrict IAE for PathParam only, but also to MatrixParam, QueryParam etc, yet I was able to run some tests passing null arg to other params without any error. Is that alright?

jamezp:
I was able to reproduce it, but really I think we shouldn't be bypassing what the JavaDoc indicates should be done here. For that reason I'll close this and reject the JIRA based on that. Thank you @petrberan for all the work on this.

PR #2880, opened 2021-08-05
[RESTEASY-2982] added security manager block

rsearls:


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ClientHttpEngineBuilder43.java
```diff
@@ -81,11 +82,22 @@ public ClientHttpEngine build()
...
+                    ServiceLoader.load(ClientConfigProvider.class).iterator();
+         } else {
+            clientConfigProviderIterator =
+                    AccessController.doPrivileged(new PrivilegedExceptionAction<Iterator<ClientConfigProvider>>() {
+               @Override
+               public Iterator<ClientConfigProvider> run() {
+                  return ServiceLoader.load(ClientConfigProvider.class).iterator();
+               }
+            });
+         }
```
jamezp, line 100: This unfortunately won't work. It's the actual iteration of the `ServiceLoader` where the error is thrown. It's a real bummer. However something like this might work.
```
private static <T> T findFirstService(final Class<T> service) {
    if (System.getSecurityManager() == null) {
        final ServiceLoader<T> loader = ServiceLoader.load(service, cl);
        if (loader.iterator().hasNext()) {
            return loader.iterator().next();
        }
        return null;
    }
    return AccessController.doPrivileged((PrivilegedAction<T>) () -> {
        final ServiceLoader<T> loader = ServiceLoader.load(service, cl);
        if (loader.iterator().hasNext()) {
            return loader.iterator().next();
        }
        return null;
    });
}
```

We have to do `null` checks or we use an `Optional`.

jamezp (changes requested):

jamezp:
I'm not sure what happened here, but it looks like the rebase got a bit weird. It's including some commits that have already been merged.

jamezp:
Just cleaning up old PR's. This has conflicts and looks out of date.

PR #2881, opened 2021-08-06
[RESTEASY-2975] Remove MicroProfile REST Client

jamezp:
Removes the MicroProfile REST Client and modules that require MicroProfile. These will be migrated to https://github.com/resteasy/resteasy-microprofile.

https://issues.redhat.com/browse/RESTEASY-2975


liweinan:
cool!

liweinan (approved):

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
```diff
@@ -420,7 +413,8 @@ public ResteasyClient build()
       if (serviceLoaderIterator.hasNext()) {
          config.register(new ClientConfigProviderFilter(serviceLoaderIterator.next()), Priorities.AUTHENTICATION);
       }
-      return createResteasyClient(engine, executor, cleanupExecutor, scheduledExecutorService, config);
+      final ContextualExecutorService executor = getExecutorService();
+      return createResteasyClient(engine, executor, !executor.isManaged(), ContextualExecutors.wrapOrLookup(scheduledExecutorService), config);
```
jimma, line 417: When create a `ResteasyClient` without setting a `ScheduledExecutorService`, the `ContextualExecutors.wrapOrLookup() `will always find the` ScheduledExecutorService` from a JNDI context after check the delegate is null. For the most ResteasyClient which is created and running in real client side instead of in container, this JNDI look up can be a lot of cost.  Should we optimize this to skip this lookup when it's not in container environment ?

jamezp: @jimma I couldn't find a good way to determine if the environment was in a Jakarta EE container or not. What I did was created a static map to hold the lookup names and a boolean of whether or not the lookup failed or succeeded. In environments that don't contain the Jakarta EE Concurrency utilities the lookup will only happen once.

jamezp:
Given there have been no objections, I'm merging this.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
jimma: @jamezp  OK. This already improved one part thing. Let's look at if there is something we can do more to improve this in the futurte.

jamezp: Sounds good to me @jimma. We can definitely improve this. It could be interesting to see if we can figure out a way to indicate whether we are running in a container or not.

PR #2882, opened 2021-08-12
[RESTEASY-2996] Calling send for each chunk with flush and completing the future

dnagarajan89:
This is an alternate approach to this [PR](https://github.com/resteasy/Resteasy/pull/2865) that is being discussed [here](https://gitter.im/reactor/reactor-netty?at=60f01fbff1274d5e504d7711). 

Instead of sending a flux of byte array, a mono of byte array is sent and flushed every time and the completion of the future is done on 'doOnSuccess' of the resulting Publisher.


server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -61,14 +62,16 @@
...
-           SinkSubscriber.subscribe(completionSink, Mono.from(reactorNettyResponse.sendByteArray(byteFlux)));
-
+           final Sinks.Many<Tuple2<byte[], CompletableFuture<Void>>> outSink =
+                   Sinks.many().unicast().onBackpressureBuffer();
+           final Flux<Void> flux = outSink.asFlux()
+                   .flatMap(tup ->
+                           Mono.from(reactorNettyResponse.send(
+                                   Mono.just(Unpooled.wrappedBuffer(tup.getT1())),
+                                   bb -> true
+                           )).doOnSuccess(v -> tup.getT2().complete(null))
```
crankydillo, line 72: You will also have to handle all the other cases, failure/cancellation/etc.

crankydillo (commented):
Did you wiretap the HTTP traffic and see if a flush was happening on each write?

crankydillo:
If we want to try to avoid the flush-on-every-asyncwrite performance problem, then:

1. We can consider making the `byte[]` array in `ProviderHelper` configurable such that we'd match the size netty uses before choosing to flush.
2. In `ChunkedOutputStream` do some accumulation of byte[] before providing it to `NettyOutbound#send`, but I don't see how that's good performance wise..

Just a few thoughts.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
crankydillo: I'm thinking we may not need `Sink` stuff if we are going to repeatedly call `NettyOutbound#send.  Seems like we can just move this to `asyncWrite` and have everything write to it.  Maybe we can be hooked into the `close` (or other 'done' hooks) method to complete `completionSink`.

server-adapters/resteasy-reactor-netty/src/test/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStreamTest.java
```diff
@@ -0,0 +1,306 @@
...
+        );
+
+        final byte[] errorBytes = "DISCARD".getBytes(StandardCharsets.UTF_8);
+        final CompletableFuture<Void> asyncWriteFuture = chunkOutputStream.asyncWrite(
+                errorBytes,
+                0,
+                errorBytes.length
+        );
+
+        StepVerifier.create(Mono.fromCompletionStage(asyncWriteFuture).then(completionSink.asMono()))
```
crankydillo, line 90: This looks nice.  My only concern is that this test has the possibility of not timing out.  There are multiple ways to deal with this.  On option is to put a timeout on the `Mono` you are passing the `StepVerifier#create`.

crankydillo (changes requested):
Looks good.  Just need to address timeout situation.  That _may_ be handled by something in Maven.  If so, just let me know.  I probably should have done a build to see, but I'm being lazy..

To test, just comment out the `doOnDiscard` in the prod code and run this test.  See if it completes.  We definitely need it to complete from the Maven build, and it would be nice if it completed from the IDE without too many hurdles (e.g. having to respect a bunch of Maven config).

dnagarajan89:
@jamezp / @asoldano - Please help reviewing this change. This is a bug that is currently affecting streaming cases with reactor-netty server adapter. Please refer the "Steps to Reproduce" section on [this JIRA issue](https://issues.redhat.com/browse/RESTEASY-2996) to reproduce the problem with the existing code.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,147 @@
...
+                    .send(Mono.just(Unpooled.wrappedBuffer(bytes)), FLUSH_ON_EACH_WRITE)
+                    .then()
+                    .doOnError(err -> completionSink.emitError(err, Sinks.EmitFailureHandler.FAIL_FAST))
+                    .doOnCancel(() -> completionSink.emitError(
+                            RESPONSE_WRITE_ABORTED_ON_CANCEL,
+                            Sinks.EmitFailureHandler.FAIL_FAST
+                    ))
+                    .doOnDiscard(
+                            ByteBuf.class,
+                            byteBuf -> completionSink.emitError(
```
crankydillo, line 130: Could you use netty technique for detecting mem leak and see if any happens?  You may need to call `ByteBuf#release` here, but I kind of doubt it..

dnagarajan89: I can see that the ByteBuf release on discard is handled in NettyOutbound#send [here](https://github.com/reactor/reactor-netty/blob/main/reactor-netty-core/src/main/java/reactor/netty/channel/ChannelOperations.java#L287).

crankydillo: The reason I think that's an acceptable answer is because of [this](https://netty.io/wiki/reference-counted-objects.html#who-destroys-it).  Otherwise, I'd say we can't rely on other's implementation (unless we must).

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,147 @@
...
+     * Used in {@link NettyOutbound#send(Publisher, Predicate)} to trigger flushing the bytes before the {@link
+     * CompletableFuture} returned from `NettyOutbound#then().toFuture()` completes.
+     */
+    private static final Predicate<ByteBuf> FLUSH_ON_EACH_WRITE = bb -> true;
+
+    private final ReactorNettyHttpResponse parentResponse;
+
+    static final Throwable RESPONSE_WRITE_ABORTED_ON_CANCEL = new WriterException("Response write aborted due to cancellation");
+
+    static final Throwable RESPONSE_WRITE_ABORTED_ON_DISCARD = new WriterException("Response write aborted due to discarded bytes");
```
jamezp, line 53: This seems a little strange to me. It would end up with an odd stack trace.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,147 @@
...
+
+    @Override
+    public void write(int b) {
+        write(new byte[]{(byte) b}, 0, 1);
+    }
+
+    @Override
+    public void write(byte[] bs, int off, int len) {
+        try {
+            asyncWrite(bs, off, len).toCompletableFuture().get();
```
jamezp, line 73: Please excuse my ignorance here as I'm not too familiar with this module :) This could potentially block forever with the `get()`. Should it have a timeout or not block at all?

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,147 @@
...
-
-        byte[] bytes = bs;
-        if (offset != 0 || length != bs.length) {
-            bytes = Arrays.copyOfRange(bs, offset, offset + length);
+    }
+
+    @Override
+    public void flush() {
+        try {
+            asyncFlush().get();
```
jamezp, line 85: Same comment as above here.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,147 @@
...
+
+    @Override
+    public CompletableFuture<Void> asyncFlush() {
+        // Everything flows through asyncWrite and we are flushing on each call, so we
+        // will treat that as a no-op for now, assuming that callers of this would have
+        // chained this onto an asyncWrite.  I hope that doesn't mess up SSE..
+        return CompletableFuture.completedFuture(null);
+    }
+
+    static class ReactorNettySendException extends RuntimeException {
```
jamezp, line 102: This should likely be moved up to the package and made public. There's no way for a user to catch this explicit exception when it's thrown.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
crankydillo: This is more or less the old code moved around, so I'll chime in on this one..  This signature comes from RestEasy's API.  We did assume it meant that the write should be completely done (at least ignoring flushing concepts) before returning.  As such, any responsibility for enforcing a timeout would be at a higher level..

crankydillo (approved):
I definitely think all others' comments should be addressed.  However, the fundamentals look OK to me.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
jamezp: Makes sense to me. I kind of had a feeling that was it, but I figured I'd at least ask :)

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
dnagarajan89: i have now used a simple error to indicate 'could not complete response write'.

https://github.com/resteasy/Resteasy/blob/3c53fd8b7cae7387697dc84dbf6227ef7f4f5602/server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java#L51

Please check this and let me know if this is clear enough to indicate something went wrong during response write operation.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -48,9 +48,7 @@
 
     private final ReactorNettyHttpResponse parentResponse;
 
-    static final Throwable RESPONSE_WRITE_ABORTED_ON_CANCEL = new WriterException("Response write aborted due to cancellation");
-
-    static final Throwable RESPONSE_WRITE_ABORTED_ON_DISCARD = new WriterException("Response write aborted due to discarded bytes");
+    static final Throwable RESPONSE_WRITE_ABORTED = new WriterException("Response write aborted abruptly");
```
jamezp, line 51: I still don't think this should be static. To me the stack trace would look very odd as it's not coming from where the exception is actually thrown.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,139 @@
...
-            byteSink = byteSinkSupplier.get();
-            parentResponse.committed();
-            started = true;
+public class ChunkOutputStream extends AsyncOutputStream {
+
+    /**
+     * This is the {@link Mono} that we return from {@link ReactorNettyJaxrsServer.Handler#handle(HttpServerRequest,
+     * HttpServerResponse)}
+     */
+    protected final Sinks.Empty<Void> completionSink;
```
jamezp, line 34: Is there a reason this can't remain private?

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,139 @@
...
-   }
-
-   @Override
-   public CompletableFuture<Void> asyncWrite(final byte[] bs, int offset, int length) {
-        final CompletableFuture<Void> cf = new CompletableFuture<>();
-        if (!started) {
-            byteSink = byteSinkSupplier.get();
-            parentResponse.committed();
-            started = true;
+public class ChunkOutputStream extends AsyncOutputStream {
```
jamezp, line 28: Is there a reason to make this public? Once it's public we can no longer make any breaking changes.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
```diff
@@ -1,156 +1,139 @@
...
+
+    private final NettyOutbound nettyOutbound;
+
+    /**
+     * Used in {@link NettyOutbound#send(Publisher, Predicate)} to trigger flushing the bytes before the {@link
+     * CompletableFuture} returned from `NettyOutbound#then().toFuture()` completes.
+     */
+    private static final Predicate<ByteBuf> FLUSH_ON_EACH_WRITE = bb -> true;
+
+    private static final String RESPONSE_WRITE_ABORTED_ERROR = "Response write aborted abruptly";
```
jamezp, line 49: This is not typically how we'd do this. There is an interface, `org.jboss.resteasy.plugins.server.reactor.netty.i18n.Messages` that could be used for this. You'd add a method like:

```
@Message(id = BASE + 10, value = "Response write aborted abruptly")
WriterException responseWriteAborted();
```

You then reference it like `Messages.responseWriteAborted()`.

jamezp (commented):
For some reason GitHub won't show me the diff for `ReactorNettyHttpResponse` so hopefully this works :)

Could we not implementation [this](https://github.com/resteasy/Resteasy/blob/ca92887376a3c61d5287a1fe500e1f8ca34862e5/server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyHttpResponse.java#L102)? I'm not sure how critical it is.

[This](https://github.com/resteasy/Resteasy/blob/ca92887376a3c61d5287a1fe500e1f8ca34862e5/server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyHttpResponse.java#L98) should be `headers.getAll(key).add(0, String.valueOf(value));`

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
dnagarajan89: There is no reason to keep this private. I will update this.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
dnagarajan89: There is no specific reason. And we understand your concern of limiting the scopes and expanding only when needed. Will fix this too!.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ChunkOutputStream.java
dnagarajan89: > For some reason GitHub won't show me the diff for `ReactorNettyHttpResponse` so hopefully this works :)
> 
> Could we not implementation [this](https://github.com/resteasy/Resteasy/blob/ca92887376a3c61d5287a1fe500e1f8ca34862e5/server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyHttpResponse.java#L102)? I'm not sure how critical it is.
> 
> [This](https://github.com/resteasy/Resteasy/blob/ca92887376a3c61d5287a1fe500e1f8ca34862e5/server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyHttpResponse.java#L98) should be `headers.getAll(key).add(0, String.valueOf(value));`

I will make this change. But I am not sure about the missing implementation of `equalsIgnoreValueOrder`.

@anilgursel Please comment on this if you remember why we didn't implemented this operation.

jamezp: The `equalsIgnoreValueOrder` really might not be a bit deal, I was more curious than anything.

dnagarajan89:
@jamezp Please let me know if you are waiting for anything from me.

dnagarajan89:
> https://issues.redhat.com/browse/RESTEASY-3012

@jamezp are we good to accept this PR? The configurable buffer [PR](https://github.com/resteasy/resteasy/pull/2891) shall wait for the above story.  Because this bug is currently affecting the streaming scenarios in our apps and we are trying to get this onboard ASAP to use this fix.

jamezp (approved):

PR #2884, opened 2021-08-20
[RESTEASY-2988] RESTEasy Galleon Feature Pack

jamezp:
This was done in two commits for now in case we want to hold onto the `resteasy-jboss-modules.zip` for now.

https://issues.redhat.com/browse/RESTEASY-2988
https://issues.redhat.com/browse/RESTEASY-2991


liweinan:
Cool!

PR #2889, opened 2021-08-25
[RESTEASY-2990] Add NoRouteToHostException to the default HttpRequest…

jamezp:
…RetryHandler for changes in Java 13+.

https://issues.redhat.com/browse/RESTEASY-2990


jamezp:
This does not fix the issue. There is something else going on.

PR #2891, opened 2021-08-31
[RESTEASY-2995] Configurable buffer size for AsyncOutputStream writer.

dnagarajan89:
As this buffer is used to write data to OutputStream, in some cases like the resteasy-netty4 server adapter, the bytes chunks are written to the response and flushed on every chunk. If we go with the default for writing a data of size 10000 bytes it has to write and flush the data ~5 times. While this can have a slight impact on the performance it has a better memory utilisation with using a less buffer size. However for the applications that this memory usage is not a bigger concern as opposed to performance, we are facilitating with an option to override this buffer size.

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
@@ -1,5 +1,7 @@
 package org.jboss.resteasy.spi;
 
+import org.jboss.resteasy.spi.config.ConfigurationFactory;
+
 import java.io.OutputStream;
 import java.util.concurrent.CompletionStage;
 
@@ -7,6 +9,13 @@ import java.util.concurrent.CompletionStage;
  * OutputStream which supports async IO operations. Use these operations if you need to support async IO.
  */
 public abstract class AsyncOutputStream extends OutputStream {
+
+    public static int ASYNC_MESSAGE_WRITE_BUFFER_SIZE = ConfigurationFactory.getInstance()
+            .getConfiguration()
+            .getOptionalValue(
+                    "resteasy.async.output.stream.write.buffer.size",
+                    Integer.class
+            ).orElse(2048);
 
     /**
      * Flushes this async output stream.
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
@@ -10,7 +10,7 @@ import java.util.concurrent.CompletionStage;
  */
 public abstract class AsyncOutputStream extends OutputStream {
 
-    public static int ASYNC_MESSAGE_WRITE_BUFFER_SIZE = ConfigurationFactory.getInstance()
+    public static final int ASYNC_MESSAGE_WRITE_BUFFER_SIZE = ConfigurationFactory.getInstance()
             .getConfiguration()
             .getOptionalValue(
                     "resteasy.async.output.stream.write.buffer.size",
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
@@ -7,6 +7,8 @@ import javax.ws.rs.core.Variant.VariantListBuilder;
 import org.jboss.resteasy.spi.AsyncOutputStream;
 
 import com.ibm.asyncutil.iteration.AsyncTrampoline;
+import org.jboss.resteasy.spi.config.Configuration;
+import org.jboss.resteasy.spi.config.ConfigurationFactory;
 
 import java.io.BufferedReader;
 import java.io.ByteArrayOutputStream;
@@ -160,7 +162,12 @@ public final class ProviderHelper
     */
    public static CompletionStage<Void> writeTo(final InputStream in, final AsyncOutputStream out)
    {
-      final byte[] buf = new byte[2048];
+      final Configuration config = ConfigurationFactory.getInstance().getConfiguration();
+      final int bufferSize = config.getOptionalValue(
+              "resteasy.async.output.stream.writer.buffer.size",
+              Integer.class
+      ).orElse(2048);
+      final byte[] buf = new byte[bufferSize];
       return AsyncTrampoline.asyncWhile(
             read -> read != -1,
             read -> out.asyncWrite(buf, 0, read).thenApply(v -> asyncRead(in, buf)),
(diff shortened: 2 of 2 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
```diff
@@ -160,7 +162,12 @@ public static void writeTo(final InputStream in, final OutputStream out) throws
     */
    public static CompletionStage<Void> writeTo(final InputStream in, final AsyncOutputStream out)
    {
-      final byte[] buf = new byte[2048];
+      final Configuration config = ConfigurationFactory.getInstance().getConfiguration();
```
crankydillo, line 165: I wonder if we have to read this every time?  If not, I'd prefer to have a member variable that holds this size.  Going the static route is another option.  Do you happen to know what Resteasy's pattern for this is?  Do they tend to respect runtime config changes?

crankydillo (commented):
Of course, we'll need some tests, but I'm definitely a fan of holding off on that until we get some feedback from the Resteasy folks.

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
```diff
@@ -8,6 +10,13 @@
  */
 public abstract class AsyncOutputStream extends OutputStream {
 
+    public static int ASYNC_MESSAGE_WRITE_BUFFER_SIZE = ConfigurationFactory.getInstance()
+            .getConfiguration()
+            .getOptionalValue(
+                    "resteasy.async.output.stream.write.buffer.size",
+                    Integer.class
+            ).orElse(2048);
+
```
jamezp, line 19: I don't think this should be static. It can't be changed without, typically, restarting the JVM. The value could come from something like a deployment descriptor too which would not work well if there are multiple deployments.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
```diff
@@ -160,7 +160,7 @@ public static void writeTo(final InputStream in, final OutputStream out) throws
     */
    public static CompletionStage<Void> writeTo(final InputStream in, final AsyncOutputStream out)
    {
-      final byte[] buf = new byte[2048];
+      final byte[] buf = new byte[AsyncOutputStream.ASYNC_MESSAGE_WRITE_BUFFER_SIZE];
```
crankydillo, line 163: There are other places in this class where `byte[]` is used in a similar fashion.  If easy, I'd try to verify whether or not similar treatment should be given to them.  _But_ I absolutely wouldn't get bogged down on that.  I'm a 'build what you need' kind of guy!  Maybe @jamezp could help us on this..

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
dnagarajan89: @jamezp is there any other better option to host this configuration ?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
jamezp: I'm not too sure what to do with the other buffers. However, the property name seems to be specific for async calls and some of the other buffers aren't explicitly used for async calls.

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
jamezp: I'm trying to think of how this could be done. It might be best to make it an instance variable with a simple getter. It would only be looked up when the a new instance of an `AsyncOutputStream` is created.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/ProviderHelper.java
crankydillo: I'm definitely fine with the if-aint-broke-dont-fix-it mentality here!

resteasy-core-spi/src/main/java/org/jboss/resteasy/spi/AsyncOutputStream.java
dnagarajan89: The `ChunkOutputStream` is created for every response. If the fetching the value from the `Configuration` should be avoided on the `ProviderHelper#asyncWrite` method, in my understanding(considering the asyncWrite also will be called just once) having an instance value should also have a similar cost. Please correct me if I am wrong.

jamezp: That does seem correct to me now that I have a closer look. Sorry, I'm still familiarizing myself with this :) 

My only concern is this could impact performance to save the cost of the memory overhead.

dnagarajan89: Actually we are trying to increase the performance with a slight memory overhead. 

As you know, we are working on a bug in the reactor-netty adapter [here](https://github.com/resteasy/Resteasy/pull/2882) where we had to write and flush each chunk(which is of size 2048 due to the ProviderHelper#asyncWrite implementation). This works fine for small data sizes. When we deal with larger data sizes, instead of flushing on every 2048 bytes(which may impact the performance), the application can choose to configure a higher size for this buffer so that the frequency of write and flush shall be minimised to get a better performance with the cost of a slight memory overhead.

jamezp: Ah okay got it. What we might need to do, and not in this change, is look for a way to keep the `Configuration` associated with the context of "deployment" rather than look it up each time which is where the expense is. We can file another JIRA for that though.

However, for this specific change we should do the lookup instead of using a static variable which is not friendly in multi-tenant environments.

dnagarajan89: I understand that you would like to go back to adding this configuration lookup in the ProviderHelper#asyncWrite implementation. But based on your previous comment on this lookup being expensive, we are concerned on how big this impact would be on the performance provided the fact that this code will be executed for every request.

jamezp: Yes that is a concern of mine as well. However, it can't be static for containers like WildFly. Possibly next week I could look at a solution to better cache a configuration, but that won't do much good now. 

The expense is in the lookup of the `ConfigurationFactory` as it uses a `ServiceLoader` for the lookup.

jamezp: The only thing really blocking this is having the value static. We can't have that and really the fact that it's not final and public is a bit of a problem too.

jamezp:

PR #2892, opened 2021-08-31
Fully upgrade RestEasy to Jakarta APIs

martin-g:


martin-g:
3000 changed files! I really sorry about this! I blame Oracle for this! :-)

martin-g:
`mvn clean install` works for all modules before the Galleon-pack one. I have no experience with Galleon and Arquillian so these will be hard for me to upgrade.

jamezp:
Thank you for all this work @martin-g. Unfortunately we cannot do this at the moment. We are working on this, but we've got some other steps we need to make first. This is coming very soon and we will be Jakarta REST 3.0 compliant. The current goal is to do this in RESTEasy 6.0.

I've actually got a local branch I tested with and it does require some updates to Arquillian. Galleon will work just fine either way :) It's what WildFly uses to produce it's Jakarta EE 9 preview.

martin-g:
@jamezp Is there a way to use RestEasy in jakarta.servlet based app ?
I see several blog posts at https://resteasy.github.io/ but IMO with latest 5.x it is still based on javax.servlet.
I saw some news that WildFly 24+ is Jakarta friendly too but I still cannot see how this could be true.

jamezp:
@martin-g For standalone RESTEasy there is currently not a way to use Jakarta REST 3.0, just the namespace change. What WildFly does is transform the bytecode to handle the package rename. I really don't think we're that far off from RESTEasy which does the namespace change. 

However, we've moved all the MicroProfile stuff and Spring stuff out of the main project and into their own projects. We want do to a final release of 5.0.0 which will still be Jakarta REST 2.1 compliant to support those two new projects until they can migrate themselves, e.g. when the MicroProfile specs migrate resteasy-microprofile will be able to migrate.

The original plan was to make 5.x Jakarta REST 3.0. It just didn't work very cleanly because with MicroProfile and Spring there were a lot of transformations needed of binaries we don't produce. Moving these dependencies to their own project seemed cleaner and more user friendly.

All this said, it sounds like I should really produce a blog post to explain all this. :) Please feel free to ask any more questions or let me know of any concerns. Any feedback, positive or negative, is welcome and helpful for us to know what the community wants :)

martin-g:
@jamezp I am trying to migrate/update Apache Wicket from javax.servlet to jakarta.servlet. 
Wicket uses [CDI Unit](https://github.com/cdi-unit/cdi-unit/issues/195) which depends on [Apache DeltaSpike](https://issues.apache.org/jira/browse/DELTASPIKE-1434) and RestEasy.
DeltaSpike provided `jakarta` jars (via Maven classifier) in addition to the default (`javax` based).
Now I tried to use RestEasy 5.x but it seems it is far from it (3000 changed files!). 5.x is still in Alpha. And we will have to wait now for 6.x :-(

Using binary transformation works only locally. I need to make it easier for the end users to point to `jakarta` versions at Maven Central and build against it.

jamezp:
@martin-g Ah okay. Maybe what I can do then is create a jakarta-rest-3.0 branch and do a 6.0.0.Alpha1 based on that with the changes. I don't expect many changes for 5.0.0.Alpha1 now. I just called it Alpha because other projects need to test consuming it. Will an Alpha release work for now?

martin-g:
Yes! Thank you!

PR #2897, opened 2021-09-09
RESTEASY-3004 Add HTTPS and HTTP2 tests to the resteasy-client-vertx

liweinan:


liweinan:
our vertx client doesn't support https and http/2 yet:

image

curl can handle http2 correctly:

495FAE9F-59CE-4E93-B7FA-9B7DC7DEA4F9

liweinan:
Add options to Vertx mays HTTPS and HTTP2 working:

image

PR #2902, opened 2021-09-15
Avoid a payload byte[] copy when using reactor-netty HTTP client engine

crankydillo:
1. Some memory copying is avoided by directly connecting resteasy's
   request payload to the reactor-netty's 'outbound'.
2. Fix a bug where headers added by things like RestEasy
   MessageBodyWriters are appropriately added to the request that
   reactor-netty sends out.
3. If empty due to thread switching, establish ResteasyContext that has
   the Provider->ClientConfiguration such that Provider field level injection
   will work in when the request body is written.

Much credit for these changes is owed to @anilgursel and @dnagarajan89.

### DIFF
--- a/resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
+++ b/resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
@@ -21,8 +21,11 @@ import javax.ws.rs.client.InvocationCallback;
 import javax.ws.rs.client.ResponseProcessingException;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Providers;
 
+import io.netty.buffer.ByteBufOutputStream;
 import io.netty.channel.group.ChannelGroup;
+import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpMethod;
 import static java.util.Objects.requireNonNull;
 
@@ -31,8 +34,13 @@ import org.jboss.resteasy.client.jaxrs.internal.ClientConfiguration;
 import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
 import org.jboss.resteasy.client.jaxrs.internal.ClientRequestHeaders;
 import org.jboss.resteasy.client.jaxrs.internal.ClientResponse;
+import org.jboss.resteasy.client.jaxrs.internal.TrackingClientRequestHeaders;
+import org.jboss.resteasy.core.ResteasyContext;
 import org.jboss.resteasy.tracing.RESTEasyTracingLogger;
 import static org.jboss.resteasy.util.HttpHeaderNames.CONTENT_LENGTH;
+
+import org.jboss.resteasy.util.CaseInsensitiveMap;
+import org.jboss.resteasy.util.TrackingMap;
 import reactor.core.publisher.Mono;
 import reactor.netty.http.client.HttpClient;
 import reactor.netty.http.client.HttpClientResponse;
@@ -111,42 +119,8 @@ public class ReactorNettyClientHttpEngine implements ReactiveClientHttpEngine {
     public <T> Mono<T> submitRx(final ClientInvocation request,
                                  final boolean buffered,
                                  final ResultExtractor<T> extractor) {
-        final Optional<byte[]> payload =
-            Optional.ofNullable(request.getEntity()).map(entity -> requestContent(request));
-
-        final HttpClient.RequestSender requestSender =
-                httpClient
-                        .headers(headerBuilder -> {
-                            final ClientRequestHeaders resteasyHeaders = request.getHeaders();
-                            resteasyHeaders.getHeaders().entrySet().forEach(entry -> {
-                                final String key = entry.getKey();
-                                final List<Object> valueList = entry.getValue();
-                                valueList.forEach(value -> headerBuilder.add(key, value != null ? value : ""));
-                            });
-
-                            payload.ifPresent(bytes -> {
-
-                                headerBuilder.set(CONTENT_LENGTH, bytes.length);
-
-                                if (log.isDebugEnabled() &&
-                                    isContentLengthInvalid(resteasyHeaders.getHeader(CONTENT_LENGTH), bytes)) {
-                                    log.debug("The request's Content-Length header is replaced " +
-                                            " by the size of the byte array computed from the request entity.");
-                                }
-                            });
-                        })
-                        .request(HttpMethod.valueOf(request.getMethod()))
-                        .uri(request.getUri().toString());
-
-        // Please see https://github.com/reactor/reactor-netty/issues/585 to see why
-        // we do not use outbound.sendObject(object) API.
-        final HttpClient.ResponseReceiver<?> responseReceiver =
-            payload.<HttpClient.ResponseReceiver<?>>map(bytes -> requestSender.send(
-                (httpClientRequest, outbound) ->
-                        outbound.sendByteArray(Mono.just(bytes)))
-            ).orElse(requestSender);
-
-        final Mono<ClientResponse> responseMono = responseReceiver
+        final Mono<ClientResponse> responseMono =
+            send(request)
                 .responseSingle((response, bytes) -> bytes
                         .asInputStream()
                         .map(is -> toRestEasyResponse(request.getClientConfiguration(), response, is))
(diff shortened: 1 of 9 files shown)


resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
```diff
@@ -196,6 +170,81 @@ public ReactorNettyClientHttpEngine(final HttpClient httpClient,
...
+                    new TrackingClientRequestHeaders(
+                        resteasyReq.getClientConfiguration(),
+                        resteasyReq.getHeaders().getHeaders()
+                    )
+                );
+
+                try {
+                    sendRequestBody(resteasyReq, byteBufOutputStream);
+                } catch (final IOException ioe) {
+                    return Mono.error(ioe);
```
crankydillo, line 201: @anilgursel `sendRequestBody` was originally wrapping `IOException` with `RuntimeException`.  Do you see any problem with the replacement I've done here?

dnagarajan89 (approved):

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/util/CaseInsentiveMapTest.java
```diff
@@ -12,6 +13,17 @@
  */
 public class CaseInsentiveMapTest {
 
+   protected CaseInsensitiveMap<String> map;
```
lalifk, line 16: There's an existing method `testMap` below that's creating a new instance of `map`. Can that method use this map instead for consistency? (wasn't able to add my comment on that line).

resteasy-client-reactor-netty/src/test/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngineTest.java
```diff
@@ -490,6 +525,88 @@ public void testThatRequestContentLengthHeaderIsOverwritten() {
...
+        final Response response = target.request().post(Entity.text("marcel sent me some text!"));
+        assertEquals(200, response.getStatus());
+        assertEquals(WRITER_ESTABLISHED_HEADER_VAL, response.readEntity(String.class));
+    }
+
+    @Test
+    public void testCaseInsensitiveHeaderReplace() {
+        final String CONTENT_TYPE_CLIENT_HEADER_VAL = "text/plain";
+        final String CONTENT_TYPE_WRITER_HEADER_VAL = "application/json";
+        final String CONTENT_TYPE_CC_HEADER_NAME = "CONTENT-TYPE";
```
lalifk, line 584: LC  stands for Camel Case. What about CC? The string value there certainly isn't Camel Case if that's what it's referring to.

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/util/CaseInsentiveMapTest.java
crankydillo: Great point.  I actually think that was the intent, and we just missed it.

resteasy-client-reactor-netty/src/test/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngineTest.java
dnagarajan89: I honestly don't remember why I used that naming here. Looking at the code what I wanted to indicate is only lower case and upper case. I will correct these.

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
jamezp: You could use an `UncheckedIOException` which extends a `RuntimeException` but is more descriptive.

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
```diff
@@ -196,6 +169,81 @@ public ReactorNettyClientHttpEngine(final HttpClient httpClient,
...
+                final TrackingMap<?> trackingMap = (TrackingMap<?>) resteasyReq.getHeaders().getHeaders();
+                trackingMap.getAddedOrUpdatedKeys()
+                    .forEach(key -> updateHeader(
+                            key,
+                            resteasyReq.getHeaders().getHeaders(),
+                            reactorReq.requestHeaders()
+                        )
+                    );
+                trackingMap.getRemovedKeys()
+                    .forEach(reactorReq.requestHeaders()::remove);
```
jamezp, line 213: I'm not really user I follow this. Could you explain the reasoning a bit?

jamezp (commented):
Overall this seems fine. I've got a couple questions on it, but not too many :)

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
crankydillo: Ultimately, we need to get all the headers established in the RestEasy client request into ReactorNetty's client request.  One of our first steps is to do this copy.  However, as we (painfully) found out, the `MessageBodyWriter` can add/remove/update headers.  So we have this 'tracking structure' that  @dnagarajan89 wrote to make adjustments to headers as a final step.

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
crankydillo: I'd like to stick with `Mono#error` over over using a nicer unchecked exception unless any of our reactor gurus complain.  I think this is considered kosher and won't have any unnecessary wrapping done solely to deal with lambda+checked-exceptions.

That said, I did not know that `UncheckedIOException` existed.  Thanks for the heads up!

jamezp: Ah I see. I thought you meant before it was doing something like `Mono.error(new RuntimeException(ioe))`. Yes, using `Mono.error` is definitely better than throwing an explicit exception.

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
jamezp: Okay got it. I think I'd rather see `TrackingClientRequestHeaders` and `TrackingMap` defined in this module rather than the `resteasy-client` and `resteasy-core` modules. Just keeps it more contained and doesn't make the public API for now :)

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
anilgursel: @crankydillo I cannot think of a problem at this moment.

resteasy-client-reactor-netty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngine.java
dnagarajan89: @jamezp we addressed your suggestion above on moving the `TrackingClientRequestHeaders` and `TrackingMap` to reactor-netty-client-integration module.

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/util/CaseInsentiveMapTest.java
```diff
@@ -12,13 +12,20 @@
  */
 public class CaseInsentiveMapTest {
 
+   /**
```
jamezp, line 15: I think this can be reverted. Other than that I think this looks fine.

crankydillo: Done!

jamezp (approved):

jamezp:
I'll get a release of out with this today as I need a 4.7.2 release for WildFly as well. First though I need to port these changes to main.

jamezp:
And crap, I just now realized this didn't have a JIRA. I'll have to file one.

PR #2934, opened 2021-10-20
RESTEASY-3031 Support `org.jboss.resteasy.http.client.engine` in resteasy-client

liweinan:


jamezp:
> @jamezp How about we keep the current `org.jboss.resteasy.http.client.engine` setting(which IMHO it's a good idea, because when I'm using it as a user by myself it's very convenient to pass setting via different configuration ways) and _also_ support `register()` method?
> 
> Meanwhile I'll start to work on implementing `register()` method.
> 
> Relative PR:
> 
> * [RESTEASY-3006 Allow microprofile-client to inject Vertx engine for HTTP/2 communication resteasy-microprofile#7](https://github.com/resteasy/resteasy-microprofile/pull/7)

@liweinan What is the advantage to using the property? I'm not saying I'm opposed to it, I'm just trying to understand :) A property requires the user to remember the property name where as `register()` you jut create the instance you want.

That said I _think_ it's legal to create use `register()`. The Javadoc didn't seem to be limit it to Jakarta RESTful Web Services types only.

jamezp:
> @jamezp Yeah seems pure prop injection has no advantage if only we can pass prop by method. I think more on the scenario, and the advantage of `prop` seems to be _non-intrusive_. So if we could provide a way that allows users to config the http engine by the config file, then this prop would become valuable.
> 
> And MP Config maybe a good way to use the prop, so I created a PoC PR here:
> 
> * [Allow HTTP Engine Injection via MP Config resteasy-microprofile#17](https://github.com/resteasy/resteasy-microprofile/pull/17)
> 
> Would you pls help to check it? Thanks :D

@liweinan let me think about this for over the weekend. I can see some advantage to that on the MicroProfile side, however we could still register it via the `ClientBuilder.register()` method.

PR #2938, opened 2021-10-25
[RESTEASY-3033] reproduction of SSE deadlock 

steven-aerts:
Reproduction scenario for RESTEASY-3033.
This is not a fix for the problem, it only reproduces the deadlock.


jamezp:
Replaced by #2962

PR #2945, opened 2021-10-29
[RESTEASY-3033]:Fix Deadlock while sending sse events when first even…

jimma:
…t is not yet send


jimma:
This is replaced with https://github.com/resteasy/resteasy/pull/2948

PR #2946, opened 2021-11-01
RESTEASY-3031 allow http client inject through registration / HTTP2 support via vertx engine

liweinan:


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
```diff
@@ -405,6 +405,13 @@ public ResteasyClient build()
          // check for proxy config parameters
          setProxyIfNeeded(config);
       }
+
+      for (Object p : getProviderFactory().getProviderInstances()) {
+         if (p instanceof ClientHttpEngine) {
+            httpEngine((ClientHttpEngine) p);
```
jamezp, line 411: This should break after this:
```java
for (Object p : getProviderFactory().getProviderInstances()) {
     if (p instanceof ClientHttpEngine) {
        httpEngine((ClientHttpEngine) p);
        break;
     }
 }
```

jamezp (changes requested):

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
liweinan: yes sir!

jamezp (approved):

PR #2948, opened 2021-11-02
[RESTEASY-3033]:Fix Deadlock while sending sse events when first event is not yet send

jimma:
* Remove the lock from  `DeferredOutputStream#onWritePossible()` and `DeferredOutputStream#onError()`. The queue is `ConcurrentLinkedQueue` and the` addToQueue()` is `synchronized`, so the queue order can be guaranteed.

* The `DeferredOutputStream`'s  `onWritePossible()` looks it adds to itself and it's executed once. When it is executed, it should not block adding queue operation and it doesn't hurt anything.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -144,7 +144,7 @@ public class HttpServletResponseWrapper implements HttpResponse
       private boolean asyncRegistered;
       private Queue<AsyncOperation> asyncQueue;
       private AsyncOperation lastAsyncOperation;
-      private boolean asyncListenerCalled;
+      private volatile boolean asyncListenerCalled;
 
       @Override
       public void write(int i) throws IOException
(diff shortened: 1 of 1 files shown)


jimma:
@steven-aerts I fixed this issue with another change. Can you please review and try if this is good ? Thanks.

jimma:
> I am wondering if this is all not related by `flushQueue` being called out of the protected/synchronized area, now that `onWritePossible` is not synchronized anymore.
> 
> I assume that you are able to reproduce all of this by running #2938, let me know if this is not the case.

Thanks for your review and analysis @steven-aerts .  I can reproduce this locally and I tried to play some ordering test and it turns out  event ordering issue is not caused by this change :  
https://github.com/jimma/Resteasy/commits/sse-event-order.  
I can always get the ordering check failures locally. This is another issue we need to have a look later.

steven-aerts:
@jimma

Is losing the last message with `.thenAccept(v -> sink.close());` issue also unrelated?  Or is is caused by `flushQueue` not being synchronized anymore?

 Thanks for the investigation.  I really appreciate it.

jimma:
> Is losing the last message with `.thenAccept(v -> sink.close());` issue also unrelated? Or is is caused by `flushQueue` not being synchronized anymore?
From my locally test , It is unrelated to the synchronized removal change.   Did you get the same result locally ?

steven-aerts:
@jimma was not able to see it without your patch, but it kept on deadlocking.  I do not mind to assume it is a separate problem.  And that the deadlock condition just happens a lot more on my issue than this close issue.

Do you want me to create separate issues in the bug-tracker for this reordering and prematurely closing of the connection?

jimma:
> Do you want me to create separate issues in the bug-tracker for this reordering and prematurely closing of the connection?

Sure. Please create an issue in [RESTEASY JIRA](https://issues.redhat.com/browse/RESTEASY).  Thanks again @steven-aerts

steven-aerts:
You are welcome.
I created [RESTEASY-3044](https://issues.redhat.com/browse/RESTEASY-3044) and [RESTEASY-3045](https://issues.redhat.com/browse/RESTEASY-3045) to track the two other issues.

jamezp:
Replaced by #2962

PR #2952, opened 2021-11-04
[RESTEASY-3037] Ignore tests that require tracing if the org.jboss.re…

jamezp:
…steasy.resteasy-tracing-api module is not present.

https://issues.redhat.com/browse/RESTEASY-3037

### DIFF
--- a/testsuite/integration-tests/pom.xml
+++ b/testsuite/integration-tests/pom.xml
@@ -118,6 +118,17 @@
             </plugins>
           </build>
         </profile>
+        <profile>
+            <id>skip.tracing.tests</id>
+            <activation>
+                <property>
+                    <name>skip.tracing.tests</name>
+                </property>
+            </activation>
+            <properties>
+                <additional.surefire.exclude.tracing.tests>,org.jboss.resteasy.category.TracingRequired</additional.surefire.exclude.tracing.tests>
+            </properties>
+        </profile>
     </profiles>
 
     <dependencies>
--- a/testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/category/TracingRequired.java
+++ b/testsuite/arquillian-utils/src/main/java/org/jboss/resteasy/category/TracingRequired.java
@@ -1,0 +1,28 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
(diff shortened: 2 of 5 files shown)


testsuite/integration-tests/pom.xml
```diff
@@ -118,6 +118,17 @@
             </plugins>
           </build>
         </profile>
+        <profile>
```
jbliznak, line 121: This profile depends on `${jboss.home}` being already set which might not be the case if `download` profile isn't processed yet.

This is IMHO the weak area of Maven, the profile processing order is somewhat undefined.

Looking into https://github.com/resteasy/resteasy/runs/4107356508?check_suite_focus=true in CI where ` mvn clean install -U -B -fae -Dserver.version=25.0.1.Final` is the command ( therefore `download` profile is activated) then eg. `ClosedResponseHandlingTest` should be executed, right? It did run in previous MR's CI https://github.com/resteasy/resteasy/runs/4086766225?check_suite_focus=true
But it seems that the excluding profile was processed first, not finding the file..

There is https://maven.apache.org/guides/introduction/introduction-to-profiles.html#profile-order so it might help if we switch the order of these two profiles in this file so that `download` is processed (if needed) before `skip.tracing.tests` but I don't understand that definition as unambiguous so who knows..

jbliznak: actually, not even switching the order is ok as maven profiles activation is evaluated only once so the unzip of downloaded server would still not happen before checking the other profile activation condition

jamezp: That is a good point. That might not be set at the time the model is created. I'll change the way this works. Thanks for checking @jbliznak!

jamezp: I went with just using the profile or system property. It seems to be the cleanest way.

jbliznak: yeah, that's probably ok.
The only other viable approach I can think of is using Assume in these tests - doing the same check as you initially proposed - presence of the tracing module file. But I like setting explicit option more - it tests what you want without magic.

jamezp: I agree. I probably change my opinion every time I think about on categories vs assumes :)

PR #2955, opened 2021-11-05
RESTEASY-3007 Jakarta RESTful Web Services 3.0 Upgrade

jamezp:
https://issues.redhat.com/browse/RESTEASY-3007

Renames packages and prepares for Jakarta EE 9.1


asoldano (commented):
thanks @jamezp , this is massive! I've made a few comments,but I think it would be great to have an additional review from someone else (or some later checks before release). In particular, I have not spent time on looking at things that might possibly have been missed here (not implying there is something missing, but just that I haven't checked that).

docbook/reference/en/en-US/modules/Asynchronous_HTTP.xml
```diff
@@ -67,9 +67,9 @@ public class SimpleResource
       </para>
       <para>
          <emphasis role="bold">NOTE:</emphasis> The old RESTEasy proprietary API for async http has been deprecated and may be removed as soon as RESTEasy 3.1.
-         In particular, the RESTEasy @Suspend annotation is replaced by <classname>javax.ws.rs.container.Suspended</classname>, and
+         In particular, the RESTEasy @Suspend annotation is replaced by <classname>jakarta.ws.rs.container.Suspended</classname>, and
          <classname>org.jboss.resteasy.spi.AsynchronousResponse</classname> is replaced by
-         <classname>javax.ws.rs.container.AsyncResponse</classname>. Note that @Suspended does not have a value field,
+         <classname>jakarta.ws.rs.container.AsyncResponse</classname>. Note that @Suspended does not have a value field,
```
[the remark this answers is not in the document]
jamezp: I think that makes sense. I might not to it in this PR. I've created a blocker JIRA to review this though https://issues.redhat.com/browse/RESTEASY-3047.

Really we might be able to get rid of any referring to 2.x and lower and make some new documentation on 3.x -> 4.x -> 5.x -> 6.x.

ronsigal:
Daaaaamn!! I didn't even know we has 3000 files. ;-)

arquillian/RESTEASY-1056-jetty-bv11/src/test/resources/arquillian.xml
```diff
@@ -4,8 +4,8 @@
    xsi:schemaLocation="http://jboss.org/schema/arquillian
         http://jboss.org/schema/arquillian/arquillian_1_0.xsd">
 
-   <!-- Force the use of the Servlet 3.0 protocol with all containers, as it is the most mature -->
-   <defaultProtocol type="Servlet 3.0" />
+   <!-- Force the use of the Server 5.0 protocol with all containers, as it is the most mature -->
+   <defaultProtocol type="Server 5.0" />
```
ronsigal, line 8: You mean "servlet"? According to https://medium.com/swlh/testing-jakarta-ee-9-components-with-arquillian-and-weld-f9e1add08895,

>By default, when arquillian-protocol-servlet-jakarta is ocurred in the classpath, the Servlet 5.0 protocol will be the default protocol, if you have declared the protocol in the existing arquillian.xml file, change it to use Servlet 5.0.

jamezp: Nice catch! I did indeed mean "Servlet 5.0".

ronsigal:
My first reaction: https://www.youtube.com/watch?v=eXrRUSRTj3k from The Quiet Man. ;-)

I downloaded the main branch with the PR applied and looked for anything that stood out. I didn't see anything terrible. If we wanted to do a complete Orwellian purge of all references to "jaxrs", then there are some things we could think about:

1. packages like org.jboss.resteasy.client.jaxrs.engines
2. an occasional use of "JAX-RS" in javadoc and messages
3. names like JaxrsInterceptorRegistryImpl

I don't think any of these are compelling. Item 1 would be somewhat consistent with the package changes in the spec API, but I'm not sure what would reasonably replace jaxrs. Item 2 would be consistent with the changes we made in the User Guide, and I think it's the most reasonable of the three. Item 3 ... meh, I don't think so.

Otherwise, thanks, James!!

jamezp:
@ronsigal I think we need to do 2, but 1 and 3 AIUI are okay because it would break compatibility. Granted, we are doing that already with the namespace change, but that's a lot less invasive than changing class names :)

jamezp:
With passing CI and at least 2 reviews I'm merging this :)

ronsigal:
+1

PR #2970, opened 2021-11-24
RESTEASY-3046 with other test fixes

jamezp:
Combined fix as the first commit seems to trigger the others failing more. Combining them just seemed correct.

Upstream #2968

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -65,10 +65,10 @@ public class HttpServletResponseWrapper implements HttpResponse
       {
          try
          {
-            stream.write(bytes, offset, length);
             // we only are complete if isReady says we're good to write, otherwise
             // we will be complete in the next onWritePossible or onError
             if(sos == null || sos.isReady()) {
+               stream.write(bytes, offset, length);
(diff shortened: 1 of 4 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -65,10 +65,10 @@ protected void doWork(ServletOutputStream sos)
       {
          try
          {
-            stream.write(bytes, offset, length);
             // we only are complete if isReady says we're good to write, otherwise
             // we will be complete in the next onWritePossible or onError
             if(sos == null || sos.isReady()) {
```
anthochristen, line 70: @jamezp This seems to fix the issue 👍, but just a question though if the condition becomes false then would it never complete as we no long complete on outside of `doWork` and we dont process in `onFlush` until `lastAsyncOperation` is complete?

jamezp: No worries at all @anthochristen. I guess things may never complete if `onWritePossible()` or `onError()` are not called. However, attempting to write outside of the check doesn't make much sense if `sos.isReady()` returns `false` as it would throw an exception since the stream is not ready.

I can't really think of a scenario where it wouldn't complete though. If you've got a scenario though I'm definitely interested and we can look at how to fix that.

PR #2983, opened 2021-12-06
[RESTEASY-3060] corrected ref to jakarta classes

rsearls:


jamezp (changes requested):
I think these `.jbang/*` files need to be removed. We might need to add `.jbang/*` to the `.gitignore` file.

jamezp (approved):

PR #2986, opened 2021-12-10
fix classname error in async doc

liweinan:


jamezp (approved):

liweinan:
@jamezp Thanks for checking!

PR #2994, opened 2021-12-15
[RESTEASY-3066]:Convert all docbook files to adoc format and generate…

jimma:
… html files from adoc


jamezp:

PR #2998, opened 2021-12-22
Enable PATCH method in metadata tooling.

crankydillo:


agaengel:
I just tried to activate resteasy-stats in one of out apps and it failed because auf the missing patch method. 
Please add this in a future version

jamezp:

crankydillo:
Hi.  Out of curiosity, what was wrong with this PR?  While it's not critical, our team could use this feature.

jamezp:
@crankydillo Nothing in particular. I was just cleaning up the PR queue is all. This was against 4.7 and the plan is to have no more 4.x releases. The only release streams will be 5.0, maybe a 5.1 if there is a good reason to continue Jakarta REST 2.1 support, and 6.2/6.3.

PR #3017, opened 2022-01-18
[RESTEASY-3079] Add support for injecting some types that are injecta…

jamezp:
…ble via @Context.

https://issues.redhat.com/browse/RESTEASY-3079

### DIFF
--- a/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
+++ b/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
@@ -3,17 +3,24 @@ package org.jboss.resteasy.cdi;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Member;
+import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.lang.reflect.Method;
+import java.util.Set;
+
 import jakarta.decorator.Decorator;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.RequestScoped;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.Any;
+import jakarta.enterprise.inject.Default;
+import jakarta.enterprise.inject.spi.AfterBeanDiscovery;
 import jakarta.enterprise.inject.spi.AnnotatedType;
 import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.BeanManager;
@@ -21,16 +28,19 @@ import jakarta.enterprise.inject.spi.BeforeBeanDiscovery;
 import jakarta.enterprise.inject.spi.Extension;
 import jakarta.enterprise.inject.spi.InjectionTarget;
 import jakarta.enterprise.inject.spi.ProcessAnnotatedType;
+import jakarta.enterprise.inject.spi.ProcessBean;
 import jakarta.enterprise.inject.spi.ProcessInjectionTarget;
 import jakarta.enterprise.inject.spi.ProcessSessionBean;
 import jakarta.enterprise.inject.spi.WithAnnotations;
 import jakarta.enterprise.util.AnnotationLiteral;
 import jakarta.ws.rs.Path;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.ClientBuilder;
 import jakarta.ws.rs.core.Application;
 import jakarta.ws.rs.ext.Provider;
-
 import org.jboss.resteasy.cdi.i18n.LogMessages;
 import org.jboss.resteasy.cdi.i18n.Messages;
+import org.jboss.resteasy.plugins.providers.RegisterBuiltin;
 import org.jboss.resteasy.util.GetRestful;
 
 /**
@@ -71,6 +81,7 @@ public class ResteasyCdiExtension implements Extension
    }
 
    private Map<Class<?>, Type> sessionBeanInterface = new HashMap<Class<?>, Type>();
+   private boolean generateClientBean = true;
 
    /**
     * Obtain BeanManager reference for future use.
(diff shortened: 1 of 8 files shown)


resteasy-cdi/src/main/resources/META-INF/beans.xml
```diff
@@ -1,6 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<beans xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
manovotn, line 2: Why do you remove this?
Asking because according to [CDI specification](https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#bean_archive), an archive with CDI extension and no `beans.xml` is _not a bean archive_.
That means no bean discovery is performed (i.e. `ContextProducers` class won't be discovered).

I think what you want to have is `beans.xml` with bean discovery mode `annotated` although I don't know every bit of your code :)

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
```diff
@@ -81,9 +87,27 @@ public static boolean isCDIActive()
...
    }
 
+
+   /**
+    * Registers a producer and disable for a {@link Client REST client}.
+    *
+    * @param event       the after bean discovery event
+    * @param beanManager the bean manager
+    */
+   public void registerClientProducer(@Observes AfterBeanDiscovery event, final BeanManager beanManager) {
```
manovotn, line 102: You should probably add a test that verifies this registration doesn't happen if the bean is user-provided.

Otherwise, to answer your email question @jamezp - the idea of conditional bean addition is fine however I wouldn't recommend using `beanManager.getBeans()` to determine whether there are such beans.

The reason is that BM is caller sensitive and in EE implements some visibility limitations. Therefore, `getBeans()` method returns whichever beans that given BM "can see" so to speak.
A more universal approach would be to add an extension method observing `ProcessBean<X>`. When there is a bean of type `X`, this observer will be notified and you can inspect the bean and store the information that you no longer need to provide a synthetic one. Here's the code:

```
private boolean generateClientBean = true;

public void observeProcessBean(@Observes ProcessBean<Client> pb) {
    // this method will get notified if there is any bean of type `Client` created
    Set<Annotation> qualifiers = pb.getBean().getQualifiers(); // you want to detect beans with no explicit qualifiers
    if (qualifiers.contains(Any.Literal.INSTANCE) && qualifiers.contains(Any.Literal.INSTANCE)) {
        generateClientBean = false;
    }
}

public void registerClientProducer(@Observes AfterBeanDiscovery event) {
    if (generateClientBean) {
        event.addBean().addTransitiveTypeClosure(Client.class)
                 .scope(ApplicationScoped.class)
                 .produceWith(instance -> ClientBuilder.newClient(RegisterBuiltin.getClientInitializedResteasyProviderFactory(getClassLoader())))
                 .disposeWith((client, instance) -> client.close());
    }
}
```

Note that calling `AfterBeanDiscovery.addBean()` will also fire a `ProcessSyntheticBean` event (subtype of `ProcessBean`) as [defined by specification](https://jakarta.ee/specifications/cdi/2.0/cdi-spec-2.0.html#after_bean_discovery). But that should do no harm in this case.

You could use similar approach for whichever type of bean you need to add conditionally.
Ask away if there is anything unclear.

manovotn (changes requested):

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
jamezp: @manovotn Excellent thanks! This makes sense to me and it didn't feel right to use the `BeanManager` there. I just wasn't sure what the best solution would be :)

I did notice though that using:
```java
public void processClientBean(@Observes final ProcessBean<Client> pb) {
   // this method will get notified if there is any bean of type `Client` created
   final Set<Annotation> qualifiers = pb.getBean().getQualifiers(); // you want to detect beans with no explicit qualifiers
   if (qualifiers.contains(Any.Literal.INSTANCE) && qualifiers.contains(Default.Literal.INSTANCE)) {
      generateClientBean = false;
   }
}
```
Didn't get triggered. It did work with something like:
```java
public <T> void processClientBean(@Observes final ProcessProducer<T, Client> pb) {
   generateClientBean = false;
}
```
I'm not sure if that is correct though. FWIW I defined my producer in my REST resource like:
```java

    @ApplicationScoped
    @jakarta.enterprise.inject.Produces
    public Client createClient() {
        return ClientBuilder.newClient();
    }
```
It seems odd to have it there for sure though so maybe that is my problem.

resteasy-cdi/src/main/resources/META-INF/beans.xml
jamezp: Added this back, but defaulted to 4.0.

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
```diff
@@ -81,9 +93,50 @@ public static boolean isCDIActive()
...
+    * @param pb the bean being processed
+    */
+   public void processClientBean(@Observes final ProcessBean<Client> pb) {
+      // this method will get notified if there is any bean of type `Client` created
+      final Set<Annotation> qualifiers = pb.getBean()
+              .getQualifiers(); // you want to detect beans with no explicit qualifiers
+      if (qualifiers.contains(Any.Literal.INSTANCE) && qualifiers.contains(Default.Literal.INSTANCE)) {
+         generateClientBean = false;
+      }
+   }
```
jamezp, line 113: @manovotn This never seems to be triggered, but the one below does.

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
```diff
@@ -81,9 +93,50 @@ public static boolean isCDIActive()
...
+
+   /**
+    * Process any producers for a client.
+    * @param producer the producer
+    * @param <T> the type of the bean class
+    */
+   public <T> void processClientProducers(@Observes final ProcessProducer<T, Client> producer) {
+      generateClientBean = false;
```
jamezp, line 121: It seems like here we might want to inspect the annotations to ensure there isn't some custom qualifier which produces it's own client.

manovotn: Note that `ProcessProducer` gets notified only for producer methods and field. Therefore, if user defined a class based bean with type `Client`, it won't trigger this observer. Same goes for a synthetic bean.

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
manovotn: This is probably because you are defining it as `ProcessBean<Client>`.
`ProcessBean<X>` is a very generic observer that gets notified for any shape of bean - producer method, producer field, class bean or synthetic bean. And `X` stands for the class of the bean which is the issue here; in your case, this is the class which contains the producer method.
Take a look at the javadoc here - https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/jakarta/enterprise/inject/spi/ProcessBean.java#L27-L39

Granted, I didn't realize it either when giving you the code snippet when I first reviewed this PR, so feel free to blame it on me :-)

You could either:
* Have several observer methods, one for each subclass of `ProcessBean` and perform the detection there
  * Has the upside of only being notified about events that you really want to know and is more performance friendly
  * But it is also a lot of code which isn't as nice to read 
* Use just `ProcessBean<?>` and inspect the bean type via something like `pb.getBean().getTypes().contains(Client.java)` to make sure it is the bean you are looking for
  * Easy to write and understand
  * Gets notified about every single bean in the deployment which might create some overhead in massive deployments

In both cases, you'll still want to add the check for `@Any` and `@Default` qualifiers.

resteasy-cdi/src/main/resources/META-INF/beans.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<beans xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-   xsi:schemaLocation="
-      http://java.sun.com/xml/ns/javaee 
-      http://java.sun.com/xml/ns/javaee/beans_1_0.xsd">
+<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd"
+       version="4.0">
```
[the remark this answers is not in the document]
jamezp: Okay. I've added `bean-discovery-mode="annotated"` now. Sorry, completely missed that part :)

resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java
jamezp: Okay done. That makes more sense to me. I've updated it to use `ProcessBean<?>` and that seems to be working. If we see deployment overhead we can make some changes.

manovotn (approved):
Can't think of anything else to pester you with @jamezp ;-)
LGTM now.

jamezp:
> Can't think of anything else to pester you with @jamezp ;-) LGTM now.

Thank you @manovotn! The help was much appreciated.

PR #3029, opened 2022-02-16
Resteasy 3078 has property

rsearls:


jamezp:
This seems like a rebase gone bad. I can't really tell what the changes are. I assume it's just https://github.com/resteasy/resteasy/pull/3029/commits/8466b14ce02ce8f0953f7ef579bf3f65da2e81eb we want though.

PR #3030, opened 2022-02-22
Make extractResult fail-fast when there's no content type

IvanVas:
This is a breaking change, it's probably a starting point. Probably a configuration is needed to opt-in for the feature.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
@@ -160,7 +160,7 @@ public class ClientInvocation implements Invocation
          {
             if (response.getMediaType() == null)
             {
-               return null;
+               throw new ProcessingException("Response has no content type");
             }
             else
             {


jamezp:
Hi @IvanVas, Can you explain the rational behind this? I've not read the spec or looked to see what happens with a `null` value, but I'd just like to understand it.

IvanVas:
@jamezp Thanks for looking into this!
If the server doesn't send the Content-Type header, the client gets null with no explanation. AFAIU, the only way to troubleshoot is debugging.

IvanVas:
@jamezp What do you think?

jamezp:
Hi @IvanVas. Sorry for the very long delay on this. I do think this makes sense. I filed https://issues.redhat.com/browse/RESTEASY-3178.

Do you know how/when the media type is null? I'm just looking for a way to reproduce it is all. However, I do agree returning `null` does not seem correct. I think, per the [documentation](https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/client/invocation#invoke()), the `[ResponseProcessingException](https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/client/ResponseProcessingException.html)` might be the correct exception to throw.

IvanVas:
> Do you know how/when the media type is null? I'm just looking for a way to reproduce it is all
The scenario: 
The server returns a response without `Content-Type`. Which is incorrect ofc, but we have to spend some time debugging in order to discover the problem. Having a fail-fast mechanism (exception) would help a lot!

If you can point me to the correct level of tests I can try to create a test which would reproduce this.

IvanVas:
> I think, per the [documentation](https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/client/invocation#invoke()), the `[ResponseProcessingException](https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/client/ResponseProcessingException.html)` might be the correct exception to throw.

Great! Will update this once we have a consensus on the approach.

jamezp:
> > Do you know how/when the media type is null? I'm just looking for a way to reproduce it is all
> 
> The scenario: The server returns a response without `Content-Type`. Which is incorrect ofc, but we have to spend some time debugging in order to discover the problem. Having a fail-fast mechanism (exception) would help a lot!
> 
> If you can point me to the correct level of tests I can try to create a test which would reproduce this.

In that case I'd almost say throwing the exception is correct. If users start seeing an issue then they could just create a `ClientResponseFilter` to set a `content-type` header.

IvanVas:
> > > Do you know how/when the media type is null? I'm just looking for a way to reproduce it is all
> > 
> > 
> > The scenario: The server returns a response without `Content-Type`. Which is incorrect ofc, but we have to spend some time debugging in order to discover the problem. Having a fail-fast mechanism (exception) would help a lot!
> > If you can point me to the correct level of tests I can try to create a test which would reproduce this.
> 
> In that case I'd almost say throwing the exception is correct. If users start seeing an issue then they could just create a `ClientResponseFilter` to set a `content-type` header.

I'll change the exception type then.

Will I be able to create an appropriate test with moderate effort?

jamezp:
@IvanVas This would be my suggestion #3176. I'm not sure how to create a test, but I'll try to come up with one.

jamezp:
I'm gonna close this in favor of #3176. Thank you @IvanVas!

PR #3034, opened 2022-02-28
[Resteasy 3096]: Resteasy new WebApplicationExceptions behavior

NicoNes:
Hi guys,

This PR is my proposal to fix inconsistency spotted by this [JIRA](https://issues.redhat.com/browse/RESTEASY-3096).

As I said in the JIRA description, the way `WebApplicationException` thrown by client in a server side context are handled, using subtypes of  `org.jboss.resteasy.client.exception.WebApplicationExceptionWrapper` to return a sanitized `jakarta.ws.rs.core.Response`  with no entity and only few important headers, looks fine to me and does not bother me.

But what bothers me more is, the insconsistency in the `WebApplicationException#getResponse()` method behavior:

- If  `WebApplicationException` is thrown by client in a server side context, then `WebApplicationException#getResponse()`  returns a sanitized reponse that does not match nor represents the remote side response.
- But if  `WebApplicationException` is thrown by client in a non server side context, then `WebApplicationException#getResponse()`  returns a reponse that match and represents the remote side response.

This inconsitency does not make sense to me and AFAIK is not written in the JAX-RS spec. More, in case of  `WebApplicationException` thrown by client in a server side context, to get the real response developper must write RESTeasy specific code and add a compile dependency to RESTeasy to invoke `WebApplicationExceptionWrapper.unwrap`. To my mind it's a bit restrictive. I mean as a developper I should not have to depend on any implementation to get a basic response.

So this proposal is about:
- making `WebApplicationException#getResponse()` consistent and always return the exact remote side response, regardless of  where the `WebApplicationException` is thrown (server side context or not).
- preserving current `WebApplicationException` thrown by client in a server side context handling.

Let me know what you think.

Thanks

Nicolas

### DIFF
--- a/resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/ResteasyBadRequestException.java
+++ b/resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/ResteasyBadRequestException.java
@@ -14,14 +14,21 @@ public class ResteasyBadRequestException extends BadRequestException implements 
 
    private static final long serialVersionUID = -6250430572164780061L;
    private final BadRequestException wrapped;
+   private final Response sanitizedResponse;
 
     ResteasyBadRequestException(final BadRequestException wrapped) {
-        super(wrapped.getMessage(), sanitize(wrapped.getResponse()), wrapped.getCause());
+        super(wrapped.getMessage(), wrapped.getResponse(), wrapped.getCause());
         this.wrapped = wrapped;
+        this.sanitizedResponse = sanitize(wrapped.getResponse());
     }
 
     @Override
     public BadRequestException unwrap() {
         return wrapped;
+    }
+
+    @Override
+    public Response getSanitizedResponse() {
+        return sanitizedResponse;
     }
 }
--- a/resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/ResteasyClientErrorException.java
+++ b/resteasy-client-api/src/main/java/org/jboss/resteasy/client/exception/ResteasyClientErrorException.java
@@ -1,6 +1,5 @@
 package org.jboss.resteasy.client.exception;
 
-
 import static org.jboss.resteasy.client.exception.WebApplicationExceptionWrapper.sanitize;
 
 import jakarta.ws.rs.ClientErrorException;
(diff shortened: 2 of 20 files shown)


jamezp (approved):

jamezp:
Thank you @NicoNes. I like this approach and it makes sense.

jamezp:
I've been doing some thinking and there is a good chance we need to revert this. It causes a regression in Quarkus and the more I think about it there is still an issue where data could be exposed that shouldn't be. Since there is a way to disable this and use the full response with a setting, I think it's appropriate to revert this until we can think more about this.

NicoNes:
Hi @jamezp,

Ok, I understand your motivation for reverting this.

> more I think about it there is still an issue where data could be exposed that shouldn't be

What are the cases you have in mind ?

jamezp:
@NicoNes The case Quarkus is testing seems valid to me. What they do is throw a redirect exception with the location, which could be somewhere internal. Then the client catches the error and rethrows it with the response. The issue I can see here is a redirect happens, adds some internal headers then some other error occurs. The client re-throws the error with the response that may container internal information.

The Quarkus test is a decent example of this IMO https://github.com/quarkusio/quarkus/blob/c2da5d77b9cd0cb6b2b732eb5f73def2397f4893/extensions/resteasy-classic/rest-client/deployment/src/test/java/io/quarkus/restclient/exception/FrontendService.java#L11. Where a client invokes an endpoint, catches the exception and re-throws with a response. I can understand how that could be a user error, but it seems since we are currently guarding against this, we should continue.

That said, I do understand how this is a pain from the user perspective of having to either turn this off with a vendor specific setting or use a vendor specific API to get the root response. I re-opened the issue because of this and I do think we can still look for a happy medium here.

NicoNes:
@jamezp to my opinion if user intentionnaly catches a **`RuntimeExeption`** such as a `WebApplicationException`, it is because he wants a special treatment for this exception. Else he will let the system handle it by default.
So, in such case user knows what he is doing.
So if user decides to catch the `WebApplicationException` deliberately to propagate the client response why should we prevent him from doing it ?
IMO vendor should only gaurantee that if a client exception is thrown in a server runtime and **is not catched by user** then the client response MUST not leak in the server reponse.

jamezp:
@NicoNes I do agree with you here. I'm not personally opposed to even adding your PR back as-is. However, I think we need to have a cross-project discussion about it to ensure we don't violate the fix for the CVE.

NicoNes:
@jamezp yes let's have this cross project discussion to find a way to handle this :+1: 
Let me know if you want me to answer here https://github.com/quarkusio/quarkus/issues/27807.

Thanks

PR #3041, opened 2022-03-11
[RESTEASY-3071] Completing the future only when the stream is ready after a write

dnagarajan89:
The current implementation where the [future is completed immediately after write(without checking if the outpustream is ready)](https://github.com/resteasy/resteasy/blob/4.7/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L71-L72) is causing incomplete writes in certain use cases where the response size is huge as seen by running this [reproducer](https://github.com/dnagarajan89/resteasy-tomcat/blob/concurreny_issue_alt/src/test/java/com/resteasy/tomcat/TomcatTest.java#L121-L134).  

So reverting back the changes done as part of [this commit](https://github.com/resteasy/resteasy/commit/f87dc2aa3dfe5a25ab2059850b6c38eb30e4231b), which introduced this change. 
Also [adding the fix](https://github.com/dnagarajan89/Resteasy/blob/RESTEASY-3071/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L250-L255) for [these tests](https://github.com/dnagarajan89/resteasy-tomcat/blob/concurreny_issue_alt/src/test/java/com/resteasy/tomcat/TomcatTest.java#L111-L119) that was there before this change. This was originally proposed in this https://github.com/resteasy/resteasy/pull/2954.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -65,10 +65,10 @@ public class HttpServletResponseWrapper implements HttpResponse
       {
          try
          {
+            stream.write(bytes, offset, length);
             // we only are complete if isReady says we're good to write, otherwise
             // we will be complete in the next onWritePossible or onError
             if(sos == null || sos.isReady()) {
-               stream.write(bytes, offset, length);
                future.complete(null);
             }
          } catch (IOException e)
@@ -97,10 +97,10 @@ public class HttpServletResponseWrapper implements HttpResponse
       {
          try
          {
+            stream.flush();
             // we only are complete if isReady says we're good to write, otherwise
             // we will be complete in the next onWritePossible or onError
             if(sos == null || sos.isReady()) {
-               stream.flush();
                future.complete(null);
             }
          } catch (IOException e)
@@ -246,8 +246,11 @@ public class HttpServletResponseWrapper implements HttpResponse
       {
          synchronized (this) {
             if (lastAsyncOperation != null) {
-               // Do not reset the lastAsyncOperation unless the current one is complete
-               if (!lastAsyncOperation.future.isDone()) {
+               lastAsyncOperation.future.complete(null);
+               // the above complete can trigger further writes inline, so the lastAsyncOperation when all
+               // those inline writes gets completed may not be the same as the one
+               // where the future is completed in the previous LOC.
+               if(!lastAsyncOperation.future.isDone()) {
                   return;
                }
                lastAsyncOperation = null;
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -51,13 +51,17 @@ public class HttpServletResponseWrapper implements HttpResponse
       private final byte[] bytes;
       private final int offset;
       private final int length;
+      private final Queue<AsyncOperation> asyncQueue;
+      private boolean dataWritten;
 
-      public WriteOperation(final OutputStream stream, final byte[] bytes, final int offset, final int length)
+      public WriteOperation(final OutputStream stream, final byte[] bytes, final int offset, final int length,
+                            final Queue<AsyncOperation> asyncQueue)
       {
          super(stream);
          this.bytes = bytes;
          this.offset = offset;
          this.length = length;
+         this.asyncQueue = asyncQueue;
       }
 
       @Override
@@ -65,11 +69,22 @@ public class HttpServletResponseWrapper implements HttpResponse
       {
          try
          {
-            stream.write(bytes, offset, length);
+            if(dataWritten) {
+               future.complete(null);
+               return;
+            }
             // we only are complete if isReady says we're good to write, otherwise
             // we will be complete in the next onWritePossible or onError
             if(sos == null || sos.isReady()) {
-               future.complete(null);
+               stream.write(bytes, offset, length);
+               dataWritten = true;
+               if(sos == null || sos.isReady()) {
+                  future.complete(null);
+               } else {
+                  // Queueing this operation to complete the future later
+                  // when the stream is ready.
+                  asyncQueue.add(this);
+               }
             }
          } catch (IOException e)
          {
@@ -97,10 +112,10 @@ public class HttpServletResponseWrapper implements HttpResponse
       {
          try
          {
-            stream.flush();
             // we only are complete if isReady says we're good to write, otherwise
             // we will be complete in the next onWritePossible or onError
             if(sos == null || sos.isReady()) {
+               stream.flush();
                future.complete(null);
             }
          } catch (IOException e)
@@ -201,7 +216,7 @@ public class HttpServletResponseWrapper implements HttpResponse
       @Override
       public CompletionStage<Void> asyncWrite(byte[] bytes, int offset, int length)
       {
-         AsyncOperation op = new WriteOperation(this, bytes, offset, length);
+         AsyncOperation op = new WriteOperation(this, bytes, offset, length, asyncQueue);
          queue(op);
          return op.future;
       }
@@ -245,16 +260,6 @@ public class HttpServletResponseWrapper implements HttpResponse
       private void flushQueue()
       {
          synchronized (this) {
-            if (lastAsyncOperation != null) {
-               lastAsyncOperation.future.complete(null);
-               // the above complete can trigger further writes inline, so the lastAsyncOperation when all
-               // those inline writes gets completed may not be the same as the one
-               // where the future is completed in the previous LOC.
-               if(!lastAsyncOperation.future.isDone()) {
-                  return;
-               }
-               lastAsyncOperation = null;
-            }
             final ServletOutputStream out;
             try {
                out = getServletOutputStream();


dnagarajan89:
@jamezp Please take a look when you have time.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -65,10 +65,10 @@ protected void doWork(ServletOutputStream sos)
       {
          try
          {
+            stream.write(bytes, offset, length);
```
crankydillo, line 68: Most async 3.1 examples write the bytes after the `ServletOutputStream#isReady` is true.  At least that's what I recall.  Can you elaborate on why this move was done and why it works?

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -246,8 +246,11 @@ private void flushQueue()
       {
          synchronized (this) {
             if (lastAsyncOperation != null) {
-               // Do not reset the lastAsyncOperation unless the current one is complete
-               if (!lastAsyncOperation.future.isDone()) {
+               lastAsyncOperation.future.complete(null);
```
crankydillo, line 249: How are we guaranteed that the last bytes have been written at this point?  I thought the point of the `if` check below was to keep things in a state that the last operation would complete successfully.

crankydillo (commented):
I haven't messed with this class much.  Just giving a bit of feedback.  My biggest ask if I were the project owners would be tests that covered servlet containers and the server adapters.

jamezp:
What OS are you testing on? I cannot get the example application to fail or hang. It passes for me every time.
```
Apache Maven 3.8.5 (3599d3414f046de2324203b78ddcf9b5e4388aa0)
Maven home: /home/jperkins/apps/maven
Java version: 11.0.14.1, vendor: Red Hat, Inc., runtime: /usr/lib/jvm/java-11-openjdk-11.0.14.1.1-5.fc35.x86_64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.16.13-200.fc35.x86_64", arch: "amd64", family: "unix"
```

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
jamezp: I agree with this. It seems odd to write before the output stream is ready.

dnagarajan89: This is the implementation which was there till Resteasy v4.7.3.Final(please check [here](https://github.com/resteasy/resteasy/blob/4.7.3.Final/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L67-L72)) until [this commit](https://github.com/resteasy/resteasy/commit/f87dc2aa3dfe5a25ab2059850b6c38eb30e4231b) which went into 4.7.4.Final. So please understand that this is not something new that I am proposing. The only change that I am having in this PR which was not there in the 4.7.3.Final version is [this check here](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L250-L255) without which we will have this [bug](https://github.com/resteasy/resteasy/pull/2954).

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: Again the same [comment](https://github.com/resteasy/resteasy/pull/3041#discussion_r826610248) will apply here too. Please find this implementation in the 4.7.3.Final tag [here](https://github.com/resteasy/resteasy/blob/4.7.3.Final/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L236).

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: My main intention to clarify the above is to convey that these are not my changes and so this is better explained by someone from resteasy team. 

But having said that, I can give an explanation based on my understanding from the debugging experience I have for this issue and [another issue](https://github.com/resteasy/resteasy/pull/2954) we reported earlier on this. 
The doWork for writing the data to an outputstream is only called from [one place which always checks if the outputstream is ready before calling it](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L265-L267). So this logic is not actually writing the bytes before the stream is ready. 
Now coming to why the future completion needs the check whether the output stream is ready, while writing large responses the outputstream is taking sometime to get to a ready state after the write execution. So in the new implementation where [future is completed immediately after write(without checking if the outpustream is ready)](https://github.com/resteasy/resteasy/blob/4.7/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L71-L72), the server completes the response(because of the future completion) before the write is completely done. So having [the check for outputstream being ready](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L101-L105) ensures that the previous write is completely done before the future completes and so we can avoid malformed chunk issues in the response.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: Here my understanding on this, the lastAsyncOperation is assigned from the queue [here](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L265-L267) and work is called on the lastAsyncOperation(when the outputstream is ready). The work always writes the data to the outputstream. But it may not complete the future due to [this condition](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L71-L73). When that happens,  the OutputStream readiness will be notified via this callback [onWritePossible](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L272-L275) which will call this flushQueue method. So this checks if the lastAsyncOperation is not null and if it is not, then it completes the future.

dnagarajan89:
> What OS are you testing on? I cannot get the example application to fail or hang. It passes for me every time.

Hello @jamezp, thats for attending to this PR. I am using a mac os. Please find the details below.
Maven home: /usr/local/apache-maven-3.6.3
Java version: 11.0.11, vendor: AdoptOpenJDK, runtime: /Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
Default locale: en_IN, platform encoding: US-ASCII
OS name: "mac os x", version: "11.5", arch: "x86_64", family: "mac"

Sorry for that, I will try to make that reproducer work on all environments.

dnagarajan89:
> The current implementation where the [future is completed immediately after write(without checking if the outpustream is ready)](https://github.com/resteasy/resteasy/blob/4.7/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L71-L72) is causing incomplete writes in certain use cases where the response size is huge as seen by running this [reproducer](https://github.com/dnagarajan89/resteasy-tomcat/blob/concurreny_issue_alt/src/test/java/com/resteasy/tomcat/TomcatTest.java#L121-L134).
> 
> So reverting back the changes done as part of [this commit](https://github.com/resteasy/resteasy/commit/f87dc2aa3dfe5a25ab2059850b6c38eb30e4231b), which introduced this change. Also [adding the fix](https://github.com/dnagarajan89/Resteasy/blob/RESTEASY-3071/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L250-L255) for [these tests](https://github.com/dnagarajan89/resteasy-tomcat/blob/concurreny_issue_alt/src/test/java/com/resteasy/tomcat/TomcatTest.java#L111-L119) that was there before this change. This was originally proposed in this #2954.

@jamezp please try running `mvn test` on this branch https://github.com/dnagarajan89/resteasy-tomcat/tree/RESTEASY-3071 and let me know if you can reproduce the issue.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
anthochristen: [This](https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/catalina/connector/CoyoteOutputStream.html#isReady()) suggests that a call `onWritePossible` implicitly guarantees that we can safely mark the task as done.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
anthochristen: Having the write after an `isReady` check is fine, is we feel strongly about it. But the key here, is as @dnagarajan89 mentioned, is that a successful `write` call per-se doesn't seem to guarantee completion of write (at-least on the CoyoteOutputStream) and so, for good measure, we would need to check again (with the isReady) post writing.

dnagarajan89:
> > The current implementation where the [future is completed immediately after write(without checking if the outpustream is ready)](https://github.com/resteasy/resteasy/blob/4.7/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L71-L72) is causing incomplete writes in certain use cases where the response size is huge as seen by running this [reproducer](https://github.com/dnagarajan89/resteasy-tomcat/blob/concurreny_issue_alt/src/test/java/com/resteasy/tomcat/TomcatTest.java#L121-L134).
> > So reverting back the changes done as part of [this commit](https://github.com/resteasy/resteasy/commit/f87dc2aa3dfe5a25ab2059850b6c38eb30e4231b), which introduced this change. Also [adding the fix](https://github.com/dnagarajan89/Resteasy/blob/RESTEASY-3071/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L250-L255) for [these tests](https://github.com/dnagarajan89/resteasy-tomcat/blob/concurreny_issue_alt/src/test/java/com/resteasy/tomcat/TomcatTest.java#L111-L119) that was there before this change. This was originally proposed in this #2954.
> 
> @jamezp please try running `mvn test` on this branch https://github.com/dnagarajan89/resteasy-tomcat/tree/RESTEASY-3071 and let me know whether the below issue is reproducible in your machine.
> 
> ```
> Tests in error: 
>   testEchoAsyncReactive(com.resteasy.tomcat.TomcatTest): org.apache.http.MalformedChunkCodingException: Unexpected content at the end of chunk
> ```

You're right. I tried running this in one of our CI's which will use a linux container to run the job and the issue is not reproducible there.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
jamezp: Thank you @dnagarajan89. I see what you're saying now. The write not being complete though seems odd as the write should be synchronous. If it's not we still have a race condition in that the future may not complete if `sos.isReady()` returns `false`. I think we're just getting lucky that it returns `true`.

The main issue I see here is we're trying to use a reactive solution in a non-reactive class. I think we need to look at this in general and possibly re-write some things to improve it.

crankydillo: Thanks @dnagarajan89 and @anthochristen .  This diff alone made this hard to see, but onWritePossible javadoc really clears up the original code quite a bit.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
crankydillo: Thanks for the link.  That really clears up a lot.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: @jamezp this issue is happening during under some extreme concurrent situation. But as you said the write is synchronus and I don't see a problem with that. The issue here is sometimes the CoyoteOutputStream is taking sometime to empty its buffer(which can be checked with [isReady](https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/catalina/connector/CoyoteOutputStream.html#isReady())). So when it happens we should not be completing the future as it causes  this exception "org.apache.http.MalformedChunkCodingException: Unexpected content at the end of chunk" on the client side. Thats why as @anthochristen had mentioned earlier, we should be completing the future only by checking isReady in the outputstream. If that is not returning `true` and if the future doesn't gets completed, as per this [contract](https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/catalina/connector/CoyoteOutputStream.html#isReady()) there will be a callback to onWritePossible which will [call the flushQueue](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L274) which completes the future that is [pending completion in the lastAsyncOperation](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L249).

jamezp: The problem I see is that the `write()` and `isReady()` are implementation specific. If the `CoyoteOutputStream` is taking too long to flush the buffer, then the write is not synchronous or we need to flush synchronously after the write. The fact that `isReady()` returns `true` if you move the `stream.write()` is just luck. The call to `flushQueue()` does not solve this as the `AsyncOperation` has already been removed from the queue so the future associated with it may still never be completed. While there is the `lastAsyncOperation` the value may change if there is more than one op in queue.

One option might be if `isReady()` returns `false` we add a new `AsyncOperation` to the queue which would simply complete the future.

liweinan:
I can pass the test in my local macos environment:

image

dnagarajan89:
> I can pass the test in my local macos environment:
> 
> image

@jamezp and @liweinan I have updated the test to get a more consistent failure for this scenario. I have tested this in a linux environment too and I can see this failing consistently. Please try with the latest change in this branch https://github.com/dnagarajan89/resteasy-tomcat/tree/RESTEASY-3071 and let me know if the problem is reproducible. If its reproducing the problem, please try changing the [resteasy version](https://github.com/dnagarajan89/resteasy-tomcat/blob/RESTEASY-3071/pom.xml#L16) to 4.7.3.Final or below and re-run to see the result.

liweinan:
@dnagarajan89 Thanks for update! I'll try it later today.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: "The fact that `isReady()` returns true if you move the `stream.write(`) is just luck." - This statement is not correct. In the failure case([reproducer](https://github.com/dnagarajan89/resteasy-tomcat/tree/RESTEASY-3071)) the `isReady()` after the `write()` is not returning `true` even after moving the `stream.write()` out. The future completion happens [here](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L248-L249), where the `flushQueue()` is called from `onWritePossible()`.

RE your concern of AsyncOperation#future may never get completed, the `lastAsyncOperation` value cannot change before the future get completed because that is the first thing we are doing in [flushQueue()](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L248-L249) before we reassign the `lastAsyncOperation` from the queue [here](https://github.com/resteasy/resteasy/blob/ea829f87d61ef60aa66b207c471cbea95b70278e/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L265).

liweinan:
I can reproduce the problem now:

image

dnagarajan89:
> I can reproduce the problem now:
> 
> image

@liweinan  this is not the correct branch. Please use this branch - https://github.com/dnagarajan89/resteasy-tomcat/tree/RESTEASY-3071. 
In this branch I have only [one test](https://github.com/dnagarajan89/resteasy-tomcat/blob/RESTEASY-3071/src/test/java/com/resteasy/tomcat/TomcatTest.java#L41-L57) whereas in your logs above, I can see three test runs.

liweinan:
@dnagarajan89 okay!

I rebased to newest branch commit:

image

And I runned the tests several times and the problem can be stable reproduced:

image

dnagarajan89:
> I rebased to newest branch commit:

> @dnagarajan89 okay!
> 
> I rebased to newest branch commit:
> 
> image
> 
> And I runned the tests several times and the problem can be stable reproduced:
> 
> image

Thanks @liweinan. this is exactly the problem that I am trying to report.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: Do you see any other scenario where this lastAsyncOperation value might change where the future is still pending completion.

crankydillo (commented):
If possible, add the test we're discussing to this PR.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
jamezp: Since the value of `lastAsyncOperation` is assigned in a loop and is synchronized on `this` then it could definitely change before `flushQueue()` is invoked again. There's definitely a race where the future may never complete.

If there are say 5 ops in the queue and the first op goes through, but `isReady()` returns `false` so the future is not complete. The loop would continue and reassign the `lastAsyncOperation`. The listener may be waiting for the lock to release on `flushQueue()`. That leaves the first ops future incomplete.

dnagarajan89:
> If possible, add the test we're discussing to this PR.

Thanks for bringing this up @crankydillo. 
@jamezp where do you think this test should go ? I couldn't find a server-adapter or any test module that uses a tomcat server in the current project structure.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
dnagarajan89: @jamezp did you get a change to look into this? Please let us know your concerns if any.

jamezp: Sorry, my week has been getting away from me. I've only had a chance to take a quick look, but I'll get to it this week.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -51,25 +51,40 @@ public void work(ServletOutputStream sos) {
...
             if(sos == null || sos.isReady()) {
                stream.write(bytes, offset, length);
-               future.complete(null);
+               dataWritten = true;
+               if(sos == null || sos.isReady()) {
+                  future.complete(null);
+               } else {
+                  // Queueing this operation to complete the future later
+                  // when the stream is ready.
+                  asyncQueue.add(this);
```
jamezp, line 86: The one problem I see here is that the queue is added to out of its lock.

In a lot of ways I really just want to re-write the whole deferred stream, but I'm not sure if I should or not :)

dnagarajan89: It is effectively locked as we call the work only from [here](https://github.com/resteasy/resteasy/blob/6f123411d72e4fe5b5b38fbc54cbb3b114f528fe/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L271) which is inside a synchronized block, right?

jamezp: Ah yeah, it would be you're right. Sorry about that :)

jamezp: I did file https://github.com/resteasy/resteasy/pull/3048 as an example too. I think it needs a few changes, but it seems to work as well.

jamezp:
Closing in favor of #3048

PR #3044, opened 2022-03-17
[RESTEASY-3097] Add a default exception mapper. Jakarta REST 3.1 sect…

jamezp:
…ion 4.4.

https://issues.redhat.com/browse/RESTEASY-3097

@marekkopecky I made some changes to some tests which made some assumptions about what is logged. These relied on log messages being logged because errors where thrown. However, this does not happen when an `ExceptionMapper` is used. In the `DefaultExceptionMapper` I do log the messages to debug which seems to be enough for the tests. But because of [RESTEASY-2056](https://issues.redhat.com/browse/RESTEASY-2056) and [RESTEASY-2106](https://issues.redhat.com/browse/RESTEASY-2106) I wanted to check if this was acceptable.

I did also remove some assumptions with logs about implementation specifics. The actual log text is implementation specific, but package names seemed odd to look for IMO.

We could log these at a level other than `DEBUG`. I was just concerned with it being too verbose, however maybe that's fine.

### DIFF
--- a/docbook/reference/en/en-US/modules/ExceptionMappers.xml
+++ b/docbook/reference/en/en-US/modules/ExceptionMappers.xml
@@ -60,6 +60,16 @@
       annotated classes, or programmatically through the
       ResteasyProviderFactory class.
    </para>
+      <para>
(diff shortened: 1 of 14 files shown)


liweinan:
The PR looks good to me.

PR #3047, opened 2022-03-25
[RESTEASY-3089] ContainerResponseFilter will not allow null entities to be set

petrberan:
https://issues.redhat.com/browse/RESTEASY-3089

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/interception/jaxrs/ContainerResponseContextImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/interception/jaxrs/ContainerResponseContextImpl.java
@@ -125,6 +125,17 @@ public class ContainerResponseContextImpl implements SuspendableContainerRespons
    public void setEntity(Object entity)
    {
       //if (entity != null) logger.info("*** setEntity(Object) " + entity.toString());
+
+      // In case the entity was null, new response is created since the response is "204 No Content"
+      // and as such the response entity would be null even when set to non-null one
+      if (!jaxrsResponse.hasEntity() && entity != null){
+         Response.ResponseBuilder builder = Response.ok(entity);
+         BuiltResponse response = (BuiltResponse)builder.build();
+
+         jaxrsResponse.setStatus(response.getStatus());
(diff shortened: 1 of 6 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/core/interception/jaxrs/ContainerResponseContextImpl.java
```diff
@@ -125,6 +125,17 @@ public Type getEntityType()
...
+
+      // In case the entity was null, new response is created since the response is "204 No Content"
+      // and as such the response entity would be null even when set to non-null one
+      if (!jaxrsResponse.hasEntity() && entity != null){
+         Response.ResponseBuilder builder = Response.ok(entity);
+         BuiltResponse response = (BuiltResponse)builder.build();
+
+         jaxrsResponse.setStatus(response.getStatus());
+         jaxrsResponse.setReasonPhrase(response.getReasonPhrase());
+      }
```
jamezp, line 137: These aren't quite right. For one they should build a new response. Another issue though is it resets the status which seems wrong. While the 204 may seem weird, the filter should handle. If, for example, this was a 404 then this resets it to 200 and gives an okay response with the entity.

One option might be to check the current status with something like:
```java
if (!jaxrsResponse.hasEntity() && entity != null && jaxrsResponse.getStatus() == Response.Status.NO_CONTENT.getStatusCode()) {
   jaxrsResponse.setStatus(Response.Status.OK.getStatusCode());
   jaxrsResponse.setReasonPhrase(Response.Status.OK.getReasonPhrase());
}
```
That said, I think we should remove this. The `ContainerResponseFilter` should be responsible for setting this IMO.

jamezp (changes requested):
I'm questioning myself on this bug now. On the `ContainerResponseFilter` if you set the status in the filter without these changes, then it passes. I'm going to post back to the forum discussion for more details.

jamezp:
I'm going to close this for now as it's a rather serious compatibility change and it seems reasonable to me that the filter would need to change the status. I'm going to reject the JIRA as well. I appreciate your work on this though @petrberan.

PR #3048, opened 2022-03-30
[RESTEASY-3071] If the ServletOutputStream is not ready after write i…

jamezp:
…s complete, we need to queue an operation to complete the future when it can be.

https://issues.redhat.com/browse/RESTEASY-3071

@dnagarajan89 what do you think of an approach like this as a fix for #304? I ran your reproducer 30 times in a loop and it passed.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -2,8 +2,9 @@ package org.jboss.resteasy.plugins.server.servlet;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.LinkedList;
+import java.util.Comparator;
 import java.util.Map;
+import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -31,11 +32,23 @@ public class HttpServletResponseWrapper implements HttpResponse
 {
    public abstract class AsyncOperation
    {
-      final CompletableFuture<Void> future = new CompletableFuture<>();
+      final CompletableFuture<Void> future;
       final OutputStream stream;
+      private final long id;
+      @Deprecated
       public AsyncOperation(final OutputStream stream)
       {
+         this(stream, new CompletableFuture<>(), (stream instanceof DeferredOutputStream ? ((DeferredOutputStream) stream).getId() : -1));
+      }
+      private AsyncOperation(final DeferredOutputStream stream)
+      {
+         this(stream, new CompletableFuture<>(), stream.getId());
+      }
+      private AsyncOperation(final OutputStream stream, final CompletableFuture<Void> future, final long id)
+      {
          this.stream = stream;
+         this.future = future;
+         this.id = id;
       }
       public void work(ServletOutputStream sos) {
          try(CloseableContext c = ResteasyContext.addCloseableContextDataLevel(contextDataMap)){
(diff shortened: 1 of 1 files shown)


dnagarajan89:
@jamezp this approach looks fine to me, but my concern [here](https://github.com/resteasy/resteasy/pull/3041#discussion_r831207953) is applicable here too. But I don't know that completion of the future in order the data is written will have an impact or not. You know better on this. Just want to bring this to your attention.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -276,10 +321,6 @@ public void onError(Throwable t)
       {
          synchronized (this) {
             asyncListenerCalled = true;
-            if (lastAsyncOperation != null) {
-               lastAsyncOperation.future.completeExceptionally(t);
-               lastAsyncOperation = null;
-            }
```
dnagarajan89, line 282: I see what you are doing here. The only problem I see is if there are two operations in the queue one was successful but the future is not completed(as sos was not ready) and the other failed write with an exception, this logic will fail all the operations in the queue. This scenario will not matter for a chunked async write like in our case. But I am not sure if this will be a problem for SSE.

jamezp:
> @jamezp this approach looks fine to me, but my concern [here](https://github.com/resteasy/resteasy/pull/3041#discussion_r831207953) is applicable here too. But I don't know that completion of the future in order the data is written will have an impact or not. You know better on this. Just want to bring this to your attention.

Yes. I thought about the order issue as well, which I do think could be an issue. This is where I keep coming back to where we may need to re-write it in general. It's just a bummer all the types are public so we have to keep compatibility for these types too.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
jamezp: Yes, that is true. I was wondering what might happen there. I guess this is part of the issue when trying to make non-reactive code reactive :)

jamezp: I've added this back as it does make sense to keep it.

jamezp:
@dnagarajan89 I think I solved the ordering issue by using an id. This seems to be working for me locally, but if you could give it a test to make sure it works for you that would be great.

dnagarajan89:
> @dnagarajan89 I think I solved the ordering issue by using an id. This seems to be working for me locally, but if you could give it a test to make sure it works for you that would be great.

Thanks @jamezp. Will check and get back.

jamezp:
Awesome thanks @dnagarajan89. If this works for then I can do a release as I need a new 4.7 release anyway.

dnagarajan89:
@jamezp I have verified this fix and it works for all my scenarios. If we can get a new patch version of 4.7 with this fix, it would be great. Thanks!.

dnagarajan89:
Also I think the test that reproduces this problem should go somewhere into the resteasy code base.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
```diff
@@ -154,9 +227,11 @@ public boolean suppressExceptionDuringChunkedTransfer() {
       private AsyncOperation lastAsyncOperation;
```
dnagarajan89, line 227: I think this is not assigned at all.

dnagarajan89: And we don't need this block - https://github.com/resteasy/resteasy/blob/9331416b193f326cfb96832f39fb564bbcf963c2/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java#L348-L351

jamezp:
> Also I think the test that reproduces this problem should go somewhere into the resteasy code base.

I'm not sure where the test would live is the problem. I don't object to introducing tests in other containers. However, we'd want a new maybe servlet only testsuite where we could add these tests and containers.

PR #3061, opened 2022-04-06
[DON'T MERGE] Quarkus changes to RestClientBuilderImpl

gsmet:
@jamezp this is to show you the changes we made in Quarkus. I tried to reduce the number of changes to the minimum.

Obviously, this is just some material for our discussions and not a proper PR.

### DIFF
--- a/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
+++ b/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
@@ -1,4 +1,4 @@
-package org.jboss.resteasy.microprofile.client;
+package io.quarkus.restclient.runtime;
 
 import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
@@ -52,8 +52,11 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.security.AccessController;
+import java.security.KeyManagementException;
 import java.security.KeyStore;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
+import java.security.SecureRandom;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -242,49 +253,27 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
-                    selectHttpProxy().ifPresent(
-                            proxyAddress -> builderDelegate.defaultProxy(proxyAddress.getHostString(), proxyAddress.getPort()));
-
-                    resteasyClientBuilder = builderDelegate;
-                }
+                resteasyClientBuilder.defaultProxy(userProxyHost, userProxyPort, userProxyScheme);
+            } else {
+                selectHttpProxy().ifPresent(
+                        proxyAddress -> resteasyClientBuilder.defaultProxy(proxyAddress.getHostString(),
+                                proxyAddress.getPort()));
```
gsmet, line 276: For that change, better have a look at the result as it's far more simple than the original code.
You can see a reference to the issues it fixes here: https://github.com/quarkusio/quarkus/pull/19261 . The problem is mostly that you consider that `http.nonProxyHosts` contains regexps, which is incorrect. Your default config of `localhost|127.*|[::1]` is by miracle valid regexps but it's not really testing things properly.

With this patch we delegate everything to the `ProxySelector` which is doing the right thing.

Whatever we decide for short term I think we need this patch in RESTEasy long term.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -321,6 +310,14 @@ public RestClientBuilder executorService(ExecutorService executor) {
...
             resteasyClientBuilder.keyStore(null, "");
         }
+
+        configureTrustAll(resteasyClientBuilder);
+        // we need to push a disabled SSL context when SSL has been disabled
+        // because otherwise Apache HTTP Client will try to initialize one and will fail
+        if (ImageInfo.inImageRuntimeCode() && !SslContextConfiguration.isSslNativeEnabled()) {
+            resteasyClientBuilder.sslContext(new DisabledSSLContext());
+        }
+
```
gsmet, line 320: This is extremely specific to Quarkus so we would need some sort of hook. Maybe a protected hook method that allows to adjust the builder just before we build it?

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -451,9 +466,6 @@ private boolean isMapperDisabled() {
                 // ignore cast exception
             }
         }
-        if (disabled) {
-            LOGGER.warnf("The default ResponseExceptionMapper has been disabled");
-        }
```
gsmet, line 456: This warning is not really nice as we override the default ResponseExceptionMapper. I think it's a perfectly valid use case and it shouldn't raise a warning.
Worst case, we can silence it with some logging config on our side.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -527,8 +539,6 @@ private void verifyBeanPathParam(Class<?> beanType, Map<String, Object> paramMap
                         : (ResteasyUriBuilder) new ResteasyUriBuilderImpl().uri(classPathAnno.value() + "/" + methodPathAnno.value());
             } else if (classPathAnno != null) {
                 template = (ResteasyUriBuilder) new ResteasyUriBuilderImpl().uri(classPathAnno.value());
-            } else {
-                template = null;
```
gsmet, line 531: That one I don't know if it's something you fixed later or if it's something we changed on purpose.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -577,18 +587,20 @@ public RestClientBuilder property(String name, Object value) {
             // Makes it possible to configure some of the ResteasyClientBuilder delegate properties
             String builderMethodName = name.substring(RESTEASY_PROPERTY_PREFIX.length());
             Method builderMethod = Arrays.stream(ResteasyClientBuilder.class.getMethods())
-                    .filter(m -> builderMethodName.equals(m.getName()) && m.getParameterTypes().length >= 1)
+                    .filter(m -> builderMethodName.equals(m.getName()) && m.getParameterCount() >= 1)
```
gsmet, line 590: This is an optimization we generalized in Quarkus: `getParameterTypes()` is quite costly.

resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/RestClientBuilderImpl.java
```diff
@@ -577,18 +587,20 @@ public RestClientBuilder property(String name, Object value) {
...
                     .findFirst()
                     .orElse(null);
             if (builderMethod == null) {
                 throw new IllegalArgumentException("ResteasyClientBuilder setter method not found: " + builderMethodName);
             }
             Object[] arguments;
-            if (builderMethod.getParameterTypes().length > 1) {
+            if (builderMethod.getParameterCount() > 1) {
```
gsmet, line 597: Same here.

PR #3098, opened 2022-05-09
[RESTEASY-2656]

ronsigal:
This is the initial pull request for the gRPC<->JAX-RS material. 

1. It needs additional functionality in HttpServletRequestImpl and HttpServletResponseImpl and related testing.
2. It needs more documentation.
3. The integrated-tests/GrpcToJaxrsTest test depends on the availability of a jar built from the RESTEasy project https://github.com/resteasy/gRPCtoJAXRS-archetype


grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
```diff
@@ -0,0 +1,979 @@
...
+         if(!dir.exists()){
+            dir.mkdir();
+         }
+      }
+      File file = new File(path + "/" + args[3] + ".proto");
+      file.createNewFile();
+      FileWriter fw = new FileWriter(file.getAbsoluteFile());
+      BufferedWriter bw = new BufferedWriter(fw);
+      bw.write(sb.toString());
+      bw.close();
```
jamezp, line 468: This could be something like:
```java
final Path file = Paths.of(path, args[3] + ".proto");
Files.writeString(file, sb, StandardCharset.UTF_8);
```

It should at least use try-with-resources and close both streams.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavabufTranslatorGenerator.java
```diff
@@ -0,0 +1,556 @@
...
+         }
+      }
+      File file = new File(root.append("/").append(translatorClass).append(".java").toString());
+      if (!file.exists()) {
+         file.createNewFile();
+      }
+      FileWriter fw = new FileWriter(file.getAbsoluteFile());
+      BufferedWriter bw = new BufferedWriter(fw);
+      bw.write(sb.toString());
+      bw.close();
```
jamezp, line 526: Same as the note above about try-with-resources or using the new file API.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/ReaderWriterGenerator.java
```diff
@@ -0,0 +1,268 @@
...
+      File file = new File(path + prefix + "MessageBodyReaderWriter.java");
+      if (file.exists()) {
+         return;
+      }
+      file.createNewFile();
+      FileWriter fw = new FileWriter(file.getAbsoluteFile());
+      BufferedWriter bw = new BufferedWriter(fw);
+      bw.write(sbHeader.toString());
+      bw.write(sbBody.toString());
+      bw.close();
```
jamezp, line 244: Same notes as above.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/AsyncContextImpl.java
```diff
@@ -0,0 +1,101 @@
...
+import jakarta.servlet.AsyncContext;
+import jakarta.servlet.AsyncListener;
+import jakarta.servlet.ServletContext;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+
+public class AsyncContextImpl implements AsyncContext {
+
+   private static ExecutorService executorService = Executors.newCachedThreadPool();
```
jamezp, line 17: IMO we should use the `ContextualExecutors.threadPool()` here so we can get the correct executor.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/AsyncMockServletOutputStream.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
+
+public class AsyncMockServletOutputStream extends MockServletOutputStream {
+
+   private CountDownLatch latch = new CountDownLatch(1);
+
+   public void await() throws InterruptedException {
+      while (true) {
```
jamezp, line 11: Is there a reason for the loop? This would be endless if interrupted.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/HttpServletRequestImpl.java
```diff
@@ -0,0 +1,670 @@
...
+import jakarta.ws.rs.core.MultivaluedMap;
+import jakarta.ws.rs.core.UriInfo;
+
+/*
+ * The http servlet request implementation. This class is not thread safe.
+ *
+ * Much of it is borrowed from io.undertow.servlet.spec.HttpServletRequestImpl,
+ * written by Stuart Douglas.
+ */
+public class HttpServletRequestImpl implements HttpServletRequest {
```
jamezp, line 53: I need to spend some time to grok this, but it seems odd we'd have to fully implement an `HttpServletRequest` without a delegate.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/HttpServletResponseImpl.java
```diff
@@ -0,0 +1,414 @@
...
+import jakarta.ws.rs.core.MultivaluedHashMap;
+import jakarta.ws.rs.core.MultivaluedMap;
+import jakarta.ws.rs.core.Response;
+
+/**
+ * Based on io.undertow.servlet.spec.HttpServletResponseImpl
+ * by Stuart Douglas and Richard Opalka
+ *
+ */
+public class HttpServletResponseImpl implements HttpServletResponse {
```
jamezp, line 34: Same not as a above for the request.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/MockServletInputStream.java
```diff
@@ -0,0 +1,38 @@
...
+   private InputStream is;
+
+   public MockServletInputStream(final InputStream is) {
+      this.is = is;
+   }
+
+   @Override
+   public boolean isFinished() {
+      try {
+         return is.available() > 0;
```
jamezp, line 19: This doesn't look right to me. There is no guarantee or contract that `InputStream.available()` must return a value other than 0 even if there is data available to be read.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/ServletConfigWrapper.java
```diff
@@ -0,0 +1,46 @@
...
+import jakarta.servlet.ServletConfig;
+import jakarta.servlet.ServletContext;
+
+import org.jboss.resteasy.plugins.server.servlet.ResteasyContextParameters;
+
+public class ServletConfigWrapper implements ServletConfig {
+
+   public static final String GRPC_JAXRS = "grpcJaxrs";
+   ServletConfig delegate;
+   String readerWriter;
```
jamezp, line 14: These should be private and final.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/util/DateUtils.java
```diff
@@ -0,0 +1,94 @@
...
+    * <p>
+    * This is the most common date format, which is why we cache it.
+    */
+   private static final ThreadLocal<SimpleDateFormat> RFC1123_PATTERN_FORMAT = new ThreadLocal<SimpleDateFormat>() {
+      @Override
+      protected SimpleDateFormat initialValue() {
+         SimpleDateFormat df = new SimpleDateFormat(RFC1123_PATTERN, LOCALE_US);
+         return df;
+      }
+   };
```
jamezp, line 34: We should really be using the new date utilities (JSR-310). This would likely not require this thread local.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/util/IteratorEnumeration.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.util.Enumeration;
+import java.util.Iterator;
+
+/**
+ * Wrapper to convert an iterator to an enumeration
+ *
+ * @author Stuart Douglas (from undertow)
+ */
+public class IteratorEnumeration<T> implements Enumeration<T> {
```
jamezp, line 11: This shouldn't be required. There is a `Collections.enumeration(Collection)` you could use instead.

testsuite/integration-tests/pom.xml
```diff
@@ -45,6 +45,12 @@
         <container.qualifier.manual.ssl.wrong>jbossas-manual-ssl-wrong</container.qualifier.manual.ssl.wrong>
         <container.qualifier.manual.ssl.wildcard>jbossas-manual-ssl-wildcard</container.qualifier.manual.ssl.wildcard>
         <container.qualifier.manual.ssl.sni>jbossas-manual-ssl-sni</container.qualifier.manual.ssl.sni>
+        
+        <!-- For grpc-bridge test -->
+        <version.jaxrs.example.grpc>0.0.2-SNAPSHOT</version.jaxrs.example.grpc>
+        <version.org.wildfly.core>18.1.0.Final</version.org.wildfly.core>
+        <version.weld.environment>5.0.0.CR2</version.weld.environment>
```
jamezp, line 52: We're already using 5.0.0.Final, we should remove this.

testsuite/integration-tests/pom.xml
```diff
@@ -45,6 +45,12 @@
...
+        <version.wildfly.weld>26.1.0.Final</version.wildfly.weld>
```
jamezp, line 53: This should be removed.

testsuite/integration-tests/pom.xml
```diff
@@ -327,6 +333,12 @@
             <groupId>org.wildfly.core</groupId>
             <artifactId>wildfly-cli</artifactId>
         </dependency>
+        
+        <dependency>
+            <groupId>org.wildfly.core</groupId>
+            <artifactId>wildfly-server</artifactId>
+            <version>${version.org.wildfly.core}</version>
+        </dependency>
```
jamezp, line 341: What do we need this fork?

testsuite/integration-tests/pom.xml
```diff
@@ -351,6 +363,87 @@
...
+                  <artifactId>javax.inject</artifactId>
+               </exclusion>
+               <exclusion>
+                  <groupId>jaxrs.example</groupId>
+                  <artifactId>jaxrs.example</artifactId>
+               </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>jaxrs.example</groupId>
```
jamezp, line 390: We definitely need to change this group id. We have to use something we can claim.

testsuite/integration-tests/pom.xml
```diff
@@ -351,6 +363,87 @@
...
+           <groupId>org.wildfly</groupId>
+           <artifactId>wildfly-weld-common-jakarta</artifactId>
+           <version>${version.wildfly.weld}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.weld.environment</groupId>
+            <artifactId>weld-environment-common</artifactId>
+            <version>${version.weld.environment}</version>
+        </dependency>
```
jamezp, line 441: What is the reason for these?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
```diff
@@ -0,0 +1,805 @@
...
+import jaxrs.example.CC1_proto.GeneralEntityMessage;
+import jaxrs.example.CC1_proto.GeneralReturnMessage;
+import jaxrs.example.CC1_proto.ServletInfo;
+import jaxrs.example.CC1_proto.gCookie;
+import jaxrs.example.CC1_proto.gHeader;
+import jaxrs.example.CC1_proto.gString;
+import jaxrs.example.CC1_proto.org_jboss_resteasy_example___CC2;
+import jaxrs.example.CC1_proto.org_jboss_resteasy_example___CC3;
+import jaxrs.example.CC1_proto.org_jboss_resteasy_example___CC4;
+import jaxrs.example.CC1_proto.org_jboss_resteasy_example___CC5;
```
jamezp, line 49: We cannot use package names like this.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
```diff
@@ -0,0 +1,805 @@
...
+      war.setManifest(new StringAsset("Manifest-Version: 1.0\n"
+//            + "Dependencies: com.google.guava services,org.jboss.resteasy.resteasy-grpc-provider services\n"));
+      + "Dependencies: com.google.guava services\n"));
+      war.merge(ShrinkWrap.createFromZipFile( WebArchive.class, TestUtil.resolveDependency("jaxrs.example:jaxrs.example.grpc:war:0.0.2-SNAPSHOT")));
+      TestUtil.addOtherLibrary(war, "org.jboss.resteasy:grpc-bridge:jar:6.1.0-SNAPSHOT");
+      TestUtil.addOtherLibrary(war, "com.github.javaparser:javaparser-symbol-solver-core:jar:3.24.2");
+      TestUtil.addOtherLibrary(war, "com.github.javaparser:javaparser-core:jar:3.24.2");
+      TestUtil.addOtherLibrary(war, "org.wildfly:wildfly-weld-jakarta:jar:26.1.0.Final");
+      TestUtil.addOtherLibrary(war, "org.wildfly:wildfly-weld-common-jakarta:jar:26.1.0.Final");
+      TestUtil.addOtherLibrary(war, "org.wildfly.core:wildfly-server:jar:18.1.0.Final");
```
jamezp, line 78: Why are these needed? Deployments should NEVER use these and they should definitely not be included in a deployment.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
```diff
@@ -0,0 +1,805 @@
...
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-api:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-context:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-core:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-netty-shaded:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-protobuf:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-protobuf-lite:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.grpc:grpc-stub:1.39.0");
+      TestUtil.addOtherLibrary(war, "io.perfmark:perfmark-api:0.23.0");
+      TestUtil.addOtherLibrary(war, "org.jboss.weld.se:weld-se-core:jar:5.0.0.CR2");
+      TestUtil.addOtherLibrary(war, "org.jboss.weld.environment:weld-environment-common:jar:5.0.0.CR2");
```
jamezp, line 89: These should not be here either.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
```diff
@@ -0,0 +1,805 @@
...
+      TestUtil.addOtherLibrary(war, "com.google.guava:failureaccess:jar:1.0.1");
```
jamezp, line 90: In general none of these versions should be hard-coded. It makes it very difficult to upgrade the component.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
```diff
@@ -0,0 +1,805 @@
...
+      WebArchive archive = (WebArchive) TestUtil.finishContainerPrepare(war, null, (Class<?>[]) null);
+      log.info(archive.toString(true));
+      archive.as(ZipExporter.class).exportTo(new File("/tmp/GrpcToJaxrs.jar"), true);
```
jamezp, line 93: Just a note we need to remember to remove this before it can be merged.

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+   <packaging>jar</packaging>
+   <name>RESTEasy gRPC to JAX-RS bridge</name>
+   <url>http://maven.apache.org</url>
+
+   <properties>
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+      <version.io.grpc>1.39.0</version.io.grpc>
+      <maven.compiler.release>11</maven.compiler.release>
+      <maven.compiler.source>11</maven.compiler.source>
+      <maven.compiler.target>11</maven.compiler.target>
```
jamezp, line 20: These should not be used and inherited from the parent.

jamezp (changes requested):
I've made several comments here. I really think we need to look at using an annotation processor. There are a lot of entry points that look like they need to invoked before this works. Plus there are a lot of assumptions that the user will be using Maven in its standard form.

For the Java source generation too we should consider using something like [jdeparser](https://github.com/jdeparser/jdeparser2) as it has a nice API to generate the source with. This part is not critical, but would be nice. There are a lot of `\n` entries which is okay, but if a user on Windows wants to look at the generated files it might look like one long line.

We should also definitely NOT be using Maven group id's that start with "jaxrs" and especially packages that start with "jaxrs".

I'd also like to understand why we're pulling the things we are from Undertow. That seems very difficult to maintain as the specification and implementations evolve.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
```diff
@@ -268,4 +268,14 @@ public void service(String httpMethod, HttpServletRequest request, HttpServletRe
...
+
+   public ResteasyProviderFactory getProviderFactory()
+   {
+      return this.providerFactory;
+   }
+
+   public void setProviderFactory(ResteasyProviderFactory providerFactory)
+   {
+      this.providerFactory = providerFactory;
+   }
```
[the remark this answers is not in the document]
ronsigal: Gone

resteasy-core/src/main/java/org/jboss/resteasy/core/ResteasyContext.java
```diff
@@ -144,4 +148,27 @@ public static Object searchContextData(Object o)
...
+
+   public static Servlet getServlet(String name)
+   {
+      return servletMap.get(name);
+   }
+
+   public static ServletContext getServletContext(Servlet servlet)
+   {
+      return servletContextMap.get(servlet);
+   }
```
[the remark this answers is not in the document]
ronsigal: Gone

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletDispatcher.java
```diff
@@ -40,13 +44,16 @@ public void init(ServletConfig servletConfig) throws ServletException
       servletContainerDispatcher = new ServletContainerDispatcher(servletConfig);
       ServletBootstrap bootstrap = new ServletBootstrap(servletConfig);
       servletContainerDispatcher.init(servletConfig.getServletContext(), bootstrap, this, this);
+      name = servletConfig.getServletName();
+      ResteasyContext.addServlet(name, this, servletConfig.getServletContext());
```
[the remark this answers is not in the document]
ronsigal: I've moved it to GrpcHttpServletDispatcher, which should be used only by grpc-bridge applications.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletDispatcher.java
```diff
@@ -24,12 +24,16 @@
 public class HttpServletDispatcher extends HttpServlet implements HttpRequestFactory, HttpResponseFactory
 {
    protected ServletContainerDispatcher servletContainerDispatcher;
+   protected String name;
```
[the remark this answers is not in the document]
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+         <artifactId>resteasy-core-spi</artifactId>
+         <version>${project.version}</version>
+         <scope>provided</scope>
+      </dependency>
+
+            <dependency>
+                <groupId>org.jboss.weld.se</groupId>
+                <artifactId>weld-se-core</artifactId>
+                <version>5.0.0.CR2</version>
+            </dependency>
```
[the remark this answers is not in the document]
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+                <groupId>org.jboss.spec.javax.servlet</groupId>
+                <artifactId>jboss-servlet-api_4.0_spec</artifactId>
+                <version>2.0.0.Final</version>
+                <scope>compile</scope>
+            </dependency-->
+    <!--dependency>
+	<groupId>org.jboss.weld.module</groupId>
+	<artifactId>weld-web</artifactId>
+	<version>3.1.4.Final</version>
+</dependency-->
```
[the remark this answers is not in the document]
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+      <!--dependency>
+         <groupId>org.wildfly</groupId>
+         <artifactId>wildfly-weld</artifactId>
+         <version>26.0.0.Final</version>
+      </dependency-->
+        <!--dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-undertow</artifactId>
+             <version>${project.version}</version>
+        </dependency-->
```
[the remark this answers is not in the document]
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+                  <phase>compile</phase>
+                  <goals>
+                     <goal>compile</goal>
+                  </goals>
+                  <configuration>
+                     <release>11</release>
+                  </configuration>
+               </execution>
+            </executions>
+         </plugin>
```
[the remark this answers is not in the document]
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+      </plugins>
+   </build>
+
+   <profiles>
+      <profile>
+         <id>i18n</id>
```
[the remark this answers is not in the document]
ronsigal: I don't remember the history of these i18n profiles. I might have created them ... don't remember. :-P

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
```diff
@@ -0,0 +1,979 @@
...
+                + "   String  id;\n"
+                + "   String  name;\n"
+                + "   byte[]  data;\n"
+                + "   long    reconnectDelay;\n"
+                + "}";
+   }
+
+   public static void main(String[] args) throws IOException
+   {
+      if (args.length != 4 && args.length != 5) {
```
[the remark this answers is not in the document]
ronsigal: > This could throw an NPE.

This surprises me, but I can't find any official statement to the contrary ...

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/GrpcCdiExtension.java
```diff
@@ -0,0 +1,20 @@
...
+   private static BeanManager beanManager;
+
+   @SuppressWarnings("static-access")
+   public void observeBeforeBeanDiscovery(@Observes BeforeBeanDiscovery event, BeanManager beanManager) {
+      this.beanManager = beanManager;
+   }
+
+   public static BeanManager getBeanManager() {
+      return beanManager;
+   }
```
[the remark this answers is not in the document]
ronsigal: I've eliminated GrpcCdiExtension.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/JaxrsImplBaseExtender.java
```diff
@@ -0,0 +1,573 @@
...
+   public JaxrsImplBaseExtender(final String[] args) {
+      servletName = args[1];
+      if (args.length == 4) {
+         inWildFly = Boolean.valueOf(args[3]);
+      }
+      parse(args[0], args[2]);
+   }
+
+   private void parse(String root, String pkg) {
+      File file = new File("./src/main/proto/" + root + ".proto");
```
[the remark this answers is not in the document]
ronsigal: Good point. For now, I'm assuming that maven is being used. I would probably enjoy adding a gradle implementation at somepoint.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
```diff
@@ -0,0 +1,979 @@
...
+   }
+
+   /**
+    * Visit all JAX-RS resource classes discovered in project's src/main/java
+    */
+   private void processClasses(String[] args, StringBuilder sb) throws IOException {
+      Log.setAdapter(new Log.StandardOutStandardErrorAdapter());
+
+      // SourceRoot is a tool that read and writes Java files from packages on a certain root directory.
+      Path path = Path.of(args[0], "/src/main/java/");
```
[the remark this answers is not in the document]
ronsigal: Ok, but I think I'll postpone fixing it for now, since the archetype creates a maven project.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
ronsigal: Ok, I've got

      final Path path = Path.of(args[0], "src", "main", "proto", args[3] + ".proto");
      Files.writeString(path, sb.toString(), StandardCharsets.UTF_8);

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavabufTranslatorGenerator.java
ronsigal: Same

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/ReaderWriterGenerator.java
ronsigal: Same

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/AsyncContextImpl.java
ronsigal: Done

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/AsyncMockServletOutputStream.java
ronsigal: Well, if the loop is interrupted, then it will go back to latch.await() and keep trying ...

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/HttpServletRequestImpl.java
ronsigal: We're coming in through the gRPC runtime, so there's no HttpServletRequest to delegate to. I looked for a way to hook into undertow to create one of its HttpServletRequests, but I didn't find a way to do it.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/HttpServletResponseImpl.java
ronsigal: Same

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/MockServletInputStream.java
ronsigal: A MockServletInputStream is always created as follows:

      ByteArrayInputStream bais = new ByteArrayInputStream(actualParam.toByteArray());
      MockServletInputStream msis = new MockServletInputStream(bais);

and available() in ByteArrayInputStream is

    public synchronized int available() {
        return count - pos;
    }

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/servlet/ServletConfigWrapper.java
ronsigal: Done

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/util/DateUtils.java
ronsigal: ServletConfigWrapper is gone

testsuite/integration-tests/pom.xml
ronsigal: Gone

testsuite/integration-tests/pom.xml
ronsigal: The last three are gone. <version.jaxrs.example.grpc> is for the archetype generated project that is used specifically in the arquillian test GrpcToJaxrsTest, so I think it makes sense to leave it here.

testsuite/integration-tests/pom.xml
ronsigal: I don't remember, but it's gone.

testsuite/integration-tests/pom.xml
ronsigal: Gone

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
ronsigal: They come from the example project and its corresponding generated project. I'll changed them.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
ronsigal: Gone

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
ronsigal: Gone

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
ronsigal: @jamezp, the latter three of these are gone. Right now, GrpcToJaxrsTest fails without perfmark-api. More generally, I need to get the right jars either in

 * the grpc-bridge-runtime module in WildFly, or
 * the generated WAR lib directory

I've been having some trouble with it and might need to consult with you.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/grpc/GrpcToJaxrsTest.java
ronsigal: Good point

grpc-bridge/pom.xml
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+      <dependencies>
+         <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-bom</artifactId>
+            <version>${version.io.grpc}</version>
+            <type>pom</type>
+            <scope>import</scope>
+         </dependency>
+      </dependencies>
+   </dependencyManagement-->
```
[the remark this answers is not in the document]
ronsigal: I've removed the comment brackets. In general, I need to go through the poms and clean them up.

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+                <groupId>org.jboss.weld</groupId>
+                <artifactId>weld-core-impl</artifactId>
+                <version>3.1.7.SP1</version>
+                <exclusions>
+                   <exclusion>
+                      <groupId>org.jboss.weld</groupId>
+                      <artifactId>weld-api</artifactId>
+                   </exclusion>
+                </exclusions>
+           </dependency-->
```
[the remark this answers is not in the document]
ronsigal: Gone

grpc-bridge/pom.xml
```diff
@@ -0,0 +1,292 @@
...
+
+      <dependency>
+         <groupId>org.jboss.logging</groupId>
+         <artifactId>jboss-logging-processor</artifactId>
+      </dependency>
+
+      <dependency>
+         <groupId>junit</groupId>
+         <artifactId>junit</artifactId>
+         <!--scope>compile</scope-->
```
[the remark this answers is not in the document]
ronsigal: There are no tests in the grpc-bridge/grpc-bridge-runtime modules, so I've removed the junit dependency.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/util/IteratorEnumeration.java
ronsigal: I just copied Stuart's code. I'll have to go back and see if there was a reason I copied the code instead of just using the classes from undertow.

testsuite/integration-tests/pom.xml
ronsigal: Yes, I'll do that.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/util/IteratorEnumeration.java
ronsigal: I've stripped out the classes I copied from undertow and call the original classes directly.

ronsigal:
I went another step in adopting the nio file / path API. For example, JavaToProtobufGenerator.writeProtoFile() is now
```
   private static void writeProtoFile(String[] args, StringBuilder sb) throws IOException {
      Path path = Files.createDirectories(Path.of(args[0], "src", "main", "proto"));
      Files.writeString(path.resolve(args[3] + ".proto"), sb.toString(), StandardCharsets.UTF_8);
   }
```

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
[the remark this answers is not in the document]
ronsigal: I've been looking at picocli. It looks nice, but I don't see how it adds much here. The command lines for the various code generating classes are pretty simple, and, in any case, they're not really meant to be used directly. They're all called from maven, which distributes the command line arguments.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletDispatcher.java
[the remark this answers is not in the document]
ronsigal: I've started looking at jdeparser2, and it looks like it would help to simplify the code here. Maybe for version 2.0?

grpc-bridge/pom.xml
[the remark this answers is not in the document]
ronsigal: Removed

jamezp:
Just closing this draft as I think it's out of date and no longer being used.

PR #3099, opened 2022-05-10
ParamConverterProvider args nullcheck

fabrii:
Fixes https://issues.redhat.com/projects/RESTEASY/issues/RESTEASY-3123

### DIFF
--- a/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
+++ b/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java
@@ -86,6 +86,10 @@ public class ProxyInvocationHandler implements InvocationHandler {
         for (Object p : providerInstances) {
             if (p instanceof ParamConverterProvider) {
 
+                if (args == null) {
+                    continue;
+                }
+
                 int index = 0;
                 for (Object arg : args) {
                     // ParamConverter's are not allowed to be passed null values. If we have a null value do not process


jamezp:
While this looks mostly correct, have you seen an issue with this? We'd need to file a JIRA and fix it on other branches if this is an issue rather than poor coding.

fabrii:
I am currently working in a project that uses the Reasteasy client Microprofile implementation.
We currently have two ways of defining client providers: 

- The first one is via annotations. For example: @RegisterProvider(QueryParamConverterProvider.class)

- The second way, is doing a global registration of providers, like specified here https://docs.jboss.org/resteasy/docs/3.15.2.Final/userguide/html/MicroProfile_Rest_Client.html (section 3. Global registration of providers).

If I register a ParamConverterProvider globally, then the methods with no params start failing.

As a workaround, I am using the annotations.

jamezp:
@fabrii Interesting. Do you have a reproducer by chance? If the `args` like this are `null` an NPE should be happening before it even makes it here. Also, if we want to check if he args are `null` we should be checking outside the loop.

fabrii:
Reading the code here: https://github.com/resteasy/resteasy/blob/237f9ac028c1070af325041ff2e34017a4201bc1/resteasy-client-microprofile-base/src/main/java/org/jboss/resteasy/microprofile/client/ProxyInvocationHandler.java#L90

In the invoke method, the args is null checked in the line 83. I checked the null inside the providers loop because we might have another type of provider that does not need args to work. Might be wrong.

I dont currently have a small reproducer for this.

I have an interface declared like this:

```
@Path("/users")
@Consumes(MediaType.APPLICATION_JSON)
@RegisterProvider(ObjectMapperContextResolver.class)
@RegisterProvider(BackendExceptionMapper.class)
public interface UserInterface extends Closeable {
    
    @GET
    @Path("/me")
    public UserDTO getMyInfo() throws Exception;
```

jamezp:
Are you seeing some kind of error or exception? If the `args` is `null` and NPE would be thrown at line 90. I will admit it looks funny because we `null` check the `args` elsewhere, but leave it possibly `null` which looks like a bug, but so far has no produced one.

fabrii:
Yes, I get a NPE here: 

org.jboss.resteasy.microprofile.client.ProxyInvocationHandler.invoke(ProxyInvocationHandler.java:90)

jamezp:
> Yes, I get a NPE here:
> 
> org.jboss.resteasy.microprofile.client.ProxyInvocationHandler.invoke(ProxyInvocationHandler.java:90)
> 
> I found it funny also because there is a `if (arg == null) {` check too just 2 lines after. smile

Ah okay, that makes sense and I would consider it a bug. However, I think we need a different change. I need to look at it closer, but at least the looping the parameter converts should be skipped if the `args` are `null`.

Yeah the `if (arg == null)` is to check that no `null` values are processed in the array (`args`).

jamezp:
Replaced by #3115

PR #2138, opened 2019-08-16
RESTEASY-2326 Allow extending resteasy-vertx classes

bentatham:
To address RESTEASY-2326, I have simply made it possible to override the require methods so we can provide our own implementation of authentication for resteasy-vertx, by allowing injection of our own `RequestDispatcher` and allowing us to override the `authenticate` method of that class.

PR #3101, opened 2022-05-19
RESTEASY-2326 Allow extending resteasy-vertx classes

bentatham:
To address RESTEASY-2326, I have simply made it possible to override the require methods so we can provide our own implementation of authentication for resteasy-vertx, by allowing injection of our own RequestDispatcher and allowing us to override the authenticate method of that class.

(this is a recreation of PR #2138 with target as main branch)

(I submit this PR under the terms of the ASL 2.0 License: http://www.apache.org/licenses/LICENSE-2.0 , and I accept those terms.)


liweinan:
@bentatham Thanks for approving the license! @asoldano @jamezp I have updated our RESTEasy contributor license agreement doc and I guess this can be merged.

jamezp (approved):

PR #3102, opened 2022-05-20
[RESTEASY-3135] Initial change for removing finalizers.

jamezp:
https://issues.redhat.com/browse/RESTEASY-3135

This is not yet complete, but wanted to get some tests running.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ManualClosingApacheHttpClient43Engine.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ManualClosingApacheHttpClient43Engine.java
@@ -23,6 +23,7 @@ import org.jboss.resteasy.client.jaxrs.i18n.Messages;
 import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
 import org.jboss.resteasy.client.jaxrs.internal.ClientResponse;
 import org.jboss.resteasy.client.jaxrs.internal.FinalizedClientResponse;
+import org.jboss.resteasy.core.ResourceCleaner;
 import org.jboss.resteasy.spi.config.ConfigurationFactory;
 import org.jboss.resteasy.util.CaseInsensitiveMap;
 
@@ -39,6 +40,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.management.ManagementFactory;
+import java.lang.ref.Cleaner;
 import java.nio.file.Path;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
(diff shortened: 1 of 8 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ManualClosingApacheHttpClient43Engine.java
```diff
@@ -47,12 +49,36 @@
...
+      private Closer(final AtomicBoolean closed, final CloseableHttpClient client) {
+         this.closed = closed;
+         this.client = client;
+      }
+
+      @Override
+      public void run() {
+         if (closed.compareAndSet(false, true)) {
+            if (client != null) {
+               LogMessages.LOGGER.closingForYou(this.getClass());
```
moghaddam, line 72: Just to make sure it would not be forgotten, following [this discussion](https://github.com/eclipse/microprofile-rest-client/issues/348) we may need to avoid this log message if connection is for a CDI-created MicroProfile rest client (depending on the way it would be clarified by the spec in future versions).

jamezp: IMO that is a bug in the MP REST Client specification. The [`Client`](https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/client/client) API specifically says:

> Client instances must be [properly closed](https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/client/client#close()) before being disposed to avoid leaking resources.

I think the MP REST Client spec needs to clarify itself. There should be a way to do it, but I've not looked at the spec itself in a while.

jamezp: Actually, it looks like resteasy-microprofile handles creating the client, so I would likely say it's a bug there. I apologize for not looking close enough. I filed https://github.com/resteasy/resteasy-microprofile/issues/50

ronsigal:
@jamezp, can I just verify that I'm understanding "cleaning" correctly? It looks like

 1. the CleanupAction has a strong reference to a Message
 2. MultipartInputImpl has a strong reference to CleanupAction
 3. java.lang.ref.Cleaner holds a phantom reference to the CleanupAction

So, when MultipartInputImpl goes away, the Message becomes phantom reachable.

jamezp:
> @jamezp, can I just verify that I'm understanding "cleaning" correctly? It looks like
> 
> 1. the CleanupAction has a strong reference to a Message
> 2. MultipartInputImpl has a strong reference to CleanupAction
> 3. java.lang.ref.Cleaner holds a phantom reference to the CleanupAction
> 
> So, when MultipartInputImpl goes away, the Message becomes phantom reachable.

@ronsigal That is how I'm hoping it works at least. The idea is that we don't keep `MultipartInputImpl` from being GC'd. I'm not 100% sure I got it right which is why I like this second set of eyes on it :)

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ManualClosingApacheHttpClient43Engine.java
```diff
@@ -681,26 +710,20 @@ private RequestConfig getCurrentConfiguration(final ClientInvocation request, fi
...
    }
 
-   public void close()
-   {
-      if (closed)
-         return;
+   @Override
+   public void close() {
+      cleanable.clean();
+   }
```
ronsigal, line 719: I'm not clear about this. cleanable.clean() will set the AtomicBoolean closed to true. Then, if the CleanupAction gets called, closed.compareAndSet(false, true) returns false and client.close() doesn't happen. What am I missing?

PR #3105, opened 2022-05-25
Adapted testsuite for jdk17

tommaso-borgato:
WIP adapting branch currently used to test JBoss EAP

### DIFF
--- a/arquillian/RESTEASY-1630-jetty-resteasy-servlet-initializer/pom.xml
+++ b/arquillian/RESTEASY-1630-jetty-resteasy-servlet-initializer/pom.xml
@@ -72,6 +72,12 @@
       <artifactId>maven-model-builder</artifactId>
       <version>${version.maven.model}</version>
       <scope>test</scope>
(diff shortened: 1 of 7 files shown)


jamezp:
This version of RESTEasy is not in EAP and will not be in EAP.

tommaso-borgato:
I see e.g. `jboss-eap-8.0/modules/system/layers/base/org/jboss/resteasy/resteasy-client/main/resteasy-client-4.7.4.Final-redhat-00001.jar` .... what is the correct version and branch to use for testing?

jamezp:
@tommaso-borgato It will be 6.1.0.Final, currently 6.1.0.Beta1 in WildFly Preview. That version is for Jakarta EE 8, 6.1 is for Jakarta EE 10.

I also do no think we should be adding these arguments. RESTEasy does not require these as they are for WildFly. While we do test on WildFly, I don't see a requirement to provide the JVM arguments required. If RESTEasy itself requires a new argument we need to know.

PR #3106, opened 2022-05-26
[RESTEASY-3138] Remove dependency on commons-io.

jamezp:
https://issues.redhat.com/browse/RESTEASY-3138
https://issues.redhat.com/browse/RESTEASY-3139

### DIFF
--- a/providers/multipart/pom.xml
+++ b/providers/multipart/pom.xml
@@ -14,6 +14,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>resteasy-core-spi</artifactId>
             <version>${project.version}</version>
--- a/providers/multipart/pom.xml
+++ b/providers/multipart/pom.xml
@@ -14,10 +14,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>commons-io</groupId>
-            <artifactId>commons-io</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>resteasy-core-spi</artifactId>
             <version>${project.version}</version>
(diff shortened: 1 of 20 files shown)


Thihup:
Would it be possible to use `InputStream.nullInputStream()` / `OutputStream.nullOutputStream()` in any of these cases?

Thihup:
Would it be possible to use `InputStream.nullInputStream()` / `OutputStream.nullOutputStream()` in any of these cases?

jamezp:
> Would it be possible to use `InputStream.nullInputStream()` / `OutputStream.nullOutputStream()` in any of these cases?

The only `NullInputStream` used was for a test and required an empty array to be initialized. I was hoping to just use `InputStream.nullInputStream()`, but the test would fail.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/EntityOutputStream.java
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * The length of the length of the content written.
+     *
+     * @return the length of the content written
+     *
+     * @throws IOException if there is an error determining the length of the content
+     */
+    long getContentLength() throws IOException {
+        final Path file = this.file;
```
jbliznak, line 198: redundant line?

jamezp: Maybe I should lock, but the idea was to get the current value as its volatile and then use that below.

providers/multipart/pom.xml
```diff
@@ -13,6 +13,10 @@
     <description/>
 
     <dependencies>
+        <dependency>
```
jbliznak, line 16: is the order on first position deliberate?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/ApacheHttpAsyncClient4Engine.java
```diff
@@ -878,4 +877,10 @@ private static IOException ioException(Exception ex) {
       return (ex instanceof IOException) ? (IOException) ex : new IOException(ex);
    }
 
+   private static void safeClose(final Closeable closeable) {
+      if (closeable != null) try {
```
jbliznak, line 881: I know it is just matter of style but using `if` without brackets could be a bit cryptic, especially with try-catch

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/engines/EntityOutputStream.java
jbliznak: I see, then lock seems safer if that matters, you already use plenty of that in this class.

PR #3109, opened 2022-06-07
[RESTEASY-2962] fix and reenable ReverseInjectionTest.testMDB

soul2zimate:
Issue: https://issues.redhat.com/browse/RESTEASY-2962 Fix and reenable ReverseInjectionTest.testMDB.

1. Fix the wrong lookup destinationName.

2. Add a timeout to the CountDownLatch `await()` method, and add a delay to `ReverseInjectionTest.testMDB()` before calling endpoint `mdb/books`.

3. Fail the test with the given message in case of exception.

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/ReverseInjectionTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/ReverseInjectionTest.java
@@ -45,8 +45,8 @@ import org.jboss.resteasy.utils.TestUtil;
 import org.jboss.shrinkwrap.api.Archive;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -75,6 +75,7 @@ import java.util.Collection;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.PropertyPermission;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.LoggingPermission;
 
 import static junit.framework.TestCase.assertEquals;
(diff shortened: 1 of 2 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/ReverseInjectionTest.java
```diff
@@ -304,6 +304,7 @@ public void testMDB() throws Exception {
          message = session.createTextMessage(book2.getName());
          producer.send(message);
          log.info("Message sent to to the JMS Provider: " + book2.getName());
+         TimeUnit.SECONDS.sleep(2);
```
jamezp, line 307: What is the sleep for?

soul2zimate: yeah, this is a hack. even thought the latch count down to zero from MDB, sometime it can't get all inserted entities back from datasource while hitting endpoint `mdb/books`, so I just add a short time to wait.

jamezp: Hmm... ...okay. I'll pull it down and have a look. I'm not a huge fan of having sleeps in tests as it _can_ be a sign of a bug.

soul2zimate: About previous comment:
>  even thought the latch count down to zero from MDB, sometime it can't get all inserted entities back from datasource while hitting endpoint mdb/books

I did some researches on it. It seems that something went wrong in h2 database between `1.3.176` and `1.4.193`. (since 1.4.177)

I tried to fix this test **_without_** that sleep in old Resteasy versions, I don't see any problem in Resteasy `4.0.0.Beta2`, but start to see described problem in Resteasy `4.0.0.Beta3`(unfortunately it was omitted due to bad exception handling for wrong lookup destinationName). 

Git bisect show that problem happens when we updated test server version from `10.1.0.Final` to `12.0.0.Final` via #1476.

More Git bisect on the WFLY server locate the behavior changed since commit [WFLY-8055 Upgrade h2 to 1.4.193 from 1.3.176 ](https://github.com/wildfly/wildfly/commit/c60339eb0ed1d505b0c7411b5062e16954c0c3f3).

I tested a local build of WFLY 26.1.0.Final with h2 version downgraded to `1.3.176`, and executed the test on the Resteasy main branch. It passed without the short sleep.

jamezp:

PR #3112, opened 2022-06-14
[RESTEASY-3146] Add resteasy-undertow-cdi module for CDI support in a…

jamezp:
…n Undertow embedded server.

https://issues.redhat.com/browse/RESTEASY-3146

### DIFF
--- a/server-adapters/resteasy-undertow-cdi/src/main/java/dev/resteasy/embedded/server/UndertowCdiEmbeddedServer.java
+++ b/server-adapters/resteasy-undertow-cdi/src/main/java/dev/resteasy/embedded/server/UndertowCdiEmbeddedServer.java
@@ -1,0 +1,179 @@
+package dev.resteasy.embedded.server;
+
+import java.util.Map;
+
+import jakarta.annotation.Priority;
+import jakarta.servlet.ServletException;
+import jakarta.ws.rs.SeBootstrap.Configuration;
+
+import io.undertow.Undertow;
+import io.undertow.server.handlers.PathHandler;
+import io.undertow.servlet.Servlets;
+import io.undertow.servlet.api.DeploymentInfo;
+import io.undertow.servlet.api.DeploymentManager;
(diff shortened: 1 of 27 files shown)


server-adapters/resteasy-undertow-cdi/src/main/java/dev/resteasy/embedded/server/UndertowCdiEmbeddedServer.java
```diff
@@ -0,0 +1,182 @@
...
+                .setClassLoader(deployment.getApplication().getClass().getClassLoader())
+                .setContextPath(contextPath)
+                .setDeploymentName("RESTEasy-" + contextPath)
+                // Set up the RESTEasy Servlet
+                .addServletContextAttribute(ResteasyDeployment.class.getName(), deployment)
+                .addServlet(resteasyServlet)
+                // Configure Weld attributes and the listener
+                .addServletContextAttribute(Listener.CONTAINER_ATTRIBUTE_NAME, container)
+                .addServletContextAttribute(WeldServletLifecycle.BEAN_MANAGER_ATTRIBUTE_NAME, container.getBeanManager())
+                .addListener(Servlets.listener(Listener.class, new ImmediateInstanceFactory<>(Listener.using(container))));
```
manovotn, line 180: I think the configuration you do here is an overkill, i.e. you shouldn't need to add two attributes and a listener which, in this case, registers the same attribute again.
Basically, registering the `Listener` and using its `Listener.using()` method will automatically add the servlet context attribute, see [this Weld code](https://github.com/weld/core/blob/94595deb7dc8432a38e93a09ffb2556023ad3f3d/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/Listener.java#L90-L93).
So in theory you could remove L178-179 and it should still work the same.

manovotn (commented):
Looks good CDI-wise, just added two small comments :)

PR #3117, opened 2022-06-22
Update to latest vertx 4.2 version

turakamou:


turakamou:
Hi,
Hum I don't understand why the checks failed, the modification is minor.
Or is it because the pull request is targeted to branch 5.0 ?

turakamou:
I've updated the related dependencies.
Now all the test should work

turakamou:
I don't have windows in order to test. So I just tried to blindly update undertow.
I would very much like if somebody could help me in this task of updating to vertx 4.2

jamezp:
The Undertow upgrade seems fine. I'm not sure when I'll have to chance to look at this, but possibly next week I can.

turakamou:
Thanks a lot @jamezp you are a life savior.
I'm completely puzzled by why the the tests are still failing on windows (but now it is another test).
It even seems to make the test suite block.

jamezp:
Ah, this is an Undertow issue. I probably didn't bring the fix over to the 5.0 branch. I'll try to get it there next week. See https://issues.redhat.com/browse/RESTEASY-3136 for the details.

turakamou:
Thanks a lot, I'm not sure I would have been able to find that.
Let me quickly try to add that to my branch to see if it is enough.

turakamou:
Seems it is now ok, I hope this can be merged.
Obviously I'll prepare another branch to upgrade to vertx4.2 in 6.0 and main branch, if it is the process?

jamezp:
I need to file some JIRA's for these upgrades before anything can be merged. There is an upstream first policy too so I'll need to prepare PR's for main and 6.0 as well. It seems a bit odd to do a minor upgrade in a micro release, but if there are no breaking changes it seems okay.

turakamou:
Ok, thanks
Please tell me if you want me to do something else, like preparing the PR's for main and 6.0.

jamezp:
Hi @turakamou. Is there a reason to jump to 4.2 instead of staying on the 4.1 release? Also why not jump to 4.3? Just trying to understand the reason for the upgrade. Especially since you'd like it in a micro release.

turakamou:
Hi @jamezp 
If we look at release dates for both Vertx and Resteasy:
When doing Resteasy 5.0.0.Final, Vertx 4.2.0 was available; so I think it makes sense to use this branch of Vertx for this branch of Resteasy.
For the same reason, I would stay on Vertx 4.2 branch for Resteasy 6.0 branch
However, for Resteasy 6.1, ie main, I would go for Vertx 4.3.0 branch ie Vertx 4.3.1.

And if we were to update older version, I would bump Resteasy 4.7 branch to Vertx 4.1.8

Here are the release dates coming from Maven.
https://mvnrepository.com/artifact/io.vertx/vertx-core
https://mvnrepository.com/artifact/org.jboss.resteasy/resteasy-vertx

Vertx
4.3.0 May, 2022
4.2.0 Oct, 2021
4.1.0 Jun, 2021
4.0.0 Dec, 2020

Resteasy
6.0.0.Final Jan, 2022
5.0.0.Final Nov, 2021
4.7.0.Final Jul, 2021
4.6.0.Final Jan, 2021
4.5.0.Final Feb, 2020

jamezp:
Thank you for the detailed response @turakamou. While normally I wouldn't do a minor upgrade in a macro, I think this makes sense. I'll file some JIRA's.

jamezp:
Thank you for all the work on this @turakamou. I'm replacing this with #3138 as I'd like the JIRA's linked.

PR #3118, opened 2022-06-22
[RESTEASY-3150] Add resteasy-rxjava3

bentatham:
Add a copy of resteasy-rxjava2 that support resteasy-rxjava3 packages. This is 100% just a copy and changing import statements, with the necessary pom additions.


bentatham:
I agree to the Apache License 2.0 terms.

jamezp:
I need to think about whether or not we should be putting these types of projects here or not. For RESTEasy 5+ we did some cleanup with moving various projects and I'm almost wondering if we should do more. I'm not objecting to this BTW, just wondering if we should consider moving it to another project like resteasy-extensions or possibly create some kind of resteasy-rxjava project. Maybe even drop rxjava2 and replace it with rxjava3 or just plain call it rxjava.... ...anyway that is my morning brain dump :)

bentatham:
Thanks -- I'm happy to move it wherever. Ideally, I'd like to backport it to Resteasy 5 as well (for javax packages), but hopefully we can upgrade to jakarta and resteasy 6 soon too, just not sure how fast we can get there. Is the 5.0 branch still used to create releases?

bentatham:
FWIW, I would not recommend removing rxjava2, or collapsing into just rxjava. RxJava themselves differentiates the maven groupIds with rxjava2 and rxjava3, so its consistent with that. Further, there will inevitably be resteasy users who may be stuck on rxjava2 and can't upgrade that right away. If you remove support, they will end up in dependency hell where they can't update resteasy as a result, and that is likely a much bigger risk (security, etc) than not updating rxjava.

jamezp:
@bentatham Here is what I'm thinking currently. I think it makes sense to likely create this as a separate project. I need to loo closer at the code to be 100% confident. The reasoning for a separate project would be to have first a `javax` namespace version, e.g. 1.0.0.Final, and a `jakarta` namespace version, e.g. 2.0.0.Final. The idea of back-porting to RESTEasy 5 means we'd have to do a 5.1.0 release and then maintain it and 5.0 possibly. A separate project also allows for say rxjava4 to be easier and releases at a different cadence.

Does this seem reasonable to you?

bentatham:
Sounds perfectly reasonable to me. Let me know what I can do to help get things setup.  (I assume this new project would still be under resteasy organization in github, and be called something like resteasy-rxjava )

jamezp:
@bentatham I've created a new project and a PR for RxJava 3 support. This currently targets RESTEasy 5.0.4.Final and Java 8. This is just a copy of resteasy-rxjava2 with some code clean up is all. If you wouldn't mind having a look at that would be much appreciated. If it looks okay and you don't think there is anything else we need to add, I'm happy to release a 1.0.0 version and then migrate to 2.0.0 which will target RESTEasy 6.1 and Jakarta REST 3.1.

jamezp:
I suppose the link to the PR would be helpful :) https://github.com/resteasy/resteasy-rxjava/pull/2

jamezp:
@bentatham Just an FYI 1.0.0.Final is released https://github.com/resteasy/resteasy-rxjava/releases/tag/1.0.0.Final.

PR #3127, opened 2022-07-06
[RESTEASY-3155] Concurrency issues in ResteasyViolationException

rmartinc:
Issue: https://issues.redhat.com/browse/RESTEASY-3155

Moving the method `convertViolations` to the constructor to avoid any race condition when calculating the lists (`convertFromString` is already called at constructor level). The lists are used now like if they were final.

I'm sending PRs for main and 3.15 branches. @jamezp If you need this any other branch just let me know.

This PR is for the main branch.
PR for 1.15: https://github.com/resteasy/resteasy/pull/3128

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/api/validation/ResteasyViolationException.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/api/validation/ResteasyViolationException.java
@@ -42,17 +42,17 @@ public abstract class ResteasyViolationException extends ConstraintViolationExce
 
    private volatile Exception exception;
 
-   private volatile List<ResteasyConstraintViolation> propertyViolations;
+   private final List<ResteasyConstraintViolation> propertyViolations = new CopyOnWriteArrayList<>();
 
-   private volatile List<ResteasyConstraintViolation> classViolations;
+   private final List<ResteasyConstraintViolation> classViolations = new CopyOnWriteArrayList<>();
 
-   private volatile List<ResteasyConstraintViolation> parameterViolations;
+   private final List<ResteasyConstraintViolation> parameterViolations = new CopyOnWriteArrayList<>();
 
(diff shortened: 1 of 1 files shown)


jamezp:
I'm not sure this is quite right. The issue from [RESTEASY-2914](https://issues.redhat.com/browse/RESTEASY-2914) could reappear here as we moved back to an `ArrayList`. For main, I might just look at re-writing this as I think there is a safer way to handle this. For the other branches we might need a different solution.

rmartinc:
@jamezp Now all the lists are initialized in the constructor and never modified again. So the `ConcurrentModificationException` is also fixed and using `CopyOnWriteArrayList` is unnecessary. I have tested this intensively with the reproducer in the JIRA and no issues detected.

Am I missing something?

jamezp:
@rmartinc In theory it is, but I see two issues. One is really pretty simple. If they are effectively final, they should be final IMO. The other is these are exposed via getters and are immutable. While users likely don't do anything with these collections, it's possible.

That said, I do think we should do 1 and make them final in the constructor instead of volatile.

jamezp:
> Argh! I wanted to delete the volatile too but I forgot to do it. Now the volatile makes no sense because we are creating all of them at construction time.
> 
> Yep, you are totally right about the getters. I had not thought about that. The developers can get the list and modify them. I suppose is very weird but possible. Then, I'm going to do the following: declare all the lists final (remove the volatile) and create them as `CopyOnWriteArrayList` since the beginning. Do we have a deal?

@rmartinc Yes, that sounds reasonable to me for sure. Thank you!

rmartinc:
@jamezp What do you think about it now? The continues working OK, no concurrent or empty string now.

If you don't like it I will just move the `convertViolations` call to the constructors without doing anything else. :wink:

jamezp (approved):

PR #3139, opened 2022-07-07
[RESTEASY-3161] Use the ASLv2 license.

jamezp:
https://issues.redhat.com/browse/RESTEASY-3161

@NicoNes Hopefully you'll see this notification :) Is this okay with you? We would just like to change the license header from LGPL to ASLv2.


NicoNes:
Hey @jamezp,

Yes  I'm fine with it :+1:

jamezp:
Thank you @NicoNes!

PR #3174, opened 2022-08-11
[RESTEASY-3177] Ensure that all async ClientInvocation's happen async…

jamezp:
…hronously.

https://issues.redhat.com/browse/RESTEASY-3177

### DIFF
--- a/resteasy-client-reactor-netty/src/test/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngineTest.java
+++ b/resteasy-client-reactor-netty/src/test/java/org/jboss/resteasy/client/jaxrs/engines/ReactorNettyClientHttpEngineTest.java
@@ -52,6 +52,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Random;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
@@ -324,7 +325,11 @@ public class ReactorNettyClientHttpEngineTest {
 
                             @Override
                             public void failed(Throwable throwable) {
-                                entity.set(throwable.getClass().getName());
+                                if (throwable instanceof CompletionException) {
+                                    entity.set(throwable.getCause().getClass().getName());
+                                } else {
+                                    entity.set(throwable.getClass().getName());
+                                }
                             }
                         });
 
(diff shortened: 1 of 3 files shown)


jamezp:
@crankydillo Just an FYI. I couldn't work out how to get this working quite right for the `ReactiveHttpEngine`. It does suffer from the same issue though where the processing of the request is not asynchronous.

crankydillo:
Hi @jamezp .  Thanks for the heads up.  So the idea here is that the filters execute 'blocking' IO-like code, which could be non-blocking?  Was there a specific use case that showed problematic performance?  I should probably ask this on the issue..

jamezp:
@crankydillo I was doing some testing on replacing the underlying HTTP client with the Java 11 HttpClient and was failing some TCK tests because it was running in the same thread. We pass currently because we don't default to a `AsyncClientHttpEngine` implementation. Because of that we submit it asynchronously https://github.com/resteasy/resteasy/blob/1f2876b4176fbef546659bfd47d13ee68b3a1673/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java#L820.

I didn't see a specific performance issue and didn't really notice any performance gains by making it asynchronous. However, it seems it should be asynchronous as that would be the assumption for the client I would think.

crankydillo:
Hi @jamezp.  Apologies for the blast from the past, but do you happen to recall the TCK test that failed for you?  This change breaks our regression.  We have some unfortunate dependencies on thread local state that this change breaks.  The only way we can skirt around this is due to you not addressing it in the 'ReactiveHttpEngine' side of things.

Ultimately, it boils down to what one thinks you'll be doing in a `ClientRequestFilter`.  Having said, popping onto another thread pool is really just kicking the can.

Anyhow, I'd really like to understand the failing scenario.  If there's any way you can get me to that without causing you too much pain, I'd love to see it:)

jamezp:
Hi @crankydillo. I'm trying to remember :) A number of the timing tests in the TCK have been changed to be ignored, so it might not even be reproducible anymore. The tests were honestly a bit goofy like kick of a request to get a `Future<Response>` then do `Assert.assertTrue(!fiture.isDone())`. If the request processed too fast it would fail :facepalm: 

Anyway, I have no objection to leave the `ReactiveHttpEngine` the way it is now. It just seemed odd to me when I looked at this in general to process the filters in the main thread and only process rest of the request asynchronously. However, maybe it makes sense to process the filters in the main thread. Again, I'm fine with whatever decision is made.

crankydillo:
This seems very similar to the reason RE came up with async versions JAX-RS `MessageBodyReader`/`MessageBodyWriter`.  The container request filter concept can leverage `SuspendableContainerRequestContext` to 'go async'.  If there's a path for this with `ClientRequestFilter`, I haven't seen it.

Regardless, I'm still thinking on the change and how/if we should adapt to it.

PR #3176, opened 2022-08-11
[RESTEASY-3178] Throw an exception instead of returning null if the c…

jamezp:
…ontent-type is null on a response.

https://issues.redhat.com/browse/RESTEASY-3178


liweinan:
Makes sense to me.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ClientInvocation.java
```diff
@@ -161,7 +162,10 @@ public static <T> T extractResult(GenericType<T> responseType, Response response
          {
             if (response.getMediaType() == null)
             {
-               return null;
+               if (response.getStatus() == Response.Status.NO_CONTENT.getStatusCode()) {
```
jamezp, line 165: Thinking about this, I'm wondering now if we should throw the exception if the status is 200 or 203. Other statuses may rightfully not have a media type.

PR #3177, opened 2022-08-11
Grpc.main.test

ronsigal:
Testing


jamezp:
This won't work on WildFly 26 because it requires Jakarta EE 10.

PR #3178, opened 2022-08-13
[RESTEASY-3071] Use a deque to simplify logic and introduce test.

crankydillo:
This is code we had laying around when we tried to work on this
RESTEASY-3071.  We never submitted because we really wanted it coupled
with tests we used to uncover the original issue, which are included
in this work.

The test is basically the same as what @dnagarajan89 was using in the
original issue.

Apologies for the lateness of this, but hopefully it's easy to understand.

I'm having a local integration test failure with `SsePublisherClientTest` locally,
but it fails with the old code as well.

 [RESTEASY-3071](https://engineering.paypalcorp.com/jira/browse/RESTEASY-3071)

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/HttpServletResponseWrapper.java
@@ -2,10 +2,9 @@ package org.jboss.resteasy.plugins.server.servlet;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.Comparator;
+import java.util.Deque;
+import java.util.LinkedList;
 import java.util.Map;
-import java.util.PriorityQueue;
-import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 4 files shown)


crankydillo:
Hmm, I thought SsePublisherClientTest was failing locally with 4.7 branch code, but that doesn't appear to be the case.  If I can't figure this out shortly, I'll close the PR.  If that happens, I'll likely reopen another with the test code.

crankydillo:
Best guess at this point is adding reactor-module stuff and/or related Maven dependencies was somehow messing up micrometer.  I just removed the reactor 'module' stuff as I don't really need that.  Hopefully, we'll get further with the build.  Even so, I'm going to move this to draft.  We have a lot of internal tests that I can run this code through.  Not sure why I didn't do that first:)  Totally understand if you don't want to review until I finish that.

crankydillo:
@jamezp I'm likely going to set the real change code aside.  For some reason, it's failing in our tests that originally uncovered some problems with that code.  I probably won't have time to figure out the flaws.  Even so, what do you think about me creating a new PR that just has `testsuite/integration-tests-tomcat-embedded`?

jamezp:
I don't really have an objection to having some Tomcat tests at all. However, I don't think we can use the change in the `HttpServletResponseWrapper` as it breaks backwards compatibility.

 Also I'm hoping to treat the 4.7 branch as a "dead" branch. The idea is to maintain 5.0 for Jakarta REST 2.1 and main (6.1) for Jakarta REST 3.1.

crankydillo:
> I don't really have an objection to having some Tomcat tests at all. 

Great, I'll likely close this PR and create new one with just the addition to `testsuite`.

>However, I don't think we can use the change in the `HttpServletResponseWrapper` as it breaks backwards compatibility.

The fundamental change is the use of a dequeue for implementation simplicity.  I was just pruning unnecessary code.  I didn't realize it was meant for overriding.  I totally blew off `protected`:P  All of that could be brought back on top of the dequeue.

On a related note, it's concerning that this dequeue change passed the build here (ignoring windows CI, maybe that is uncovering the issue!), but fails in our testbed.  But such is life.  Hard to have bulletproof tests.  That's one reason I was hoping a code simplification could help.  The dequeue stuff is easier to reason about, but clearly there is a flaw.

Regardless, I'm doubtful I'll come back to it, but you never know!

> 
> Also I'm hoping to treat the 4.7 branch as a "dead" branch. The idea is to maintain 5.0 for Jakarta REST 2.1 and main (6.1) for Jakarta REST 3.1.

Thanks for the heads up.  I'll take this back to my team.

crankydillo:
Closing this.  It is concerning to me that this simpler implementation failed.  I had a colleague look at it and he didn't see a flaw.  It makes me wonder if the test is uncovering some other issue.  Unfortunately, I won't be able to spend more time on it:(

PR #3236, opened 2022-09-01
[RESTEASY-2693] remove use of a thread group from NettyJaxrsServer

rsearls:


jamezp:
@rsearls Can you explain a bit on why the changes for the `EventExecutor`? I'm not a Netty expert, but it does seem to make some sense. The other option _might_ be to add the `eventExecutor` to the two other request handlers. I'll have to read up on this though.

jamezp:
This test seems to keep hanging. It's not hanging in my Windows VM, but I want to do some more testing.

jamezp:
@rsearls Can you pull down the latest main and rebase this again? It passed for me onall my local runs and on CI after I rebased it.

jamezp (approved):

PR #3247, opened 2022-09-09
[RESTEASY-2656] gRPC to Jakarta REST

ronsigal:


ronsigal:
Hey @jamezp,

I've finally started to address your many comments on PR https://github.com/resteasy/resteasy/pull/3098. My responses are over there.

In many cases, the comments address code that has since disappeared. In others, I just made exactly the changes you suggested. In all of these cases, I left a comment and resolved the conversation.

In other cases, I made the suggested change (or didn't) and left a comment for further discussion.

For some other comments, I have to look a little deeper.

-Ron

ronsigal:
At this point, there are a few things still to do.

1. Now that my tests are passing on github, I'm at the point of working on finer improvements. For example,
    A. Cleaning up the poms
    B. Arranging for the dependencies to be stored in the right places, e.g., the grpc-bridge-runtime module or the WEB-INF/lib directory of the generated project. I seem to have some blind spots in getting this stuff to work.
    C. Look into adding more tests

2. See if I can use the undertow utility classes directory instead of copying the code.
3. Extend functionality
    A. Unimplemented methods in HttpServletRequestImpl and HttpServletResponseImpl
    B. Implement SSE

grpc-bridge-runtime/d1
```diff
@@ -0,0 +1,71 @@
+[WARNING]
```
jamezp, line 1: What is this file?

grpc-bridge-runtime/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+         <groupId>org.jboss.resteasy</groupId>
+         <artifactId>resteasy-cdi</artifactId>
+         <version>${project.version}</version>
+         <scope>provided</scope>
+         <exclusions>
+            <exclusion>
+               <groupId>org.jboss.weld</groupId>
+               <artifactId>weld-api</artifactId>
+            </exclusion>
+         </exclusions>
```
jamezp, line 50: Somewhat thinking out loud here. I'm not sure we should be excluding. At this current point for resteasy-cdi to work the `org.jboss.weld:weld-api` is required.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/AsyncContextImpl.java
```diff
@@ -0,0 +1,102 @@
...
+import jakarta.servlet.AsyncContext;
+import jakarta.servlet.AsyncListener;
+import jakarta.servlet.ServletContext;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+
+public class AsyncContextImpl implements AsyncContext {
+
+   private static ExecutorService executorService = ContextualExecutors.threadPool();
```
jamezp, line 18: This shouldn't be static. There are potential leaks of using the wrong executor in multi-tenant environments like WildFly.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/AsyncContextImpl.java
```diff
@@ -0,0 +1,102 @@
...
+import org.jboss.resteasy.concurrent.ContextualExecutors;
+
+import jakarta.servlet.AsyncContext;
+import jakarta.servlet.AsyncListener;
+import jakarta.servlet.ServletContext;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+
+public class AsyncContextImpl implements AsyncContext {
```
jamezp, line 16: Something to consider here is this won't work for Quarkus. I don't think Quarkus uses servlet at all or at least it's not required.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/AsyncContextImpl.java
```diff
@@ -0,0 +1,102 @@
...
+
+   private static ExecutorService executorService = ContextualExecutors.threadPool();
+
+   private ServletRequest servletRequest;
+   private ServletResponse servletResponse;
+   private Set<AsyncListener> listeners = new HashSet<AsyncListener>();
+   long timeout;
```
jamezp, line 23: Should this be private? It should probably at least be volatile.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/GrpcHttpServletDispatcher.java
```diff
@@ -0,0 +1,58 @@
...
+import jakarta.servlet.Servlet;
+import jakarta.servlet.ServletConfig;
+import jakarta.servlet.ServletContext;
+import jakarta.servlet.ServletException;
+
+public class GrpcHttpServletDispatcher extends HttpServlet30Dispatcher {
+
+   private static final long serialVersionUID = -7323100224345687064L;
+   private static final Map<String, Servlet> servletMap = new HashMap<String, Servlet>();
+   private static final Map<Servlet, ServletContext> servletContextMap = new HashMap<Servlet, ServletContext>();
```
jamezp, line 17: We probably need to think about this and how it works in environments where this would be shared across multiple deployments. The servlet names could be the same which could cause some issues.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/HttpServletResponseImpl.java
```diff
@@ -0,0 +1,440 @@
...
+   public static final String GRPC_ASYNC = "grpc-async";
+
+   public enum ResponseState {
+      NONE,
+      STREAM,
+      WRITER
+   }
+
+   private static final String RFC1123_PATTERN = "EEE, dd MMM yyyy HH:mm:ss z";
+   private static final SimpleDateFormat SDF = new SimpleDateFormat(RFC1123_PATTERN, Locale.US);
```
jamezp, line 48: We should use the new JSR 310 formatters so it's thread safe by default.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/HttpServletResponseImpl.java
```diff
@@ -0,0 +1,440 @@
...
+
+   @Override
+   public void setContentType(String type) {
+      if (type == null) {
+         return;
+      }
+      ContentTypeInfo ct = parseContentType(type);
+      contentType = ct.getContentType();
+      boolean useCharset = false;
+      if(ct.getCharset() != null && writer == null && !isCommitted()) {
```
jamezp, line 150: Is the `!isCommitted()` needed here? It always returns `false`.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/HttpServletResponseImpl.java
```diff
@@ -0,0 +1,440 @@
...
+
+   @Override
+   public boolean isCommitted() {
+      // TODO Auto-generated method stub
+      return false;
+   }
+
+   @Override
+   public void reset() {
+      msos.getDelegate().reset();
```
jamezp, line 192: Can this just be `msos.reset()`? It's weird this would have to know it's a delegate.

grpc-bridge-runtime/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+   xsi:schemaLocation="
+      http://java.sun.com/xml/ns/javaee 
+      http://java.sun.com/xml/ns/javaee/beans_1_0.xsd">
+</beans>
```
jamezp, line 6: Minor, but we should use the newer XSD here.

grpc-bridge/ResteasyGrpcProviderFactoryImpl.java
```diff
@@ -0,0 +1,17 @@
+package org.jboss.resteasy.grpc.protobuf;
```
jamezp, line 1: This looks strange. The path and package do not match.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
```diff
@@ -0,0 +1,1032 @@
...
+   }
+
+   /**
+    * Visit all JAX-RS resource classes discovered in project's src/main/java
+    */
+   private void processClasses(String[] args, StringBuilder sb) throws IOException {
+      Log.setAdapter(new Log.StandardOutStandardErrorAdapter());
+
+      // SourceRoot is a tool that read and writes Java files from packages on a certain root directory.
+      Path path = Path.of(args[0], "/src/main/java/");
```
jamezp, line 349: Something to consider is this will only work with a standard maven build with a path like this.

jamezp (commented):
I think overall this looks okay. I didn't read the documentation in detail, but made some comments on the code.

I'd like to understand the test  bit better, but it looks like the PR is out of date :(. It's feels strange to me to have a test depend on two outside project builds. But again I need to understand it better :)

I think before you rebase we should make the main branch 6.3.0.Alpha1-SNAPSHOT so this can be merged at some point.

grpc-bridge-runtime/d1
ronsigal: Ooops. Gone.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/AsyncContextImpl.java
ronsigal: Done

grpc-bridge-runtime/src/main/resources/META-INF/beans.xml
ronsigal: Done

grpc-bridge/ResteasyGrpcProviderFactoryImpl.java
ronsigal: I just moved it out of the source tree instead of deleting it. Gone.

grpc-bridge/src/main/java/org/jboss/resteasy/grpc/protobuf/JavaToProtobufGenerator.java
ronsigal: True. I'll work on it later.

grpc-bridge-runtime/src/main/java/org/jboss/resteasy/grpc/runtime/servlet/HttpServletResponseImpl.java
ronsigal: Done

ronsigal:
This pull request is now superseded by https://github.com/resteasy/resteasy/pull/3302

jamezp:
Closing this PR as I think it's out of date and the branch is no longer being used.

PR #3255, opened 2022-09-12
[RESTEASY-3210] changed parsson version to match yasson

rsearls:


jamezp:
We shouldn't downgrade a dependency here. We need to exclude the transitive dependency where appropriate instead.

PR #3278, opened 2022-09-22
[RESTEASY-3225] Allow the default exception mapper to be disabled. Wh…

jamezp:
…en it's not disabled, log a better exception message as an error message.

Disable the default exception mapper for some tests.

https://issues.redhat.com/browse/RESTEASY-3225


resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
```diff
@@ -103,7 +103,7 @@ protected Response handleApplicationException(HttpRequest request, ApplicationEx
          return jaxrsResponse;
       }
       jaxrsResponse = unwrapException(request, e, logger);
-      if (jaxrsResponse == null) {
+      if (jaxrsResponse == null && providerFactory.isDefaultExceptionManagerEnabled()) {
          jaxrsResponse = providerFactory.getThrowableExceptionMapper().toResponse(e.getCause());
```
gsmet, line 107: Do I misunderstand or are you disabling all the exception mappers if we disable the default one by doing that?

It doesn't seem like the behavior we want. We want to disable the default one but also allow the users to install an exception mapper.

resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
```diff
@@ -325,6 +325,14 @@ public interface LogMessages extends BasicLogger
    @Message(id = BASE + 370, value = "Failed to load: %s")
    void failedToLoad(@Cause Throwable cause, String resourceName);
 
+   @LogMessage(level = Level.ERROR)
+   @Message(id = BASE + 375, value = "Handling an uncaught exception with the default exception mapper.")
+   void defaultExceptionMapper(@Cause Throwable cause);
+
+   @LogMessage(level = Level.ERROR)
+   @Message("Handling an uncaught exception with the default exception mapper on resource %s.")
```
gsmet, line 333: You don't want a message id here?

resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
```diff
@@ -325,6 +325,14 @@ public interface LogMessages extends BasicLogger
    @Message(id = BASE + 370, value = "Failed to load: %s")
    void failedToLoad(@Cause Throwable cause, String resourceName);
 
+   @LogMessage(level = Level.ERROR)
+   @Message(id = BASE + 375, value = "Handling an uncaught exception with the default exception mapper.")
```
gsmet, line 329: I wonder if making this message so long has value. My point is that the important information for the user is that an error occurred. Do you really care about all the details?

Note that I'm not very familiar with RESTEasy/JAX-RS so it might actually have value.

gsmet (commented):
Added a few comments, some minor but one crucial if I understood correctly what you're doing.

resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
jamezp: Maybe it doesn't make much sense actually. Maybe the message really should just be `cause.getLocalizedMessage()` in this case.

resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
jamezp: It's inherited from the `void defaultExceptionMapper(@Cause Throwable cause);`

resteasy-core/src/main/java/org/jboss/resteasy/core/ExceptionHandler.java
jamezp: That's a good catch. It will break if the user disables the default exception mapper, but adds their own `ExceptionMapper<Throwable>`.

jamezp: This actually does work and can be kept like this. I'll add some comments for clarify in the source though.

resteasy-core-spi/src/main/java/org/jboss/resteasy/resteasy_jaxrs/i18n/LogMessages.java
jamezp: I changed it a bit, but ended up keeping some kind of message indicating it came from RESTEasy.

resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/DefaultExceptionMapper.java
```diff
@@ -20,19 +20,24 @@
...
 import org.jboss.logging.Logger;
+import org.jboss.resteasy.resteasy_jaxrs.i18n.LogMessages;
 import org.jboss.resteasy.spi.ApplicationException;
+import org.jboss.resteasy.spi.ResteasyProviderFactory;
 
 /**
  * The default {@link ExceptionMapper} for RESTEasy.
  *
  * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
  */
```
ronsigal, line 38: Shouldn't this have a minimal priority?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionThrowableMapperMapperTest.java
```diff
@@ -0,0 +1,109 @@
...
+                .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    /**
+     * Tests that the default exception mapper does not exist.
+     */
+    @Test
+    public void defaultExceptionMapper() {
+        final ExceptionMapper<?> mapper = providers.getExceptionMapper(RuntimeException.class);
+        Assert.assertNotNull("Expected expected our default exception mapper", mapper);
```
ronsigal, line 69: But you don't expect the default exception mapper. Actually, I think this test is made redundant by the next one.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionThrowableMapperMapperTest.java
```diff
@@ -0,0 +1,109 @@
...
+        Assert.assertTrue(String.format("Expected mapper %s to be instance of %s.", mapper, ThrowableExceptionMapper.class),
+                mapper instanceof ThrowableExceptionMapper);
```
ronsigal, line 71: Wouldn't this be the case even if the DefaultExceptionMapper were enabled, given that ThrowableExceptionMapper isn't builtin?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionMapperTest.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@ServerSetup(DisableDefaultExceptionMapperSetupTask.class)
+public abstract class DisabledDefaultExceptionMapperTest {
+
+    @Inject
+    protected Providers providers;
+    @Inject
+    protected Client client;
```
ronsigal, line 55: For my own education, I'm not sure where to look for the Client producer ...

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionMapperTest.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+
+    /**
+     * Tests that a defined exception mapper is used
+     *
+     * @throws Exception if an exception occurs
+     */
+    @Test
+    public void defaultExceptionMapperNotUsed() throws Exception {
```
ronsigal, line 80: I'm not sure how to think about this test, in the sense that it should pass whether or not the DefaultExceptionMapper is enabled. I guess it's a contribution to testing RESTEasy's default exception mapper behavior, independent of this particular pull request.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/whitelist/WhiteListPolymorphicTypeValidatorManualOverrideTest.java
```diff
@@ -83,7 +83,7 @@ public void testAircraftFailure() throws Exception {
         String response = sendPost(new TestPolymorphicType(new Aircraft()));
         logger.info("response: " + response);
         Assert.assertNotNull(response);
-        Assert.assertTrue(response.contains("Response code: " + HttpResponseCodes.SC_BAD_REQUEST));
+        Assert.assertTrue(response.contains("Response code: " + HttpResponseCodes.SC_INTERNAL_SERVER_ERROR));
```
ronsigal, line 86: This seems kind of debatable. If a request entity fails the whitelist, isn't that a client problem?

resteasy-core/src/main/java/org/jboss/resteasy/core/providerfactory/DefaultExceptionMapper.java
jamezp: It doesn't have a priority because it is never looked up.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionMapperTest.java
jamezp: This was something I had added because there is talk about removing the `@Context` annotation. The bulk of them are created here https://github.com/resteasy/resteasy/blob/f8e03aabbca93fbaeecb21919343aa4484ae7ea1/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ContextProducers.java#L45. The client itself is specifically done here https://github.com/resteasy/resteasy/blob/f8e03aabbca93fbaeecb21919343aa4484ae7ea1/resteasy-cdi/src/main/java/org/jboss/resteasy/cdi/ResteasyCdiExtension.java#L87-L116. The benefit injecting the client is that it's automatically closed for you.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionMapperTest.java
jamezp: That's a good point. I copied it from another test, but it really doesn't make sense here.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionThrowableMapperMapperTest.java
jamezp: You're right. This one can be removed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/DisabledDefaultExceptionThrowableMapperMapperTest.java
jamezp: Yes, it would. I guess I kept it to just ensure we get the correct one.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/providers/jackson2/whitelist/WhiteListPolymorphicTypeValidatorManualOverrideTest.java
jamezp: I don't fully disagree with you. The only reason I did this was because the spec suggests to return 500. While that isn't required, I think it's best to do what the spec suggests there. However, I do agree that is up for debate and really the spec should state a "MUST" or not state it at all IMO.

PR #3296, opened 2022-10-07
Use the request's Host header when setting RE's URI object.

crankydillo:
We want to make a greater effort at capturing the actual hostname used by the client.


jamezp:
These Windows failures are caused by this change.

crankydillo:
@jamezp Something I changed is causing the windows break?  If so, the error doesn't make that obvious.  Any tips on what I should do?  Building on Windows will be problematic for me.

crankydillo:
Ah, I see it now.  I didn't scroll through all the logs.  I should be able to address this.  Out of curiousity, is there some way to get to a more human-friendly test report.  Something that makes it easy to see the errors?

jamezp:
@crankydillo I'll have to look into that. It feels like there should be a way to do that. I'll do some research on that.

crankydillo:
@jamezp What do you think about this PR?  Think it can be merged into the 5 and 6 branches?

jamezp:
Oops, sorry. Let me have a look again. It fell off my radar.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyJaxrsServer.java
```diff
@@ -379,11 +380,14 @@ private ResteasyUriInfo extractUriInfo(final HttpServerRequest req, final String
       if (uri.startsWith(req.scheme() + "://")) {
          uriString = uri;
       } else {
+         String host = req.requestHeaders().get(HttpHeaderNames.HOST);
+         if (host == null || "".equals(host.trim())) {
+            host = "unknown";
```
jamezp, line 385: Should we fallback here to something like:
```java
final InetSocketAddress hostAddress = req.hostAddress();
if (hostAddress != null) {
    host = hostAddress.getHostString();
    port = hostAddress.getPort);
} else {
    host = "unknown";
}
```

crankydillo: I think the main issue with that it may not be correctly represent what the client-used URI.  The reason I'm changing this is because a customer of ours is complaining about an injected UriInfo having invalid info based on `hostAddress#getHostString` (current code).  Is that worse than `unknown`?  Probably not:)  Anyhow, I think there was some other InetSockAddress option that is more likely to return a correct value, but it comes with a performance hit.  But it has been a bit since I thought about this..

Anyhow, I will think on it, but I'm not opposed to you making a call either.   Our customers probably won't be unhappy with your proposal.

jamezp: My argument for including this is the value here is what is used in any injected `UriInfo` for the request. If a user were to inject this they could do something like:
```java
final URI newUri = uriInfo.getRequestUriBuilder().path("append").buld();
```
That could result in something like "http://unknown/rest/api/append" and give some weird errors. Even worse if unknown actually resolves to something.

Also with HTTP/2 there is not `Host` header.

crankydillo: I think that makes sense.  I'm embarrassed I didn't think about 'unknown' resolving to an IP!  I will make this change.  Just a heads up though, this code exists in other server adapters.  There's some duplication amongst these, quite a bit in the tests.  I'm guilty of it:(

crankydillo: @jamezp Apologies for going radio silent.  I will get this in within the next few days.

crankydillo: Hi @jamezp .  I pushed the change you suggested.

server-adapters/resteasy-reactor-netty/src/test/java/org/jboss/resteasy/plugins/server/reactor/netty/UriInfoTest.java
```diff
@@ -41,32 +43,87 @@ public static void end() throws Exception
...
+    private String uriInfoRequestUri(final String uri) throws IOException {
+        final Map<String, String> additionalHeaders = new HashMap<>();
+        additionalHeaders.put("Host", TestPortProvider.getHost() + ":" + TestPortProvider.getPort());
+        return uriInfoRequestUri(uri, additionalHeaders);
+
+    }
+
+    /**
+     *
+     */
```
jamezp, line 111: Nit, but this should probably be removed.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyJaxrsServer.java
```diff
@@ -374,15 +375,20 @@ private ResteasyUriInfo extractUriInfo(final HttpServerRequest req, final String
...
          uriString = uri;
       } else {
          String host = req.requestHeaders().get(HttpHeaderNames.HOST);
          if (host == null || "".equals(host.trim())) {
-            host = "unknown";
+            final InetSocketAddress hostAddress = req.hostAddress();
+            if (hostAddress != null) {
+               host = hostAddress.getHostString() + ":" + hostAddress.getPort();
+            } else {
+               host = "unknown"; // Do we even want this?  Should we assert that this never executes?
```
jamezp, line 390: I almost feel like we should make an assertion here. If we don't have a host, allowing the builder to build a URI seems wrong.

server-adapters/resteasy-reactor-netty/src/main/java/org/jboss/resteasy/plugins/server/reactor/netty/ReactorNettyJaxrsServer.java
```diff
@@ -369,26 +373,41 @@ private SslContext toNettySSLContext(final SSLContext sslContext)
...
+         final String uriString;
+
+         // If we have an absolute URL, don't try to recreate it from the host and request line.
+         if (uri.startsWith(req.scheme() + "://")) {
+            uriString = uri;
+         } else {
+            String host = req.requestHeaders().get(HttpHeaderNames.HOST);
+            if (host == null || "".equals(host.trim())) {
+               final InetSocketAddress hostAddress = req.hostAddress();
+               if (hostAddress == null) {
```
crankydillo, line 390: What do you think @jamezp , this or:
```
assert hostAddress != null : "Host address not accessible from request";
```
Also, a netty4 adapter test failed in CI.  I hope that's not related to my change!

jamezp: My only preference for the `IllegalArgumentException` is just because assertions aren't usually used in production.

> Also, a netty4 adapter test failed in CI. I hope that's not related to my change!
Nah, that test is just flaky for some reason. I've been meaning to look at it, but haven't found the time.

crankydillo: Makes sense to me.  Is there anything else I need to do to get this merged?

jamezp: Sorry, I meant a PR for main as well :)

jamezp:
replaced by #3361

PR #3310, opened 2022-10-13
[RESTEASY-3241] expose jakarta.inject-api in resteasy-undertow-cdi

rsearls:


server-adapters/resteasy-undertow-cdi/pom.xml
```diff
@@ -42,6 +42,12 @@
         <dependency>
             <groupId>jakarta.enterprise</groupId>
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>jakarta.inject</groupId>
+                    <artifactId>jakarta.inject-api</artifactId>
+                </exclusion>
+            </exclusions>
```
jamezp, line 50: Thinking out loud here. I wonder if we should exclude this from the resteasy-dependency-bom and ensure anything that uses CDI imports both dependencies.

jamezp (approved):

PR #3320, opened 2022-10-27
[RESTEASY-3109] close SseEventSinkInterceptor on exception; Test case…

rsearls:
… provided.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/interception/jaxrs/PreMatchContainerRequestContext.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/interception/jaxrs/PreMatchContainerRequestContext.java
@@ -1,5 +1,7 @@
 package org.jboss.resteasy.core.interception.jaxrs;
 
+import org.jboss.logging.Logger;
+import org.jboss.resteasy.core.PostResourceMethodInvoker;
 import org.jboss.resteasy.core.ResteasyContext;
 import org.jboss.resteasy.core.SynchronousDispatcher;
 import org.jboss.resteasy.core.ResteasyContext.CloseableContext;
@@ -8,6 +10,7 @@ import org.jboss.resteasy.specimpl.ResteasyHttpHeaders;
 import org.jboss.resteasy.spi.ApplicationException;
 import org.jboss.resteasy.spi.HttpRequest;
 import org.jboss.resteasy.tracing.RESTEasyTracingLogger;
+import org.jboss.resteasy.core.PostResourceMethodInvokers;
 
 import jakarta.ws.rs.container.ContainerRequestFilter;
 import jakarta.ws.rs.core.Cookie;
@@ -35,6 +38,8 @@ import java.util.function.Supplier;
  */
 public class PreMatchContainerRequestContext implements SuspendableContainerRequestContext
 {
+   private static final Logger LOG = Logger.getLogger(PreMatchContainerRequestContext.class);
+
    protected final HttpRequest httpRequest;
    protected Response response;
    private ContainerRequestFilter[] requestFilters;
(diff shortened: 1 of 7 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/exception/RESTEASY3109Test.java
```diff
@@ -0,0 +1,64 @@
...
+                String s = t.getMessage();
+                String error="";
+                Assert.assertTrue(s.contains("HTTP 500 Internal Server Error"));
+            }, () -> {
+                // do nothing
+                String nothing = "";
+            });
+            source.open();
+        }
+    }
```
jamezp, line 63: This should look something more like:
```java
@Test(timeout = 8000)
public void testException () throws Exception {
    WebTarget target = client.target(generateURL("/sse"));
    final CountDownLatch latch = new CountDownLatch(1);
    try (SseEventSource source = SseEventSource.target(target).build()) {
        source.register(evt -> {
            Assert.fail("Should not have seen any results");
        }, t -> {
            String s = t.getMessage();
            Assert.assertTrue(s.contains("HTTP 500 Internal Server Error"));
        }, latch::countDown);
        source.open();
        Assert.assertTrue(latch.await(5, TimeUnit.SECONDS));
    }
}
```

resteasy-core/src/main/java/org/jboss/resteasy/core/interception/jaxrs/PreMatchContainerRequestContext.java
```diff
@@ -368,4 +370,17 @@ public boolean startedContinuation()
...
+
+   private void cleanupPostResourceMethodInvokers() {
+      PostResourceMethodInvokers postResourceMethodInvokers =
+              ResteasyContext.getContextData(PostResourceMethodInvokers.class);
+      // close PostResourceMethodInvokers and clear array list
+      if (postResourceMethodInvokers != null) {
+         postResourceMethodInvokers.getInvokers().forEach(e -> e.close());
+      }
+      // finally
+      if (postResourceMethodInvokers != null) {
```
jamezp, line 382: These two lines could just be deleted. It might make sense to use a try/catch for each invoker too and log the failures. Otherwise we could end up with invokers in weird states.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/exception/resource/RESTEASY3109ExceptionRequestFilter.java
```diff
@@ -0,0 +1,19 @@
...
+
+import jakarta.ws.rs.ext.Provider;
+import jakarta.annotation.Priority;
+import jakarta.ws.rs.container.ContainerRequestContext;
+import jakarta.ws.rs.container.ContainerRequestFilter;
+import java.io.IOException;
+
+@Provider
+@Priority(Integer.MAX_VALUE)
+//@Priority(2000)
```
jamezp, line 11: This could be removed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/exception/resource/RESTEASY3109ExceptionRequestFilter.java
```diff
@@ -0,0 +1,19 @@
...
+public class RESTEASY3109ExceptionRequestFilter  implements ContainerRequestFilter {
+
+    @Override
+    public void filter(final ContainerRequestContext requestContext) throws IOException {
+        throw new IOException("Thrown on purpose");
+        //return;
```
jamezp, line 17: This could be removed.

jamezp (changes requested):

rsearls:
All changes complete

jamezp (approved):

PR #3338, opened 2022-11-18
Add defaults for media type per RFC-7578

jim-krueger:
Add defaults for media type per RFC-7578 to avoid NPE if not provided when EntityPart is built.


jim-krueger:
Accidentally created the PR from my main branch.  Closing.

PR #3348, opened 2022-12-06
RESTEASY-3259 Proxy Framework fails to Produce Mono<T> at runtime whe…

soft4rchitecture:
…n proxying a java Interface

The Providers loading mechanism used by RestEasy client actually goes through the Template Parameters types of The Invokers and guessed that for MonoRxInvoker to be called, return type from a remote Interface method shoudl be reactivestreams.Publisher and not Mono<T>. This is because `MonoRxInvoker extends PublisherRxInvoker` which is RxInvoker<Publisher>, instead of `MonoRxInvoker extends RxInvoker<Mono<?>>` like is already the case for resteasy-rxjava2 plugin for example. This causes the Providers loading mechanism to load MonoRxInvoker only if the remote interface signature has Publisher as a return type...

### DIFF
--- a/resteasy-reactor/src/main/java/org/jboss/resteasy/reactor/MonoRxInvokerImpl.java
+++ b/resteasy-reactor/src/main/java/org/jboss/resteasy/reactor/MonoRxInvokerImpl.java
@@ -3,146 +3,158 @@ package org.jboss.resteasy.reactor;
 import jakarta.ws.rs.client.Entity;
 import jakarta.ws.rs.core.GenericType;
 import jakarta.ws.rs.core.Response;
-
 import org.jboss.resteasy.client.jaxrs.internal.ClientInvocationBuilder;
 import org.jboss.resteasy.client.jaxrs.internal.PublisherRxInvokerImpl;
+import org.reactivestreams.Publisher;
 import reactor.core.publisher.Mono;
 
+import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 
-public class MonoRxInvokerImpl extends PublisherRxInvokerImpl implements MonoRxInvoker {
+public class MonoRxInvokerImpl implements MonoRxInvoker {
+
+
+    private final MonoPublisherInvoker monoInvoker;
+
+
+    static class MonoPublisherInvoker extends PublisherRxInvokerImpl {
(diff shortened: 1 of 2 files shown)


liweinan:
relative discussion: https://github.com/resteasy/resteasy/discussions/3346

jamezp:
Thank you @soft4rchitecture! We require all contributions to be made under the terms of the ASL 2.0 License:[ http://www.apache.org/licenses/LICENSE-2.0](http://www.apache.org/licenses/LICENSE-2.0) 
Do you agree to these terms?

resteasy-reactor/src/main/java/org/jboss/resteasy/reactor/MonoRxInvokerImpl.java
```diff
@@ -3,146 +3,158 @@
...
+
+
+    static class MonoPublisherInvoker extends PublisherRxInvokerImpl {
+        MonoPublisherInvoker(final ClientInvocationBuilder builder) {
+            super(builder);
+        }
+
+        @Override
+        protected <T> Publisher<T> toPublisher(CompletionStage<T> completable) {
+            return Mono.fromFuture(completable.toCompletableFuture());
```
jamezp, line 27: Probably not critical, but shouldn't `Mono.fromCompletionStage(completable)` work here?

soft4rchitecture: fixed it in next commit

soft4rchitecture:
I agree that my contributions is made under the terms of the ASL 2.0 License:[ http://www.apache.org/licenses/LICENSE-2.0](http://www.apache.org/licenses/LICENSE-2.0)

jamezp:
> I agree that my contributions is made under the terms of the ASL 2.0 License:[ http://www.apache.org/licenses/LICENSE-2.0](http://www.apache.org/licenses/LICENSE-2.0)

Excellent, thank you @soft4rchitecture. I plan on processing pull requests this week. Last week I got pulled into an issue that require immediate attention.

jamezp:
I'm going to merge this as I do think it will be useful for 6.2.2.Final. However, I think the whole thing needs an overhaul I'll handle in 6.3. There is a lot of reactive leaking into the standard client which I think we need to remove.

PR #3350, opened 2022-12-07
[RESTEASY-3256] Add check for known interfaces to ensure they are con…

jamezp:
…figured as CDI beans.

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/interceptors/resource/NameBoundCDIProxiesApplication.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/interceptors/resource/NameBoundCDIProxiesApplication.java
@@ -1,6 +1,5 @@
 package org.jboss.resteasy.test.cdi.interceptors.resource;
 
-import java.util.HashSet;
 import java.util.Set;
 
 import jakarta.ws.rs.ApplicationPath;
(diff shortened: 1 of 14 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/interceptors/resource/NameBoundCDIProxiesApplication.java
```diff
@@ -10,15 +9,8 @@
...
-      Set<Class<?>> set = new HashSet<Class<?>>();
-      set.add(NameBoundCDIProxiesResource.class);
-      return set;
-   }
-
-   public Set<Object> getSingletons() {
-      Set<Object> set = new HashSet<Object>();
-      set.add(new NameBoundCDIProxiesInterceptor());
-      return set;
+      return Set.of(NameBoundCDIProxiesResource.class, NameBoundCDIProxiesInterceptor.class);
```
jamezp, line 14: There is another way to fix this test too by annotating `NameBoundCDIProxiesInterceptor` with `@Vetoed`. Since these had `CDI` in the name though, I opted to treat it as a CDI bean. The reason it requires this change is `getSingletons()` returns what are supposed to be singletons which is not a CDI bean.

liweinan:
looks great!

PR #3352, opened 2022-12-09
[RESTEASY-3260] Add a custom constructor for the YAML provider which …

jamezp:
…only allows specific types.

https://issues.redhat.com/browse/RESTEASY-3260

### DIFF
--- a/providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
+++ b/providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
@@ -4,9 +4,13 @@ package org.jboss.resteasy.plugins.providers;
 import org.jboss.resteasy.plugins.providers.yaml.i18n.LogMessages;
 import org.jboss.resteasy.plugins.providers.yaml.i18n.Messages;
 import org.jboss.resteasy.spi.ReaderException;
+import org.jboss.resteasy.spi.ResteasyConfiguration;
+import org.jboss.resteasy.spi.ResteasyProviderFactory;
 import org.jboss.resteasy.spi.WriterException;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.BaseConstructor;
 import org.yaml.snakeyaml.constructor.CustomClassLoaderConstructor;
+import org.yaml.snakeyaml.nodes.Node;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.Produces;
@@ -19,7 +23,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.net.URLClassLoader;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -38,7 +50,18 @@ import java.util.Set;
 @Produces({"text/yaml", "text/x-yaml", "application/x-yaml"})
 @Deprecated
 public class YamlProvider extends AbstractEntityProvider<Object> {
+   private static final String ALLOWED_LIST = "resteasy.yaml.deserialization.allowed.list.allowIfBaseType";
+   private static final String DISABLE_TYPE_CHECK = "resteasy.yaml.deserialization.disable.type.check";
+   private static final Collection<String> DENY_LIST = Arrays.asList(
+           "javax.script.ScriptEngineManager",
+           URLClassLoader.class.getName(),
+           Object.class.getName()
+   );
+   private final Collection<String> allowedList;
 
+   public YamlProvider() {
+      allowedList = createAllowedList();
+   }
    // MessageBodyReader
 
    public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
@@ -52,9 +75,19 @@ public class YamlProvider extends AbstractEntityProvider<Object> {
       try {
          LogMessages.LOGGER.debugf("Provider : %s,  Method : readFrom", getClass().getName());
          if (isValidInternalType(type)) {
-            return new Yaml().load(entityStream);
+            // Use the old behavior of trusting everything
+            if (isTypeCheckDisabled()) {
+               return new Yaml().loadAs(entityStream, type);
+            }
+            final BaseConstructor constructor;
+            if (genericType instanceof ParameterizedType) {
+               constructor = new TypeSafeConstructor((ParameterizedType) genericType, getClassLoader(type), allowedList);
+            } else {
+               constructor = new TypeSafeConstructor(type, getClassLoader(type), allowedList);
+            }
+            return new Yaml(constructor).loadAs(entityStream, type);
          } else {
-            CustomClassLoaderConstructor customClassLoaderConstructor = new CustomClassLoaderConstructor(type.getClassLoader());
+            CustomClassLoaderConstructor customClassLoaderConstructor = new CustomClassLoaderConstructor(type, getClassLoader(type));
             return new Yaml(customClassLoaderConstructor).loadAs(entityStream, type);
          }
       } catch (Exception e) {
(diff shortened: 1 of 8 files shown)


providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +141,110 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+      } else {
+         value = AccessController.doPrivileged((PrivilegedAction<String>) () -> System.getProperty(ALLOWED_LIST));
+      }
+      return value == null ? Collections.emptyList() : Arrays.asList(value.split(","));
+   }
+
+   private static boolean isTypeCheckDisabled() {
+      final String value;
+      if (System.getSecurityManager() == null) {
+         value = System.getProperty(DISABLE_TYPE_CHECK);
```
ronsigal, line 176: Same comment about System.getProperty()

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +141,110 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+      } else {
+         value = AccessController.doPrivileged((PrivilegedAction<String>) () -> System.getProperty(DISABLE_TYPE_CHECK));
```
ronsigal, line 178: Same

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +141,110 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+            return false;
+         }
+         final String name = type.getName();
+         // Check the allowed types first
+         if (allowedList.contains(name)) {
+            return false;
+         }
+         // Allow all primitives
+         if (type.isPrimitive()) {
+            return true;
```
ronsigal, line 228: If you want to allow all primitives, shouldn't it be "false" here?

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -52,9 +73,19 @@ public Object readFrom(Class<Object> type, Type genericType, Annotation[] annota
       try {
          LogMessages.LOGGER.debugf("Provider : %s,  Method : readFrom", getClass().getName());
          if (isValidInternalType(type)) {
```
ronsigal, line 75: This block, down to 87, apply only to lists, etc. So lists, etc., also have to be on the allowedList (unless isTypeCheckDisabled() is true). Is that what you want? 

Also, isTypeCheckDisabled() applies only for lists, etc. Again, is that what you want?

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -52,9 +73,19 @@ public Object readFrom(Class<Object> type, Type genericType, Annotation[] annota
...
+            final BaseConstructor constructor;
+            if (genericType instanceof ParameterizedType) {
+               constructor = new TypeSafeConstructor((ParameterizedType) genericType, getClassLoader(type), allowedList);
+            } else {
+               constructor = new TypeSafeConstructor(type, getClassLoader(type), allowedList);
+            }
+            return new Yaml(constructor).loadAs(entityStream, type);
          } else {
-            CustomClassLoaderConstructor customClassLoaderConstructor = new CustomClassLoaderConstructor(type.getClassLoader());
+            CustomClassLoaderConstructor customClassLoaderConstructor = new CustomClassLoaderConstructor(type, type.getClassLoader());
```
ronsigal, line 88: Why not a call to getClassLoader() here?

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -52,9 +73,19 @@ public Object readFrom(Class<Object> type, Type genericType, Annotation[] annota
...
+               return new Yaml().loadAs(entityStream, type);
+            }
+            final BaseConstructor constructor;
+            if (genericType instanceof ParameterizedType) {
+               constructor = new TypeSafeConstructor((ParameterizedType) genericType, getClassLoader(type), allowedList);
+            } else {
+               constructor = new TypeSafeConstructor(type, getClassLoader(type), allowedList);
+            }
+            return new Yaml(constructor).loadAs(entityStream, type);
          } else {
```
ronsigal, line 87: If we get here, we're bypassing the use of TypeSafeConstructor.  Why is that?

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +141,110 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+         if (DENY_LIST.contains(name)) {
+            return true;
+         }
+         // We should trust known types
+         if (name.startsWith("java") || name.startsWith("javax")) {
+            return false;
+         }
+         // Finally check the known types
+         boolean denied = true;
+         for (Class<?> allowed : types) {
```
ronsigal, line 240: types contains either the type passed into readFrom() or the parameter types thereof. I don't understand why they have any special status. Shouldn't they have to be primitive or be on allowedList?

ronsigal:
I haven't had much interaction with YAML, so maybe I'm missing some things.

ronsigal:
I think "resteasy.yaml.deserialization.allowed.list.allowIfBaseType" should be mentioned in the Users Guide.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +141,110 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+            result = type.getClassLoader();
+         }
+         return result;
+      });
+   }
+
+   private static Collection<String> createAllowedList() {
+      final String value;
+      if (System.getSecurityManager() == null) {
+         value = System.getProperty(ALLOWED_LIST);
```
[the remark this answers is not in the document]
jamezp: That is a good  point. I should use that.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
jamezp: Yes, nice catch.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
jamezp: I only left it because that's how it 'was", but yeah it likely makes sense to change that too.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
jamezp: Just because that is how it was and I didn't want to change the behavior too much. It also seems like the type itself is not constructed as it uses the actual return type or parameter.

jamezp:
> I think "resteasy.yaml.deserialization.allowed.list.allowIfBaseType" should be mentioned in the Users Guide.

The only reason I left it out is this will only go into 3.15 which is only for JBoss EAP 7.4. The provider was removed in 4.x.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
jamezp: They pass because they start with the package "java.". This does seem to be what we want because the issue only seems to appear in generic types.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
jamezp: They only have special status for convenience. The type parameters are pulled from the method and should be safe. If a user allows a specific type, it's currently not required to be in the allowed list. e.g. `List<Customer>` should be okay without adding `Customer` to the allowedList.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +143,131 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+   }
+
+   private static Collection<String> createAllowedList() {
+      final String value = getProperty(ALLOWED_LIST);
+      return value == null ? Collections.emptyList() : Arrays.asList(value.split(","));
+   }
+
+   private static boolean isTypeCheckDisabled() {
+      final String value = getProperty(DISABLE_TYPE_CHECK);
+      return value != null && (value.isEmpty() || value.equalsIgnoreCase("true"));
```
jimma, line 172: This accepts an empty value `""` as true.  Should it only accept not empty "true" value as true ?

jamezp: We could require it be set to `true` explicitly. However, this allows something like `-Dresteasy.yaml.deserialization.disable.type.check` to be passed.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
ronsigal: Ok, cool

JLLeitschuh:
Sorry, external security researcher here: If this fixed a real security vulnerability, RestEasy should either update the existing CVE to announce that this project is also patched, or issue a separate CVE.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -110,4 +143,131 @@ public void writeTo(Object t, Class<?> type, Type genericType, Annotation[] anno
...
+            return false;
+         }
+         // Denied types are not safe
+         if (DENY_LIST.contains(name)) {
+            return true;
+         }
+         // We should trust known types
+         if (name.startsWith("java") || name.startsWith("javax")) {
+            return false;
+         }
```
JLLeitschuh, line 260: Here's a payload to bypass this check:

```yaml
!!java.rmi.server.UnicastRemoteObject [6666]
```

jamezp:
> Sorry, external security researcher here: If this fixed a real security vulnerability, RestEasy should either update the existing CVE to announce that this project is also patched, or issue a separate CVE.

The current stream of RESTEasy does not provide a YAML provider and therefore is not affected. This 3.15 stream is only used in JBoss EAP 7.4 and will not be released.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
jamezp: This is why deny lists are not great :) However, creating an allow list with every Java type is also not ideal. TBH we should probably deny any `java.rmi` type. I see no reason why that would be a generic type.

JLLeitschuh:
Aren't EAPs supposed to be tested by end users. That's why EAPs are released are they not?

jamezp:
> Aren't EAPs supposed to be tested by end users. That's why EAPs are released are they not?

@JLLeitschuh Yes, sorry that sounded more dismissive than I meant it to :) I'm trying to think of how to fix this for RESTEasy 3.15 correctly. I wanted to avoid an allow only list, but it might be our best option TBH. It is for sure the safest option.

jamezp:
@JLLeitschuh Here is my follow up https://github.com/resteasy/resteasy/pull/3372. I think this should be okay. Thanks for looking at this BTW.

JLLeitschuh:
I'll take a look. Any updates on the CVE for the Snake Yaml vulnerability that was present prior to this PR's attempt to fix it?

jamezp:
> I'll take a look. Any updates on the CVE for the Snake Yaml vulnerability that was present prior to this PR's attempt to fix it?

Not to my knowledge which is one reason we're fixing it here. We also have the benefit of knowing the generic type and since we test the class name itself, I _think_ it should be safe.

JLLeitschuh:
What is the best way to unit test this. I don't think it was safe as it was originally.

jamezp:
There is a unit test in the project itself. I find it easiest to run it in two steps. 

### Step 1
```
mvn clean install -DskipTests
````

### Step 2
```
cd testsuite
mvn clean test -am -pl integration-tests -Dtest=BadActorYamlProviderTest
```

PR #3372, opened 2022-12-19
[RESTEASY-3260] Harden the checks to an allow list only and don't jus…

jamezp:
…t trust java and javax namespaces.

https://issues.redhat.com/browse/RESTEASY-3260

### DIFF
--- a/providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
+++ b/providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
@@ -25,9 +25,10 @@ import java.io.OutputStream;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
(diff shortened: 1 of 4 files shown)


providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -52,10 +53,20 @@
...
+           Byte.class.getName(),
+           Double.class.getName(),
+           Float.class.getName(),
+           Integer.class.getName(),
+           List.class.getName(),
+           Long.class.getName(),
+           Map.class.getName(),
+           Set.class.getName(),
+           Short.class.getName(),
+           String.class.getName()
```
ronsigal, line 69: `Character` ...

jamezp: Ah, yes. That is what I missed. Thank you!

PR #3373, opened 2022-12-19
ResteasyProviderFactoryImpl: try to copy default exception mapping option from parent

stevenschlansker:
Filed issue: https://issues.redhat.com/browse/RESTEASY-3269

ResteasyProviderFactoryImpl: try to copy default exception mapping option from parent

Calling e.g. `WebTarget.path` ends up creating a new web target, which creates a new configuration, which creates a new provider impl, which looks up this option, which scans the classpath, which triggers disk IO and unzip'ing.

On a JFR profile of a load harness, the majority of allocations in a pure HTTP network request ends up being byte buffers loading classpath data to determine the state of this option, over and over again.

Can we try to copy the parent's setting instead?


jamezp:
Closing as the JIRA is now closed [RESTEASY-3269](https://issues.redhat.com/browse/RESTEASY-3269)

PR #3379, opened 2022-12-29
RESTEASY-3271 ResteasyServletInitializer should use correct Applicati…

liweinan:
…on class name

### DIFF
--- a/arquillian/RESTEASY-1056-jetty-bv11/src/test/resources/web.xml
+++ b/arquillian/RESTEASY-1056-jetty-bv11/src/test/resources/web.xml
@@ -7,7 +7,7 @@
         <servlet-name>Resteasy</servlet-name>
         <servlet-class>org.jboss.resteasy.plugins.server.servlet.HttpServletDispatcher</servlet-class>
         <init-param>
-            <param-name>jakarta.ws.rs.Application</param-name>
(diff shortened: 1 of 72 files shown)


liweinan:
This solution seems not work, maybe it's because the `jakarta.ws.rs.core.Application` name has conflict with loading process.

However the naming of `jakarta.ws.rs.Application` seems weird, because this does not exists in `jakarta.ws.rs` namespace. Maybe we can use a different name like `resteasy.Application`.

jamezp:
@liweinan It should be `jakarta.ws.rs.Application` per the spec. I agree it seems odd, but that is what the spec says.

From https://jakarta.ee/specifications/restful-ws/3.1/jakarta-restful-ws-spec-3.1.html#servlet

> If not using the Servlet 3 framework pluggability mechanism (e.g. in a pre-Servlet 3.0 container), the servlet-class or filter-class element of the web.xml descriptor SHOULD name the JAX-RS implementation-supplied servlet or filter class respectively. The Application subclass SHOULD be identified using an init-param with a param-name of `jakarta.ws.rs.Application`.

liweinan:
@jamezp ah okay. I should read the spec more carefully :p

Thanks for clarification!

PR #3396, opened 2023-01-24
[RESTEASY-3280] Do not load the ThreadContext's from a service loader…

jamezp:
… each time the contexts are required.

https://issues.redhat.com/browse/RESTEASY-3280

Upstream #3395

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
@@ -24,7 +24,6 @@ import java.security.PrivilegedAction;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.ServiceLoader;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
```diff
@@ -323,20 +322,13 @@ private static Runnable runnable(final Map<ThreadContext<Object>, Object> contex
...
-        }
         // Load any registered providers
-        final ThreadContexts threadContexts = ResteasyProviderFactory.getInstance().getContextData(ThreadContexts.class);
-        if (threadContexts != null) {
-            for (ThreadContext<Object> context : threadContexts.getThreadContexts()) {
-                contexts.put(context, context.capture());
-            }
+        ThreadContexts threadContexts = ResteasyProviderFactory.getInstance().getContextData(ThreadContexts.class);
+        if (threadContexts == null) {
+            threadContexts = new ThreadContexts();
```
crankydillo, line 328: Hi @jamezp .  Did the customer say this fixed the issue?  For us, we still see it going through the serviceloader for each request that flows through this.   `threadContexts` always appears to be `null`.  What am I missing?

jamezp: @crankydillo No, it didn't end up fixing it, https://issues.redhat.com/browse/RESTEASY-3313. It cut down on it, but didn't fix it. I'm going to have to see if there is a better place to put it.

I'm currently trying to see if it's reasonable to rewrite the entire context and `ResteasyProviderFactory` stuff now as I think there are better ways to handle it.

crankydillo: As always, thanks for the fast-feedback!  Apologies for not searching JIRA.  I should have done that!

jamezp: No worries. I was looking at it yesterday so I had the JIRA up already in a tab :)

PR #3398, opened 2023-01-25
[RESTEASY-3272] moved class to resolve duplicate pkg name between mod…

rsearls:
…ules

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/ResteasyThreadContext.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/ResteasyThreadContext.java
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/ResteasyThreadContext.java
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-package org.jboss.resteasy.concurrent;
+package org.jboss.resteasy;
```
jamezp, line 20: I think we should move it to `org.jboss.resteasy.core.concurrent` to ensure we don't split packages elsewhere.

jamezp (changes requested):

PR #3409, opened 2023-02-02
[RESTEASY-3286] Use the new NIO file utilities to create temporary files

jamezp:
https://issues.redhat.com/browse/RESTEASY-3286


AndreasIgelCC:
@jamezp: for which version of resteasy will that be solved (do not have any access to resteasy-Jira-project) and when will that be released? So will there be a fix for 4.7.X too? I am asking because there is a vulnerability linking to this commit here: https://nvd.nist.gov/vuln/detail/CVE-2023-0482

jamezp:
@AndreasIgelCC For the following versions:
* 4.7.8.Final
* 5.0.6.Final
* 6.2.3.Final

AndreasIgelCC:
Thank you for your response, that helps a lot!!

PR #3418, opened 2023-02-09
JettyClientEngine: elide buffering

stevenschlansker:
As originally implemented, the JettyClientEngine could not extract responses without buffering the entire response body, since the MessageBodyReader is sync only. Now, RESTEasy has an asynchronous Executor available to the engine.

So, instead of buffering the full response, let's push the body reader to the async executor pool and feed it content as it becomes available. This should play well with the upcoming virtual threads support in the JVM. This also allows us to implement proper flow control: instead of buffering the whole response, we tell Jetty when we have consumed its buffers, so it will limit the amount of in-flight data.

As an added bonus, the request content send is restructured so the content is also streamed out without buffering the entire request body in a ByteArrayOutputStream.

In the best case, this should remove an entirely unnecessary byte[] copy of both request and response data.
I have completed some light testing and will continue to run the code over the coming days to make sure there's no obvious problems.


jamezp:
Just an FYI, this is not being ignored. I've just been away at meetings all week so I likely won't have time to look at this until next week.

stevenschlansker:
No rush! We haven't observed any problems with the new code yet, but will continue testing.

stevenschlansker:
@jamezp anything I can do to help this get reviewed? :)

jamezp:
@stevenschlansker Sorry, the last two weeks got away from me. I'm out of the office tomorrow, but I'll try to have a look on Monday.

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
```diff
@@ -112,112 +97,75 @@ public <T> CompletableFuture<T> submit(ClientInvocation request, boolean buffere
...
         if (invocation.getEntity() != null) {
-            final ByteArrayOutputStream os = new ByteArrayOutputStream();
-            try {
-                invocation.writeRequestBody(os);
-            } catch (IOException e) {
-                future.completeExceptionally(e);
-                if (callback != null) {
-                    callback.failed(e);
+            final OutputStreamRequestContent contentOut = new OutputStreamRequestContent(
+                    invocation.getHeaders().getMediaType().toString());
```
jamezp, line 118: The `getMediaType().toString()` has potential to throw an NPE. We should check the `MediaType` is not `null` before invoking `.toString()`.

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
```diff
@@ -112,112 +97,75 @@ public <T> CompletableFuture<T> submit(ClientInvocation request, boolean buffere
...
+                try {
+                    invocation.writeRequestBody(contentOut.getOutputStream());
+                } catch (Exception e) {
+                    future.completeExceptionally(e);
+                    if (callback != null) {
+                        callback.failed(e);
+                    }
+                } finally {
+                    contentOut.close();
                 }
```
jamezp, line 129: We may as well use try-with-resources here.

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
```diff
@@ -112,112 +97,75 @@ public <T> CompletableFuture<T> submit(ClientInvocation request, boolean buffere
...
-                copy.limit(buf.remaining());
-                copy.put(buf);
-                copy.flip();
-                stream.offer(copy, new ReleaseCallback(bufs, copy));
-            }
-
-            @Override
-            public void onSuccess(Response response) {
-                if (buffered) {
+                asyncExecutor.submit(() -> {
```
jamezp, line 146: Should this be done on the `onHeader()`? It seems odd to me. `onComplete()` or `onSuccess()` seems to make more sense to me. However, I'm not a Jetty expert.

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
```diff
@@ -112,112 +97,75 @@ public <T> CompletableFuture<T> submit(ClientInvocation request, boolean buffere
...
                     try {
-                        complete();
+                        if (buffered) {
+                            cr.bufferEntity();
+                        }
+                        complete(extractor == null ? (T) cr : extractor.extractResult(cr));
                     } catch (Exception e) {
+                        try {
+                            inputStream.close();
```
jamezp, line 154: It looks like the stream is only ever closed if there is an error. Shouldn't it also be closed on complete? Or at least after it's buffered.

jamezp (changes requested):

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
stevenschlansker: I think neither complete nor success handlers would work - both of these callbacks are fired only _after_ the response body is processed. (It's in the javadoc.)

We might be able to defer this work to an onContent callback, but since that will be called many times, it requires a bit more work to keep track of state.

Do note that onHeader*s* is called after all headers are received and parsed - it is not called for each header - so the very next thing coming is body content.

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
stevenschlansker: I think it does not matter much - all the close is doing is releasing any buffers it still has. On success, there will be no buffers left. But you're right, let's be defensive in case the impl changes later.

stevenschlansker:
Ok thanks for the review @jamezp , I have made the suggested changes. The try-with-resources is a bit odd style since I wanted to make sure to catch any exception thrown by `close` (unlikely). I think the onHeaders callback should work as is, let me know if you have further ideas for how that could be better.

I'll build this and start testing the new version internally. I doubt it'll be any different, but you never know...

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
stevenschlansker: Oh, I see why this is this way. It is entirely allowable to call `readEntity(InputStream.class)` or otherwise process the data in a streaming manner - so we cannot close the input stream here, otherwise the client cannot access the streamed data in the ClientResponse.

stevenschlansker: I added a close after buffering the entity, since in that case we won't need the stream. But otherwise I think we have to leave it open here.

stevenschlansker: And, in the other case - where the engine is not buffering - the close is taken care of by JettyClientResponse->BuiltResponse.releaseConnection-> `is.close()`.

stevenschlansker:
I don't understand the Javadoc failure, probably not related to this change?

```
 Error:  Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.8.2:site (default-site) on project resteasy-jaxrs-all: failed to get report for org.apache.maven.plugins:maven-jxr-plugin: Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M3:enforce (ban-bad-dependencies) on project resteasy-client-reactor-netty: Execution ban-bad-dependencies of goal org.apache.maven.plugins:maven-enforcer-plugin:3.0.0-M3:enforce failed: org.apache.maven.shared.dependency.graph.DependencyGraphBuilderException: Could not resolve following dependencies: [org.jboss.resteasy:resteasy-jackson2-provider:jar:6.3.0.Alpha1-SNAPSHOT (test)]: Could not resolve dependencies for project org.jboss.resteasy:resteasy-client-reactor-netty:jar:6.3.0.Alpha1-SNAPSHOT: Could not find artifact org.jboss.resteasy:resteasy-jackson2-provider:jar:6.3.0.Alpha1-SNAPSHOT in jboss-public-repository-group (https://repository.jboss.org/nexus/content/groups/public/) -> [Help 1]
```

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
stevenschlansker: Oh, closing it after buffering is redundant too. the `bufferEntity` implementation already calls `releaseConnection` for us.

jamezp: Ah, right. Sorry, you're right.

jamezp:
The JavaDoc issue isn't related to this PR. It's also an issue on main. The other two timeouts might be. I'm looking into that part now.

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
jamezp: This seems okay for now. What's not clear is if there could be any potential race issues with `onHeader()` if the content itself is not yet available. However, so far that doesn't seem to be an issue so I might be overthinking it.

jamezp:
@stevenschlansker One last request, If you could squash the commits and prefix the commit message with `[RESTEASY-3297]`, I think this is ready to merge. If you don't have time to deal with that I can absolutely do it.

https://issues.redhat.com/browse/RESTEASY-3297

resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
stevenschlansker: All it is doing is dispatching the deserializer work to a thread here. It will attempt to read from the content input stream, which the InputStreamResponseListener takes care of synchronizing. So the first read will block until content is available.

stevenschlansker:
thanks @jamezp , I reworded and squished the commits.

jamezp (approved):

stevenschlansker:
Awesome, thank you!

PR #3428, opened 2023-02-17
[RESTEASY-3290] Null check dir on temporary file creation

nikosmoum:


jamezp:
Thank you so much for this fix @nikosmoum. I've created a new PR because I noticed something else that needed a change too. This will be replaced by #3435

PR #3431, opened 2023-02-22
Revert - [RESTEASY-2845] ResteasyWebTarget.proxy(Class) problem with questionmark in @Path with Regex

jamezp:
https://issues.redhat.com/browse/RESTEASY-2845


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -142,29 +138,11 @@ protected Object invokeSync(Object[] args) {
 
     protected ClientInvocation createRequest(Object[] args) {
         WebTarget target = this.webTarget;
+        for (int i = 0; i < processors.length; i++) {
```
ilx, line 141: @jamezp processors.length looks suspicious. If processors can be null then it might blow up here,
OTOH if it cannot be null then why check for null on the next line?

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientInvoker.java
```diff
@@ -177,12 +155,11 @@ protected ClientInvocation createRequest(Object[] args) {
         if (accepts != null) {
             clientInvocation.getHeaders().accept(accepts);
         }
-        if (processors != null) {
-            for (int i = 0; i < processors.length; i++) {
-                if (processors[i] instanceof InvocationProcessor) {
-                    InvocationProcessor processor = (InvocationProcessor) processors[i];
-                    processor.process(clientInvocation, args[i]);
-                }
+        for (int i = 0; i < processors.length; i++) {
```
ilx, line 158: I guess same comment about processors potentially being null applies here

PR #3508, opened 2023-03-02
[RESTEASY-3291] added 2 new methods to process regex expressions and …

rsearls:
…2tests to verify changes

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilderImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilderImpl.java
@@ -179,18 +179,36 @@ public class ResteasyUriBuilderImpl extends ResteasyUriBuilder {
                 this.host = host;
             }
         }
-        if (match.group(5) != null) {
-            String group = match.group(5);
+
+        HashMap<String, String> pathComponentsMap;
+        if (match.group(6) == null && match.group(8) == null) {
+            pathComponentsMap = new HashMap<>();
+            pathComponentsMap.put("path", match.group(5).toString());
+        } else {
+            StringBuilder sb = new StringBuilder();
+            sb.append(match.group(5) == null ? "" : match.group(5));
+            sb.append(match.group(6) == null ? "" : match.group(6));
+            sb.append(match.group(8) == null ? "" : match.group(8));
+            pathComponentsMap = pathSegementParser(sb.toString());
+        }
+
+        if (pathComponentsMap.get("path") != null) {
+            String group = pathComponentsMap.get("path");
             if (!scheme && !"".equals(group) && !group.startsWith("/") && group.indexOf(':') > -1 &&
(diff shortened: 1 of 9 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilderImpl.java
```diff
@@ -607,6 +604,8 @@ protected StringBuilder replaceParameter(Map<String, ? extends Object> paramMap,
             }
             Object value = paramMap.get(param);
             String stringValue = value != null ? value.toString() : null;
+            String regexValue = regexEval(string, stringValue, param);
+            stringValue = regexValue;
```
jamezp, line 608: Any reason not to just do:
```java
String stringValue = value != null ? regexValue(string, value.toString(), param) : null;
```

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilderImpl.java
```diff
@@ -1088,4 +1087,86 @@ public UriBuilder resolveTemplatesFromEncoded(Map<String, Object> templateValues
...
+                if (group5.regionMatches(posSlash + 1, "{", 0, 1)) {
+                    tmpGroup5 = group5 + group6.substring(0, posBracketClose + 1);
+                    String tmpTmpGroup6 = group6.substring(posBracketClose + 1);
+
+                    int posQmark = tmpTmpGroup6.indexOf("?");
+                    if (posQmark != -1) {
+                        tmpGroup6 = tmpTmpGroup6.substring(posQmark + 1);
+                    } else {
+                        tmpGroup6 = "";
+                    }
```
jamezp, line 1122: This breaks the `UriBuilderTest`. The pattern `http://{host}/x/y/{path}?{q}={qval}` is not parsed correctly. The `tmpGroup6` ends up being set to "" leaving off the `?` and `={qval}`. 

It might also be helpful to name the parameters something other than "groupX" so it's more self documenting.

jamezp (changes requested):

resteasy-core/src/main/java/org/jboss/resteasy/specimpl/ResteasyUriBuilderImpl.java
rsearls: an IllegalArgumentException should be thrown if value is null.  regexEval catches this.

jamezp:
The TCK now fails on this :( https://ci.wildfly.org/buildConfiguration/WildFlyProject_RESTEasy_PullRequests_RESTEasyTckRunnerJava11/356260?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

rsearls:
@jamezp  How to kick off the tests again?  There is 1 tck failure with jdk-11 however jdk-17 passes.   I don't see that the failure is related to my code change.  Would like the tests to run again.

jamezp (approved):

PR #3519, opened 2023-03-15
RESTEASY-3302 Make snakeyaml accept tags during parsing.

TomasHofman:
Whitelisting of classes that are allowed to be instantiated is done on the resteasy level instead.

https://issues.redhat.com/browse/RESTEASY-3302

### DIFF
--- a/providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
+++ b/providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
@@ -55,7 +55,7 @@ import static java.security.AccessController.doPrivileged;
 @Produces({"text/yaml", "text/x-yaml", "application/x-yaml"})
 @Deprecated
 public class YamlProvider extends AbstractEntityProvider<Object> {
-   private static final String ALLOWED_LIST = "resteasy.yaml.deserialization.allowed.list.allowIfBaseType";
+   protected static final String ALLOWED_LIST = "resteasy.yaml.deserialization.allowed.list.allowIfBaseType";
    private static final String DISABLE_TYPE_CHECK = "resteasy.yaml.deserialization.disable.type.check";
    // Setting this property tells snakeyaml to allow all tags during parsing. The tags will be instead whitelisted by resteasy
    // provided constructor.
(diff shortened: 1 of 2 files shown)


TomasHofman:
@soul2zimate @jamezp can you review?

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
```diff
@@ -84,6 +90,11 @@ public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotati
    public Object readFrom(Class<Object> type, Type genericType, Annotation[] annotations, MediaType mediaType,
                            MultivaluedMap<String, String> httpHeaders, InputStream entityStream) throws IOException,
             WebApplicationException {
+      // Set "org.yaml.snakeyaml.allow-all-tags" to true to make snakeyaml parsing work in permissive mode, allowing all tags
+      // to be parsed. The tags will be checked later by a resteasy provided constructor (TypeSafeConstructor).
+      // This is only relevant for snakeyaml 1.33.SP2+, which is Red Hat fork of snakeyaml 1.33.
```
iweiss, line 95: I would remove this sentence and point to the upstream discussion on this

soul2zimate (approved):

TomasHofman:
@wolfc your PR works with newly introduced API from snakeyaml 1.33.SP / snakeyaml 2.0, using that API would mean resteasy becomes incompatible with the original snakeyaml 1.33. I thought it would make matters lot easier if we stay compatible with both 1.33 / 1.33.SP.

We would also probably need to release community snakeyaml 1.33.SP, so resteasy could switch to it.

TomasHofman:
@jamezp I added a fix for reading the whitelisted classes from system properties.

providers/yaml/src/main/java/org/jboss/resteasy/plugins/providers/YamlProvider.java
soul2zimate: Yes, that's a shame. Could we point to https://bitbucket.org/snakeyaml/snakeyaml/pull-requests/44 ?

providers/yaml/src/test/java/org/jboss/resteasy/plugins/providers/YamlProviderTestCase.java
jamezp, line 1: This needs a license header.

jamezp (changes requested):

PR #3521, opened 2023-03-16
RESTEASY-3302: put SnakeYAML into permissive mode by using trusted tag inspector

wolfc:
https://issues.redhat.com/browse/RESTEASY-3302

Note that this won't pass the testsuite as it uses WildFly 24.0.1 to test against.


jamezp:
This is being handled elsewhere therefore I'm closing this.

PR #3522, opened 2023-03-17
[RESTEASY-3300] Do not append single '&' with empty query parameters

Dkafetzis:
Fixed this Jira Issue (https://issues.redhat.com/browse/RESTEASY-3300) and added an integration test for this case


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/RequestNamedQueryParameterTest.java
```diff
@@ -0,0 +1,186 @@
...
+                return null;
+            }
+
+            @Override
+            public void close() {
+            }
+        }).build();
+
+        try {
+            client.target("http://a.place.org").proxy(SomeResource.class).methodWithLists(List.of(), List.of(), List.of());
```
jamezp, line 95: We should use localhost in all the places here. While this shouldn't be resolved given the above, this is a potential exploit point.

jamezp (changes requested):
We could actually test this with standard API's rather than internal ones. We could use a `ContainerRequestFilter` and get the URI from there to check. In some ways this would be the preferred way IMO.

If we want to stick with using internal API's we should have a method like:
```java
private static Client createClient(final String expectedQuery) {
    ResteasyClientBuilder builder = (ResteasyClientBuilder) ClientBuilder.newBuilder();
    ResteasyClient client = builder.httpEngine(new ClientHttpEngine() {
        @Override
        public Response invoke(Invocation request) {
            Assert.assertEquals(expectedQuery, ((ClientInvocation) request).getUri().getQuery());
            return new AbortedResponse(null, new ServerResponse());
        }

        @Override
        public SSLContext getSslContext() {
            return null;
        }

        @Override
        public HostnameVerifier getHostnameVerifier() {
            return null;
        }

        @Override
        public void close() {
        }
    }).build();
}
```
This would avoid us repeating similar patterns in the test.

This is a little OT for this PR, but I want to remove the usages of `ResteascyClientBuilder` and the `ResteasyClient` in the test suite where we can. We should be using standard API's for this as they are available.

Dkafetzis:
@jamezp Changes implemented, waiting for review.

Dkafetzis:
@jamezp should be fine now.

Dkafetzis:
@jamezp All tests passed, I think we can merge this.

jamezp (approved):

jamezp:
replaced by #3570

PR #3523, opened 2023-03-17
RESTEASY-3302 Implement deserialization whitelist using new snakeyaml API

TomasHofman:
This is an alternative solution to https://github.com/resteasy/resteasy/pull/3519/.


TomasHofman:
I'm still about to run all the integration tests, and this is also going to fail on CI until we release community artifacts for snakyaml 1.33.SP1.

TomasHofman:
This PR may be unsuitable as it changes behavior compared to an already released patch. (Need to verify.)

PR #3542, opened 2023-04-05
[RESTEASY-3303] Partial revert of RESTEASY-2874. We should not overri…

jamezp:
…de the Content-Type in proxies where the header possibly set via other means. Otherwise, we should allow the entity to always override the Content-Type.

https://issues.redhat.com/browse/RESTEASY-3303

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/servlet/ServletConfigTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/servlet/ServletConfigTest.java
@@ -75,7 +75,7 @@ public class ServletConfigTest {
     public void testNullJaxb() throws Exception {
         Response response = client.target(generateURL("/my/null")).request().header("Content-Type", "application/xml")
                 .post(Entity.text(""));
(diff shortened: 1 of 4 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/servlet/ServletConfigTest.java
```diff
@@ -75,7 +75,7 @@ public void testCount() throws Exception {
     public void testNullJaxb() throws Exception {
         Response response = client.target(generateURL("/my/null")).request().header("Content-Type", "application/xml")
                 .post(Entity.text(""));
-        Assert.assertEquals(HttpResponseCodes.SC_BAD_REQUEST, response.getStatus());
+        Assert.assertEquals(HttpResponseCodes.SC_UNSUPPORTED_MEDIA_TYPE, response.getStatus());
```
jamezp, line 78: This was changed in #2794 ([RESTEASY-2874](https://issues.redhat.com/browse/RESTEASY-2874)) and reverted back here. This should likely not have been changed.

The test itself might be incorrect, but it shouldn't have been changed.

PR #3626, opened 2023-05-10
[RESTEASY-3314] Use a global executor service to avoid creating too m…

jamezp:
…any executors.

https://issues.redhat.com/browse/RESTEASY-3314

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/GlobalContextualExecutorService.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/GlobalContextualExecutorService.java
@@ -1,0 +1,98 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/GlobalContextualScheduledExecutorService.java
```diff
@@ -0,0 +1,95 @@
...
+    public boolean isTerminated() {
+        return false;
+    }
+
+    @Override
+    ScheduledExecutorService getDelegate() {
+        if (delegate == null) {
+            synchronized (this) {
+                if (delegate == null) {
+                    final int poolSize = SecurityActions.getCoreThreads("resteasy.async.timeout.scheduler.min.pool.size");
```
jamezp, line 75: Maybe default to unbounded.

resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/GlobalContextualExecutorService.java
```diff
@@ -0,0 +1,96 @@
...
+    public boolean isTerminated() {
+        return false;
+    }
+
+    @Override
+    ExecutorService getDelegate() {
+        if (delegate == null) {
+            synchronized (this) {
+                if (delegate == null) {
+                    final int poolSize = SecurityActions.getCoreThreads("dev.resteasy.concurrent.core.pool.size");
```
jamezp, line 76: Maybe default to unbounded.

PR #3628, opened 2023-05-11
[RESTEASY-3314] Use a global executor service to avoid creating too m…

jamezp:
…any executors.

https://issues.redhat.com/browse/RESTEASY-3314

Upstream #3626


liweinan:
Looks great!

PR #3632, opened 2023-05-15
[RESTEASY-2992] Fix the ClientConfigProviderTestJarHelper to work on Windows

Dkafetzis:
Jira issue link https://issues.redhat.com/browse/RESTEASY-2992

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
@@ -9,11 +9,14 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.jar.Attributes;
 import java.util.jar.JarEntry;
 import java.util.jar.JarOutputStream;
 import java.util.jar.Manifest;
+import java.util.regex.Pattern;
 
 /**
  * Contains utility methods used for creating, running and getting results of jars meant to test ClientConfigProvider
@@ -41,7 +44,7 @@ class ClientConfigProviderTestJarHelper {
 
     private static final String PACKAGE_NAME = "org.jboss.resteasy.test.security.testjar";
     private static final String PACKAGE_PATH = "org/jboss/resteasy/test/security/testjar/";
-    private static final String JAR_NAME = "client-config-provider-test.jar";
+    private static final String JAR_NAME = "clientconf.jar";
 
     private static ConfigProviderProperties bearerJarConfigProperties = new ConfigProviderProperties();
     private static ConfigProviderProperties basicAuthJarConfigProperties = new ConfigProviderProperties();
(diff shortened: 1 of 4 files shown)


Dkafetzis:
@jamezp I think the format check is stuck on this one.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
```diff
@@ -120,13 +122,39 @@ static Process runClientConfigProviderBearerTestJar(TestType testType, String ja
...
+        for (String s : paths) {
+            final Path path = Path.of(s);
+            if (pattern.matcher(path.toString()).matches()) {
+                continue;
+            }
+            newCp.append(path).append(File.pathSeparator);
+        }
+        newCp.deleteCharAt(newCp.length() - 1);
+        // use quotation marks for classpath on windows because folder names can have spaces
+        if (System.getProperty("os.name").contains("indows")) {
```
jamezp, line 142: This should use the `TestUtil.isWindows()`. TBH we probably don't even need to test it as it's valid to surround it in quotes for Linux too.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
```diff
@@ -120,13 +122,39 @@ static Process runClientConfigProviderBearerTestJar(TestType testType, String ja
...
+        if (args[1].equals("TEST_BEARER_TOKEN_IS_USED") || args[1].equals("TEST_BEARER_TOKEN_IGNORED_IF_BASIC_SET_BY_USER")) {
+            pb = new ProcessBuilder("java", "-ea", "-cp", newCp.toString(), ClientConfigProviderTestJarHelper.PACKAGE_NAME + "."
+                    + args[0], args[1]);
+        } else if (args[1].equals("TEST_SSLCONTEXT_USED")) {
+            pb = new ProcessBuilder("java", "-cp", newCp.toString(), ClientConfigProviderTestJarHelper.PACKAGE_NAME + "."
+                    + args[0], args[1], args[2], args[3]);
+        } else {
+            pb = new ProcessBuilder("java", "-cp", newCp.toString(), ClientConfigProviderTestJarHelper.PACKAGE_NAME + "."
+                    + args[0], args[1], args[2]);
+        }
```
jamezp, line 156: Why this different behavior? I understand we have to customize the command to be an array instead of a string, but this looks overly complicated.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
```diff
@@ -120,13 +122,39 @@ static Process runClientConfigProviderBearerTestJar(TestType testType, String ja
...
+        final String[] paths = cp.split(Pattern.quote(File.pathSeparator));
+        final StringBuilder newCp = new StringBuilder();
+        for (String s : paths) {
+            final Path path = Path.of(s);
+            if (pattern.matcher(path.toString()).matches()) {
+                continue;
+            }
+            newCp.append(path).append(File.pathSeparator);
+        }
+        newCp.deleteCharAt(newCp.length() - 1);
```
jamezp, line 140: Why are we doing this?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
```diff
@@ -120,13 +122,39 @@ static Process runClientConfigProviderBearerTestJar(TestType testType, String ja
...
     static Process runClientConfigProviderTestJar(String jarPath, String[] args) throws IOException {
-        // use quotation marks for classpath on windows because folder names can have spaces
-        String classPath = System.getProperty("os.name").contains("indows")
-                ? "\"" + jarPath + ";" + System.getProperty("java.class.path") + "\""
+        final Pattern pattern = Pattern.compile(
+                ".*(maven|plexus|sisu|xnio|jsoup|compress|shrinkwrap|IntelliJ|mockito|arquillian|aether|wildfly|glassfish|hibernate|netty|xerces|xalan|hamcrest|io7m|slf4j|objenesis|sshd|marshalling|remoting|bytebuddy).*");
+
+        final String cp = System.getProperty("os.name").contains("indows")
+                ? jarPath + ";" + System.getProperty("java.class.path")
                 : jarPath + ":" + System.getProperty("java.class.path");
```
jamezp, line 130: Why not simply:
```java
final var cp = jarPath + File.pathSeparator + System.getProperty("java.class.path");
```

jamezp (changes requested):

jamezp:
> @jamezp I think the format check is stuck on this one.

Sorry @Dkafetzis, it just required an approval to run for some reason.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
Dkafetzis: For some reason when the command is provided with the quotes on process builder on Linux it fails to load the class path. I know that Linux normally supports them but something is not right with the way the ProccessBuilder parses the command.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
Dkafetzis: Correct, this will be corrected

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
Dkafetzis: Originally the above method left a trailing path separator in the new Class Path, after more testing I think it work even with the trailing separator so we can remove this

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/security/ClientConfigProviderTestJarHelper.java
Dkafetzis: Apparently Process Builder handles by itself when to put quotes and when not to. I removed the quotes on windows too and the tests passed, so if the tests don't fail here too it is probably ok to remove the extra quotes.

Dkafetzis:
@jamezp changes applied, waiting for another review

jamezp (changes requested):
It looks like some formatting was missed before the commit. Just make sure you compile before committing. You could just do something like the following to fix this:
```bash
mvn clean install -DskipTests
git commit -a --amend
```

Dkafetzis:
@jamezp I don't know why the TCK test is failing now, I also rerun the tests locally and everything is passing

jamezp (approved):

jamezp:
@Dkafetzis this is a known intermittent failure.

PR #3665, opened 2023-06-12
* feat(resteasy-guice): add back the Google Guice integration

desaerun:
Guice 7.0.0 is fully on the jakarta namespace now.
 Reverts 8e64b0650e0066ac1b73299affac3facd182a988

### DIFF
--- a/distribution/assembly.xml
+++ b/distribution/assembly.xml
@@ -41,6 +41,7 @@
                 <include>org.jboss.resteasy:resteasy-json-p-provider</include>
                 <include>org.jboss.resteasy:resteasy-json-binding-provider</include>
                 <include>org.jboss.resteasy:resteasy-jackson2-provider</include>
+                <include>org.jboss.resteasy:resteasy-guice</include>
                 <include>org.jboss.resteasy:resteasy-cdi</include>
                 <include>org.jboss.resteasy:resteasy-jsapi</include>
                 <include>org.jboss.resteasy:resteasy-links</include>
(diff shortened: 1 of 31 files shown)


jamezp:
For now I'll leave this open, but I don't think we want Guice back in RESTEasy. We could make a separate project or add it to the [resteasy-extensions](https://github.com/resteasy/resteasy-extensions). I don't think it belongs in RESTEasy though. 

Another thing to note is in Jakarta REST 4.0 CDI is going to be replacing the `@Context` injection. This will likely cause a refactor of how we use injection. Part of the plan is to hopefully make it more plugable for other DI frameworks.

RWSalexjarvis:
Hey, no problem! I was bumping the dependencies on an old project and ended up using a hacked-together version of this to get resteasy and guice working together again, so I figured with Guice being on jakarta it might be useful to add `resteasy-guice` back in (since the rationale for removing it in the first place was that they weren't migrating to the jakarta namespace fast enough).

Thanks for your review in any case.

desaerun:
Hey, no problem! I was bumping the dependencies on an old project and ended up using a hacked-together version of this to get resteasy and guice working together again, so I figured with Guice being on jakarta it might be useful to add resteasy-guice back in (since the rationale for removing it in the first place was that they weren't migrating to the jakarta namespace fast enough).

Thanks for your review in any case.

jamezp:
My apologies for the delay on replying to this. This is something we could consider adding support again for. However, it should be it's own separate project which we can create. If we do this, there will be a need for community support. This is something I'm not too familiar with and do not use in any other projects.

jamezp:
This was moved to https://github.com/resteasy/resteasy-guice

PR #3673, opened 2023-06-25
Reactor-Netty client: Make sure most exceptions are ProcessingExceptions

crankydillo:
The JAX-RS API does not use checked exceptions, so it's important to covert to `ProcessingException` early so that callers have reliable exception behavior.  For example, IOExceptions are propagated as the causes of many ProcessingExceptions and callers may want to retry these.


crankydillo:
@jamezp Do you think this lines up with how the JAX-RS api wants us to convey exceptions like `IOException`?

jamezp:
@crankydillo I think this makes sense. I filed https://issues.redhat.com/browse/RESTEASY-3349 for it. We'll need to port this fix to the main and 6.2 branches as well.

jamezp (approved):

crankydillo:
> We'll need to port this fix to the main and 6.2 branches as well.

Thanks @jamezp .  Would you like me to do this?

crankydillo:
@jamezp Would you like me to make any changes on this or the other PRs (#3708 and #3709)?

jamezp:
> @jamezp Would you like me to make any changes on this or the other PRs (#3708 and #3709)?

Hello @crankydillo, sorry I'm just back from PTO today. I'm hopping to have a look at all of these by the end of this week.

crankydillo:
No worries and thanks.  Hope you enjoyed your PTO!

PR #3705, opened 2023-07-13
SslSniHostNamesTest : Using TLSv1.2 for all jdks ...

panossot:
SslSniHostNamesTest : Using TLSv1.2 for all jdks ...


jamezp:
We are no longer supporting the 3.15 branch. However, given this is a test only change it seems reasonable.

jamezp (approved):

PR #3710, opened 2023-07-18
[RESTEASY-3324] Enable redirects using property.

The-Huginn:
https://issues.redhat.com/browse/RESTEASY-3324

### DIFF
--- a/resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
+++ b/resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
@@ -50,6 +50,8 @@ public abstract class ResteasyClientBuilder extends ClientBuilder {
      */
     public static final String PROPERTY_PROXY_SCHEME = "org.jboss.resteasy.jaxrs.client.proxy.scheme";
 
+    public static final String PROPERTY_FOLLOW_REDIRECTS = "dev.resteasy.client.follow.redirects";
+
     /**
      * Changing the providerFactory will wipe clean any registered components or properties.
      *
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
@@ -383,6 +383,11 @@ public class ResteasyClientBuilderImpl extends ResteasyClientBuilder {
             }
         }
 
+        final Object localFollowRedirects = config.getProperty(PROPERTY_FOLLOW_REDIRECTS);
+        if (localFollowRedirects != null) {
+            this.followRedirects = Boolean.parseBoolean(String.valueOf(localFollowRedirects));
+        }
+
         ClientHttpEngine engine = httpEngine != null ? httpEngine
                 : new ClientHttpEngineBuilder43().resteasyClientBuilder(this).build();
         if (resetProxy) {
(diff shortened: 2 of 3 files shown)


resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/i18n/Messages.java
```diff
@@ -163,4 +163,7 @@ public interface Messages {
     @Message(id = BASE + 194, value = "No content type found in response. Cannot extract the response value.")
     @Signature(messageIndex = 1, value = { Response.class, String.class })
     ResponseProcessingException noContentTypeFound(@Param Response response);
+
+    @Message(id = BASE + 195, value = "Unable to set follow redirects policy.")
+    String unableToSetFollowRedirects();
```
jamezp, line 168: Given the comment below we likely don't need this.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
```diff
@@ -667,6 +667,25 @@ public ResteasyClientBuilder setFollowRedirects(boolean followRedirects) {
...
+                if (enabled instanceof Boolean) {
+                    this.followRedirects = (Boolean) enabled;
+                } else if (enabled instanceof String) {
+                    this.followRedirects = Boolean.parseBoolean((String) enabled);
+                }
+            }
+        } catch (Exception e) {
+            // catch possible exceptions (in this case we do not set follow-redirects at all)
+            LogMessages.LOGGER.warn(Messages.MESSAGES.unableToSetFollowRedirects(), e);
+        }
```
jamezp, line 685: This isn't quite right. We need to remove this whole method and check the property in the `build()` method. In the `build()` method it should just look something like:

```java
final Object localFollowRedirects = config.getProperty(PROPERTY_FOLLOW_REDIRECTS);
if (localFollowRedirects != null) {
    followRedirects = Boolean.parseBoolean(String.valueOf(localFollowRedirects));
}
```

resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/ResteasyClientBuilder.java
```diff
@@ -50,6 +50,8 @@ public enum HostnameVerificationPolicy {
      */
     public static final String PROPERTY_PROXY_SCHEME = "org.jboss.resteasy.jaxrs.client.proxy.scheme";
 
+    public static final String PROPERTY_FOLLOW_REDIRECTS = "org.jboss.resteasy.jaxrs.client.follow-redirects";
```
jamezp, line 53: We have started to change the prefix we use. This should be something like `dev.resteasy.client.follow.redirects`.

jamezp (changes requested):
This would also need to be added to the documentation in the `RESTEasy_Client_Framework.xml` in the `docbook` directory as well.

The-Huginn:
Currently I assume, the property has higher priority. Is that expected or do I add additional logic to allow overriding property by `setFollowRedirects()`?

jamezp:
> Currently I assume, the property has higher priority. Is that expected or do I add additional logic to allow overriding property by `setFollowRedirects()`?

It's a tough one, but yes we are saying the property wins. If a user did something like:
```
ResteasyClientBuilder().setFollowRedirects(false)
    .property("dev.resteasy.client.follow.redirect", "true");
```

The latter would win. However, I think that's okay as then the user doesn't need to use the `ResteasyClientBuilder` directly.

jamezp (approved):

PR #3711, opened 2023-07-19
[Resteasy 3304] Return null upon missing value of request parameter

The-Huginn:
https://issues.redhat.com/browse/RESTEASY-3304

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
@@ -801,6 +801,9 @@ public class StringParameterInjector {
                     throw ((WebApplicationException) targetException);
                 }
                 LogMessages.LOGGER.unableToExtractParameter(e, getParamSignature(), strVal, target);
+                if (strVal.isBlank()) {
+                    return null;
+                }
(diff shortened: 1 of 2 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -801,6 +801,9 @@ public Object extractValue(String strVal) {
                     throw ((WebApplicationException) targetException);
                 }
                 LogMessages.LOGGER.unableToExtractParameter(e, getParamSignature(), strVal, target);
+                if (strVal.equals("")) {
+                    return null;
+                }
```
jamezp, line 806: This pretty significantly changes behavior. I'm not sure we want to do this.

resteasy-core/src/main/java/org/jboss/resteasy/core/StringParameterInjector.java
```diff
@@ -801,6 +801,9 @@ public Object extractValue(String strVal) {
                     throw ((WebApplicationException) targetException);
                 }
                 LogMessages.LOGGER.unableToExtractParameter(e, getParamSignature(), strVal, target);
+                if (strVal.equals("")) {
```
jamezp, line 804: We should just use `strValue.isBlank()`

jamezp (approved):

PR #3712, opened 2023-07-19
[RESTEASY-3358] Drain SseEventSink queue on close with flush

radcortez:
When calling `close` on `SseEventSink`, it is possible to miss events that are still in the queue, because they are still waiting on output completion before being drained. Currently, the `close` method clears all events from the queue.

The proposal is to just call `drainQueue` instead of clearing the events on a close with flush.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/providers/sse/SseEventOutputImpl.java
@@ -387,20 +387,21 @@ public class SseEventOutputImpl extends GenericType<OutboundSseEvent> implements
         if (state.getAndSet(CLOSED) != CLOSED) {
             if (error != null) {
                 drainQueue(error);
-            } else {
-                synchronized (lock) {
-                    events.clear();
-                }
             }
             if (flushBeforeClose && responseFlushed) {
                 try (CloseableContext c = ResteasyContext.addCloseableContextDataLevel(contextDataMap)) {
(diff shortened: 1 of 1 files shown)


jamezp:
This looks reasonable to me. The TCK passed and I don't see anything specific in the spec on what should be done with queued events on close. The only possible issue I could see is if the event processing takes some time, but I'm not sure that's a real issue.

Before this is merged, we will need an upstream PR too for the main branch and we'll need a JIRA. Both steps can be done by the merger though. I'm on PTO until 2023-07-26, but I can have a look when I get back if this is still here.

radcortez:
> This looks reasonable to me. The TCK passed and I don't see anything specific in the spec on what should be done with queued events on close. The only possible issue I could see is if the event processing takes some time, but I'm not sure that's a real issue.

Yes, I noticed this by running the TCK on Quarkus. It seems the Vert.x writer gets deferred and in the meanwhile the close from `try` kicks in and clears the events. With the chance I was able to pass these TCKs in Quarkus.

> Before this is merged, we will need an upstream PR too for the main branch and we'll need a JIRA. Both steps can be done by the merger though. I'm on PTO until 2023-07-26, but I can have a look when I get back if this is still here.

No worries. I can do it.

Thanks!

jamezp:
@jimma WDYT?

jimma:
@jamezp  It's reasonable to me too. For the even processing time, is it possible to set some allowed time span to drain the queue if there are a large number of events in the queue ?

radcortez:
JIRA: https://issues.redhat.com/browse/RESTEASY-3358
PR to main: https://github.com/resteasy/resteasy/pull/3715

PR #3721, opened 2023-07-25
[RESTEASY-3205] Change Content-Type to text/plain for Security filter

The-Huginn:
https://issues.redhat.com/browse/RESTEASY-3205


jamezp:
@ronsigal What do you think about this? It could be considered a breaking change, but it feels more correct to me.

jamezp:
> Hey @jamezp,
> 
> Sorry for the delay. Technically, "Access forbidden: role not allowed" isn't correct HTML, so using "text/html" just isn't correct. I agree that switching to "text/plain" might be a breaking change, but haven't we always accepted a breaking change in the interest of correctness?
> 
> -Ron

Thanks Ron! Yes, I agree a correctness is important and this can be changed. I was just looking for your opinion on it as well :)

PR #3728, opened 2023-08-02
Fix client-side ParamConverter retrieval

coder-hugo:
(Not sure if it's necessary to create an issue for this small change so I just created a PR)

At the moment the `resteasy-client` doesn't retrieve `ParamConverter`s for objects different from Collection and array if the `ParamConverterProvider` requires the `genericType` or the `annotations`. This PR fixes the issue.
I haven't found any tests for the retrieval of `ParamConverter`s on client-side. If there are any could you just point me to them so that I can add a test case for this.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/processors/AbstractCollectionProcessor.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/processors/AbstractCollectionProcessor.java
@@ -57,6 +57,10 @@ public abstract class AbstractCollectionProcessor<T> {
                 target = apply(target, arr);
             }
         } else {
+            ParamConverter<Object> paramConverter = config.getParamConverter(object.getClass(), type, annotations);
+            if (paramConverter != null) {
+                object = paramConverter.toString(object);
+            }
             target = apply(target, object);
         }
         return target;


jamezp:
We will need a JIRA for this, really for any bug. However, I can create that. For now we'll let CI run to see how it goes :)

Do you agree to these terms?

coder-hugo:
Yes, I agree to the terms of the ASL 2.0 License.

jamezp:
@coder-hugo Do you have an example use-case by chance?

coder-hugo:
If you use for example this `ParamConverterProvider`:
```java
@Provider
public class UnmodifiedSinceParamConverterProvider implements ParamConverterProvider {

    @Override
    public <T> ParamConverter<T> getConverter(Class<T> rawType, Type genericType, Annotation[] annotations) {
        if (rawType.isAssignableFrom(OffsetDateTime.class) && Optional.ofNullable(annotations).stream().flatMap(Stream::of).anyMatch(this::match)) {
            return (ParamConverter<T>) new UnmodifiedSinceParamConverter();
        }
        return null;
    }

    private boolean match(Annotation annotation) {
        return annotation instanceof HeaderParam headerParam && "If-Unmodified-Since".equals(headerParam.value());
    }

    private static class UnmodifiedSinceParamConverter implements ParamConverter<OffsetDateTime> {

        @Override
        public OffsetDateTime fromString(String value) {
            return value == null ? null : OffsetDateTime.parse(value, DateTimeFormatter.RFC_1123_DATE_TIME);
        }

        @Override
        public String toString(OffsetDateTime value) {
            return value == null ? null : value.format(DateTimeFormatter.RFC_1123_DATE_TIME);
        }

    }

}
```
for this interface:
```java
@Path("/")
public interface TestApi {

    @POST
    void update(@HeaderParam("If-Unmodified-Since") OffsetDateTime ifUnmodifiedSince);

}
```
resteasy will register in the current version the `UnmodifiedSinceParamConverter` for the `update` method on server side but not on client side.

jamezp (approved):

jamezp:
Thank you @coder-hugo. I'll be back-porting this fix to 6.2 and 5.0 as well.

PR #3730, opened 2023-08-02
Make sure most exceptions are ProcessingExceptions.

jamezp:
The JAX-RS API does not use checked exceptions, so it's important to early on do the conversion to `ProcessingException` because that's what client's of the JAX-RS API have to catch in order to look for IOExceptions (in the cause) that they may want to retry.

Upstream: #3729


jamezp:
Duplicate of #3709

PR #3736, opened 2023-08-04
[RESTEASY-3362] Do not depend on Xerces for the JAXP implementation. …

jamezp:
…This required test changes as when enabling the feature secure processing, this does not allow overriding external connections in the JDK's JAXP implementation.

https://issues.redhat.com/browse/RESTEASY-3362


ropalka:
Thanks for sorting this issue out @jamezp !

jamezp:
IMO this belongs in main for sure, which is going to be 7.0.0.Alpha1 soon. If approved by others, we will back port to 6.2.

PR #3741, opened 2023-08-08
Add the ability to use the Java HttpClient as the default HttpClientE…

jamezp:
…ngine.


jamezp:
> This is great!

Thank you. I'm not sure why these tests are failing as they pass, at least on Linux, for me locally. I'm on a new laptop so I need to get a new Windows VM setup to test there.

liweinan:
It seems like some random async problems.

jamezp:
> It seems like some random async problems.

Yes, in some cases it looks like the connection gets closed so it could be a real issue. I just have been able to reproduce it.

jamezp:
Please also see the discussion https://github.com/resteasy/resteasy/discussions/3743

PR #3742, opened 2023-08-09
fixing issue with proxybuild

ChrisL113:
Hi I just recently had an issue with this package I would like that you take a look at my changes since this is the solution for this comeup error 
image

### DIFF
--- a/resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/ProxyBuilder.java
+++ b/resteasy-client-api/src/main/java/org/jboss/resteasy/client/jaxrs/ProxyBuilder.java
@@ -12,30 +12,17 @@ public abstract class ProxyBuilder<T> {
 
     @SuppressWarnings({ "rawtypes", "unchecked" })
     public static <T> ProxyBuilder<T> builder(Class<T> iface, WebTarget webTarget) {
-        try {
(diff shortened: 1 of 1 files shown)


ChrisL113:
i'm running the latest image of keycloak, here https://hub.docker.com/r/jboss/keycloak/

jamezp:
Do you by any chance have an example deployment? I would need to see what's happening. However, we really need to default to the TCCL. Given the `KeycloakConnectionService` in that stack trace, it might be possible to setup the TCCL (`Thread.currentThread().setContextClassLoader()`).

ChrisL113:
I will ask to my team tomorrow and i will get back to you

ChrisL113:
here there are some things I changed tho
image

ChrisL113:
hello ? ?????

jamezp:
Sorry, I was on PTO. I'd like to understand how to reproduce the issue here, but I'm not sure how to do that. We can't do the change as that will break how RESTEasy works in containers like WildFly.

jamezp:
I'm closing this as we need to use the TCCL in most containers. If this is still an issue, please feel free to file an issue on [JIRA](https://issues.redhat.com/browse/RESTEASY) with some details on how to reproduce the issue. Thank you for your patience on this overdue PR.

PR #3775, opened 2023-08-23
[RESTEASY-3111] Properly inherit @Cache annotation for base class.

The-Huginn:
https://issues.redhat.com/browse/RESTEASY-3111

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
@@ -19,10 +19,11 @@ public class CacheControlFeature implements DynamicFeature {
     @Override
     public void configure(ResourceInfo resourceInfo, FeatureContext configurable) {
         final Class<?> declaring = resourceInfo.getResourceClass();
-        final Method method = resourceInfo.getResourceMethod();
 
-        if (declaring == null || method == null)
+        if (declaring == null || resourceInfo.getResourceMethod() == null)
             return;
+
+        final Method method = findInterfaceBasedMethod(declaring, resourceInfo.getResourceMethod());
         if (!method.isAnnotationPresent(GET.class))
             return;
 
@@ -73,5 +74,26 @@ public class CacheControlFeature implements DynamicFeature {
         for (String field : value.fields())
             cacheControl.getNoCacheFields().add(field);
         return cacheControl;
+    }
+
+    private Method findInterfaceBasedMethod(Class<?> root, Method method) {
+        if (method.getAnnotation(Cache.class) != null || method.getAnnotation(NoCache.class) != null) {
+            return method;
+        }
+
+        if (method.getDeclaringClass().isInterface() || root.isInterface()) {
+            return method;
+        }
+
+        for (Class<?> intf : root.getInterfaces()) {
+            try {
+                return intf.getMethod(method.getName(), method.getParameterTypes());
+            } catch (NoSuchMethodException ignored) {
+            }
+        }
+
+        if (root.getSuperclass() == null || root.getSuperclass().equals(Object.class))
+            return method;
+        return findInterfaceBasedMethod(root.getSuperclass(), method);
     }
 }
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/CacheControlAnnotationTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/CacheControlAnnotationTest.java
@@ -5,12 +5,15 @@ import jakarta.ws.rs.client.ClientBuilder;
 import jakarta.ws.rs.client.WebTarget;
 import jakarta.ws.rs.core.CacheControl;
 import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.RuntimeDelegate;
 
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.test.core.basic.resource.CacheAnnotationInheritance;
 import org.jboss.resteasy.test.core.basic.resource.CacheControlAnnotationResource;
+import org.jboss.resteasy.test.core.basic.resource.CacheControlAnnotationResourceInheritance;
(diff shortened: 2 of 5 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/resource/CacheAnnotationInteritance.java
jamezp, line 1: This will need a license header.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/resource/CacheControlAnnotationResourceInheritance.java
jamezp, line 1: Same as above, we need a license header here.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
```diff
@@ -74,4 +74,20 @@ protected CacheControl initCacheControl(NoCache value) {
             cacheControl.getNoCacheFields().add(field);
         return cacheControl;
     }
+
+    private Method findInterfaceBasedMethod(Class<?> root, Method method) {
+        if (method.getDeclaringClass().isInterface() || root.isInterface())
```
jamezp, line 79: The `method.getDeclaringClass().isInterface()` could throw an NPE. The `Class.getDeclaringClass()` may return `null`.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/CacheControlAnnotationTest.java
```diff
@@ -108,4 +113,20 @@ public void testResourceCompositeNoCache() {
...
+    public void testInheritedResourceValid() {
+        WebTarget base = client.target(generateURL("/inheritance"));
+        Response response = base.request().get();
+
+        Assert.assertEquals(HttpServletResponse.SC_OK, response.getStatus());
+        CacheControl cc = CacheControl.valueOf(response.getHeaderString("cache-control"));
+        Assert.assertFalse("Cache should not be private", cc.isPrivate());
+        Assert.assertEquals("Wrong age of cache", 3600, cc.getMaxAge());
+
+        response.close();
```
jamezp, line 130: May as well use try-with-resources since we will likely not back port this change.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
```diff
@@ -74,4 +74,20 @@ protected CacheControl initCacheControl(NoCache value) {
...
+            return method;
+
+        for (Class<?> intf : root.getInterfaces()) {
+            try {
+                return intf.getMethod(method.getName(), method.getParameterTypes());
```
jamezp, line 84: We need to look at the call stack here and determine whether or not this should be a privileged action if the security manager is being used.

jamezp (changes requested):
We should add `@Inherited` to the `@Cache` annotation as well. It acts as documentation plus for non-interfaces it will work. We should look at the lookup algorithm too as we should only need to be recursive on interfaces. Classes should automatically inherit the annotation.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
The-Huginn: I am not sure if it can return null. However `resourceInfo.getResourceMethod() or resourceInfo.getResourceClass()` could do that. I have updated the code accordingly to that

resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
The-Huginn: Just to add, I have used the same approach as for example `@GET` annotation retrieves inheritance. Might be useful to update that as well then.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
jamezp: Sorry, I think I mixed something up in my head :)

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/CacheControlAnnotationTest.java
```diff
@@ -60,14 +66,14 @@ public void after() throws Exception {
...
-        Assert.assertEquals("Wrong age of cache", 3600, cc.getMaxAge());
 
-        response.close();
+        try (Response response = base.request().get()) {
+            Assert.assertEquals(HttpServletResponse.SC_OK, response.getStatus());
+            CacheControl cc = RuntimeDelegate.getInstance().createHeaderDelegate(CacheControl.class)
+                    .fromString(response.getHeaderString("cache-control"));
+            Assert.assertFalse("Cache should not be private", cc.isPrivate());
+            Assert.assertEquals("Wrong age of cache", 3600, cc.getMaxAge());
+        }
```
jamezp, line 76: This is just an FYI. While I don't mind changes like this, they should be in a separate commit. In this specific case, I don't think a JIRA is needed. You don't need to make that change for this PR, but in the future changes for formatting, patterns or removing deprecated code uses should be done separate from the main change being performed.

jamezp (changes requested):

resteasy-core/src/main/java/org/jboss/resteasy/plugins/interceptors/CacheControlFeature.java
```diff
@@ -74,4 +75,20 @@ protected CacheControl initCacheControl(NoCache value) {
...
+            } catch (NoSuchMethodException ignored) {
+            }
+        }
+
+        if (root.getSuperclass() == null || root.getSuperclass().equals(Object.class))
+            return method;
+        return findInterfaceBasedMethod(root.getSuperclass(), method);
+    }
```
jamezp, line 93: This isn't quite right. It should look at the method and see if it has the `@Cache` annotation before walking down the dependency chain looking for one to inherit. In other words, if the annotation is on the implementation, that should override the annotation that's on the interface method.

jamezp (changes requested):

jamezp (approved):

PR #3803, opened 2023-09-15
[RESTEASY-3384] Improve parsing of malformed MediaTypes in MediaTypeHeaderDelegate

thomasdarimont:
Previously, a "broken" MIME-type could trigger an StringIndexOutOfBoundsException instead of the more suitable IllegalArgumentException.

Example: "Accept: x; /x"

This PR now throws an IllegalArgumentException in case of a broken MIME-type.

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
@@ -93,7 +93,11 @@ public class MediaTypeHeaderDelegate implements RuntimeDelegate.HeaderDelegate<M
         } else {
             major = type.substring(0, typeIndex);
             if (paramIndex > -1) {
-                subtype = type.substring(typeIndex + 1, paramIndex);
+                int beginIndex = typeIndex + 1;
+                if (beginIndex > paramIndex) {
+                    throw new IllegalArgumentException(Messages.MESSAGES.failureParsingMediaType(type));
+                }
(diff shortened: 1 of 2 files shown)


thomasdarimont:
It would be great if this could be backported to earlier versions. Other projects like Keycloak would benefit from a backport to resteasy 4.7.9+.

jamezp:
Thank you @thomasdarimont. I'm on PTO for another week, but will look in more detail when I'm back. We are currently not doing any more 4.7.x releases. Version 5.0.x should effectively be the same. The main difference there is things like the MicroProfile Client and Spring were pulled out and moved to separate projects. Google Guice support was removed as well. Everything else should be the same though.

Do you agree to these terms?

thomasdarimont:
Thanks for your feedback!

Yes, I agree with the ASL terms.

The 5.x upgrade story (~seamless upgrade from 4.7.9) sounds interesting, I'm going

thomasdarimont:
Sorry for closing, "fat-fingered" the wrong button on my mobile :-D

thomasdarimont:
@jamezp I just noticed that the same issue is present in the [MediaTypeHeaderDelegate](https://github.com/quarkusio/quarkus/blob/main/independent-projects/resteasy-reactive/common/runtime/src/main/java/org/jboss/resteasy/reactive/common/headers/MediaTypeHeaderDelegate.java

) used in resteasy-reactive project used by quarkus.
See: https://gist.github.com/thomasdarimont/cd81991674357220510bbb732d5f517f

jamezp:
@thomasdarimont Yes, some, (a lot?), of resteasy-reactive was pulled from RESTEasy.

jamezp (approved):

PR #3808, opened 2023-10-01
[RESTEASY-3387] Use java.util.Locale to parse the languages from the Accept-Language header.

nderwin:
Updated the unit tests to check the header with case insensitivity.

This is a fix for https://issues.redhat.com/projects/RESTEASY/issues/RESTEASY-3387. 

Signed-off-by:Nathan Erwin <nathan.d.erwin@gmail.com>

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/util/LocaleHelper.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/util/LocaleHelper.java
@@ -11,14 +11,7 @@ public class LocaleHelper {
         int q = lang.indexOf(';');
         if (q > -1)
             lang = lang.substring(0, q);
-        String[] split = lang.trim().split("-");
-        if (split.length == 1)
-            return new Locale(split[0].toLowerCase());
-        else if (split.length == 2)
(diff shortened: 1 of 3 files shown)


nderwin:
If this is accepted, would someone mind putting the hacktoberfest-accepted label on this PR?

jamezp:
I think overall this makes sense, but I need to think about it as it does introduce a behavior change.

jamezp:
I've converted this to a draft for now as it causes the TCK to fail. I need to look at the spec to see what it says about this.

nderwin:
@jamezp Any updates with respect to the spec?

jamezp:
@nderwin Sorry, I have not. I know it fails the TCK though :) I'm hoping to review the PR queue this week or next though.

jamezp:
Sorry for the extremely long delay on this @nderwin. This seems to be passing the TCK so I'm not sure what the previous issue was.

Do you agree to these terms?

nderwin:
@jamezp Yes, I agree to the terms.

The previous issue was that the parsing of the locale tag was not taking into account the more advanced forms with extensions, private use tags, etc.  If the TCK has been updated to account for that too, that's awesome.

PR #3812, opened 2023-10-03
[RESTEASY-2212] Provide getUri() of client proxy method

Dkafetzis:
https://issues.redhat.com/browse/RESTEASY-2212


jamezp:
IMO we don't want to do this. I"m going to reject the JIRA as there are other ways around this as mentioned in the JIRA.

PR #3835, opened 2023-10-10
[RESTEASY-3399] Require a minimum of Java 21.

jamezp:
https://issues.redhat.com/browse/RESTEASY-3399


jamezp:
Converting this to a draft as Temurin 21 is not available on [Windows](https://github.com/adoptium/temurin/issues/8) yet. The builds and processing are still running.

jamezp:
This also needs to hold off until we have a wildfly-runtime image for Java 21.

PR #3852, opened 2023-10-16
add support for returning vertx future type

craig-day:
Adds support for a MP rest client implementation to return a Vert.x `Future`.

### DIFF
--- a/resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/internal/VertxFutureRxInvokerImpl.java
+++ b/resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/internal/VertxFutureRxInvokerImpl.java
@@ -1,0 +1,182 @@
+package org.jboss.resteasy.client.jaxrs.internal;
+
+import jakarta.ws.rs.HttpMethod;
+import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.client.SyncInvoker;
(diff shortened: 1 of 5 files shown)


jamezp:
We will end up needing a JIRA for this, but the concept makes sense to me.

resteasy-client-vertx/src/main/java/org/jboss/resteasy/client/jaxrs/internal/VertxFutureRxInvokerImpl.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Override
+    public Future<Response> method(String name) {
+        Context vertxContext = Vertx.currentContext();
+        if (vertxContext != null) {
+            return Future.fromCompletionStage(createClientInvocation(name, null).submitCF(), vertxContext);
+        } else {
+            return Future.fromCompletionStage(createClientInvocation(name, null).submitCF());
+        }
```
craig-day, line 130: @vietj I'm curious if you have thoughts on this implementation. Do you think it's good enough to just wrap the completion stage in a context-aware way, or is it worth rebuilding some of `ClientInvocation` to leverage things like `executeBlocking` or maybe even try to stay on the event loop thread?

craig-day:
Closing as outdated

jamezp:
I'm sorry that this completely fell off my radar. I am however working on updating resteasy-vertx now.

PR #3895, opened 2023-10-30
Only test main with WildFly 30. Limit the number of jobs ran as much …

jamezp:
…as possible. Additional updates to Java versions used.


jamezp:
It's a little odd merging this before #3894. However, I'd like to test that the PR triggers the cloud jobs.

PR #3927, opened 2023-11-28
[RESTEASY-3357] Endpoints returned a status code of 500 if Accept header had badly formatted value

khermano:
Issue: https://issues.redhat.com/browse/RESTEASY-3357

### DIFF
--- a/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
+++ b/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
@@ -201,6 +201,11 @@ public class ServletContainerDispatcher {
             try {
                 headers = ServletUtil.extractHttpHeaders(request);
                 uriInfo = ServletUtil.extractUriInfo(request, servletMappingPrefix);
+            } catch (IllegalArgumentException e) {
+                response.sendError(HttpServletResponse.SC_NOT_ACCEPTABLE);
+                // made it warn so that people can filter this.
+                LogMessages.LOGGER.failedToParseRequest(e);
+                return;
             } catch (Exception e) {
                 response.sendError(HttpServletResponse.SC_BAD_REQUEST);
                 // made it warn so that people can filter this.
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/exception/ClientErrorTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/client/exception/ClientErrorTest.java
@@ -199,7 +199,7 @@ public class ClientErrorTest {
         Response response = null;
         try {
             response = builder.get();
-            Assertions.assertEquals(HttpServletResponse.SC_BAD_REQUEST, response.getStatus());
+            Assertions.assertEquals(HttpServletResponse.SC_NOT_ACCEPTABLE, response.getStatus());
         } catch (Exception e) {
             throw new RuntimeException(e);
         } finally {
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/InvalidMediaTypeTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/core/basic/InvalidMediaTypeTest.java
@@ -50,6 +50,9 @@ public class InvalidMediaTypeTest {
         ResteasyClient client = (ResteasyClient) ClientBuilder.newClient();
         Invocation.Builder request = client.target(generateURL("/test")).request();
 
+        // Missing slash
+        doTest(request, "invalid");
+
         // Missing type or subtype
         doTest(request, "/");
         doTest(request, "/*");
@@ -82,7 +85,7 @@ public class InvalidMediaTypeTest {
         Response response = request.get();
         logger.info("mediaType: " + mediaType + "");
         logger.info("status: " + response.getStatus());
-        Assertions.assertEquals(HttpResponseCodes.SC_BAD_REQUEST, response.getStatus());
+        Assertions.assertEquals(HttpResponseCodes.SC_NOT_ACCEPTABLE, response.getStatus());
         response.close();
     }
 }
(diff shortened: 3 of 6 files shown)


resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
```diff
@@ -87,7 +88,7 @@ private static MediaType internalParse(String type) {
                 major = major.substring(0, paramIndex);
             }
             if (!MediaType.MEDIA_TYPE_WILDCARD.equals(major)) {
-                throw new IllegalArgumentException(Messages.MESSAGES.failureParsingMediaType(type));
+                throw new NotAcceptableException(Messages.MESSAGES.failureParsingMediaType(type));
```
jamezp, line 91: This isn't the correct place for this to be thrown. The java doc for `HeaderDelegate` states:
```quote
[IllegalArgumentException](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/IllegalArgumentException.html) - if the supplied string cannot be parsed or is null.
```
Reference https://jakarta.ee/specifications/restful-ws/3.1/apidocs/jakarta.ws.rs/jakarta/ws/rs/ext/runtimedelegate.headerdelegate

jamezp (changes requested):
We need to figure out where the parsing exception is happening and possibly catch the error there.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/delegates/MediaTypeHeaderDelegate.java
khermano: Hello @jamezp , thank you very much for the review.
I looked into it, and the place where the MediaTypeHeaderDelegate.fromString() is called is the MediaType.valueOf() method, which is from an external library jakarta.ws.rs-api.
Can I ask? What do you think I should do in such a case? Is this still an issue regarding the RESTEASY project?

jamezp: I think what we need to do is create a reproducer. The original issue is against RESTEasy Reactive which is part of Quarkus. Chunks of RESTEasy Reactive were pulled from RESTEasy.  I believe this is how it was fixed in Quarkus, https://github.com/quarkusio/quarkus/pull/34874. However, we need to attempt to reproduce it in RESTEasy, likely with WildFly, and then we can track down the best place to fix it.

khermano: Thank you very much for the suggestion @jamezp ! :)
So, I created a reproducer with quickstart/spring-resteasy. I uploaded war into WildFly and used `curl -i -H "Accept: invalid" -X GET http://localhost:8080/spring-resteasy/basic`. From the server log https://pastebin.com/XFjCAn5Q, I discovered that MediaTypeHeaderDelegate.parse() could be the method for applying changes. I force-pushed new changes, but I had to change some tests. What do you think about that?

khermano: I don't know what I thought when the parse method provided output for the fromString method. I apologize.
Yes sure. I'll send you the war via GChat because it can't be uploaded here. 
I uploaded the war to WildFly and then used `curl -i -H "Accept: invalid" -X GET [http://localhost:8080/spring-resteasy/basic`](http://localhost:8080/spring-resteasy/basic). Hopefully, it is the right reproducer for the issue.
But I was so fixed on details that I only now realized that the endpoint is not returning 500 as issue description claims, but 400. ServletContainerDispatcher catches the IllegalArgmuentException.service here: https://github.com/khermano/resteasy/blob/main/resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java#L205, so I am not sure if the issue is valid.

khermano: Hello @jamezp ,
I force-pushed some changes, but as I mentioned in the previous comment, I think the issue is no longer valid because I discovered that using the "invalid" Accept header returns 400 even before any change is added.
What do you think about that?

khermano:
Hello @jamezp, I forced-pushed some changes. As I described in the JIRA comment, I am not sure if they are necessary, so if you have time, could you please look at this and let me know what you think about it? Thank you very much.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
```diff
@@ -201,6 +201,11 @@ public void service(String httpMethod, HttpServletRequest request, HttpServletRe
             try {
                 headers = ServletUtil.extractHttpHeaders(request);
                 uriInfo = ServletUtil.extractUriInfo(request, servletMappingPrefix);
+            } catch (IllegalArgumentException e) {
+                response.sendError(HttpServletResponse.SC_NOT_ACCEPTABLE);
+                // made it warn so that people can filter this.
+                LogMessages.LOGGER.failedToParseRequest(e);
+                return;
```
jamezp, line 208: We can't do this here as there could be other reasons an `IllegalArgumentException` is thrown.

jamezp (changes requested):
FWIW I actually think we could close the JIRA too. The 400 response seems correct to me.

resteasy-core/src/main/java/org/jboss/resteasy/plugins/server/servlet/ServletContainerDispatcher.java
khermano: Thank you very much for the feedback. I will close the PR and the issue as Obsolete.

jamezp: Thank you for all the research on this @khermano!

khermano: You're welcome. Thank you for all your help and information.

PR #3930, opened 2023-11-29
[RESTEASY-3380 ]Introduce an ExceptionMapper which handles JsonProces…

jamezp:
…singException's, logs the error then returns an innocuous response.

https://issues.redhat.com/browse/RESTEASY-3380

Upstream #3928


jamezp:
I'm not sure why these jobs are hanging, but it doesn't have to do with this PR. It's passing for me locally.

PR #3931, opened 2023-11-30
[RESTEASY-3380] Introduce an ExceptionMapper which handles JsonProces…

jamezp:
…singException's, logs the error then returns an innocuous response.

https://issues.redhat.com/browse/RESTEASY-3380

Upstream: #3928


jamezp:
The failure is due to file names being too long for Windows. Nothing to be concerned about for now as this is an unsupported branch.

PR #3966, opened 2023-12-31
Upgrade `dep.arquillian-bom.version` to `1.8.0.Final`

liweinan:


jamezp:
This is failing and we'll need to look at why.

liweinan:
@jamezp Yeah it seems the upgrade of Arquillian is not backward compatible. I'll check it.

jamezp:
I'm closing this for now as we will need a wildfly-arquillian release and upgrade before we should do this.

PR #4009, opened 2024-02-02
[RESTEASY-3443] In the JsonProcessingExceptionMapper currently return…

jamezp:
…s a 500 status. However, it's more correct to return a 400 as it's a processing error. This is not backwards compatible. It can easily be overridden by creating a custom ExceptionMapper.

https://issues.redhat.com/browse/RESTEASY-3443


liweinan:
looks good.

PR #4015, opened 2024-02-06
[RESTEASY-3443] In the JsonProcessingExceptionMapper currently return…

jamezp:
…s a 500 status. However, it's more correct to return a 400 as it's a processing error. This is not backwards compatible. It can easily be overridden by creating a custom ExceptionMapper.

https://issues.redhat.com/browse/RESTEASY-3443

Upstream: #4009


steinsag:
@jamezp Can we please trigger a 6.2.8 release now?

jamezp:
@steinsag I'm just back from PTO today. I'll try to get a release out this week or early next week.

jamezp:
@steinsag I've had a few things pop up stopping me from releasing. I will definitely do a release early next week and if a few things miss the window, they miss the window. I'm sorry for the delay.

PR #4021, opened 2024-02-08
[RESTEASY-3432] Add a channel for overriding the version of RESTEasy …

jamezp:
…provisioned in WildFly.

https://issues.redhat.com/browse/RESTEASY-3432


wildfly/resteasy-manifest/pom.xml
```diff
@@ -0,0 +1,269 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-jaxrs-all</artifactId>
+        <version>7.0.0.Alpha2-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>resteasy-manifest-7.0</artifactId>
```
liweinan, line 32: Is this version in the `artifactId` mandatory for the manifest to work? I mean for each release, do we need to maintain a different version of manifest like this?

liweinan:
Cool!

testsuite/integration-tests/pom.xml
```diff
@@ -88,13 +88,16 @@
...
                                         </excluded-packages>
                                     </feature-pack>
-                                    <feature-pack>
-                                        <groupId>${resteasy.test.feature.pack.groupId}</groupId>
-                                        <artifactId>${resteasy.test.feature.pack.artifactId}</artifactId>
-                                        <version>${resteasy.test.feature.pack.version}</version>
-                                        <inherit-configs>false</inherit-configs>
-                                    </feature-pack>
                                 </feature-packs>
+                                <channels>
```
spyrkob, line 92: If you're able to use newer WildFly (>= 32.0.0.Final) it might be worth adding another channel with the WildFly manifest (`org.wildfly.channels:wildfly-ee`). Server provisioning should be much faster this way.

wildfly/resteasy-manifest/pom.xml
spyrkob: The version makes it possible to create a separate manifest stream for different releases. This way channels can detect the latest manifest in each stream to e.g. perform update.

I wonder if it would be possible to create a property in top level pom and use it in other poms so it's easier to manage.

wildfly/resteasy-manifest/pom.xml
```diff
@@ -0,0 +1,269 @@
...
+    <name>RESTEasy: Channel Manifest</name>
+    <packaging>pom</packaging>
+
+    <properties>
+        <generated.manifest.file>${project.build.directory}/${project.artifactId}-${project.version}-manifest.yaml</generated.manifest.file>
```
spyrkob, line 37: Is this property used anywhere?

jamezp: Oops, no. Nice catch. I had it when I was drying to dynamically generate the artifact ID which I ended up not doing because there were too many possible holes in trying to cheat Maven :)

wildfly/resteasy-manifest/pom.xml
jamezp: @spyrkob I had it working like that, e.g. a dynamic `artifactId`. The problem was you have to override the `maven-deploy-plugin` and the `maven-install-plugin` to skip the default `deploy` and `install` goals. You then need an execution for the `deploy-file` and `install-file` goals. While those seem to work, they seem fragile and could break other plugins.

One option I thought of is to keep the artifact ID generic, e.g. `resteasy-manifest` and create an additional manifest for the release stream and have the one with the version left off always be the latest.

jamezp:
I've moved this back to a draft as I'm going to try the dynamic approach of the manifest and may end up changing the groupId to something like `dev.resteasy.channel`.

jamezp:
The current cloud-tests failures is likely a real issue, see https://wildfly.zulipchat.com/#narrow/stream/184003-galleon/topic/Feature.20Pack.20location.20vs.20GAV.

docbook/reference/en/en-US/modules/Installation_Configuration.xml
```diff
@@ -247,48 +247,44 @@
 	        <title>Upgrading RESTEasy within WildFly</title>
 	        <para>
                 RESTEasy is bundled with WildFly. However you may wish to upgrade to the latest version. With
-                <ulink url="https://docs.wildfly.org/24/Galleon_Guide.html">Galleon</ulink> this makes upgrading RESTEasy
+                <ulink url="https://docs.wildfly.org/31/Galleon_Guide.html">Galleon</ulink> this makes upgrading RESTEasy
```
TomasHofman, line 250: Here the Wildfly version doesn't match what used father in the code - 31 vs 32.

PR #4027, opened 2024-02-14
[RESTEASY-3388] code changes supporting junit5.

rsearls:
This is the 1st of several (future) PRs in migrating the testsuite
to junit5.  Each module in testsuite will have a separate PR.  This
PR MUST be committed 1st.  The other modules will rely on changes 
made to module, arquillian-utils.  testsuite/pom.xml has been edited
to disable all other testsuite modules except this one, so that
the project will compile and run successfully.

### DIFF
--- a/testsuite/arquillian-utils/pom.xml
+++ b/testsuite/arquillian-utils/pom.xml
@@ -60,12 +60,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-
-        <dependency>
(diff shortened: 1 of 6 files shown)


jamezp:
I don't think we can commit this first as it disables all testing. We can have more than one commit in the PR, but I don't think we should commit this with disabled testing.

jamezp:
Included in #4028

PR #4028, opened 2024-02-14
Resteasy 3388 junit5 migration integration tests

rsearls:
This is the 2nd PR in migrating the testsuite to junit5.  This PR requires 
https://github.com/resteasy/resteasy/pull/4027 be merged 1st as it relies 
on junit5 changes in arquillian-utils. testsuite/pom.xml has been edited
to disable all other testsuite modules except arquillian-utils and
integration-tests, so that the project will compile and run successfully.


testsuite/pom.xml
```diff
@@ -149,7 +149,9 @@
             <modules>
                 <module>test-utils</module>
                 <module>arquillian-utils</module>
+                <!-- rls tmp disable (start)
```
marekkopecky, line 152: @jamezp Is there some plans about missing unit-tests module?

asoldano: @marekkopecky see https://github.com/resteasy/resteasy/pull/4033

marekkopecky: Thanks!

PR #4043, opened 2024-02-22
[RESTEASY-3449] Allow the SizeUnitTestCase to pass in all locales.

jamezp:
https://issues.redhat.com/browse/RESTEASY-3449

Upstream #4042


christf:
That was quick, thank you.

PR #4060, opened 2024-03-04
6.2.4.final  bootable jar

tommaso-borgato:
This PR is to be changed because it has to against some branch derived from the 6.2.4.Final tag (e.g. 6.2.4.SPx)

This PR:
-  reintroduces bootable jar tests
- adds option to run bootable jar tests with a channel manifest

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
     <groupId>org.jboss.resteasy</groupId>
     <artifactId>resteasy-jaxrs-all</artifactId>
-    <version>6.2.4.Final</version>
+    <version>6.2.4.Final--bootable-jar</version>
     <packaging>pom</packaging>
 
     <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -21,12 +21,12 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <!-- Dependency versions -->
-        <dep.arquillian-bom.version>1.7.0.Alpha14</dep.arquillian-bom.version>
+        <dep.arquillian-bom.version>1.7.2.Final</dep.arquillian-bom.version>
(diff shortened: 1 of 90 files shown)


pom.xml
```diff
@@ -53,6 +53,7 @@
         <version.org.jboss.galleon>5.1.0.Final</version.org.jboss.galleon>
         <version.org.wildfly.galleon-plugins>6.4.2.Final</version.org.wildfly.galleon-plugins>
         <version.org.wildfly.plugins.wildfly-maven-plugin>4.0.0.Final</version.org.wildfly.plugins.wildfly-maven-plugin>
```
fabiobrz, line 55: I guess this should be bumped as well.

pom.xml
```diff
@@ -53,6 +53,7 @@
         <version.org.jboss.galleon>5.1.0.Final</version.org.jboss.galleon>
         <version.org.wildfly.galleon-plugins>6.4.2.Final</version.org.wildfly.galleon-plugins>
```
fabiobrz, line 54: And maybe these Galleons dependency will need to be updated according to those of the wildfly-maven-plugin and wildfly-jar-maven-plugin? Not sure though, but I think this should be worth checking.

fabiobrz (changes requested):
Hi @tommaso-borgato - I went through the changes and dropped a couple of minor comments.
BTW - Looking at the original [PR that introduced the changes originally](https://github.com/resteasy/resteasy/pull/2558) I noticed you've not documented the changes in the README, and IIUC also the filtered tests seem to miss some comments about the reason for the exclusion. WDYT?

tommaso-borgato:
We are going to use https://github.com/tommaso-borgato/Resteasy/tree/6.2.4.Final--bootable-jar

tommaso-borgato:
We are going to use https://github.com/tommaso-borgato/Resteasy/tree/6.2.4.Final--bootable-jar

PR #4080, opened 2024-03-14
Fix usage of defaultProxy

steffen19w23:
Set this.defaultProxy before createDefaultHttpClient, cf. #6167e1534f6e2379cf0ad1074b34f87d13045e84


steffen19w23:
In case somebody asks: contribution intentionally submitted

jamezp:
Sorry, looks like I somehow missed this for 2 months now. While this looks reasonable to me, did you see an error because of this?

steffen19w23:
Haven't see an error because proxy was no use-case for me, but setting `this.defaultProxy` after `createDefaultHttpClient()` seems to be an obvious bug

jamezp (approved):

PR #4091, opened 2024-03-19
[RESTEASY-3463] Create an API for getting the EntityPart's from the m…

jamezp:
…ultipart project rather than the HttpServletRequest.getParts().

https://issues.redhat.com/browse/RESTEASY-3463

### DIFF
--- a/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/EntityPartFilter.java
+++ b/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/EntityPartFilter.java
@@ -1,0 +1,221 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2024 Red Hat, Inc., and individual contributors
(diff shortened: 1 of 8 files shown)


providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/EntityPartFilter.java
```diff
@@ -0,0 +1,198 @@
...
+import org.jboss.resteasy.plugins.server.Cleanables;
+import org.jboss.resteasy.spi.EntityOutputStream;
+import org.jboss.resteasy.spi.multipart.MultipartContent;
+import org.jboss.resteasy.spi.util.Types;
+
+/**
+ * @author <a href="mailto:jperkins@redhat.com">James R. Perkins</a>
+ */
+@Provider
+@Consumes(MediaType.MULTIPART_FORM_DATA)
```
jamezp, line 49: I forgot to read the docs to see if this even works. I will do that before this is merged, assuming we want to go with this approach.

PR #4092, opened 2024-03-19
[RESTEASY-3450] removed Hamcrest statements.  Replaced with junit5 As…

rsearls:
…sertions

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/ValidationXMLTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/ValidationXMLTest.java
@@ -14,8 +14,6 @@ import jakarta.ws.rs.client.Entity;
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
-import org.hamcrest.MatcherAssert;
-import org.hamcrest.Matchers;
 import org.jboss.arquillian.container.test.api.Deployment;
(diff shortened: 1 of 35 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/ValidationXMLTest.java
```diff
@@ -438,4 +432,15 @@ private static <T> Collection<String> resolveValues(final Collection<T> c, final
...
+    private static boolean hasItems(Collection<String> items, String... controlList) {
+        for (String c : items) {
+            for (String s : controlList) {
+                if (c.equals(s)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
jamezp, line 445: This doesn't look quite right as it will exit on the first match. This could probably just be:
```java
private static boolean hasItems(Collection<String> items, String... controlList) {
    return items.containsAll(List.of(controlList));
}
```

jamezp (changes requested):

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/validation/ValidationXMLTest.java
rsearls: good catch.  Fixed.

jamezp (approved):

PR #4107, opened 2024-03-23
RESTEASY-3469: Installed updated jacoco profile

ronsigal:


pom.xml
```diff
@@ -312,6 +312,40 @@
...
+                            </execution>
+                        </executions>
+                    </plugin>
+					<plugin>
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-surefire-plugin</artifactId>
+						<configuration>
+							<argLine>@{argLine}</argLine>
+						</configuration>
+					</plugin>
```
jamezp, line 345: Nit pick, but please change the tabs to spaces :)

jamezp (changes requested):
It doesn't have to happen in this PR, but we should look at removing the old jacoco configurations.

ronsigal:
Aaaarrrrrggggghhhhh :(

Of course. My bad.

PR #4119, opened 2024-04-01
[RESTEASY-3490] Client proxy: invoke default methods rather than failing with RESTEASY004530

stevenschlansker:
Hi RESTEasy devs,

We use the client proxy. When calling a default method, we get an error RESTEASY004530 that the method isn't recognized.

Is it possible to extend the proxy implementation to respect default methods?

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.client.jaxrs.internal.proxy;
 
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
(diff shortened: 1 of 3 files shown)


stevenschlansker:
Thank you very much! Please let me know of any necessary updates I should make :)

jamezp:
Yes, those are not related. I've never been able to figure out what the issue is there. At some point it should be prioritized :) But nothing for you to worry about.

jamezp (approved):
I think overall this looks good and makes sense. We will need to do a security manager check and likely handle some issues there. I can do this in a follow up PR though.

stevenschlansker:
Yes, I agree there will likely be followup changes, also JPMS can cause issues too. But since it 100% doesn't work now it isn't a regression :)

jamezp:
> Yes, I agree there will likely be followup changes, also JPMS can cause issues too. But since it 100% doesn't work now it isn't a regression :)

Yes and technically it's not in a JPMS module anyway. We'll have a lot to solve to make that happen :)

stevenschlansker:
Closed in favor of #4125

PR #4125, opened 2024-04-04
[RESTEASY-3490] Allow default methods on proxy clients to be used

jamezp:
https://issues.redhat.com/browse/RESTEASY-3490

@stevenschlansker does this seems okay to you?

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
@@ -1,5 +1,6 @@
 package org.jboss.resteasy.client.jaxrs.internal.proxy;
 
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.util.Map;
@@ -52,6 +53,11 @@ public class ClientProxy implements InvocationHandler {
                 return this.toString();
             } else if (method.getName().equals("as") && args.length == 1 && args[0] instanceof Class) {
                 return ProxyBuilder.proxy((Class<?>) args[0], target, config);
+            } else if (method.isDefault()) {
+                return MethodHandles.privateLookupIn(clazz, MethodHandles.lookup())
+                        .unreflectSpecial(method, clazz)
+                        .bindTo(o)
+                        .invokeWithArguments(args);
             }
         }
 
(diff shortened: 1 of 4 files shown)


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
```diff
@@ -52,6 +56,34 @@ public Object invoke(Object o, Method method, Object[] args)
...
+                            return MethodHandles.privateLookupIn(clazz, MethodHandles.lookup())
+                                    .unreflectSpecial(method, clazz)
+                                    .bindTo(o)
+                                    .invokeWithArguments(args);
+                        } catch (Throwable e) {
+                            throw new RuntimeException(e);
+                        }
+                    });
+                } catch (PrivilegedActionException e) {
+                    final Exception real = e.getException();
```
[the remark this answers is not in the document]
jamezp: Because I missed the part in the JavaDoc that said `getCaused()` is preferred :). Thank you @fjuma!

stevenschlansker (approved):
LGTM, thanks :)

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
```diff
@@ -52,6 +56,34 @@ public Object invoke(Object o, Method method, Object[] args)
...
+                            .invokeWithArguments(args);
+                }
+                try {
+                    return AccessController.doPrivileged((PrivilegedExceptionAction<Object>) () -> {
+                        try {
+                            return MethodHandles.privateLookupIn(clazz, MethodHandles.lookup())
+                                    .unreflectSpecial(method, clazz)
+                                    .bindTo(o)
+                                    .invokeWithArguments(args);
+                        } catch (Throwable e) {
```
stevenschlansker, line 76: It looks like this could catch something like `OutOfMemoryError`, `StackOverflowError` or `VirtualMachineError` and wrap it in `RuntimeException`. Should this only wrap Exception types and rethrow other error types as-is?

stevenschlansker: It would seemingly also wrap unchecked exception types like `jakarta.ws.rs.NotFoundException` which callers might expect to observe directly.

stevenschlansker:
If possible, we would appreciate a backport to 6.2

jamezp:
Thank you for the review. I'll double check the exception handling and I'll definitely backport this to 6.2 once we have a final version of it.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/proxy/ClientProxy.java
jamezp: Yes. It's likely better to only wrap `Exception` types and re-throw any `WebApplicationException`'s. Nice, catch, thank you!

PR #4142, opened 2024-04-19
[RESTEASY-3492] Upgrade WildFly Arquillian to 5.1.0.Beta2. Migrate te…

jamezp:
…st utilities to new WildFly Arquillian dependencies.

Also sneaked in a minor test commit to ensure @ManagedBean annotated tests work. Note this test may end up going as we upgrade the Jakarta Annotations API. However, we can move the test to WildFly when we do that.

https://issues.redhat.com/browse/RESTEASY-3492


jamezp:
A 6.2 PR is still required, but I wanted to run this through CI first. Note that when back poring we may want to consider keeping some of the deleted utilities. That said, those are not meant to be used outside of RESTEasy so breaking compatibility is not as big of a deal. The common WildFly Arquillian utilities should really be used.

PR #4250, opened 2024-07-31
Fix DynamicFeatureTest (TestUtil had wrong log folder for bootable jar)

tommaso-borgato:
When using bootable jar, the standalone directory is not identified correctly and test `DynamicFeatureTest` fails because it cannot find the server.log file


jamezp:
@tommaso-borgato Can you show me how to reproduce the issue? It seems to be working for me.

PR #4282, opened 2024-08-16
[RESTEASY-3532] Ensure there is a boundary set for headers on the Jetty Client Engine.

jamezp:
https://issues.redhat.com/browse/RESTEASY-3532

### DIFF
--- a/resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
+++ b/resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
@@ -2,8 +2,12 @@ package org.jboss.resteasy.client.jaxrs.engines.jetty;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.time.Duration;
+import java.util.List;
 import java.util.Objects;
+import java.util.UUID;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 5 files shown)


resteasy-client-jetty/src/main/java/org/jboss/resteasy/client/jaxrs/engines/jetty/JettyClientEngine.java
```diff
@@ -248,6 +288,43 @@ private static RuntimeException clientException(Throwable ex, jakarta.ws.rs.core
...
+            }
+            return cl == null ? ClassLoader.getSystemClassLoader() : cl;
+        });
+    }
+
+    private static boolean isMultipart(final ClientInvocation invocation) {
+        return MULTIPART_WILDCARD.isCompatible(invocation.getHeaders().getMediaType());
+    }
+
+    private static boolean canSetBoundary(final Object entity) {
```
mqchen-ws, line 312: should this return something like `Optional<String>` as the boundary, provided by `MultipartOutput.getBoundary()`, so it can be set on the header instead of a random UUID on line 146? Otherwise wouldn't we end up with a different boundary in the header vs. the body?

I'm not totally sure how `EntityPart` writes the boundary. Not sure if this is in the path but I do see lists of `EntityPart`  being converted into `MultipartFormDataOutput` in `MultipartEntityPartWriter#create`. If so then the boundary uuid would get assigned as a random uuid there when `new MultipartFormDataOutput()` is called, and that would also need to match this header?

jamezp: The `AbstractMultipartWriter` [checks the header](https://github.com/resteasy/resteasy/blob/2f82f7db7df5ac5dc654d5c8785224ded854092e/providers/multipart/src/main/java/org/jboss/resteasy/plugins/providers/multipart/AbstractMultipartWriter.java#L52-L54) before it gets the boundary from the `MultipartOutput`.

The `Entitypart` does end up using the `AbstractMultipartFormDataWriter`.

IoW, the `MultipartOutput.getBoundary()` shouldn't be used at all if the header has been set.

stevenschlansker:
Thanks for the fix! :)

stevenschlansker:
I'm sure you thought of this, but it sucks that the engine / entity interface requires you to "peek through" the abstraction boundary. But I'm sure fixing that is well outside the scope of this fix.

jamezp:
> I'm sure you thought of this, but it sucks that the engine / entity interface requires you to "peek through" the abstraction boundary. But I'm sure fixing that is well outside the scope of this fix.

Yeah. I wanted to keep it simple for now. I had some reflection calls to avoid adding the multipart provider dependency on the client. This just seemed like the simplest solution for now.

I've also been really tempted to create a new implementation of the multipart provider that doesn't use mime4j.

PR #4302, opened 2024-08-30
[RESTEASY-3538] Update the way the media type for a response is deter…

jamezp:
…mined. Allow the previous negotiation to work with a configuration property.

https://issues.redhat.com/browse/RESTEASY-3538


jamezp:
I think this might need to be reversed and opt-in vs opt-out. Though it's closer to what the specification indicates should be happening, it's a rather large change in the way the media type is resolved for RESTEasy.

PR #4375, opened 2024-11-06
[RESTEASY-2567] Added new test coverage for JSON merge patch

honza-kasik:
Mainly negative test cases, see test documentation for details.

See https://issues.redhat.com/browse/RESTEASY-2567

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
@@ -1,0 +1,228 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ *
+ * Copyright 2024 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.resteasy.test.resource.patch;
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.ws.rs.HttpMethod;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.client.ClientBuilder;
+import jakarta.ws.rs.client.Entity;
+import jakarta.ws.rs.client.WebTarget;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+
+import org.jboss.arquillian.container.test.api.Deployer;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.OperateOnDeployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit5.ArquillianExtension;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.logging.Logger;
+import org.jboss.resteasy.utils.PortProviderUtil;
+import org.jboss.resteasy.utils.TestUtil;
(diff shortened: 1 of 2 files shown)


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
jamezp, line 1: This needs a copyright header.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
```diff
@@ -0,0 +1,198 @@
...
+    @AfterEach
+    public void afterEach() {
+        deployer.undeploy(PATCH_DEPLOYMENT);
+    }
+
+    @Deployment(managed = false, name = PATCH_DEPLOYMENT)
+    public static Archive<?> createDeployment() {
+        WebArchive war = TestUtil.prepareArchive(PATCH_DEPLOYMENT);
+        return TestUtil.finishContainerPrepare(war, null, StudentResource.class, Student.class);
+    }
```
jamezp, line 78: Why are we not using a managed deployment?

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
```diff
@@ -0,0 +1,198 @@
...
+public class StudentJsonMergePatchTest {
+
+    private static final Logger log = Logger.getLogger(StudentJsonMergePatchTest.class);
+
+    @ArquillianResource
+    private Deployer deployer;
+
+    static Client client;
+    static final String PATCH_DEPLOYMENT = "Patch";
+    static final Student testStudentEntity = new Student()
```
jamezp, line 47: Minor nit, there should be no reason for these not to be `private`.

jamezp (changes requested):

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
honza-kasik: The deployment is using a `Map` to store entities. To make tests as independent as possible we need the `Map` to be empty for each test. So either we have to have a deployment for each test or reuse deployment with refreshed state. To achieve the initial state with empty `Map` I decided to use non-managed deployment with redeployment in `each` JUnit lifecycle.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
honza-kasik: Thanks, fixed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
jamezp: Hmm... ...okay. Not suggesting we need to do this, but we could add a method to clear the map. I"m not sure if it's faster to redeploy or send a request to clear the map. 

This shouldn't block this PR though IMO.

honza-kasik: Thanks for the feedback. This solution in my opinion is the simplest available - we don't need to make additional checks whether the map is really empty after emptying it somehow and introduce any new methods for the deployment.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java
honza-kasik: Added, thanks

jamezp:
@honza-kasik It looks like the format check failed.
```
Error:  Failed to execute goal net.revelc.code:impsort-maven-plugin:1.11.0:check (check-import-sort) on project resteasy-integration-tests: Imports are not sorted in /home/runner/work/resteasy/resteasy/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/StudentJsonMergePatchTest.java -> [Help 1]
```

This can be resolved by simply running `mvn clean process-test-classes`.

Also, can you squash the commits into a single commit?

jamezp (approved):

jamezp:
Upstream #4383

PR #4410, opened 2024-11-17
[RESTEASY-3558] Drop MavenUtil and unused Aether dependencies, use ShrinkWrap Resolvers instead

jbliznak:
https://issues.redhat.com/browse/RESTEASY-3558


jamezp:
@jbliznak Yes they are. I'm going to look at those failures this morning. They're failing on the nightly jobs as well.

PR #4590, opened 2025-05-26
Added Support for POST APIs with SSE Client

Rajashri-KN:
The SSE Client previously only supported invoking GET APIs. Support for POST APIs has now been added.

### DIFF
--- a/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
+++ b/resteasy-client/src/main/java/org/jboss/resteasy/plugins/providers/sse/client/SseEventSourceImpl.java
@@ -50,8 +50,14 @@ import org.jboss.resteasy.resteasy_jaxrs.i18n.Messages;
 
 public class SseEventSourceImpl implements SseEventSource {
     public static final long RECONNECT_DEFAULT = 500;
+    private static final String HTTP_REQUEST_TYPE_GET = "GET";
+    private static final String HTTP_REQUEST_TYPE_POST = "POST";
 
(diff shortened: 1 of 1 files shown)


jamezp:
I assume this is wanted to work with langchain4j. It should already be possible to do this using the `org.jboss.resteasy.plugins.providers.sse.client.SseEventSourceImpl.open(String lastEventId, String verb, Entity<?> entity, MediaType... mediaTypes)` method.

Rajashri-KN:
You're right that org.jboss.resteasy.plugins.providers.sse.client.SseEventSourceImpl.open(String lastEventId, String verb, Entity<?> entity, MediaType... mediaTypes) allows specifying the HTTP verb, request body, and media types for SSE connections, which can be useful in advanced use cases.

However, this open method is not part of the standard SseEventSource interface and is intended for internal use within the implementation. The recommended approach is to configure the HTTP verb and request body using the SourceBuilder before building the event source. Calling this method directly from client code breaks encapsulation and bypasses the intended configuration flow.

To align with best practices, I updated the builder to set the request type and body, and then used the standard open() method to initiate the connection.

jamezp:
This builder is also not a standard API. It's a RESTEasy only API and you're still using the `org.jboss.resteasy.plugins.providers.sse.client.SseEventSourceImpl` to get to the builder.

What we really need to do is propose this to the Jakarta RESTful Web Services specification so the specification API would support this.

Rajashri-KN:
We are using the SseEventSource.build() method to create an instance of SseEventSourceImpl. I will close this PR, and we can propose the update in SseEventSource class instead.

PR #4617, opened 2025-07-22
RESTEASY-2962 - The ReverseInjectionTest.testMDB test does correct work

jasondlee:
https://issues.redhat.com/browse/RESTEASY-2962

- Decouple MDB, EJB, and Rest Resource
- Remove use of latch to determine when to read back from DB 
- Move "business logic" into the EJB and out of the Rest layer 
- Modify test to poll w/timeout for results from server

### DIFF
--- a/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/MDBInjectionTest.java
+++ b/testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/MDBInjectionTest.java
@@ -1,6 +1,11 @@
 package org.jboss.resteasy.test.cdi.injection;
 
 import java.net.URI;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Objects;
+import java.util.function.Consumer;
 
 import jakarta.annotation.Resource;
 import jakarta.ws.rs.client.Client;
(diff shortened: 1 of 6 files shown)


jamezp (dismissed):

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/MDBInjectionTest.java
```diff
@@ -121,19 +125,34 @@ public void testMDB() throws Exception {
...
+
+        while (Instant.now().isBefore(endTime) && !success) {
+            try (Response response = client.target(uri).request().get()) {
+                assertionConsumer.accept(response);
+                success = true;
+            } catch (AssertionError assertionError) {
+                log.debug("waitAndAssert(..) validation failed - retrying.");
+                Thread.onSpinWait();
+            }
+        }
```
jamezp, line 156: We need to add a line here something like:
```java
Assertions.assertTrue(success, "Did not get response within 30 seconds");
```

We probably need to catch the assertionError too and append the error message. Otherwise, we don't really know why it failed.

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/ReverseInjectionTest.java
```diff
@@ -279,54 +302,57 @@ public void testEJBHolderInResource() throws Exception {
...
+
+        while (Instant.now().isBefore(endTime) && !success) {
+            try (Response response = client.target(uri).request().get()) {
+                assertEquals(HttpResponseCodes.SC_OK, response.getStatus());
+                assertionConsumer.accept(response.readEntity(new GenericType<>(BookCollectionType)));
+                success = true;
+            } catch (AssertionError assertionError) {
+                log.debug("waitAndAssert(..) validation failed - retrying.");
+                Thread.onSpinWait();
             }
```
jamezp, line 356: Same as above here.

jamezp (changes requested):

testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/cdi/injection/MDBInjectionTest.java
jasondlee: Ah, right. I just realized I removed some pretty vital logic from the WildFly test I lifted this from. I've updated both methods. Hopefully that will address your concerns.

jamezp (approved):

jamezp:
Thank you @jasondlee

PR #4641, opened 2025-08-19
[RESTEASY-3617] avoiding blocking call in reactive flow

radhakrishnan10192:
ResteasyJackson2Provider initializes the default ObjectMapper from [this](https://github.com/resteasy/resteasy/pull/4325) change, which is initialized using `ObjectMapper.findAndRegisterModules()` this internally trying to load the classes using service loader, which is a blocking call. This is a blocking call in reactive flow which is flagged by Blockhound. 

This fix would make the initialization only once and reused without affecting the reactive flow.


radhakrishnan10192:
@jamezp I have tried this in local and it seems to be fixing the blocking execution in the reactor flow. WDYT about this?

jamezp:
@radhakrishnan10192 What does the issue seem to be? I see from the JIRA that it's blocking, but I'm not following why that would be an issue.

PR #4649, opened 2025-09-03
Module org.bouncycastle has been removed in WFLY-16576

tommaso-borgato:
Module org.bouncycastle has been removed in WFLY-16576: we need to replace its references with individual modules where needed


jamezp:
This looks reasonable to me. We'll need a JIRA other than just the WildFly one. I do think there is an issue with the module dependencies too as the user shouldn't have to add these ,but that's a different issue.

A little OT, but I've also filed https://issues.redhat.com/browse/RESTEASY-3431 to get rid of BouncyCastle all together if we can. I think that would be really useful.

PR #4650, opened 2025-09-04
Bootable jar tests

tommaso-borgato:
These tests are already excluded in the regular testing pipeline; this way we can avoid messing with the surefire.excludes parameter


tommaso-borgato:
Replaced by https://github.com/resteasy/resteasy/pull/4651

PR #4652, opened 2025-09-04
[6.2] Exclude additional tests in resteasy-integration-tests

tommaso-borgato:
*** back-port of https://github.com/resteasy/resteasy/pull/4651 to branch 6.2 ***

In other CI environments other than Github, we need to be able to exclude additional tests without overriding existing excludes;


tommaso-borgato:
@jamezp looks like checks do not start here ... is that expected?

tommaso-borgato:
@jamezp I see there are errors in resteasy-vertx ... they seem unrelated to this PR

PR #4701, opened 2025-10-07
Draft: Save double byte[] copies on Jackson

franz1981:
Jackson usually configure different form of buffer pooling (see https://github.com/FasterXML/jackson-core/blob/e19615bb99f6127230793e2c7766bbb3beaee6fc/src/main/java/com/fasterxml/jackson/core/io/IOContext.java#L63) and perform a single `write(byte[] b, int offset, int lengh)` vs the provided output stream during serialization, using its internally pooled buffer.

In short, this enable an async write to:
- expect a single encoded byte[] 
- size the receiver output stream exactly to it
- save the additional copy for the async completion to happen

image

This is just an idea as the subsequent `AsyncOutputStream ::asyncWrite` can "backfire" performing an additional copy, see 3408cf1f148f4b065e3179b156b877297e6edaf5


jamezp:
@franz1981 Is this something we should do more generally for the {{org.jboss.resteasy.spi.AsyncMessageBodyWriter}}?

In general looking at that, writing everything to memory before writing it to the async stream like it could be a potential issue for large objects.

franz1981:
In Quarkus I have implemented a thing called AppendBuffer in the vertx output stream, for Restleasy reactive, which:
- allow appending encoded chunks in batches
- once the buffer capacity (sum of all chunks) is exhausted, it flush it to http as chunked response, and keep on filling it with the rest of the stream

In this way you both have batchy flush and some predictable memory usage.

This specific Jackson writer here looks to be just in memory, I can tell; which as you said, could be dangerous under heavy load due to memory footprint: I've indeed found this improvement because due to the existing implementation, it was so memory hungry to appear in our profiling data in the performance lab

franz1981:
Good point @jamezp yep, in theory should work similarly

jamezp:
Perfect. I may copy it over and look at the other chunking separately.

Also, thank for you this. I really want to learn more about performance testing, but it seems like an art form that there either isn't a lot of information on or I'm not able to find it :)

franz1981:
Reach out to me and my team and I can help with it @jamezp no worries :)

PR #4877, opened 2026-06-02
Added support of Record type

neenapj:


resteasy-core/src/main/java/org/jboss/resteasy/core/FormInjector.java
```diff
@@ -20,22 +21,130 @@ public class FormInjector implements ValueInjector {
...
 
+        System.out.println("  No injectable constructor found, will use property injection");
+        return null; // No injectable constructor found
+    }
+
+    /**
+     * Check if the class is a Record (Java 16+).
+     * Uses reflection to avoid compilation issues on older Java versions.
+     */
+    private boolean isRecord(Class<?> clazz) {
```
jim-krueger: I think you can use the following to check the java version rather than reflection:
`if (System.getProperty("java.version") >= 16)`

resteasy-core/src/main/java/org/jboss/resteasy/core/FormInjector.java
```diff
@@ -20,22 +21,130 @@ public class FormInjector implements ValueInjector {
...
         this.type = type;
         Constructor<?> constructor = null;
 
-        try {
-            constructor = type.getDeclaredConstructor();
-            constructor.setAccessible(true);
-        } catch (NoSuchMethodException e) {
-            throw new RuntimeException(Messages.MESSAGES.unableToInstantiateForm());
+        // Debug: Log class being processed
+        System.out.println("=== FormInjector for: " + type.getName() + " ===");
```
jim-krueger: Rather than using System.out.println() you should be using RESTEasyTracingLogger.  You can see examples in other classes within org.jboss.resteasy.core such as SynchronousDispatcher.java

jim-krueger (changes requested):

resteasy-core/src/main/java/org/jboss/resteasy/core/FormInjector.java
jamezp: We'll always be greater than 21 because that will end up being the minimum requirement.

That said, I don't think we need to use reflection here at all. Even if a class is compiled to a lower level, the method will still exist at runtime.

resteasy-core/src/main/java/org/jboss/resteasy/core/FormInjector.java
jamezp: I assumed these are just for testing purposes, but they'll definitely need to be removed before this can be merged. If we want debug logging, we should use `LogMessages.LOGGER.debuf("FormInjector for: %s", type.getName())`.

resteasy-core/src/main/java/org/jboss/resteasy/core/FormInjector.java
neenapj: Done

resteasy-core/src/main/java/org/jboss/resteasy/core/FormInjector.java
neenapj: Done

PR #4883, opened 2026-06-08
3732 jsonmergepatch issue

neenapj:


testsuite/integration-tests/src/test/java/org/jboss/resteasy/test/resource/patch/JsonMergePatchParameterTest.java
```diff
@@ -0,0 +1,217 @@
...
+                .add("lastName", "Smith")
+                .add("school", "MIT")
+                .build();
+
+        // Apply patch using JsonMergePatch parameter
+        Response response = patchStudentWithJsonMergePatch(id, patchJson.toString(), DEPLOYMENT_NAME);
+
+        // Print error details if not successful
+        if (response.getStatus() != 200) {
+            System.out.println("Response status: " + response.getStatus());
```
jim-krueger: I think you need to use `java.util.logging.Logger` rather than `System.out.println`.  See other test files for examples.

jim-krueger (changes requested):
One change suggestion for sure.  I may have more later.

PR #4909, opened 2026-06-30
Replace wildfly jar maven plugin

tommaso-borgato:


tommaso-borgato:
replaced by https://github.com/resteasy/resteasy/pull/4917

PR #4933, opened 2026-07-22
[RESTEASY-3754] & [RESTEASY-3755] Update CDI support

jamezp:
### [RESTEASY-3755](https://redhat.atlassian.net/browse/RESTEASY-3755)

Enable injection of @*Param annotated parameters with CDI. Ensure invalid CDI resource beans are ignored by the CDI extension and processed by RESTEasy.

#### CDI Integration Changes

A new CDI extension, `ResteasyParamCdiExtension`, registers Jakarta REST parameter annotations (`@CookieParam`, `@FormParam`, `@HeaderParam`, `@MatrixParam`, `@PathParam`, `@QueryParam`, and `@BeanParam`) as CDI qualifiers. This allows Jakarta REST resources to receive parameter values via CDI injection — including constructor injection — without requiring a public no-arg constructor (provided
the class is still proxyable).

For any proxyable Jakarta REST resource or provider:
- `@Inject` is implicitly added to `@Context`-annotated fields, setter methods, and constructors (`ResteasyCdiExtension`)
- `@Inject` is implicitly added to `@*Param`-annotated fields, setter methods, and constructors (`ResteasyParamCdiExtension`)
- Synthetic CDI beans are registered to produce parameter values at request time by delegating to `ParameterExtractors`

Resources that are not proxyable (final class, final methods, or no non-private no-arg constructor) are excluded from CDI management and handled by RESTEasy directly. This resolves the deployment regression in RESTEASY-3752 where resources with only a parameterized constructor would fail.

#### Other Changes

- `StringParameterConverter` — new extraction/conversion system used by the CDI synthetic producers
- `CdiProxyUnwrapper` — utility for unwrapping Weld and OpenWebBeans client proxies via reflective method handles
- `PropertyInjectorImpl` — skips @Inject-annotated members to avoid double injection when CDI is managing the bean

### [RESTEASY-3754](https://redhat.atlassian.net/browse/RESTEASY-3754)

This fixes a bug in the `PriorityServiceLoader` which did not include all services if the services had the same priority.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -73,6 +73,38 @@ jobs:
           name: server-logs-${{ matrix.os }}-${{ matrix.java }}
           path: '**/server.log'
 
+  additional-profiles:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        profile:
+          - '-Denhanced.cdi.disabled'
+
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up JDK 25
+        uses: actions/setup-java@v5
+        with:
+          java-version: 25
+          distribution: 'temurin'
+          cache: 'maven'
+      - name: Test with ${{ matrix.profile }} - JDK 25
+        run:  |
+          mvn clean install -U -B -DskipTests
+          cd testsuite
+          mvn verify -U -B -fae ${{ matrix.profile }}
+      - name: Upload test logs for failed run
+        uses: actions/upload-artifact@v7
+        if: failure()
+        with:
+          name: failure-reports${{ matrix.profile }}
+          path: |
+            **/surefire-reports/**
+            **/failsafe-reports/**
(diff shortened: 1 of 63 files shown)


jamezp:
We need to decide if we should create a 7.1 version this goes into or if we're okay with this going into 7.0.3. It's kind of a big change, but should be backwards compatible for the most part. However, I'm seeing issues in WildFly where modules within WildFly need an explicit dependency on `org.jboss.resteasy.resteasy-cdi` so the `ContextProducers` is visible to them.

What I want to avoid is supporting a long term 7.0 if we do a 7.1, but that might be the way we need to go. This CDI integration will really help user to be able to start migrating to using CDI now rather than having to wait for Jakarta REST 5.0.

jamezp:
I've decided to keep this in 7.0 with the ability to disable it. The default is what we're referring to as enhanced CDI is enabled by default. This will be back-ported to 6.2 as well, but for 6.2 we will disable it by default.

PR #4941, opened 2026-07-28
Make wildfly-maven-plugin GAV configurable

tommaso-borgato:
Make `wildfly-maven-plugin` GAV configurable with one notable exception: the `<goal>execute-commands</goal>` execution since the productized `wildfly-maven-plugin` doesn't have the `execute-commands` goal


tommaso-borgato:
@jamezp I am not sure why some tests are failing ... this PR doesn't actually touch any line of code and the changes to the pom.xml files shouldn't change anything either: the changes in this PR are just to add the option to to override the wildfly-maven-plugin GAV in case it's needed

jamezp (approved):

PR #5006, opened 2026-08-31
Implemented ClientListener API

neenapj:


resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientBuilderImpl.java
```diff
@@ -75,6 +76,7 @@ public class ResteasyClientBuilderImpl extends ResteasyClientBuilder {
     protected boolean disableAutomaticRetries = false;
     protected boolean followRedirects;
     private boolean useAsyncHttpClient;
+    protected List<ClientListener> listeners = new ArrayList<>();
```
jamezp: This should be `private` and `final`.

resteasy-client/src/main/java/org/jboss/resteasy/client/jaxrs/internal/ResteasyClientImpl.java
```diff
@@ -93,6 +108,9 @@ public void close() {
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
+        for (ClientListener listener : listeners) {
+            listener.closed(this);
+        }
```
jamezp, line 113: Something we might want to think of for the API part too. I wonder if we should have an overloaded:
```java
ClientListener.close(Client, Throwable);
```

Then we can notify users if the close fails too.

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/client/ClientListenerTest.java
```diff
@@ -0,0 +1,160 @@
...
+import org.jboss.resteasy.client.jaxrs.ResteasyClient;
+import org.jboss.resteasy.client.jaxrs.internal.ResteasyClientBuilderImpl;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * @tpSubChapter Resteasy-client
+ * @tpChapter Unit tests
+ * @tpTestCaseDetails Tests for ClientListener support in ResteasyClientBuilderImpl and ResteasyClientImpl
+ * @tpSince RESTEasy 7.0
```
jamezp: This will end up being 8.0.

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/client/ClientListenerTest.java
```diff
@@ -0,0 +1,160 @@
...
+    public void listenerReturnsBuilder() {
+        ResteasyClientBuilderImpl builder = (ResteasyClientBuilderImpl) ClientBuilder.newBuilder();
+        ClientListener listener = new ClientListener() {
+        };
+
+        ClientBuilder returned = builder.listener(listener);
+
+        Assertions.assertSame(builder, returned,
+                "listener() must return the same builder instance");
+        builder.build().close();
```
jamezp: I understand why the test is written like this as we have a lot of tests written like this. However, new tests shouldn't do this IMO. It should probably look more like;
```java
ClientBuilder builder = ClientBuilder.newBuilder();
try {
    ClientListener listener = new ClientListener() {
    };

    ClientBuilder returned = builder.listener(listener);

    Assertions.assertSame(builder, returned,
            "listener() must return the same builder instance");
} finally {
    builder.build().close();
}
```

I'm not actually sure we need to close it, but it won't hurt.

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/client/ClientListenerTest.java
```diff
@@ -0,0 +1,160 @@
...
+        ClientListener listener = new ClientListener() {
+            @Override
+            public void closed(Client client) {
+                called.set(true);
+                closedClient[0] = client;
+            }
+        };
+
+        Client client = ClientBuilder.newBuilder().listener(listener).build();
+        client.close();
```
jamezp: This should just be:
```java
try (Client client = ClientBuilder.newBuilder().listener(listener).build()) {
}
```

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/client/ClientListenerTest.java
```diff
@@ -0,0 +1,160 @@
...
+     * @tpTestDetails Register a listener on the builder, build a first client, add a second listener, then
+     *                build a second client. Verifies that listeners are snapshotted at build() time so that
+     *                additions to the builder after build() do not affect already-built clients.
+     * @tpPassCrit Listener added to builder after build() does not fire on the already-built client
+     * @tpSince RESTEasy 7.0
+     */
+    @Test
+    public void listenerSnapshotTakenAtBuildTime() {
+        AtomicBoolean firstListenerCalled = new AtomicBoolean(false);
+        AtomicBoolean secondListenerCalled = new AtomicBoolean(false);
```
jamezp: We should probably use an `AtomicInteger` here. Then increment the count on each close. The `firstListenerCalled` should be invoked twice, while the `secondListenerCalled` should be invoked once after the final close.

neenapj: Done

testsuite/unit-tests/src/test/java/org/jboss/resteasy/test/client/ClientListenerTest.java
neenapj: Done

PR #5014, opened 2026-09-03
Cancel async timeout watchdogs and stop keeping them past shutdown

stevenschlansker:
Note: I tried to file a RESTEASY-XXXXX issue for this, but somehow seem to have lost my Jira Reporter status since I last visited here. Apologies if I somehow missed on this one. Thanks for considering this report!

**Observed**: a service with long-poll `@Suspended` endpoints (30-second timeouts, a few requests per second) took a constant extra 10.1 seconds on every deployment stop; a thread dump shows the stopping thread parked in `ContextualExecutorService.awaitTermination` with uncancelled watchdogs in the queue.

`Servlet3AsynchronousResponse` schedules a timeout watchdog per suspended request and only cancels it when a timeout is re-set. `resume()`, `complete()`, `cancel()`, and `close()` leave the watchdog queued until its delay elapses. With a deployment in context, the watchdogs accumulate on the deployment's shared scheduled executor; without one, each response's private pool thread survives until the watchdog fires.

That pool also keeps delayed tasks after shutdown() (the `ScheduledThreadPoolExecutor` default), so `ResteasyDeploymentImpl.stop()` waits out its full 10-second awaitTermination on any deployment that recently served suspended requests, and only a debug log mentions it.

Cancel the watchdog in `close()`, the funnel for every terminal transition, and create the pool with
`executeExistingDelayedTasksAfterShutdownPolicy(false)` and `removeOnCancelPolicy(true)`.

(Analysis produced using Claude Fable)

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
@@ -242,8 +242,10 @@ public class ContextualExecutors {
      */
     public static ContextualScheduledExecutorService newScheduledThreadPool() {
         final int poolSize = CoreThreads.getCoreThreads("resteasy.async.timeout.scheduler.min.pool.size");
-        final ScheduledExecutorService delegate = Executors.newScheduledThreadPool(poolSize,
+        final ScheduledThreadPoolExecutor delegate = new ScheduledThreadPoolExecutor(poolSize,
                 new ContextualThreadFactory("contextual-scheduled-pool"));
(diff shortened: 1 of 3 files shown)


jamezp:
Thank you @stevenschlansker. I created https://redhat.atlassian.net/browse/RESTEASY-3795 for this. No worries on the Jira account issues. I think they get auto-disabled after 90 days or something like that now.

jamezp:
I'll have a closer look at this, but a test does seem to be hanging with this change.

stevenschlansker:
Thanks, test hang should be fixed with the latest push.

PR #5015, opened 2026-09-03
[RESTEASY-3313] Reuse the provider factory's ThreadContexts when a thread has none

stevenschlansker:
`ContextualExecutors.getContexts()` created a new `ThreadContexts`, and with it a full `ServiceLoader` scan of the `ThreadContext` providers, each time it wrapped a task on a thread whose context data holds no `ThreadContexts` entry. A client application that submits requests from its own threads pays that classpath scan on every request submission (RESTEASY-3313).

Cache a lazily created `ThreadContexts` on the `ResteasyProviderFactory` instance and use it as the fallback source of contexts for threads whose context class loader is the loader that defined the factory. The cached providers load from that defining loader, so a factory shared across deployments cannot cache one deployment's provider view for the others, and cannot retain an undeployed application's class loader. A thread with a different context class loader keeps the previous behavior, a fresh `ThreadContexts` per wrapped task. The cache follows the lifecycle of the factory, and no static classloader-keyed state is added. `ResteasyProviderFactoryDelegate` and
`ThreadLocalResteasyProviderFactory` forward `getThreadContexts()` to their delegate, so the cache lives on the real factory.

The new accessor also gives applications a supported point to register a `ThreadContext` that all same-loader threads can see. Before, a registration was visible only to threads that hold a `ThreadContexts` entry in their context data.

Fix built with Claude Fable.

### DIFF
--- a/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
+++ b/resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
@@ -420,16 +420,31 @@ public class ContextualExecutors {
     private static Map<ThreadContext<Object>, Object> getContexts() {
         final Map<ThreadContext<Object>, Object> contexts = new LinkedHashMap<>();
         // Load any registered providers
-        ThreadContexts threadContexts = ResteasyProviderFactory.getInstance()
-                .getContextData(ThreadContexts.class);
-        // Create a new ThreadContexts which will load at least the ones from services
+        final ResteasyProviderFactory factory = ResteasyProviderFactory.getInstance();
+        ThreadContexts threadContexts = factory.getContextData(ThreadContexts.class);
         if (threadContexts == null) {
-            threadContexts = new ThreadContexts();
+            threadContexts = fallbackContexts(factory);
         }
         for (ThreadContext<Object> context : threadContexts.getThreadContexts()) {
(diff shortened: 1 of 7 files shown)


resteasy-core-spi/src/main/java/org/jboss/resteasy/concurrent/ContextualExecutors.java
```diff
@@ -420,18 +420,33 @@ private static Runnable runnable(final Map<ThreadContext<Object>, Object> contex
...
+     * others.
+     */
+    private static ThreadContexts fallbackContexts(final ResteasyProviderFactory factory) {
+        final ClassLoader tccl = Thread.currentThread().getContextClassLoader();
+        final ClassLoader factoryLoader = factory.getClass().getClassLoader();
+        if (tccl == factoryLoader || (tccl == null && factoryLoader == ClassLoader.getSystemClassLoader())) {
+            return factory.getThreadContexts();
+        }
+        return new ThreadContexts();
+    }
```
jamezp, line 448: I'm not sure I'm following this. Is here any reason we shouldn't simply always use the `ResteasyProviderFactory.getThreadContexts()`? Or what might even be better is to load it once and place it in the context data. We already do it partially in the `ResteasyProviderFactoryImpl.processProviderInstanceContracts()`. We should probably just add it by default to the `ResteasyContext`.

PR #5021, opened 2026-09-12
Support records as @MultipartForm parameter containers

jnbdz:
`MultipartFormAnnotationReader` instantiates the `@MultipartForm` parameter container with `Class.newInstance()` and then assigns the annotated fields reflectively. Records have no no-argument constructor and their fields are final, so a record used as a parameter container has always failed with:

```
java.lang.NoSuchMethodException: ...UploadRecord.<init>()
```

This was reported against Quarkus in quarkusio/quarkus#28143 (RESTEasy Classic there), where it was traced back to this reader.

Records are now read by resolving each `RecordComponent` against the form data and invoking the canonical constructor. Since an annotation placed on a record component is propagated to the field, accessor or constructor parameter depending on its `@Target`, the annotation is looked up on the component, its accessor and its backing field. Components without a matching part keep the default value for their type, and `InputPart`/`InputStream` components still keep the input open, as for classes.

`MultipartFormRecordTest` covers an explicit `@FormParam("name")`, the RESTEasy `@FormParam` with an implicit name, conversion to `int`, an absent object part and an absent primitive part. Without the change it fails with the `<init>()` error above.

There is no Jira issue for this yet — @jamezp if you would like one, I am happy to file it or update the title with whichever key you assign.


theigl:
The same would be great for `@BeanParam`.

